PR #399, opened 2017-01-17
Hotkey Mode

totallynotmattroks:
Adds the standard hotkey mode from Bay. Shouldn't have any issues. As a ghost, page-up and page-down should allow you to move up and down a z-level in hotkey mode as well.


fluorescent-net:
https://github.com/Endless-Horizon/CEV-Eris/issues/385

PR #437, opened 2017-01-31
Stairs

LethalGhost:


LethalGhost:
Ghosts, AI-eye, robots, humans, etc can use ladders/stairs via bump or click (robots still can't use ladders).
Using ladders take some time.
Climbing show log on both part (sides?) of ladder.

PR #584, opened 2017-02-26
Clothing fixes

SpadesSlick:
Removing nipples from First Officer back, fixin wedding dress on males, and other questionable stuff.


MyLittleLoly:
сlosed due commit duplicate of #585

PR #600, opened 2017-03-07
[Ready To Merge] The Holopad changes: Hologram call

Quardbreak:
### Changes:

- Now you can call someone via holopad. Both you and receiver of call should be standing near holopads. Incoming call should be approved by receiver.
- Holopads work on all Z-levels, thereby giving you ability to call 5th level from 3rd.

If there are no players near holopads, the call starter will talk to himself (holopad will repeat all his phrases).

Haven't tested some features because haven't got enough people. If there will be any testers - we will quickly test this PR and I will try to fix any bugs (if present).
I think character animation should also work, but haven't tested yet.

If this PR will be approved - it will be quite good to add more holopads on the ship.


Quardbreak:
Я не знаю, почему СЕЙЧАС ничего не работает, вчера вечером всё офигенно работало :\

MyLittleLoly:
Когда местный трап просит вытащить из брига, а вы с пацанами лампово сидите в баре.
image

PR #641, opened 2017-03-16
Openspace icon record lowered turf dir

LethalGhost:
сlose #636;


fluorescent-net:
Please, provide proper issue number for the one that you want to close.

PR #681, opened 2017-03-27
catwalks destroy fix

top4secret5drvn:


fluorescent-net:
https://github.com/Endless-Horizon/CEV-Eris/issues/497

PR #698, opened 2017-04-02
Forbid augmentations selection for christians in preferences setup & sound loop initial

Camper12345:


code/game/sound_loop.dm
```diff
@@ -0,0 +1,127 @@
...
+	update_listener(M, S)
+	M << S
+
+
+datum/sound_loop/proc/remove_listener(var/mob/M)
+	var/sound/S = listeners[M]
+	S.status = SOUND_PAUSED | SOUND_UPDATE
+	S.repeat = FALSE
+	qdel(S)
+	listeners -= M
```
fluorescent-net: Please prefer using .Add() and .Remove() methods for byond lists.

code/game/sound_loop.dm
```diff
@@ -0,0 +1,127 @@
...
+	playing = TRUE
+	src.force_update = TRUE
+
+
+datum/sound_loop/proc/stop()
+	playing = FALSE
+	src.force_update = TRUE
+
+
+datum/sound_loop/proc/set_volume(var/newvol)
```
fluorescent-net: Please use "snake-case" variables style and avoid abbreviations (var/new_volume should be nice). Same comments for other datum methods.

Issue #738, opened 2017-04-10
Gravity generator spamming with announcements if it's offline


Sharmount:
[Common] CEV Eris System Announcer states, "Priority Announcement: The gravity generator was brought offline."
You will hear it every 5 seconds, while it's offline.

Issue #729, opened 2017-04-09
Running into walls continues to play walk sound


mattroks101:
Run into a wall and hold the direction you're running. You won't move but the walk sound will persist.

PR #750, opened 2017-04-12
Gravity generator and footstep fix

Camper12345:
* Gravity generator no more flood off announcemets
* Footsteps not playing, if you run in the wall


Camper12345:
#729 #738

PR #774, opened 2017-04-23
Bluespace harpoon rework

Camper12345:
* Harpoon now will not teleport you to contents of space cleaner
* Some rebalance
* Transforming now work correctly


Camper12345:
#131

PR #785, opened 2017-04-25
Organs3

LethalGhost:


code/modules/organs/subtypes/robotic.dm
```diff
@@ -11,6 +11,14 @@
 	var/list/forced_children = null
 	var/attack = 0 			//Attack mode
```
Camper12345: If variable must contain bool value, use TRUE / FALSE, if variable must contain object, use null to clear and specify variable type

code/modules/organs/internal/eyes.dm
```diff
@@ -0,0 +1,85 @@
...
+		target.update_eyes()
+	..()
+
+/obj/item/organ/eyes/proc/update_colour()
+	if(!owner)
+		return
+	eyes_color = owner.eyes_color
+
+
+obj/item/organ/eyes/New(var/mob/living/carbon/holder,var/datum/organ_description/OD)
```
Camper12345: Insert space between the arguments and "/" before the proc path.

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -2,7 +2,7 @@
 	Global associative list for caching humanoid icons.
 	Index format m or f, followed by a string of 0 and 1 to represent bodyparts followed by husk fat hulk skeleton 1 or 0.
 	TODO: Proper documentation
-	icon_key is [species.race_key][g][husk][fat][hulk][skeleton][s_tone]
+	icon_key is [species.race_key][husk][fat][hulk][skeleton]
```
LethalGhost: Порядок нарушен.

PR #787, opened 2017-04-28
Fix fixes

Camper12345:
* Fixed mob deafness in closets
* Fixed catwalk footstep sounds shit

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -530,3 +530,6 @@ its easier to just keep the beam vertical.
 /turf/Entered(var/atom/movable/AM, var/atom/old_loc, var/special_event)
 	return ..(AM, old_loc, 0)
 
+/atom/proc/get_footstep_sound()
+	return
+
--- a/code/game/turfs/turf.dm
+++ b/code/game/turfs/turf.dm
@@ -233,12 +233,13 @@ var/const/enterloopsanity = 100
 	return
 
 /turf/get_footstep_sound(var/mobtype)
-	if(!footstep_sounds || !footstep_sounds.len)
-		return
 
-	var/sound = pick(footstep_sounds[mobtype])
+	var/sound
 
-	if(locate(/obj/structure/catwalk) in src)
-		sound = pick('sound/effects/footstep/catwalk1.ogg', 'sound/effects/footstep/catwalk2.ogg', 'sound/effects/footstep/catwalk3.ogg', 'sound/effects/footstep/catwalk4.ogg', 'sound/effects/footstep/catwalk5.ogg')
+	var/catwalk = locate(/obj/structure/catwalk) in src
+	if(catwalk)
+		sound = safepick(catwalk.footstep_sounds[mobtype])
+	else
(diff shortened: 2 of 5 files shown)


code/game/atoms.dm
```diff
@@ -530,3 +530,6 @@ its easier to just keep the beam vertical.
 /turf/Entered(var/atom/movable/AM, var/atom/old_loc, var/special_event)
 	return ..(AM, old_loc, 0)
 
+/atom/proc/get_footstep_sound()
```
LethalGhost: Make `get_footstep_sound()` turf proc?

code/game/turfs/turf.dm
```diff
@@ -231,3 +231,14 @@ var/const/enterloopsanity = 100
 
 /turf/proc/update_blood_overlays()
 	return
+
+/turf/get_footstep_sound(var/mobtype)
+	if(!footstep_sounds || !footstep_sounds.len)
+		return
+
+	var/sound = pick(footstep_sounds[mobtype])
```
LethalGhost: May be replace with `safepick()`

code/game/turfs/turf.dm
```diff
@@ -231,3 +231,14 @@ var/const/enterloopsanity = 100
...
+
+	if(locate(/obj/structure/catwalk) in src)
```
LethalGhost: Can rewrite `pick` in 239. Better check that first?

code/modules/mob/living/carbon/human/human_movement.dm
```diff
@@ -110,42 +110,41 @@
...
+
+			if(!has_gravity(src))
+				if(step_count % 3) //this basically says, every three moves make a noise
+					return //1st - none, 1%3==1, 2nd - none, 2%3==2, 3rd - noise, 3%3==0
+
+			if(species.silent_steps)
+				return //species is silent
+
+			playsound(T, S, volume, 1, range)
+			return
```
LethalGhost: Hm. Better move checks like that (can stop proc): 
```
if(species.silent_steps)
	return //species is silent
```
upper than setting up sound.

code/modules/mob/living/say.dm
```diff
@@ -242,13 +242,10 @@ proc/get_radio_key_from_channel(var/channel)
...
 		var/list/hear = hear(message_range,T)
-		var/list/hearturfs = list()
 
-		for(var/I in hear)
-			if(ismob(I))
-				var/mob/M = I
-				listening += M
-				hearturfs += M.locs[1]
+		for(var/mob/M in mob_list)
+			if(M && M.locs.len && M.locs[1] in hear)
```
LethalGhost: `M` can not be `null`. 
`for(var/TYPE/VAR in LIST)`  implicitly call `istype(VAR, TYPE)`

code/game/atoms.dm
Camper12345: footstep_sounds list defined in atom, so this proc too.

LethalGhost: Move both?

PR #795, opened 2017-05-13
Showcase quick-fix.

LethalGhost:
Objs damage procs req rework?


code/game/machinery/doors/blast_door.dm
```diff
@@ -178,7 +178,7 @@ obj/machinery/door/blast/regular
 	maxhealth = 600
 	block_air_zones = 1
 
-obj/machinery/door/blast/regular/open
+/obj/machinery/door/blast/regular/open
 	icon_state = "pdoor0"
 	density = 0
 	opacity = 0
```
Camper12345: TRUE / FALSE instead of 1 / 0

code/game/machinery/doors/blast_door.dm
```diff
@@ -178,7 +178,7 @@ obj/machinery/door/blast/regular
 	maxhealth = 600
 	block_air_zones = 1
```
Camper12345: TRUE / FALSE instead of 1 / 0

PR #802, opened 2017-05-15
A fix for the constant Supermatter delamination.

### LINKED ISSUE #801: Supermatter core stuck on the shutter when ejected
Yeah, i tried to eject core before it explodes(about 30% integrity, air appeared from somewhere[and most likely didn't got filtered by gas disposal system, but i'm not sure] and heat was going up) but when i pressed button and went down to check, it was suck on a shutter door which is leading outside.

astralenigma:
Fixes the engine pipe and adds a button to vent excess atmosphere in the SM chamber and allows to eject the SM successfully.
Also gives a buff to the actual main SMES.
Fixes #801.


SpadesSlick:
Sound good, will check it tomorrow.

PR #834, opened 2017-06-15
[WIP] Fixes "я" issue in notes

KuzyXD:
+ I made another one \n for beauty.


KuzyXD:
Пока не принииать.

KuzyXD:
После чека - можно.

Issue #865, opened 2017-06-26
Recharges dont work


SpadesSlick:
Looks like new battery system broke it. We need to finish that and than fix that.

PR #881, opened 2017-07-02
Slightly fixed batteries bug.

KuzyXD:
#865


LethalGhost:
No no no.
We need rewrite `/obj/item/weapon/cell/big/proc/give` to `/obj/item/weapon/cell/proc/give` not copy-past it.
How that pass Travis also?

PR #890, opened 2017-07-06
Fix Stun if Entered on Underplating Floor

Penguinboy95:
Now when you entering on Underplating Floor, stun is comes instantly


code/game/turfs/simulated/floor_types.dm
```diff
@@ -131,6 +131,7 @@
 	if(M.m_intent == "run")
 		if(prob(75))
 			M.adjustBruteLoss(5)
+			M.slip()
```
LethalGhost: Effect of default `stun_on_slip` and `weak_on_slip` will be added. 
Confirm that it is intended, please.

PR #891, opened 2017-07-07
Fix #766 Alden-Saraspova Counter Missing Sprite

### LINKED ISSUE #766: Alden-Saraspova counter is missing sprite

Penguinboy95:
Static sprite. After using an item turns on an animation which lasts for 2 seconds. After that a sprite goes back to a static.
image
image


LethalGhost:
Recreate with #894

PR #903, opened 2017-07-12
Fix and simplify radio-chanels declaration

LethalGhost:


LethalGhost:
All ways work fine now: cyrillic and latin, upper and lowercase.

PR #913, opened 2017-07-19
Reworking power cells and devices they are using. Close #865, close #794.

### LINKED ISSUE #794: Replaceable batteries for energy weapons.
This is already made on some level. We need to use new batteries for that, with new sprites.

They should be standardized for two sizes:
1. Small one, for pistols.
2. Big one, for rifles.
 
Within those standards, however, old gradation of hyper, super capacity still can exist.

### LINKED ISSUE #865: Recharges dont work

SpadesSlick:
New size standards - small, medium, big.
Mesons, thermals, analyzers and other things now use cells.
Cell from energy guns now can be removed and replaced.


code/modules/clothing/glasses/thermal.dm
```diff
@@ -0,0 +1,90 @@
...
+	prescription = 1
+	vision_flags = SEE_MOBS
+	see_invisible = SEE_INVISIBLE_NOLIGHTING
+	flash_protection = FLASH_PROTECTION_REDUCED
+	active = 0
+	var/tick_cost = 0.5
+	var/obj/item/weapon/cell/cell = null
+	var/suitable_cell = /obj/item/weapon/cell/small
+
+	emp_act(severity)
```
fluorescent-net: Please provide full method path (i.e. /obj/item/clothing/glasses/thermal/emp_act()).

code/modules/clothing/glasses/thermal.dm
```diff
@@ -0,0 +1,90 @@
...
+		if(ishuman(src.loc))
+			var/mob/living/carbon/human/M = src.loc
+			M << "<span class='danger'>The Optical Thermal Scanner overloads and blinds you!</span>"
```
fluorescent-net: Looks like this output should be after next 'if' condition. Also please use [src.name] instead of object name hardcoding.

code/modules/projectiles/guns/energy.dm
```diff
@@ -83,22 +78,39 @@
...
+		update_held_icon()
+
+/obj/item/weapon/gun/energy/MouseDrop(over_object)
+	if(!self_recharge)
+		if((src.loc == usr) && istype(over_object, /obj/screen/inventory/hand) && eject_item(cell, usr))
+			cell = null
+			update_icon()
+	else
+		usr << "<span class='warning'>[src] is a self-charging gun, its batteries cannot be removed!.</span>"
+		return
```
fluorescent-net: Return statement seems redundant here and in line#116.

LethalGhost:
Also. PR should be linked with #865.

PR #897, opened 2017-07-09
Basic Economy - Payday

Rodial-Admiral:
Adds new type of events - economic events. Currently this adds additional funds to each active account every 30 minutes, depending on account's owner's job. Also replaces thalers with credits.

PR #929, opened 2017-07-23
Payday (by Rodial-Admiral). Close #824. Close  #825.

### LINKED ISSUE #824: Basic Economy: Payday
Upon every 30 minutes of round everyone must receive a fixed amount of credits, so people can waste them in vendors and buy things. Later it will be more complicated.

We also need to announce to radio every time payday arrives. same way it works for meteors, radiation and other things.

### LINKED ISSUE #825: Basic Economy: Replace thalers with credits.
Why that's even a thing.

LethalGhost:


LethalGhost:
Replace #897

LethalGhost:
Don not forget.
"the word credits should be made plural and non-capitalized (credits, 1000 credits, pile of credits) in every case, so we won't get mess of words with different capitalization and plurality."

PR #934, opened 2017-07-24
New cells are now can be researched and produced in RnD.

SpadesSlick:
image
image


fluorescent-net:
Okay, we'll refactor it later.

PR #964, opened 2017-08-07
Admin permissions loading/setting refactoring

fluorescent-net:


fluorescent-net:
Deleted player notes should be reworked in https://github.com/discordia-space/CEV-Eris/issues/970.

PR #975, opened 2017-08-16
Improvement radio usage

LethalGhost:
Allow using multiple headsets simultaniously
Allow using radios (with :r/:l) in hands while wearing headset/s.


LethalGhost:
Ported from Old World Blue (ru comm)

PR #991, opened 2017-09-04
Fix asteroid lights. Fix #821

### LINKED ISSUE #821: Light broken on mines, again
Thats how light works on generated asteroid level:
image
And thats how it looks like if we dont generate it, by disabling it from config:
image

We fix that in the past, but looks like we miss something.

LethalGhost:


code/modules/lighting/lighting_system.dm
```diff
@@ -41,5 +41,5 @@
 /world/New()
 	. = ..()
 
-	create_all_lighting_corners()
-	create_all_lighting_overlays()
\ No newline at end of file
+	create_all_lighting_overlays()
+	create_all_lighting_corners()
```
fluorescent-net: Please add empty line at the end of file.

fluorescent-net: Please add empty line at the end of file.

PR #1036, opened 2017-10-16
Rework and refactor admin verbs/permissions 

fluorescent-net:
This is a stripped version of https://github.com/discordia-space/CEV-Eris/pull/964 without schema-related changes.

-  Remove a bunch of redundant/duplicate admin verbs
-  Replace admin verbs categories to more relatable
-  Refactor and simplify admin permissions system
-  Remove shitty player notes functionality (should be reworked in #970)
-  Slightly improve code style


code/modules/virus2/admin.dm
```diff
@@ -32,8 +33,9 @@
 /datum/admins/var/datum/virus2_editor/virus2_editor_datum = new
 /client/proc/virus2_editor()
 	set name = "Virus Editor"
-	set category = "Admin"
-	if(!holder || !check_rights(R_SPAWN)) return // spawn privileges to create viruses
+	set category = "Fun"
+	if(!holder || !check_rights(R_FUN))
+		return // spawn privileges to create viruses
```
LethalGhost: Proc req R_ADMIN right for showing up.

code/__defines/admin.dm
```diff
@@ -22,21 +22,13 @@
...
-#define R_MENTOR        0x4000
-#define R_HOST          0x8000 //higher than this will overflow
+#define R_FUN           0x1
+#define R_SERVER        0x2
+#define R_DEBUG         0x4
+#define R_PERMISSIONS   0x8
+#define R_MENTOR        0x10
+#define R_MOD           0x20
+#define R_ADMIN         0x40
+#define R_HOST          0x80
```
LethalGhost: R_HOST only allow delete admin memo created by other admins and used as default required right for admin secret but overrided in each case.

fluorescent-net: So should we get rid of this flag?

LethalGhost: Think so. We had remove `R_BUILDMODE` because of same reason hadn't we?

PR #1042, opened 2017-10-21
Deleted one else input() form and optimisation!

seekin4u:
Deleted one else input() form and optimisation!

When you placing item
in deconstructor - you need to accept your
choise(useless shit
yea).
Besides, checks like ```item.null_variable == null``` replaced on
BYOND
macro
```
view == null 10000 = 33.4 sec
view isnull() 10000 = 32.4
sec << it's a liiiitle faster
```
and also markdown in file.


seekin4u:
Alt text

PR #1068, opened 2017-10-27
Say/Me button call typing indicator showing

LethalGhost:


code/modules/mob/say.dm
```diff
@@ -6,19 +6,46 @@
 	set category = "IC"
 	return
 
+
+/mob/verb/say_wrapper()
+	set name = "Say verb"
+	set category = "IC"
+
+	set_typing_indicator(1)
+	hud_typing = 1
```
fluorescent-net: Could you please use TRUE/FALSE as boolean values? Applies to the rest of PR as well.

code/modules/mob/say.dm
```diff
@@ -6,19 +6,46 @@
...
+	var/message = input("","say (text)") as text
```
fluorescent-net: Please add space between function arguments.

code/modules/mob/say.dm
```diff
@@ -6,19 +6,46 @@
...
 	usr.say(message)
 
+
+/mob/verb/me_wrapper()
+	set name = "Me verb"
+	set category = "IC"
+
+	set_typing_indicator(1)
+	hud_typing = 1
+	var/message = input("","me (text)") as text
```
fluorescent-net: Please add space between function arguments.

PR #1333, opened 2017-12-29
alt click for knife boot

TheReduce:
alt click for knife boot


code/modules/clothing/clothing.dm
```diff
@@ -312,7 +312,9 @@ BLIND     // can't see anything
 	update_icon()
 	return
 
-
+/obj/item/clothing/shoes/AltClick()
+	if(src in usr)
+		draw_knife()
```
LethalGhost: Add 
```
else
    ..()
```
To keep original behavior

TheReduce: done

PR #1334, opened 2017-12-29
Cyrillic language

seekin4u:
New chance-based language to all ```/datum/job/```
See commits' desc to know more.

Added cyrillirc to technomants with 100% and much less to all another crew (15 - 20 %)

Chancing depends of also_known_languages associative list with languages NAME and chance of spawning. 
Functions located in job_controller.dm as it is a part of job system, so it gain as like as job equipment and clothing.
(Small addition to it - mind now have ```assigned_job``` assigned to it when binding of mind and ```/datum/job``` happens. Before it was just null)


code/game/jobs/job/job.dm
```diff
@@ -130,6 +131,26 @@
 
 	return 1
 
+datum/job/proc/add_additiional_language(var/mob/living/carbon/human/target)
```
LethalGhost: Do not skip `/`.

code/game/jobs/job/job.dm
```diff
@@ -130,6 +131,26 @@
...
+	if(!target)
+		return 0
+
+	if(!ishuman(target))
+		return 0
```
LethalGhost: Use TRUE/FALSE instead 1/0

code/game/jobs/job/job.dm
```diff
@@ -130,6 +131,26 @@
 
 	return 1
 
+datum/job/proc/add_additiional_language(var/mob/living/carbon/human/target)
+	if(!target)
+		return 0
+
+	if(!ishuman(target))
```
LethalGhost: Previous check included in that.

code/game/jobs/job/job.dm
```diff
@@ -130,6 +131,26 @@
...
+		return 0
+
+	var/mob/living/carbon/human/H = target
+
+	if(!H.mind || !H.mind.assigned_job || !also_known_languages.len)
```
LethalGhost: Checks except `!also_known_languages.len` are redundant.

code/modules/mob/language/outsider.dm
```diff
@@ -94,3 +94,24 @@
 	colour = "cult"
 	key = "y"
 	flags = RESTRICTED | HIVEMIND
+
+/datum/language/russian
```
LethalGhost: "Normal" languages defination should be placed in `/code/modules/mob/language/station.dm`

PR #1378, opened 2018-01-14
Refactored scanners. 

seekin4u:
Say NO to copypasta + shitcode and replace it by half-shitcode.


code/game/objects/items/devices/scanners.dm
```diff
@@ -7,36 +7,65 @@ GAS ANALYZER
...
-	..()
-	if(!cell && suitable_cell)
-		cell = new suitable_cell(src)
+	var/mode = 1
 
-/obj/item/device/healthanalyzer/attack(mob/living/M, mob/living/user)
-	if(!cell || !cell.checked_use(3))
-		user << SPAN_WARNING("[src] battery is dead or missing.")
+/obj/item/device/scanner/healthanalyzer/attack(mob/living/M, mob/living/user)
+	cell_use_check(3)
```
deathride58: This won't cancel out a scan if there's a dead or missing battery, it'll just display the "It's out of battery" warning and let the scan go through anyhow.

Replace it with if(cell_use_check(3)), and make cell_use_check() return either TRUE or FALSE depending on if the check failed or not.

You could also move this check, along with the user.incapacitated() check to the /obj/item/device/scanner type, and call it in the subtype's attack procs with
`. = ..()
if(!.)
    return`

PR #1460, opened 2018-03-13
Disk copying machine

Camper12345:
There is a map compile error because of old disk type on the map.

### DIFF
--- a/code/game/machinery/autolathe.dm
+++ b/code/game/machinery/autolathe.dm
@@ -12,7 +12,7 @@
 	var/obj/item/weapon/disk/autolathe_disk/disk = null
 
 	var/list/stored_material =  list()
-	var/list/storage_capacity = list(MATERIAL_STEEL = 0, MATERIAL_GLASS = 0)
+	var/list/storage_capacity = list(MATERIAL_STEEL = 2500, MATERIAL_GLASS = 0)
 	var/show_category = "All"
 
 	var/hacked = 0
@@ -208,9 +208,12 @@
 		make(making_recipe.type,making_left)
 
 	nanomanager.update_uis(src)
-	return
+
 
 /obj/machinery/autolathe/attack_hand(mob/user as mob)
+	if(..())
+		return
+
 	user.set_machine(src)
 	ui_interact(user)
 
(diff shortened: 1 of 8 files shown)


code/game/machinery/autolathe.dm
```diff
@@ -12,7 +12,7 @@
 	var/obj/item/weapon/disk/autolathe_disk/disk = null
 
 	var/list/stored_material =  list()
-	var/list/storage_capacity = list(MATERIAL_STEEL = 0, MATERIAL_GLASS = 0)
+	var/list/storage_capacity = list(MATERIAL_STEEL = 2500, MATERIAL_GLASS = 0)
```
LethalGhost: Разборка/сборка будет восстанавливать запас ресурсов.

code/game/machinery/autolathe.dm
```diff
@@ -103,49 +103,111 @@
...
+		return
+
+	user.set_machine(src)
+	ui_interact(user)
+
+/obj/machinery/autolathe/Topic(href, href_list)
 
-	if(istype(I,/obj/item/weapon/disk/autolathe_disk))
+	if(..())
+		return
```
LethalGhost: Better keep returning value (or just TRUE) for possible subtype overrides.

code/game/machinery/autolathe.dm
```diff
@@ -103,49 +103,111 @@
...
+			var/material = href_list["eject_material"]
+			var/material/M = get_material_by_name(material)
+
+			if(!M.stack_type)
+				return
+
+			var/obj/item/stack/material/sheetType = M.stack_type
+			var/perUnit = initial(sheetType.perunit)
+
+			var/num = input("Enter sheets count to eject. 0-[round(stored_material[material]/perUnit)]","Eject",0) as num
```
LethalGhost: User can move away from src.

code/game/machinery/autolathe.dm
```diff
@@ -103,49 +103,111 @@
...
+	var/obj/item/eating = user.get_active_hand()
+
+	if(!istype(eating))
+		return
+
+	if(istype(eating,/obj/item/weapon/disk/autolathe_disk))
 		if(disk)
 			user << SPAN_NOTICE("There's already a disk inside the autolathe.")
 			return
 		user.drop_item()
```
LethalGhost: Better use 
`user.unEquip(item, new location)` 
instead
```
user.drop_item()
item.forceMove(new location)
```
user.drop_item() - should be used only for situation where user must drop any item (no matter what it was) from hands.

code/game/machinery/autolathe_disk_cloner.dm
```diff
@@ -0,0 +1,204 @@
...
+        // for a list of parameters and their descriptions see the code docs in \code\modules\nano\nanoui.dm
+		ui = new(user, src, ui_key, "autolathe_disk_cloner.tmpl", "Autolathe disk cloner", 480, 555)
+		// when the ui is first opened this is the data it will use
+		ui.set_initial_data(data)
+		// open the new ui window
+		ui.open()
+
+/obj/machinery/autolathe_disk_cloner/Topic(href, href_list)
+	if(..())
+		return
```
LethalGhost: Same as before

code/game/machinery/autolathe.dm
Camper12345: Fix is below line 156.

PR #1516, opened 2018-05-28
Each head of staff that joins round, gets piece of pass code from ship's nuke bomb.

0xA48:
<details><summary>Looks like this in-game (click):</summary>

d1fanuk
</details>
Rest info in commit message.


code/game/machinery/nuclear_bomb.dm
```diff
@@ -23,11 +23,22 @@ var/bomb_set
...
 	var/datum/wires/nuclearbomb/wires = null
 
+	var/eris_ship_bomb = FALSE           // if TRUE (1 in map editor), then Heads will get parts of code for this bomb. Obviously used in map editor. Single mapped bomb supported.
+	var/ship_nuke_code_rotation_part = 1 // Which part of code next spawned Head gets. See "get_next_nuke_code_part()" proc.
+
 /obj/machinery/nuclearbomb/New()
 	..()
-	r_code = "[rand(10000, 99999.0)]"//Creates a random code upon object spawn.
+	if(eris_ship_bomb)
+		r_code = "[rand(100000, 999999)]" // each time new Head spawns, s/he gets 2 numbers of code.
```
LethalGhost: 6 digit pass can't be entered (Check Topic())

code/game/machinery/nuclear_bomb.dm
```diff
@@ -23,11 +23,22 @@ var/bomb_set
 	var/previous_level = ""
 	var/datum/wires/nuclearbomb/wires = null
 
+	var/eris_ship_bomb = FALSE           // if TRUE (1 in map editor), then Heads will get parts of code for this bomb. Obviously used in map editor. Single mapped bomb supported.
+	var/ship_nuke_code_rotation_part = 1 // Which part of code next spawned Head gets. See "get_next_nuke_code_part()" proc.
```
LethalGhost: Unused var?

ZVee: used by copytext() in get_next_nuke_code_part() proc below, it guides procedure which digits we need to copy from password. i can add this in comment if needed.

code/game/machinery/nuclear_bomb.dm
ZVee: Good catch, 2 ways i see for myself on how to resolve this:
1) eris_ship_bomb var check in topic
2) 6-digit codes as standard for all bombs.

Or maybe you have something in mind.
(I'l fix this in couple of hours).

code/game/machinery/nuclear_bomb.dm
LethalGhost: As far as i can see `get_next_nuke_code_part()` is using gametickers variable of same name, not that one.

PR #1523, opened 2018-06-07
Resolved most critical problems with initialization.

0xA48:
* Example: pipes had no parent calls, so they weren't added in processing list and no gas was processed thru them. Computers. Anything else in machinery with same symptoms. Areas were initialized twice (manual call in old controller, then again in new one).
* Is something new now broken? Maybe. Depends on how properly initialization port were done before. I actually touched only most obvious stuff.

### DIFF
--- a/code/modules/research/rdconsole.dm
+++ b/code/modules/research/rdconsole.dm
@@ -108,7 +108,7 @@ won't update every console in existence) but it's more of a hassle to do. Also, 
 		C.files.RefreshResearch()
 
 /obj/machinery/computer/rdconsole/New()
-	..()
+	. = ..()
 	files = new /datum/research(src) //Setup the research data holder.
 	if(!id)
 		for(var/obj/machinery/r_n_d/server/centcom/S in SSmachines.machinery)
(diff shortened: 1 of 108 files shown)


code/modules/research/rdconsole.dm
```diff
@@ -108,14 +108,15 @@ won't update every console in existence) but it's more of a hassle to do. Also,
 		C.files.RefreshResearch()
 
 /obj/machinery/computer/rdconsole/New()
-	..()
+	. = ..()
 	files = new /datum/research(src) //Setup the research data holder.
 	if(!id)
 		for(var/obj/machinery/r_n_d/server/centcom/S in SSmachines.machinery)
 			S.Initialize()
 			break
```
LethalGhost: Mark 1

code/modules/research/server.dm
```diff
@@ -15,10 +15,6 @@
 	req_access = list(access_rd) //Only the R&D can change server settings.
 	circuit = /obj/item/weapon/circuitboard/rdserver
 
-/obj/machinery/r_n_d/server/New()
-	..()
-	Initialize()
-
```
LethalGhost: RDconsole (see Mark 1) will not find server in machinery list. Use `INITIALIZE_HINT_LATELOAD` in RDconsole

0xA48: Done.

PR #1593, opened 2018-06-23
Fixing runtimes related to organ seperation.

### LINKED ISSUE #1554: runtime error: Cannot read null.organ_tag
runtime error: Cannot read null.organ_tag
proc name: attack (/obj/item/stack/medical/attack)
  source file: medical.dm,26

### LINKED ISSUE #1555: runtime error: Cannot read null.open
runtime error: Cannot read null.open
proc name: attack (/obj/item/stack/medical/advanced/bruise_pack/attack)
  source file: medical.dm,189

LudwigVonChesterfield:
Mostly adding null checks after doing get_organ().
- fixes #1554
- fixes #1555


code/game/objects/items/stacks/medical.dm
```diff
@@ -23,6 +23,9 @@
 		var/mob/living/carbon/human/H = M
 		var/obj/item/organ/external/affecting = H.get_organ(user.targeted_organ)
 
+		if(!affecting)
+			to_chat(user, SPAN_WARNING("What [user.targeted_organ]?"))
```
LethalGhost: `return` missed

code/modules/mob/living/carbon/brain/brain_item.dm
```diff
@@ -54,18 +54,21 @@
...
-	var/mob/living/simple_animal/borer/borer = owner.has_brain_worms()
+	name = "[owner_.real_name]'s brain"
+
+	var/mob/living/simple_animal/borer/borer = owner_.has_brain_worms()
 
 	if(borer)
 		borer.detatch() //Should remove borer if the brain is removed - RR
 
 	var/obj/item/organ/internal/brain/B = src
-	if(istype(B) && istype(owner))
```
LethalGhost: `owner` is organ var. Why it have been replaced with proc argument (`owner_`)?

code/modules/mob/living/carbon/human/death.dm
```diff
@@ -38,9 +38,10 @@
...
 	var/mob/living/simple_animal/borer/B
 
-	for(var/I in head.implants)
-		if(istype(I,/mob/living/simple_animal/borer))
-			B = I
+	if(head)
+		for(var/I in head.implants)
+			if(istype(I,/mob/living/simple_animal/borer))
+				B = I
 	if(B)
```
LethalGhost: Probably should be inside `if(head) {...}` block

code/modules/mob/living/carbon/brain/brain_item.dm
LudwigVonChesterfield: Because when head is seperated from the body, it loses it's owner var(as well as the brain). However this entire code is made for the seperated head, which just contradicts itself.

code/modules/mob/living/carbon/human/death.dm
LudwigVonChesterfield: What if somebody finding a borer elsewhere?

code/game/objects/items/stacks/medical.dm
LudwigVonChesterfield: Fixed.

code/modules/organs/external/_external.dm
```diff
@@ -95,9 +95,12 @@
 		for(var/obj/item/organ/O in internal_organs)
 			qdel(O)
 
+	if(owner)
+		owner.organs -= src
+		owner.organs_by_name -= src.name
```
LethalGhost: Not `src.name`, but `src.organ_tag`
Organs with name "left leg" and one with "robotic left leg" has same `organ_tag` (L_LEG)

code/game/objects/items/stacks/medical.dm
LudwigVonChesterfield: <code>var/mob/living/carbon/human/victim = owner_ ? owner_ : owner // Used to keep and put into other's removed procs.</code>

code/modules/mob/living/carbon/brain/brain_item.dm
LudwigVonChesterfield: <code>var/mob/living/carbon/human/victim = owner_ ? owner_ : owner // Used to keep and put into other's removed procs.</code>

code/modules/organs/external/_external.dm
```diff
@@ -137,22 +140,25 @@
 	if(module)
 		module.organ_installed(src, owner)
 
-/obj/item/organ/external/removed(mob/living/user, redraw_mob = TRUE)
-	if(!owner)
+/obj/item/organ/external/removed(mob/living/user, redraw_mob = TRUE, mob/living/carbon/human/owner_)
+
+	var/mob/living/carbon/human/victim = owner_ ? owner_ : owner // Used to keep and put into other's removed procs.
```
LethalGhost: Why we might need to override owner?

LudwigVonChesterfield: "Because when head is seperated from the body, it loses it's owner var(as well as the brain). However this entire code is made for the seperated head, which just contradicts itself."

code/modules/mob/living/carbon/brain/brain_item.dm
LethalGhost: Also. 
`var/mob/living/carbon/human/victim = owner_ ? owner_ : owner`
would not affect `owner_` in `.../organ/external/head/remove()`.

LudwigVonChesterfield: It would, <code>if(internal_organs)
		for(var/obj/item/organ/organ in internal_organs)
			organ.removed(owner_ = victim)</code>

PR #1618, opened 2018-07-22
New evacuation actions from comm programm and command console

ziz11:


ziz11:
Also, there a trouble about default comm consoles, which can't cancell new evac procedures with pods (only can only initiate), and there should be a bug with admin panel, where you probably can't call or cancell evac too. (Probably wrong procs in code)

PR #1626, opened 2018-08-07
More critical fails for tools.

SpadesSlick:
Explosion effect are now more rare. You also can drop tool, stuck it in your hand, or throw it away by accident.


NanakoAC:
wait this PR is over a month old, why is it suddenly active

Issue #1636, opened 2018-08-10
Point arrow is on wrong layer


NanakoAC:
When pointing at something, the arrow doesnt always draw ontop. it should be on a high layer

PR #1642, opened 2018-08-10
Pointer Plane Fix

NanakoAC:
This PR Fixes pointers being on the wrong plane and thus hidden behind most objects


NanakoAC:
Fixes #1636

Issue #1648, opened 2018-08-11
Grammar errors


silentenemy:
- [x] NeoTheology vendomat: **Teo**Mat
- [x] Kitchen dinnerware vendomat: cat **lunc**box
- [x] Railing: A standar**t** steel railing. Prevents from human stupidity.
- [x] Asters Miscellaneous Pack: `disk for autolathe designs.`

If you want to add some more, please, put `- [ ] ` before the text so there'll be a checkbox.

praisenarsie:
![https://i.imgur.com/zWDYvVD.png](https://i.imgur.com/zWDYvVD.png)
circuits

silentenemy:
- [ ] This is Bulletproof Vest.
capitalized somewhy

praisenarsie:
![https://i.imgur.com/7Nx5FQv.png](https://i.imgur.com/7Nx5FQv.png)

removes

Voldirs:
He has a robot Hesphaistos right leg. It hasa lot of dents!

SpadesSlick:
Closed by #1659.

Voldirs:
This is Alliance LG "Cog". 
It is a normal-sized item.
Alliance cheap laser gun. In our deep past - this was main weapon of low-rank police force across all Milky Way, b**ili**ons co**pys** of this gun was made. They still can be found alm**oust** everywhere.

@SpadesSlick that almost made me cry

mistyLuminescence:
As of 21/7/19, all fixed.

PR #1659, opened 2018-08-12
Grammar fixes

praisenarsie:
Fixes some of the grammar issues in #1648


code/game/machinery/vending.dm
```diff
@@ -992,7 +992,7 @@
 
 
 /obj/machinery/vending/teomat
-	name = "NeoTheology Teo-Mat."
+	name = "NeoTheology Teo-Mat"
```
NanakoAC: shouldn't it be Theo-Mat ?

code/game/objects/structures/railing.dm
```diff
@@ -1,6 +1,6 @@
 /obj/structure/railing
 	name = "railing"
-	desc = "A standart steel railing. Prevents from human stupidity."
+	desc = "A standard steel railing. Prevents stupid people from falling into their impending doom."
```
NanakoAC: This one is still not gramatically correct
An appropriate version would be

"A standard steel railing. Prevents stupid people from falling to their doom."

"into" isnt quite right here, and "impending" has no place here at all

code/game/objects/items/weapons/autolathe_disks.dm
```diff
@@ -310,7 +310,7 @@
 	..()
 
 /obj/item/weapon/disk/autolathe_disk/nonlethal_ammo
-	name = "Frozen Star Nonlethal Ammo Pack"
```
NanakoAC: I think Nonlethal all as one word is fine here.
This is an item name, which means space is at a premium

PR #1671, opened 2018-08-12
Reducing the amount of material when disassembling the modular console

S-konoplich:


code/modules/modular_computers/computers/machinery/modular_console.dm
```diff
@@ -13,7 +13,7 @@
 	base_idle_power_usage = 100
 	base_active_power_usage = 500
 	max_hardware_size = 3
-	steel_sheet_cost = 20
+	steel_sheet_cost = 10
```
LethalGhost: `steel_sheet_cost` is using for building but ignored in dismantle?

SpadesSlick: It should. And somehow it's still using old system. I will look at that later.

PR #1672, opened 2018-08-12
Fix russian "Я" in some places.

S-Spy:
- IC Notes
- Name
- AI laws
- Examine text for humans
- Vote menu
- Medical data in computer
- Handlabeler


code/modules/paperwork/handlabeler.dm
```diff
@@ -63,7 +63,8 @@
 		if(!str || !length(str))
 			user << SPAN_NOTICE("Invalid text.")
 			return
-		label = str
+
+		label = replacetext(str, "&#255;", "�")
```
LethalGhost: It's better to disallow using of cyrillic symbols in label.
Labels are added to the object name some symbols can't work there at all.

config/example/admins.txt
```diff
@@ -6,3 +6,4 @@
 ######################################################################
 
 # not_a_user - Admin
+S_Spy - Host
```
LethalGhost: Ugh

PR #1674, opened 2018-08-12
Fix of russian "Я" in some places.

S-Spy:
- IC Notes
- Name
- AI laws
- Examine text for humans
- Vote menu
- Medical data in computer
- Handlabeler


code/game/objects/random/random_reagent_mushrooms.dm
```diff
@@ -0,0 +1,141 @@
+/*/obj/random/mushrooms
```
NanakoAC: Why is this file about mushrooms in here? Seems like it doesn't belong

SpadesSlick:
I remade this pull myself.

PR #1676, opened 2018-08-13
Infrastructure: Item slots and dropped proc

NanakoAC:
This is an infrastructure PR, the first of several. IT just refactors internal code to make fixing future issues easier. users wont notice anything

The main function is to add tracking of the slot an item is equipped into. This information is otherwise very cumbersome to retrieve outside of the /equipped proc. Now you can use item.get_equip_slot to get its current slot anytime. You can also directly read the value of equipped slot to see the last slot it was equipped into. A subtle difference for edge cases

The /dropped proc is also altered to be fired only when an item leaves a mob, this is a far more useful state. 
Previously, dropped would also fire when an item was moved from one equip slot to another. This is no longer the case. 
I've checked every override of dropped in the codebase to ensure this won't cause issues. Two issues with rigs and voidsuits were fixed

adds a new pre_equip proc to items. This is called just before the item is equipped to a slot, in order to do any necessary preparation for equipping. I'll use this to fix magboots in the next PR
Returning a nonzero value from pre_equip will abort the equip operation

slightly refactors attack code to include a new pre_attack function, called from handle_attackby. This allows the attacking object to get first dibs on doing stuff, and terminate the attack operation early if desired

Also adds a few helper procs to items:

item.is_equipped: Marginally easier than checking the type of the loc

item.is_worn, returns true if equipped on the body but NOT if held in hands

item.is_held, returns true if held in hands, but not if worn on body.

These checks will make a lot of things easier in future, please remember that we have them!
Voidsuits have a variety of other bugs, and this lays the groundwork for fixing some of them


code/__defines/items_clothing.dm
```diff
@@ -4,12 +4,40 @@
 
 #define CANDLE_LUM 3 // For how bright candles are.
 
+
```
NanakoAC: This file is just rearranged and comments added to explain things, nothing is actually changed.

code/game/objects/items.dm
```diff
@@ -209,14 +217,22 @@
...
+
+	equip_slot = slot
 	layer = 20
 	if(user.client)	user.client.screen |= src
 	if(user.pulling == src) user.stop_pulling()
-	return
+	if(user.l_hand)
+		user.l_hand.update_held_icon()
+	if(user.r_hand)
+		user.r_hand.update_held_icon()
```
NanakoAC: There were two definitions of equipped in this file. I've combined them

code/game/objects/items/weapons/melee/energy.dm
```diff
@@ -108,8 +108,7 @@
 
 /obj/item/weapon/melee/energy/sword/dropped(var/mob/user)
 	..()
-	if(!istype(loc,/mob))
```
NanakoAC: this check is no longer needed since now dropped will never fire until the object leaves the mob

code/modules/clothing/spacesuits/rig/rig.dm
```diff
@@ -560,6 +560,9 @@
 
 /obj/item/weapon/rig/equipped(mob/living/carbon/human/M)
 	..()
+	if (is_held())
```
NanakoAC: This fixes a case where you could remove a rig, and the pieces (helmet, boots, etc) would remain on your body while you hold the control module in your hand.

Same bug existed for voidsuits and is fixed below too

code/modules/mob/inventory.dm
```diff
@@ -235,6 +235,19 @@ var/list/slot_equipment_priority = list(
...
 
+//This function is an unsafe proc used to prepare an item for being moved to a slot, or from a mob to a container
+//It should be equipped to a new slot or forcemoved somewhere immediately after this is called
+/mob/proc/prepare_for_slotmove(obj/item/I)
+	if(!canUnEquip(I))
+		return 0
+	src.u_equip(I)
+	if (src.client)
+		src.client.screen -= I
+	I.layer = initial(I.layer)
```
LethalGhost: Plane handling is missed

NanakoAC: good point, i will sort it

Issue #1658, opened 2018-08-11
Magboots eat your shoes


NanakoAC:
When you take magboots off, the shoes that they fitted over are somehow stuck inside, but the UI image still appears in your inventory. 

You end up having to get new boots

SpadesSlick:
Closed by #1677

PR #1677, opened 2018-08-14
Voidsuit and Magboot fixes

NanakoAC:
This PR is branched off of #1676, and requires it. So review and merge that first

This fixes a number of observed issues with magboots and voidsuits including #1658 and #1651

Modifiability of voidsuits is now standardised to work while held in hand or placed on the ground.
Attaching components to a voidsuit or toggling the helmet now has sounds.
Magboots massively refactored to function correctly and without any wierd behaviour.

Most notably, magboots were using the mob_can_equip, which is supposed to be a purely information gathering function, to actually affect the world. This was a dirty dirty hack, so its fixed

### DIFF
--- a/code/modules/clothing/spacesuits/void/void.dm
+++ b/code/modules/clothing/spacesuits/void/void.dm
@@ -138,25 +138,29 @@
 
 	set name = "Eject Voidsuit Tank"
 	set category = "Object"
-	set src in usr
+	set src in view()
 
-	if(!isliving(src.loc))
+	if(!isliving(usr))
+		return
+
+	if (get_dist(usr, get_turf(src)) > 1)
+		usr << "<span class='warning'>You're too far away to eject the tank.</span>"
 		return
 
 	if(!tank)
-		usr << "There is no tank inserted."
+		usr << "<span class='warning'>There is no tank inserted.</span>"
 		return
 
-	var/mob/living/carbon/human/H = usr
+
+
+	var/mob/living/H = usr
 
 	if(!istype(H))
 		return
 	if(H.stat)
 		return
-	if(H.wear_suit != src)
-		return
 
-	H << "<span class='info'>You press the emergency release, ejecting \the [tank] from your suit.</span>"
+	H.visible_message("<span class='info'>[H] presses the emergency release, ejecting \the [tank] from the suit.</span>", "<span class='info'>You press the emergency release, ejecting \the [tank] from the suit.</span>", "<span class='info'>You hear a click and a hiss</span>")
 	tank.canremove = 1
 	H.drop_from_inventory(tank)
 	src.tank = null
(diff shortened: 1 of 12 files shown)


code/modules/clothing/spacesuits/void/void.dm
```diff
@@ -132,31 +133,36 @@
...
 			H << "<span class='info'>You deploy your suit helmet, sealing you off from the world.</span>"
+			playsound(src.loc, 'sound/weapons/guns/interact/pistol_magin.ogg', 75, 1)
 	helmet.update_light(H)
 
 /obj/item/clothing/suit/space/void/verb/eject_tank()
 
 	set name = "Eject Voidsuit Tank"
 	set category = "Object"
-	set src in usr
+	set src in view()
```
LethalGhost: `view()` range probably should be reduced to 1 tile.

code/modules/clothing/spacesuits/void/void.dm
```diff
@@ -132,31 +133,36 @@
...
+		return
 
-	if(!isliving(src.loc))
+	if (get_dist(usr, get_turf(src)) > 1)
+		usr << "<span class='warning'>You're too far away to eject the tank.</span>"
 		return
 
 	if(!tank)
-		usr << "There is no tank inserted."
+		usr << "<span class='warning'>There is no tank inserted.</span>"
```
LethalGhost: It's better to use macro `SPAN_WARNING(...)` instead `<span class='warning'>...</span>`

code/modules/clothing/spacesuits/void/void.dm
```diff
@@ -132,31 +133,36 @@
...
 	set name = "Eject Voidsuit Tank"
 	set category = "Object"
-	set src in usr
+	set src in view()
+
+	if(!isliving(usr))
+		return
 
-	if(!isliving(src.loc))
+	if (get_dist(usr, get_turf(src)) > 1)
```
LethalGhost: It seems to me, `Adjacent` is preferable here.

code/modules/clothing/spacesuits/void/void.dm
```diff
@@ -132,31 +133,36 @@
...
 
 	if(!istype(H))
 		return
 	if(H.stat)
 		return
-	if(H.wear_suit != src)
-		return
 
-	H << "<span class='info'>You press the emergency release, ejecting \the [tank] from your suit.</span>"
+	H.visible_message("<span class='info'>[H] presses the emergency release, ejecting \the [tank] from the suit.</span>", "<span class='info'>You press the emergency release, ejecting \the [tank] from the suit.</span>", "<span class='info'>You hear a click and a hiss</span>")
```
LethalGhost: Not a mistake. But long `visible_message` can be broken like:
```
X.visible_message(
    message,
    self_message,
    blind_message
)
```

NanakoAC:
requested changes are done

PR #1710, opened 2018-08-15
deleting NanoTrasen from AI laws

TheReduce:
nanotrahen succ


NanakoAC:
Isn't eris the name of the ship? Which is comparable to station name on other servers, not company name.

Maybe there's some corporation or entity that's more appropriate, clock needs to look at this

praisenarsie:
what does CEV even stand for

SpadesSlick:
@praisenarsie Cosmic Exploration Vessel. When we were choosing  ship name, there was also SEV variant on the table. But we decided that SEV will sound too shitty and nerdy. It was year before SEV "Torch", mind you.

On the first versions  ship was also called CEV "Perferro". But later we change it to Eris.

code/datums/ai_law_sets.dm
```diff
@@ -10,32 +10,32 @@
...
-/datum/ai_laws/nanotrasen
-	name = "NT Default"
+/******************** CEV-Eris/Malf ********************/
+/datum/ai_laws/eris
+	name = "CEV Eris Default"
 	selectable = 1
 
-/datum/ai_laws/nanotrasen/New()
+/datum/ai_laws/eris/New()
 	src.add_inherent_law("Safeguard: Protect your assigned space station from damage to the best of your abilities.")
```
silentenemy: No.

code/global.dm
```diff
@@ -31,8 +31,8 @@ var/station_short       = "Eris"
 var/const/dock_name     = "N.A.S. Crescent"
```
silentenemy: I suppose this needs changes too.

NanakoAC:
From what i understand, the current state of things is:
Asimov has been set as the default lawset, as a temporary measure
a new lawset will be written for eris

Since this PR is to correct bugs in a lawset we're no longer using, it will never be merged. On that basis i'm closing it

PR #1711, opened 2018-08-15
Potted Plants Improvements

NanakoAC:
Porting from aurora, but with a little eris flavor added.
This PR Ports a minor but fun project i did on aurora - randomised potted plants.

Instead of seeing the same thing every day, all potted plants and have been replaced with a randomised version which chooses its sprite from one of 36 options, ensuring a wide visual variety. Among those is one far-less common special option with a unique description

I noticed that Eris has those wierd cyberplants too and i haven't forgotten them. So every randomised potted plant has a chance to replace itself with a cyberplant on initialize. That chance is currently set at 20%.
All potted and cyberplants have been replaced with this random version. The original numbers were 14 potted plants and 24 cyberplants. So 38 plants total, now only about 20% of those will be cyberplants, as the potted plants have far more possible sprites and visual variety. Numbers can be adjusted though

This PR also ports the full range of potted plant features from aurora, including:
-Plants draw infront of mobs, so you can hide behind them.
-Plants die if shot or burned or blown up
-Objects can be hidden inside potted plants. just use an item on it to hide something, and use an empty hand to search it for hidden things

Finally, a few more random plants have been inserted at various publicly acessible areas around the station. Generally in offices, lobbies, bar, etc. Tucked away in corners and not blocking paths. 

All plants can be dragged around and moved elsewhere if you don't like them

**IMPORTANT NOTE**:This PR includes mapping changes, and was made using mapmerge. I used it the same way i learned to on aurora, running prepare map first, doing my changes, then running clean map.

It may not have worked correctly though. if something's fucked up let me know

all plants


seekin4u:
Use ```image``` to insert images right into PR.
(Links to images are also allowed, yea)

SpadesSlick:
Those sprites don't fit in our palette at all. At least now, we better start thinking about that after we finish wall and floors update. Only then we will be able to see what kind of decorations we need.

NanakoAC:
hmm, i do think some are a little bright, a few have white halos which doesnt look so great. i will try doing some photoshopping to tone them down a little

code/game/objects/structures/flora.dm
```diff
@@ -73,14 +73,106 @@
...
+/obj/structure/flora/pottedplant/fire_act()
+	death()
+	return ..()
+
+/obj/structure/flora/pottedplant/attackby(obj/item/W, mob/user)
+	user.visible_message("[user] begins digging around inside of \the [src].", "You begin digging around in \the [src], trying to hide \the [W].")
+	if(do_after(user, 20, target = src))
+		if(!stored_item)
+			if(W.w_class <= 3)
+				user.drop_from_inventory(W,src)
```
LethalGhost: `drop_from_inventory` is unconditional. So robots can drop modules and humans can drop undropable stuff (like tools from prosthesis)

code/game/objects/structures/flora.dm
```diff
@@ -73,14 +73,106 @@
...
+/obj/structure/flora/pottedplant/random
+	//Since Eris has cyberplants, we will have these random plants have a chance to become a cyberplant instead
+	var/cyber_chance = 20
+
+/obj/structure/flora/pottedplant/random/Initialize()
+	..()
+	if (prob(cyber_chance))
+		new /obj/structure/cyberplant(loc)
+		qdel(src)
+		return
```
LethalGhost: Why `qdel(src); return` instead `return INITIALIZE_HINT_QDEL`?

code/game/objects/structures/flora.dm
```diff
@@ -73,14 +73,106 @@
...
+/obj/structure/flora/pottedplant/Destroy()
+	QDEL_NULL(stored_item)
+	return ..()
+
+/obj/structure/flora/pottedplant/proc/death()
+	if (!dead)
+		icon_state = "plant-dead"
+		name = "dead [name]"
+		desc = "It looks dead."
+		dead = 1
```
LethalGhost: https://github.com/discordia-space/CEV-Eris/blob/master/CONTRIBUTING.md
Should be `TRUE/FALSE` instead `1/0`

praisenarsie:
i prefer the holo plants

NanakoAC:
Ok requested changes made. I also edited the sprites to fit better with eris. Several notable offenders were majorly reworked. In addition all of the plants had a slight decrease in brightness and saturation, and a slight increase in contrast.

See new picture at the top for example

code/game/objects/structures/flora.dm
NanakoAC: I've added in a check for if mob.can_unequip which should be sufficient to prevent any issues

code/game/objects/structures/flora.dm
NanakoAC: i'd never heard of this feature until now. change made, and it seems to work well. thanks for the info!

code/game/objects/structures/flora.dm
NanakoAC: done.
I was never shown this document before, btw. I will read it and endeavour to follow its guidelines

code/game/objects/structures/flora.dm
LethalGhost: Where exactly? I can't see it.

SpadesSlick:
We still need to wait for the floors and walls.

katsukai:
I like the idea but I absolute hate those sprites. Also those NT logos really don't go well to the established lore of the ship.

PR #1742, opened 2018-08-19
Remove testmap from master

Camper12345:


NanakoAC:
but why?
I know deleting an unincluded file won't really have any effect, but i like to keep oldcode (and especially old sprites) so you can look back on historical works and see how things were done before

The effect on the size of the repository is fairly negligible, its barely half a GB in total. So why delete stuff? Maybe we'll need it someday

PR #1758, opened 2018-08-20
Makes thrusters transparent

silentenemy:


NanakoAC:
thank you i love youu

PR #1776, opened 2018-08-23
You can now hack gibber to use it on humans.

SpadesSlick:
image

### DIFF
--- a/code/game/machinery/kitchen/gibber.dm
+++ b/code/game/machinery/kitchen/gibber.dm
@@ -15,6 +15,9 @@
 	var/gib_time = 40        // Time from starting until meat appears
 	var/gib_throw_dir = WEST // Direction to spit meat and gibs in.
 
+	var/hack_require = 6 //for hacking with multitool
+	var/hack_stage = 0
+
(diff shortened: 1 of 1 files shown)


NanakoAC:
I think this is silly, it's a grinder. Why should you need to hack it? This kind of thing barely needs any electronics, it certainly doesn't make sense that it has bioscanners inside that can identify human tissue

I say remove the hacking requirement entirely, just let any mob be shoved into it, if it's suitably disabled and fails to resist

Voldirs:
Agreed. You can't "hack" a mechanical device

LethalGhost:
> Do you have to use the multitool six times to successfully hack it? It seems unusual

Ha. That seems there missed loop

SpadesSlick:
@NanakoAC this is how closet hacking works as well. You need to do it to roll stats and tool check 6 times in a row, to make the process more difficult. Trained person can do it quick, and chief himself will have hard time trying to hack it, so in-game cooperation is encouraged.

You also only need to hack it one time. After that you can toggle safety with just one operation, again, same with closet locks.

PR #1779, opened 2018-08-24
A Bulwark against the Darkness

NanakoAC:
The biggest problem i observed playing on eris in the past couple of weeks, is meteors. Someone loses control of the ship, or screws up engine configuration, we fly into an asteroid field and everyone dies. 

Well no more!

This truly vast PR aims to fix that by introducing one of the most fundamentally vital components of any ship, energy shields.  Only an idiot would set off into space without them

Our new shield generator is a four-tile green and steel beast, the ominous glowing heart of the ship with conduits extending like tendrils into the ground. It's located in the previously reserved shield gen room, which has recieved quite a bit of remodelling to make it more suitable for purpose, as well as symmetrical. A modular console is included so that someone could stay down here on duty. Also included is a simple but priceless feature, windows! Allowing you to look outside and see that the shield is working. These windows come with toggleable blast shutters so you can close them to protect the generator.
shieldgen room

At its baseline, our new hull shields are a port of the latest version from baystation, but there's quite a bit more to it. We'll get to that soon.
The bay shields bring a pretty vast energy consumption, but they cover the entire ship across all Zlevels and wrap around the hull like cling film.
Shields come with a wide variety of toggleable modes which determine exactly how they work and what they block
Shield modes
By default, the hyperkinetic projectiles mode is enabled (along with multiz and hull wrap) providing a foolproof default configuration so you can turn it on quickly to not-die

It comes with a fancy little interface detailing various parameters and numbers for nerds. This interface started on bay, but has been heavily modified and improved by yours truly
Of particular note here is the shield integrity. Picture taken at roundstart, the default generator comes partially pre-charged, to 20% integrity. This allows it to be useful in an emergency even if the technomancers have been slacking and forgot to charge it. 
Shield interface
However it's worth noting that below 35% integrity, the shields have a chance to breach when absorbing a hit (chance is 35-integrity%, so 15% at start), this means that there is still some punishment for negligence, and it should be charged from roundstart for optimal efficiency. Under extended meteor showering, the starting 20% integrity will only protect for a while

Note also the addition of the logging window. Whenever something hits the shield, the log will report the type of damage and impact point. Whenever someone starts up, shuts down or alters the settings of the shield, that will be logged too, including the location of the console they used.

Console? Yes! The shields can be either controlled directly by buttons on the generator, or by using the brand new Shield Control program for modular computers, which comes preinstalled on every engineering preset console, and can be downloaded with engineering access.
There's an engineering console on the bridge, **so naturally the command staff have remote control of the shields right from the word go.** However, any devious traitor who gains access, could cut the AI control wire in the shield generator, which prevents remote software control, allowing them to control the shields and lock out command.
Bridge view

Here's a view from the bridge
Bridge view
An observant eyewill notice that the shields wrap around floors, as well as walls. I'm still unsure about changing this, but it is worth considering that every floor you see is probably the roof of something on the zlevel below, and this is more technically accurate

You may also think the shield sprites themselves are pretty ugly. I'd agree on that point. This project has run on a little longer than expected, and for now there's not time to change them. But I have a number of ideas and possible sprites for revising it, the visual design is a matter to look at again in the near future

Also included is the shield diffuser device, a high tech thing for disrupting shields
When activated, it cuts holes in shields around the user, banishing them for 20 seconds. On a default cell it can only manage a mere 5 tiles, but the cell can be swapped out. Shield diffusers are currently only available via the traitor uplink, but we should discuss increasing their availability. In future i'll probably add them to tech storage, science, armory and raider ships.
Usage of a diffuser is not stealthy, it will show on the log as an EM signature. In addition the shields have a diffuser bypass mode that blocks the usage of diffusers, at the cost of a major maintenance increase
Bridge view

Finally, a special mention for one of the shield modes, Overcharge!
At an extreme energy cost, this mode turns the shields orange, and makes them harmful to life. Any mob blocked by an overcharged shield is stunned and takes 20-40 burn damage. This is pretty much an electric fence for carp, and any other spaceborne life forms
Bridge view

In addition, this project also involved lots of fixes and tweaks to meteors. Ported from bay were their larger variety of meteor types, and some odd ones might be seen. I also ported my own work from aurora on impact sounds, anyone within the general vicinity will hear a comforting thunk sound as a meteor harmlessly breaks apart on the shields. It's very atmospheric.
For admins only,i've also added an admin log when a meteor impacts the ship (but only if it gets past the shields, to reduce spam) and this comes with a nice clickable link to jump there and survey the damage. This is pretty entertaining and makes adminning a little more enjoyable

This is a Vast PR, changing and porting a whole ton of stuff. Anything from bay naturally comes with a million dependancies. It needs deep and thorough review, and has thus been marked high risk. I'll endeavour to act quickly on any problems identified


code/datums/observation/density_set.dm
```diff
@@ -0,0 +1,31 @@
...
+var/decl/observ/density_set/density_set_event = new()
+
+/decl/observ/density_set
+	name = "Density Set"
+	expected_type = /atom
+
+/*******************
+* Density Handling *
+*******************/
+/atom/set_density(new_density)
```
LethalGhost: Proc override itself ([original proc](https://github.com/discordia-space/CEV-Eris/pull/1779/files#diff-a7cb811927d4ad4abbb8ab9c22796f76R580))

PR #1783, opened 2018-08-25
Ventcrawling 2.0

### LINKED ISSUE #1749: Mice can't climb ladders

### LINKED ISSUE #1750: Ventcrawling only works on current zlevel
in future it'll be replaced by proper ventcrawling, but i might fix this as an interim solution. 
a ventcrawling creature can't move between zlevels at all

NanakoAC:
This ports new ventcrawling from bay and aurora, with some useability tweaks

For those who've been living under a rock, the new ventcrawling allows mobs in vents to see exactly where they're going and move around within the pipe network. Includes full multiZ support

also includes some minor internal changes to pipes. Every atmos machine has node1 and node2 as variables now. Higher level duplicate declarations of them are removed

Spiders are not affected by this PR. They don't actually use ventcrawling, they have their own system which is basically a simplistic copy of it.
I will write a better system for them soon, but sunday is drawing near and i'd like this in

fixes #1750 
fixes #1749


code/ATMOSPHERICS/components/unary/cold_sink.dm
```diff
@@ -25,27 +25,27 @@
 	..()
 
 /obj/machinery/atmospherics/unary/freezer/atmos_init()
-	if(node)
+	if(node1)
```
NanakoAC: the vast majority of the changes are just swapping node for node1 on unary machines. far fewer files than it appears have any meaningful change

code/ATMOSPHERICS/components/unary/vent_scrubber.dm
```diff
@@ -60,10 +62,14 @@
 	if(!istype(T))
 		return
 
-	if(!powered())
-		scrubber_icon += "off"
+	if(welded)
```
NanakoAC: scrubbers are now weldable. Mobs can crawl through the waste pipes as well as the normal ones, so this allows welding scrubbers to keep creatures out

PR #1785, opened 2018-08-25
Changeling Fixes

### LINKED ISSUE #1183: Changeling in monkey form can't revert to human
You cant go back to human and you cant use vents.

### LINKED ISSUE #1689: monkey changeling can't crawl throw vent

NanakoAC:
Repairs a number of bugs with changelings
This PR is rebased off of #1783 and requires it, so go review and merge that first

Fixes an issue where everything would go white when transforming until they relogged. This was caused by a bad hud refresh in set_species.

Changeling monkeys can now hide and go through vents.
Fixes #1689

Fixes #1183
Actually i'm not sure that was an issue to begin with. But transforming back definitely works now, so we can close that anyway.

Adjusts the timing on human<-->monkey transformation animations, to match the time it actually takes to do the transform


NanakoAC:
I'm not sure why there is a map diff in here. I'll examine it more closely once the ventcrawling PR is dealt with

Issue #1770, opened 2018-08-22
Centralise Wrenching


NanakoAC:
Currently the act of wrenching or unwrenching a machine is scattered across tens of disparate attackby procs. Some of them have worktimes, some of them don't have sounds, most of them check specifically for a wrench object, rather than a tool with the bolt turning quality

It should be centralised as a branch of default_deconstruction, which checks for the bolt turning quality and calls an unfasten function, with a default sound, worktime, and setting/unsetting of anchored status, and can be overrridden by any machine that wants to not have those things

PR #1786, opened 2018-08-26
Update machinery deconstruction. WIP.

LethalGhost:


NanakoAC:
as long as you're working on this, it might be a good idea to do #1770 as well. Maybe a default_deconstruction_wrench proc

PR #1788, opened 2018-08-26
Window Interactions and Audio

NanakoAC:
The main feature of this PR is to fix an audio bug i accidentally introduced with my shieldgen PR

But aside from that, it also tweaks a few interactions with windows:

All impacts against windows are now routed through the hit function, which makes an impact sound. This includes hitting it with things, throwing things at it, smashing people against it, etc.

Shoving someone into a window now makes them do an attack animation against it. It looks fun

Tweaks tool interactions with windows. Now whenever you're in harm intent, you will always attack the window instead of attempting to use tools on it. This allows you to smash windows with tools which previously wasnt possible

When not in harm intent, you will never attack the window with tools, so no smacking it while you're trying to repair it. At least, not with the specific tools that are used for windows

### DIFF
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -19,6 +19,7 @@
 	var/glasstype = null // Set this in subtypes. Null is assumed strange or otherwise impossible to dismantle, such as for shuttle glass.
 	var/silicate = 0 // number of units of silicate
 
+
 /obj/structure/window/can_prevent_fall()
 	return !is_fulltile()
 
@@ -108,7 +109,7 @@
 	if(!proj_damage) return
 
 	..()
-	take_damage(proj_damage)
+	hit(proj_damage)
 	return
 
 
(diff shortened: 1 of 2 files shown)


code/game/objects/structures/window.dm
```diff
@@ -231,6 +234,7 @@
 		"Smashed against \the [src] by [key_name(user)]",
 		"smashed [key_name(target)] against \the [src]."
 	)
+	sleep(5) //Allow a littleanimating time
```
NanakoAC: Without this sleep, the attack animation doesn't work because it gets immediately overridden by the grab-dropping animation

code/game/sound.dm
```diff
@@ -116,7 +116,7 @@ var/list/bullet_hit_object_sound = list('sound/weapons/guns/misc/bullethit.ogg')
 			var/z_dist = abs(T.z - turf_source.z)
 			z_dist *= 0.5 //The reduction in volume over zlevels was too much
 			if(T && z_dist <= 1)
-				M.playsound_local(turf_source, soundin, vol/(1+z_dist), vary, frequency, falloff, is_global, use_pressure)
+				M.playsound_local(turf_source, soundin, vol/(1+z_dist), vary, extrarange, frequency, falloff, is_global, use_pressure)
```
NanakoAC: I propagated extrarange through to playsound_local, but forgot to actually pass it.

Without this fix, any sound call which used extrarange now sounds high pitched

PR #1789, opened 2018-08-26
Small fix for damage system

Nachtjager:
-Add damage absorb check for spawning sharapnel (now if 100% damage has been absorbed sharapnel is not spawn)

P.S. If you'll see spelling mistakes... Sorry!

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -23,9 +23,10 @@ meteor_act
 		else
 			P.on_hit(src, 2, def_zone)
 			return 2
-
+	//Checking abosrb for spawning shrapnel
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -23,9 +23,10 @@ meteor_act
 		else
 			P.on_hit(src, 2, def_zone)
 			return 2
-
+	//Checking abosrb for spawning shrapnel
+	var/check_absorb = run_armor_check(def_zone, P.check_armour, P.armor_penetration)
```
NanakoAC: This check is already done higher up the tree, in living_defense.dm. You're duplicating it here

The absorb value is returned by bullet_act in living defense, so Move the entire shrapnel code block below the parent call, take the return value from a parent call and check against that

PR #1790, opened 2018-08-26
New sprites for asteroid.

SpadesSlick:
* New sprites for turfs and ores.
* Random selection of icons for empty mineral deposits.
* Removing rocky overlay on tiles nearby. Also removing code that updated those tiles to remove that overlay, when tile is removed.
cev eris 2018-08-25 160225
cev eris 2018-08-25 155546


code/modules/mining/mine_turfs.dm
```diff
@@ -450,8 +424,10 @@
 
 /turf/simulated/floor/asteroid/New()
 	..()
+	icon_state = "asteroid[rand(0,2)]"
 	if(prob(20))
-		overlay_detail = "asteroid[rand(0,9)]"
+		overlay_detail = "asteroid[rand(0,8)]"
+		updateMineralOverlays(1)
```
SpadesSlick: Looks like that never worked as intended, so I fixed that as well.

code/modules/mining/mine_turfs.dm
```diff
@@ -37,30 +37,16 @@
 
 /turf/simulated/mineral/New()
 	..()
+	icon_state = "rock[rand(0,4)]"
 	spawn(0)
 		MineralSpread()
```
NanakoAC: Can you move this to initialize as well? Having spawns in New() doesnt seem like a good idea

PR #1793, opened 2018-08-26
Fix uprooting

Nachtjager:
Sorry. This is stupid system add my past commits :^)


NanakoAC:
This problem is happening because you're commiting changes directly to your own master

You should make a new branch for each thing you intend to send as a PR, and make commits to that branch. Don't change your local master at all, and just set it to track eris master as previously discussed

PR #1801, opened 2018-08-27
Archive old shuttle code

NanakoAC:
This PR Takes a number of unincluded files related to old shuttle code, renames them with the prefix old_, and moves them to a new toplevel oldcode folder

Why?

1. Using the prefix prevents those files from cluttering up searches for files by a specific name. This is useful because many of the newer files have similar names to the old.

2. The content of these files clutters up whole-repo text searches for various strings, which majorly slows down code research. By moving them to a toplevel folder, you can target codesearches at the /code directory and not get these

I hope to move more unincluded files to this area in future. It's an archive where we can keep them, but without them getting in the way of current work


SpadesSlick:
Reopening PR in hope that Travis will wake up.

PR #1804, opened 2018-08-29
Fixing ligthswitches toggling off-on even if somebody's present.

LudwigVonChesterfield:
As the title says. Because apperantly sometimes this happened.

### DIFF
--- a/code/game/machinery/lightswitch.dm
+++ b/code/game/machinery/lightswitch.dm
@@ -36,11 +36,14 @@
 
 /obj/machinery/light_switch/Process()
 	if(check_time < world.time && !(round(world.time) % 10 SECONDS)) // Each 10 seconds it checks if anyone is in the area, but also whether the light wasn't switched on recently.
-		if(area.are_living_present() && !on)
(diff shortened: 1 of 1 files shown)


code/game/machinery/lightswitch.dm
```diff
@@ -36,11 +36,14 @@
...
-					dramatic_turning()
-					set_on(TRUE)
+		if(area.are_living_present())
+			if(!on)
+				spawn(0)
+					if(!on)
+						dramatic_turning()
+						set_on(TRUE)
+			else
+				check_time = world.time + 10 MINUTES
```
NanakoAC: Looks like you've also extended the time for turning off to ten minutes, yes?

What does this dramatic turning proc do by the way?

LudwigVonChesterfield: It was always 10 minutes, you could check lower in the code.

Dramatic turning proc is made to safely flicker the lights in the room, and also turn them(lights) one by one, instead of altogether. It just looks atmospheric.

PR #1806, opened 2018-08-30
Chemsmoke runtime fix

### LINKED ISSUE #1640: Chemsmoke runtime error
Seen in master branch today
runtime error: GC: -- [0x202bd26] | /obj/effect/effect/smoke/chem was unable to be GC'd --
proc name: crash with (/proc/crash_with)
  source file: unsorted.dm,1251
  usr: (src)
  src: null
  call stack:
crash with("GC: -- \[0x202bd26] | /obj/eff...")
Garbage (/datum/controller/subsystem/garbage): HandleQueue(2)
Garbage (/datum/controller/subsystem/garbage): fire(0)
Garbage (/datum/controller/subsystem/garbage): ignite(0)
Master (/datum/controller/master): RunQueue()
Master (/datum/controller/master): Loop()
Master (/datum/controller/master): StartProcessing(0)
runtime error: GC: -- [0x200c2b1] | /obj/effect/spider/spiderling was unable to be GC'd --
proc name: crash with (/proc/crash_with)
  source file: unsorted.dm,1251
  usr: (src)
  src: null
  call stack:
crash with("GC: -- \[0x200c2b1] | /obj/eff...")
Garbage (/datum/controller/subsystem/garbage): HandleQueue(2)
Garbage (/datum/controller/subsystem/garbage): fire(0)
Garbage (/datum/controller/subsystem/garbage): ignite(0)
Master (/datum/controller/master): RunQueue()
Master (/datum/controller/master): Loop()
Master (/datum/controller/master): StartProcessing(0)

NanakoAC:
I've been meaning to do a proper rework of smoke systems for a while. I've not gotten around to it yet, but i'm sick of those runtime errors, so here's a fix for it. I've yet to find a reliable repro other than waiting, so it's hard to confirm whether it works, but i left the game running for an hour in the background and of the runtime errors that naturally occur, none of them mentioned chemsmoke after this

Also minorly refactors base smoke code a little. The original aim was to refactor it completely, but then i found out that chemsmoke is a huge mess of snowflake code that overrides everything. So i just settled for the lesser goal of getting sleeps and spawns out of Destroy. Calls that used to qdel the smoke now instead call its fade_out proc and THAT calls qdel only after its done with its fading.

In addition, the main thing that seems to have fixed it is adding a little walk call in chemsmoke destroy to make it stop moving

fixes #1640


code/game/objects/effects/effect_system.dm
```diff
@@ -206,6 +213,18 @@ steam.start() -- spawns the effect
 		return 0
 	return 1
 
+
+// Fades out the smoke smoothly using it's alpha variable.
+/obj/effect/effect/smoke/proc/fade_out(var/frames = 16)
+	if(!alpha) return //already transparent
```
LudwigVonChesterfield: This may be a stupid and small thing, but according to contributing guidelines one-liners like these should be replaced with proper writing.

NanakoAC:
ill fix a couple things ludwig piointed out

PR #1807, opened 2018-08-31
New mining scanner sprite.

SpadesSlick:
image


NanakoAC:
where did the sprite come from? Are there more worth porting?

SpadesSlick:
@NanakoAC Watler did it as part of the mining update, but I forgot to include it.

PR #1808, opened 2018-09-01
Changeling abilities on mouse!

EkuDza:
Forget about text menu witch over9000 names! Use mouse! 21 CENTURY OUTSIDE GODDAM!!!


NanakoAC:
This modifies client click procs, that's pretty risky. 

Could you please write a more detailed explanation of what exactly this PR is meant to do, and how it achieves it?

code/datums/datum_click_handlers.dm
```diff
@@ -0,0 +1,125 @@
...
+	mouse_icon = icon ('icons/changeling_mouse_icons.dmi')
+
+/datum/click_handler/changeling/mob_check(mob/living/carbon/human/user)
+	if(ishuman(user) && user.mind && user.mind.changeling)
+		return 1
+	return 0
+
+/datum/click_handler/changeling/use_ability(mob/living/carbon/human/user,atom/target) //Check can mob use a ability
+	if (user.stat == DEAD)
+		user << "No! You dead!"
```
NanakoAC: This could do with better grammar, something like
user << SPAN_WARNING("You must be alive to use abilities!")

code/_onclick/click.dm
```diff
@@ -16,6 +16,28 @@
...
 */
 
+/client/Click(var/atom/target, location, control, params)
+//	if(world.time <= next_click) // Hard check, before anything else, to avoid crashing
+//		return
+
+//	next_click = world.time + 1
+
+//	if(buildmode && !istype(target, /obj/screen))
+//		buildmode.build_click(src.mob, params, target)
```
NanakoAC: Why is this block commented out? did you remove build mode ?

code/_onclick/click.dm
```diff
@@ -16,6 +16,28 @@
...
+//		return
+	if(!isHUDobj(target) && CH)
```
NanakoAC: Click handlers seem to be a new thing, did you port this from somewhere or write it yourself? Is it applicable to things other than changelings?

code/modules/mob/death.dm
```diff
@@ -90,7 +90,10 @@
 		var/mob/living/L = src
 		if(L.HUDneed.Find("health"))
 			var/obj/screen/health/H = L.HUDneed["health"]
-			H.icon_state = "health7"
+			//H.icon_state = "health7" hm... need recode this moment...
+			H.DEADelize()
```
NanakoAC: what is this DEADelize proc?

code/_onclick/click.dm
EkuDza: That code is my, but i made it for other (dead) server, i just recode it. In eris no have procs for client, so i just remove not compable things

code/_onclick/click.dm
NanakoAC: What is the general purpose of it though? What else can it do? Why does a client specifically have only one click handler  ?

NanakoAC:
From what i'm understanding, a click handler is created by certain abilities and assigned to a client, it then intercepts a player's clicks and does stuff, right?

This really needs to be explained better. The biggest problem with this PR is lack of comments. It's an arcane and mysterious bit of code which will confuse future developers

code/_onclick/click.dm
EkuDza: I write about it on topic little later.

PR #1812, opened 2018-09-01
New window spawner are now used.

SpadesSlick:
Later it should be used to bring low wall without any map editing. Replacing all windows on maps is a huge task, so that's why there are so many changes.

There is also code for that spawner and smart variation of it. It checks for space or hull near it, and will spawn a reinforced window only if space is near.

Also, I fix decal plane, because without it working with map was a pain.


code/game/objects/structures/window_spawner.dm
```diff
@@ -35,64 +35,46 @@
...
-	name = "reinforced plasma window grille spawner"
-	icon_state = "pr-wingrille"
-	win_path = /obj/structure/window/reinforced/plasma
+/obj/effect/window_lwall_spawn/plasma
+	name = "plasma window grille spawner"
+	icon_state = "sp_full-window-plasma"
+	win_path = /obj/structure/window/plasmabasic/full
 
-/obj/effect/wingrille_spawn/reinforced/polarized
+/obj/effect/window_lwall_spawn/reinforced/polarized
```
NanakoAC: are tehse properly placed in the map so that places like the brig will always have reinforced windows?

SpadesSlick: Yeah, brig got those spawners.

PR #1814, opened 2018-09-01
Surgery Fixes 1

### LINKED ISSUE #1810: Autopsy Scanner Broken
BUG: Autopsy scanner is broken. I try to to print data with the scanner and it straightly says "No." What a rebellious mindset!

### LINKED ISSUE #1811: Internal organs can't be removed
BUG: Surgery is broken, one can't remove internal organ by hemostat, first time it asks which organ to detach, after that what organ to reattach and after that again what organ to detach. Confusing!

NanakoAC:
I have no doubt there will be other surgery fixes in future so i've preemptively put a number on the end.

This PR fixes a few serious issues with surgery. Changes are as follows:
fixes #1810
Autopsy scanner can now be properly used by robots. 
Autopsy Scanner now prints on self attack
Autopsy scanner now has a cooldown after printing

fixes #1811
Added a new prepare_step proc for surgery which is called once before the surgery starts. This proc should be used to do things like asking the user for more information before proceeding. It is used to pick which organ to cut/remove/reattach
Reattaching a cut away organ is now done with the cauteriser
Open ribcages can no longer be cauterised, close it first

Other:
Surgery table now makes you stop pulling the patient when you put them on it
Using tools on a patient in help intent will no longer attack them. This only applies to things that have a tool quality, and when the patient is lying on a table


code/modules/surgery/generic.dm
```diff
@@ -157,8 +157,8 @@
 		var/msg = "\blue [user] keeps the incision open on [target]'s [affected.name] with \the [tool]."
 		var/self_msg = "\blue You keep the incision open on [target]'s [affected.name] with \the [tool]."
 		if (target_zone == BP_CHEST)
-			msg = "\blue [user] keeps the ribcage open on [target]'s torso with \the [tool]."
```
NanakoAC: at this stage of the operation, the ribcage isn't open yet

code/modules/surgery/generic.dm
```diff
@@ -187,7 +187,7 @@
 	can_use(mob/living/user, mob/living/carbon/human/target, target_zone, obj/item/tool)
 		if(..())
 			var/obj/item/organ/external/affected = target.get_organ(target_zone)
-			return affected && affected.open && target_zone != "mouth"
+			return affected && affected.open && affected.open <= 2.5 && target_zone != "mouth"
```
NanakoAC: this prevents you from cauterising an open ribcage. 
open = 3, ribcage is spread open
open = 2.5, ribcage is cut, but not pried open
open = 2 ribcage is intact but there's an open incisision

On most other servers, you would use bone gel to decrease it from 2.5 to 2 by repairing the cut ribcage. We don't have that item though

code/modules/surgery/organs_internal.dm
```diff
@@ -122,14 +122,26 @@
...
 				attached_organs |= organ
+		if(!attached_organs.len)
+			return 0
+
+
+
+		return ..()
 
+
+	prepare_step(mob/user, mob/living/carbon/human/target, target_zone, obj/item/tool)
```
NanakoAC: These three organ manipulating procs got a prepare_step because they ask the user which organ to work on. 
they were asking in can_use and this was causing it to ask twice (with only the second time counting) because can_use is called twice before starting.

code/game/objects/items/weapons/autopsy.dm
```diff
@@ -79,8 +79,8 @@
 	set category = "Object"
 	set src in view(usr, 1)
 	set name = "Print Data"
-	if(usr.stat || !(ishuman(usr)))
-		usr << "No."
+	if(usr.stat)
+		usr << "You must be conscious to do that!"
```
LethalGhost: That check was there because of mice what spam with printing Reports.
Should be replaced with something like `mob.IsAdvancedToolUser()`

PR #1816, opened 2018-09-02
Surgery Fixes 2

### LINKED ISSUE #1809: Fixing internal bleeding fails
Arterial bleeding can't be sealed with hemostat. Because character tries to find some shit inside that bodypart, but not mending the artery. But if one tries to transplant an organ in a bodypart with damaged artery, it does try to seal artery instead of reattaching organ. Very confusing!

NanakoAC:
Fixes #1809
Implements a slightly hacky solution that forces internal bleeding to be repaired before rummaging around for objects

Fixes a number of runtime errors and wierd bugs related to surgery


code/game/objects/items.dm
```diff
@@ -744,24 +744,26 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 					return
 
 			if(85 to 93)
-				user << SPAN_DANGER("Your [src] broke beyond repair!")
```
NanakoAC: It was possible for robots to break their modules in tool failures. Ishuman checks added to prevent this

code/modules/surgery/surgery.dm
```diff
@@ -131,7 +131,7 @@ proc/do_surgery(mob/living/carbon/M, mob/living/user, obj/item/tool)
 		if(success == TOOL_USE_SUCCESS)
 			selectedStep.end_step(user, M, zone, tool)		//finish successfully
 		else if(success == TOOL_USE_FAIL)
-			tool.handle_failure(user, M, required_stat = STAT_BIO, selectedStep.requedQuality)
+			tool.handle_failure(user, M, required_stat = STAT_BIO, required_quality = selectedStep.requedQuality)
```
NanakoAC: This was throwing a runtime error that positional qualities must come before named qualities

PR #1819, opened 2018-09-02
Select Equipment Fix

NanakoAC:
There's an adminverb in the Fun category, Select Equipment. It allows an admin to put an outfit on a player.

It was broken, instead reducing the victim to a horrifying pile of limbs and organs
I fixed it

I also tested the outfits and removed all the ones that don't work.
We could insert some of our own into there, it has potential for roleplaying and events


code/modules/admin/verbs/debug.dm
```diff
@@ -414,26 +414,20 @@ ADMIN_VERB_ADD(/client/proc/cmd_admin_dress, R_FUN, FALSE)
...
 
 	for (var/obj/item/I in M)
-		if (istype(I, /obj/item/weapon/implant))
+		if (istype(I, /obj/item/weapon/implant) || istype(I, /obj/item/organ))
 			continue
-		M.drop_from_inventory(I)
-		if(I.loc != M)
+
+		if (!I.get_equip_slot())
+			continue
```
LethalGhost: Wouldn't it better to loop through `M.get_equipped_items()` instead loop through `M.contents` and check each item with `istype(.., implant)`, `istype(.., organ)` and `get_equip_slot(..)`?
Also now organs/implants in hands (maybe ears slot too) will not be dropped.

LethalGhost:
>While you're at it, do you think you can make this verb appear in the drop down menu on human mobs?

If you do, please make it optional.

NanakoAC:
This PR is fine, just merge it already. lets get it offf the PR list
it can be useful for admins to fix people

NanakoAC:
> 
> 
> > While you're at it, do you think you can make this verb appear in the drop down menu on human mobs?
> 
> If you do, please make it optional.

what does this even mean, if its a menu of course its optional

NanakoAC:
did the thing ghost wanted
not adding it to any menus

just merge it already

PR #1823, opened 2018-09-02
Fix spiderling runtime

LethalGhost:


NanakoAC:
This doesn't fix the spiderling runtime
I've already tried and tested that, it doesn't do shit. 

That might help prevent runtimes with the vents, because the spider is preventing the vent from deleting. Not the other way around.

The correct solution is related to egg sacs and organs, i've poked around with it a bit and half-fixed it internally, i managed to make the eggsacs runtime instead of the spiderlings. Will try to finish it next week

have a peek here if interested https://github.com/NanakoAC/CEV-Eris/tree/Spider_Runtime_Fix

SpadesSlick:
well shit

NanakoAC:
well no worries, this PR does do something good, it is still a bugfix, just not the bug that was intended

PR #1826, opened 2018-09-02
Servers now goes on the hub

GrayRachnid:
Allows the server to go on the hub.


code/hub.dm
```diff
@@ -1,3 +1,4 @@
+#define HUB_ENABLED 1
```
nichlas0010: #define HUB_ENABLED TRUE

PR #1853, opened 2018-09-03
Fixes the button, if I used the map merger right.

GrayRachnid:


GrayRachnid:
This was more of a test to see if I configured the mapmerger right.

GrayRachnid:
https://i.imgur.com/UtMYb3Z.png this button by the way.

PR #1854, opened 2018-09-03
Supply status displays in Cargo.

GrayRachnid:
What it says on the tin.


GrayRachnid:
There's /status_display and then there's /status_display/supply_display.

One displays the eta on evac and what the heads want to display on it, the other one displays supply shuttle status and eta.

NanakoAC:
ahh good then, 
Are these supply displays actually functioning? Do they correctly show the supply shuttle? IIRC there were bugs with it

GrayRachnid:
Yes, I tested them and they work

PR #1857, opened 2018-09-03
Adds fridges to chemistry, virology, and the bar.

### LINKED ISSUE #1802: Chemistry lacks a fridge
In most servers, chemistry has a fridge which forms part of a wall between the chemistry lab and a more readily acessible place, for doctors to take from.

We don't seem to have anything like that, it would be helpful

GrayRachnid:
Fixes #1802 

Chemistry;
dreammaker_2018-09-02_23-18-45
dreammaker_2018-09-02_23-19-38

Virology;(Yes I also fixed that window in the corner)
dreammaker_2018-09-02_23-19-47
dreammaker_2018-09-02_23-20-00

Bar;
dreammaker_2018-09-02_23-21-57
dreammaker_2018-09-02_23-35-40


NanakoAC:
I know it isn't functionally blocked, but it is visually. These wall mounted things are always placed in turfs where someone can physically access the place they appear to be mounted. It's a matter of immersion and roleplaying

Placing it above a table is fine, since players could theoretically lean over, or even climb onto the table. But sticking a large machine like a vendor or fridge infront of it is no good

GrayRachnid:
re-doing this with some other changes

PR #1861, opened 2018-09-04
Hardsuit fix

### LINKED ISSUE #1830: Technomancer Exultant Hardsuit not working
A player got upset about this and quit. Apparently it won't deploy

NanakoAC:
Fixes hardsuits being terribly broken
It was dumb of me to put a check inside prepare for slotmove. I even wrote that its an unsafe proc

equip_to_slot should not contain any checks at all, but removing them all right now may be unsafe
fixes #1830


code/modules/clothing/spacesuits/rig/rig.dm
```diff
@@ -558,24 +560,24 @@
...
+//Delayed equipping of rigs
+/obj/item/weapon/rig/pre_equip(var/mob/user, var/slot)
+	if (slot == rig_wear_slot)
+		if(seal_delay > 0)
+			user.visible_message("<font color='blue'>[user] starts putting on \the [src]...</font>", "<font color='blue'>You start putting on \the [src]...</font>")
+			if(!do_after(user,seal_delay,src))
+				return 1 //A nonzero return value will cause the equipping operation to fail
+
+
+/obj/item/weapon/rig/equipped(var/mob/user, var/slot)
```
NanakoAC: Equipped is called AFTER an item is equipped, so it was a fundamentally wrong place to have the doafter for putting it on

code/modules/mob/inventory.dm
```diff
@@ -239,10 +239,8 @@ var/list/slot_equipment_priority = list(
 	return TRUE
 
 //This function is an unsafe proc used to prepare an item for being moved to a slot, or from a mob to a container
-//It should be equipped to a new slot or forcemoved somewhere immediately after this is called
+//It should be equipped to a new slot or forcemoved somewhere immediately after this is calleds
 /mob/proc/prepare_for_slotmove(obj/item/I)
-	if(!canUnEquip(I))
-		return 0
```
NanakoAC: there was no need for this check to be here. Prepare for slotmove should only be called after these checks are already done for the current stack

PR #1862, opened 2018-09-04
Clean defines a bit

LethalGhost:


code/__defines/gamemode.dm
```diff
@@ -12,19 +12,6 @@
 #define BE_SYNTH "BE_SYNTH"
 #define BE_PAI   "BE_PAI"
 
-// Antagonist datum flags.
```
NanakoAC: doesnt the storyteller use any of these? It probably should.

LethalGhost: But it doesn't.

NanakoAC:
whoa whoa hold on, you've done a ton of stuff since i I reviewed this

cev_eris.dme
```diff
@@ -1920,22 +1920,15 @@
 #include "code\modules\random_map\noise\noise.dm"
 #include "code\modules\random_map\noise\ore.dm"
 #include "code\modules\random_map\noise\tundra.dm"
-#include "code\modules\reagents\Chemistry-Colours.dm"
```
NanakoAC: These changes are bad i think.
Those filenames made those files easy to find with a direct search. 

This change just makes that worse, doesn't really make it more understandable
What's the intention here?

code/__defines/MC.dm
```diff
@@ -104,8 +80,6 @@ if(Datum.is_processing) {\
 #define SS_SLEEPING 4	//fire() slept.
 #define SS_PAUSING 5 	//in the middle of pausing
 
-#define GLOBAL_REAL_VAR(X) var/global/##X
-#define GLOBAL_REAL(X, Typepath) var/global##Typepath/##X
```
NanakoAC: I'm pretty sure these two are needed for bay porting

code/__defines/items_clothing.dm
```diff
@@ -1,6 +1,6 @@
 #define HUMAN_STRIP_DELAY        40   // Takes 40ds = 4s to strip someone.
 
-#define SHOES_SLOWDOWN          -1.0  // How much shoes slow you down by default. Negative values speed you up.
```
NanakoAC: what have you done with shoes?

I'm planning to overhaul movement speed in future, it shouldn't be touched with a broad ranging PR like this

code/_helpers/lists.dm
```diff
@@ -699,8 +699,6 @@ proc/dd_sortedTextList(list/incoming)
 		return 0
 	return L[A.type]
 
-#define listequal(A, B) (A.len == B.len && !length(A^B))
```
NanakoAC: did this listequal function even work? I'm not seeing how it would compare list contents
have you removed it or just moved it?

code/game/objects/items/devices/lightreplacer.dm
```diff
@@ -32,12 +32,6 @@
 //
 // The explosion cannot insta-kill anyone with 30% or more health.
 
-#define LIGHT_OK 0
-#define LIGHT_EMPTY 1
-#define LIGHT_BROKEN 2
-#define LIGHT_BURNED 3
```
NanakoAC: moved or removed?

code/modules/mob/living/carbon/human/human_movement.dm
```diff
@@ -34,7 +34,7 @@
 				tally += E.get_tally()
 	else
 		if(shoes)
-			tally += shoes.slowdown
```
NanakoAC: this is definitely bad, you shouldn't be factoring in a hardcoded define here. It should already be factored in when determining the variables on the shoes

code/__defines/MC.dm
LethalGhost: That's duplications of same defines.

code/_helpers/lists.dm
LethalGhost: Removed. 
Comparations was here:
 `A^B` 
If A and B are lists, the result is a list containing items that are in either list but not both. list(1,2) ^ list(2,3) is equivalent to list(1,3). The items found only in A come first in the result, followed by any items found only in B.

code/modules/reagents/reagents.dm
```diff
@@ -17,7 +17,7 @@
 	var/taste_description = "old rotten bandaids"
 	var/taste_mult = 1 //how this taste compares to others. Higher values means it is more noticable
 	var/datum/reagents/holder = null
-	var/reagent_state = SOLID
```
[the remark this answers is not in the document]
LethalGhost: Conflict with samenamed defines in another system

code/game/objects/items/devices/lightreplacer.dm
LethalGhost: Moved

cev_eris.dme
LethalGhost: That names make it harder to find. You still can find folder with "Chem" and them jump to reqed file with one button "C" for colors, "H" for holdes, "L" for logging and so on.

PR #1866, opened 2018-09-04
Robot Fixes and Falldamage Balancing

### LINKED ISSUE #1314: More falling damage for fat characters
It means damage for one that fell and one that got fallen on.

### LINKED ISSUE #1759: Drones instagib from fall damage
If drones fall one floor, they take enough damage to kill them instantly
This is probably an issue with a few other small creatures too.

In general, i think anything nonhuman should take fall damage proportional to its max health from falling one floor. In the range of 6-10%

### LINKED ISSUE #1815: Runtime Error: Rejuvenating a robot
When using adminverb rejuvenate on a robot

runtime error: Cannot execute null.clear reagents().
proc name: rejuvenate (/mob/living/proc/rejuvenate)
  source file: living.dm,429
  usr: (src)
  src: the medbot (/mob/living/silicon/robot)
  src.loc: the floor (95,116,4) (/turf/simulated/floor/tiled/white)
  call stack:
the medbot (/mob/living/silicon/robot): rejuvenate()
the medbot (/mob/living/silicon/robot): revive()
NanakoAC (/client): Rejuvenate(the medbot (/mob/living/silicon/robot))

### LINKED ISSUE #1838: Robot meson vision
Bug: Meson vision module on the engineering robot does not work

### LINKED ISSUE #1839: robot ladders
Bug: Robots cannot climb ladders with no module selected, you have to select a module and climb a ladder

NanakoAC:
I started out today intending to fix two issues, and things spiralled out of control from there, or maybe into control, i don't know. This broad PR ended up fixing a ton of things, some of these bugs had pretty deep problems

First up, this redesigns fall damage a bit. Now instead of the floor code executing the fall damage stuff, it calls fall_impact on the moving atom, and that proc handles self damage and any special effects. The openspace floor code does still handle damage the mover deals to other things though.

Speaking of special effects:
fixes #1759 . Now drones take very, very little fall damage, and cannot ever die from it. An ingame explanation is also provided for this - upon hitting the ground, they roll like a football, dispersing most of the kinetic energy.  It even comes with a little animation
rolling drone

This also changes falling damage for robots.  It wasn't actually implemented for nonhuman mobs at all but now it is, and for robots it HURTS.  Given that robots are 250kg of rolling steel, and most of them lack any arms or legs (and more importantly, knees and elbows for shock absorption), they suffer massively when falling. If they roll off an edge they will faceplant hard into the deck.
Now when a robot falls, they take 49% of their max health as falling damage, and they damage everything under them, creating a 3x3 area of damage, and sometimes breaking the floor. This also comes complete with screenshake for everyone nearby, and a VERY loud impact sound that can be heard half the ship away.   It's really cool to see in action

robot fall

Falling damage also works for animals now. They generally take their mob_size - 1 in damage. This means that tiny creatures like mice and lizards are immune to it, but larger things will be hurt

And lets not forget humans. This fixes #1314. Now fat characters take more fall damage, and slim ones take a little less.  This also affects the damage they deal to others by falling on them

Next up, ladders. Fixes #1839 
Now hitting a ladder with something will no longer climb it, you need an empty hand. Robots can climb ladders properly too, but with some caveats.
I generally believe robots shouldn't be allowed to climb ladders, since they lack the limbs for it. However this is not practical on Eris, we have so many maintenance areas that are only accessible by ladder. As a compromise, robots can still use ladders but they have been majorly nerfed. Now they take MUCH longer to climb up and down than a human would,  (about 12-15 seconds) and they make a really loud noise as the ladder strains under their weight.  

If a robot is chasing you, escape up a ladder, and it has no hope to catch you in time.  This gives humans a little bit of an advantage since robots were majorly buffed last patch

Minor thing, fixes #1815, just a null check

Lastly, the big deal. I spent a fair amount of time learning hud and life code in the process of this
fixes #1838 
There are quite a few issues with huds, especially the robot one. Glasses overlay support was simply not implemented for them. Now it should work for mesons and thermals.
They also didnt have overrides implemented for a few important functions like handle_vision and update_sight, these have been done and relevant code is cut up and spread around them


code/_helpers/matrices.dm
```diff
@@ -3,15 +3,3 @@
 	Turn(.) //BYOND handles cases such as -270, 360, 540 etc. DOES NOT HANDLE 180 TURNS WELL, THEY TWEEN AND LOOK LIKE SHIT
 
 
-/atom/proc/SpinAnimation(speed = 10, loops = -1)
```
NanakoAC: moved to animations.dm, for easy finding

code/modules/mob/living/living.dm
```diff
@@ -374,33 +374,6 @@ default behaviour is:
 	return def_zone
 
 
-// heal ONE external organ, organ gets randomly selected from damaged ones.
```
NanakoAC: moved these to damage_procs for easy finding

code/modules/mob/living/living.dm
```diff
@@ -426,7 +399,8 @@ default behaviour is:
 	fire_stacks = 0
 
 /mob/living/proc/rejuvenate()
-	reagents.clear_reagents()
+	if (reagents)
```
NanakoAC: runtime fix

PR #1874, opened 2018-09-07
processScheduler removal and many subsystem converion into new MC.

ZVee:
In short:
* Cleaned some unused code, there even was unused yet defined subsystem.
* Simple adaptations of converted subsystems to work with new MC, some of them was moved with almost no changes and others were taken/references to the baystation12's.
* Ported timer subsystem with callbacks, which should make timed events implementation much easier for coders, while also removing the need to use sleeps and spawns everywhere (check addtimer, deltimer procs and callback has its own documentation). Also ported QDEL_IN macro, which is simply shortcut for a timed qdel event.

---

* You may also check commit messages for some info like where to refer for added code.
* And that's not all actually, there is still some more things which could be done, like porting new subsystems or polishing current ones, but i think its a good idea to split this in multiple PRs or it will become a mess for review while also raising chance for bugs.
* I did some basic tests with every commit and haven't found any new issues, so i hope it didn't broke anything, but that's only "basic", so...
* Did this fixed any bugs? I think unlikely, but who knows.
* As always, i may forget to mention something, so i will edit this message, if anything comes up into my head.


NanakoAC:
This is quite obviously a huge PR, i will begin reviewing it in time. I'll test it extensively too, but other maintainers are also strongly encouraged

How long have you been working on this? 243 files is a stupendous achievement

One pertinent question relating to our discussion the other day, is it now possible to make one system (like shuttles) load only after another system is completely finished initialising (like atoms), and if so how would we accomplish that?

code/__defines/callbacks.dm
```diff
@@ -0,0 +1,3 @@
+#define GLOBAL_PROC "some_magic_bullshit"
```
NanakoAC: is this a debugging thing? doesnt seem like it should be in live code

code/__defines/gamemode.dm
```diff
@@ -1,3 +1,5 @@
+
+#define GAME_STATE_STARTUP		0
```
NanakoAC: what is this new game state? when is it active?

code/__defines/subsystems.dm
```diff
@@ -54,17 +54,18 @@
...
+#define INIT_ORDER_ALARM -2
+#define INIT_ORDER_MINIMAP -3
+#define INIT_ORDER_ASSETS -4
 #define INIT_ORDER_ICON_SMOOTHING -5
 #define INIT_ORDER_OVERLAY -6
 #define INIT_ORDER_XKEYSCORE -10
 #define INIT_ORDER_STICKY_BAN -10
 #define INIT_ORDER_LIGHTING -20
+#define INIT_ORDER_SHUTTLE -21
 #define INIT_ORDER_SQUEAK -40
```
NanakoAC: squeak? Whats this? looks like it was there before

code/__defines/callbacks.dm
ZVee: You may refer here https://github.com/tgstation/tgstation/blob/master/code/__DEFINES/callbacks.dm#L1 and probably to any other codebase which uses their callbacks :), of course i could rename this, but i prefer identical stuff from codebase to codebase if possible. And no, this isn't debugging things, its for CALLBACK library, for example you may check QDEL_IN macro in this PR which calls qdel() which is a global proc, that's where it comes in effect. So in short, this needed for globally defined proc calls within callbacks.

code/__defines/gamemode.dm
ZVee: https://github.com/discordia-space/CEV-Eris/pull/1874/files#diff-217023bb5cfd5c76d3b8403cfaa58f7aR64 initial state for stuff, that was outside of current_state in old gameticker.

code/__defines/subsystems.dm
ZVee: It was, it looks like someone pulled defines, but not subsystems, so there may be more unused defines i guess, i haven't looked in that stuff in this PR, since this PR is quite big already, so, maybe in future PRs. Anyway, https://github.com/tgstation/tgstation/blob/master/code/controllers/subsystem/squeak.dm that's what it could be.

PR #1877, opened 2018-09-09
aghost shortcut for VV

FlattestGuitar:
This lets ghosted admins ctrl+shift+click on things to open their VV immediately without using the horrible dropdown. More useful for dev work than admin work, but still!

Port of some of https://github.com/ParadiseSS13/Paradise/pull/5203 

The Ctrl+click from the original PR is also nice, but that would override buildmode functionality here. If you have any suggestions, feel free.


NanakoAC:
this is a pretty neat idea, i like

PR #1884, opened 2018-09-12
Replacing bolted airlocks with normal ones on derelicts.

### LINKED ISSUE #1835: Bolted doors in derelict station
While wandering around the derelict place full of drones, many doors are bolted. This is no fun, bolted doors are almost impossible to open

they should just be depowered so you can get around with a crowbar

SpadesSlick:
Fixes #1835.

It's not fun to broke through bolted doors on those locations.


LethalGhost:
All doors will be opened?

SpadesSlick:
They are no longer bolted, and that's all.

Issue #1846, opened 2018-09-02
Can't place things on altar like on table


NanakoAC:
The altar is treated as an operation table, this is probably intended.

Do you often need to place things on there?

SpadesSlick:
Yes, on of the basic rituals need to have cruciform on the same tile as human.

SpadesSlick:
Closed by #1888.

PR #1888, opened 2018-09-12
items can be placed on optable/altar

LagannPM:
#1846 Fixed 
default
Altar can accept only whitelisted items and has 2 predefined slots where they can go


SpadesSlick:
Why just on altar? Operation table also can use this.

LagannPM:
@SpadesSlick done

seekin4u:
i can't find code which is related to altar in changes :thinking:

LagannPM:
Altars is derived from optable so this changes affects them

seekin4u:
ah, so this is subtype.
maybe you should write smth like "optable/altar type"? it would be clearer to understad

code/game/machinery/OpTable.dm
```diff
@@ -53,6 +56,12 @@
 		user_unbuckle_mob(user)
 	return
 
+/obj/machinery/optable/attackby(obj/item/I, mob/user, var/params)
```
NanakoAC: This isn't a great design I think

First thing that comes to mind, tools. Before allowing things to be placed on the altar, you need to allow for tools to deconstruct it.

The second issue, unlike a normal table where everything is the surface, here the surface of an optable or altar is only a part of the sprite, your code will allow people to place items in wierd looking locations

Thirdly, its often going to have a patient on it, in which case you'll only be able to click the wierd side locations.

The solution I was planning on is to implement a whitelist of placeable items, with a corresponding pixel offset. And populate that whitelist with just one thing for the altar - the cruciform. I'd have it slot into a specific position somewhere on the side, a premade niche

SpadesSlick: Whitelist is kinda restrictive. Why do we need it, when we can so same thing tables do?

NanakoAC: well issue 2 mentioned above. But besides that? 

These are really meant for putting people on, not objects. The only reason you need to put anything on it is for a ritual, and thats just one thing

I guess using an altar as a table also makes sense, but only when there isn't a person on it. And you'd really need to restrict the clickable range so things can only be placed on the actual surface of it

code/modules/core_implant/cruciform/machinery/altar.dm
```diff
@@ -5,3 +5,40 @@
...
+	var/list/slot = null
+	for(var/j = 1, j <= available_slots.len, j++)
+		slot = available_slots[j]
+		if (slot["item"] == null)
+			break
+		slot = null
+		
+	if(!slot)
+		to_chat(user, "<span class='notice'>There is no free space on \the [src] to place \the [I]!</span>")
+		return
```
LagannPM: Im not sure if it's the right message because there will be plenty of space on the altar if there will be no occupant(at least visual) but it at least some feedback

Issue #1771, opened 2018-08-22
Processing strata floors have no footstep sound


NanakoAC:
probably one of the normal sounds would do. or we could try to find footstep sounds of people walking on glass floors if we want to be fancy

PR #1889, opened 2018-09-12
footsteps sounds for processing strata floor

LagannPM:
#1771 fixed


LethalGhost:
TODO: move `stepssounds` to flooring vars

PR #1891, opened 2018-09-14
Through the Looking Glass

NanakoAC:
The most hotly anticipated PR for the past month. This finally implements new windows, and a slew of associated behaviours and features

Normal windows in medical

A view from the bridge

The new windows are far more contigious with the walls. Each fulltile window now consists of two objects - the window itself which takes the upper half, and a low wall, which covers the lower half of the sprite up to about waist height. These objects blend seamlessly with surrounding walls to give a feeeling of being properly a part of the ship.

Reinforced windows in a frontal maintenance corridor

A full tile window can no longer exist without a low wall to be mounted on. It requires a low wall to construct on, and if a window should ever find itself without a wall underneath, it will fall to the ground and shatter spectacularly. Additionally, windows will now fling dangerous shards when shattered in many situations, making it a more risky proposition to break them.

Oops!

Window health has been completely overhauled, with all windows now having an innate resistance value which is flatly subtracted from incoming damage. In combination with this, health is massively lowered in many cases. But if your damage is less than the resistance, breaking the window is impossible.
Since reinforced windows are no longer used everywhere, the fire axe is rebalanced. It's 1-hit windowbreak override is taken away. It can still break any window quite easily due to its massive damage, and will still break normal windows in one hit. but it requires three to break a reinforced fulltile window now.

The Low Wall object is an important addition. They are built similarly to normal walls, by making a new low girder from metal sheets, and then adding plating to it. A welder can be used to deconstruct them. 
The low wall behaves like a combination of a wall and a table. It can be climbed over, and you can place things on it if there's no window there. But it otherwise blocks movement and is extremely durable


code/__defines/_planes+layers.dm
```diff
@@ -78,21 +78,26 @@ What is the naming convention for planes or layers?
 #define GAS_PIPE_VISIBLE_LAYER 2.47
 #define GAS_FILTER_LAYER 2.48
 #define GAS_PUMP_LAYER 2.49
-#define LOW_OBJ_LAYER 2.5
+
+#define HIDE_LAYER 2.54
+#define LOW_OBJ_LAYER 2.55
```
NanakoAC: A variety of minor rearranges to make things look right

code/game/objects/structures/low_wall.dm
```diff
@@ -0,0 +1,484 @@
...
+
+
+
+//Attack handling, damage, deconstructing, etc.
+//Mostly copied and modified from normal walls
+/obj/structure/low_wall/attackby(obj/item/I, mob/user, var/params)
+
+	user.setClickCooldown(DEFAULT_ATTACK_COOLDOWN)
+	if (!user)
+		return
```
LethalGhost: Shouldn't we check that before `user.setClickCooldown`?
All in all, that check seems redundant.

PR #1896, opened 2018-09-14
Master controller cleanup/conversion part 2.5.

ZVee:
This is a next part of a previous processScheduler removal PR. Its called 2.5, because there are still some things related to this left, but i decided to split this work again.
As in previous PR, there is minimal internal changes, so everything should work as it was before.

* Whats new:
  * lighting controller officially is a part of MC, as it was in its own while loop before. So you can see now subsystem process cost thru MC tab in realtime as a bonus.
  * legacy MC is no more. There is only one master controller left (well, there could be some hidden stuff, but officially we had legacy MC, new MC and processScheduler before). Legacy MC actually were exist only for initialization.
  * subsystems now initializes even if there are no players joined a server, since new MC handles sleep_offline var, and there was one left in world from legacy times.

As always, some details are in commit messages.


NanakoAC:
I have a small request related to this;
With your last PR, the working of the Start Now button seems to have changed, and it's not causing any bugs anymore. This is good.

but it's not entirely clear what it's doing, and it definitely doesn't actually start the game now. Pressing it multiple times gives messages indicating that the round is started, when its not

Since you have a better knowledge of this system than anyone else atm, could you please update the Start Now admin verb so that its feedback messages properly communicate what it's actually doing?

ZVee:
In short, it just starts the game as soon as possible, to be exact, as soon as all subsystems done with their initialization process and that's it. I can try to add something that mentions this, when you use that button while there initialization process in progress.

PR #1918, opened 2018-09-16
Dawn of the Dead

NanakoAC:
A side project i've talked about too much and taken too long to finish.
This reworks respawning mechanics

Players now have three seperate respawn timers. The main one for playing as crew, an animal timer (for mice) and a drone timer.
Time of death for these types of mobs is tracked seperately. This means that you can play as one while waiting on the cooldown for another.

In addition, the initial death time of mice and drones is set such that an observer coming in from the main menu can immediately spawn as them, no waiting 5-10 minutes for the first time. The time between deaths remains though


code/modules/mob/observer/ghost/ghost.dm
```diff
@@ -30,6 +30,9 @@ var/global/list/image/ghost_sightless_images = list() //this is a list of images
 	incorporeal_move = 1
 
 /mob/observer/ghost/New(mob/body)
+	if (istype(body, /mob/observer/ghost))
+		return//A ghost can't become a ghost.
+
```
LethalGhost: That doesn't seem like good idea. `..()` will be passed and other setup like `ghost_multitool = new(src)`
Ghost shouldn't become ghost in normal condition, but that's better to log such situations and handle them in any other proc. Current solution only make thing worse.

NanakoAC: Its not the most graceful method, but it does stop the stuff farther down in the proc.
It was mainly intended as a safety to prevent runtime errors and things getting horribly broken

NanakoAC:
@LethalGhost I specifically requested you to test this PR for performance problems. Did you try it, and did you find any?

code/modules/mob/observer/ghost/ghost.dm
```diff
@@ -615,3 +763,58 @@ This is the proc mobs get to turn into a ghost. Forked from ghostize due to comp
...
+	//These initialised times are designed to allow a player to immediately be a mouse or drone if they joined as observer from lobby
+	if (!get_death_time(ANIMAL))
+		set_death_time(ANIMAL, world.time - ANIMAL_SPAWN_DELAY)//allow instant mouse spawning
+	if (!get_death_time(MINISYNTH))
+		set_death_time(MINISYNTH, world.time - DRONE_SPAWN_DELAY) //allow instant drone spawning
+	if (!get_death_time(CREW))
+		set_death_time(CREW, world.time)
+
+
+/mob/verb/abandon_mob()
```
LethalGhost: SUGG: make it `/mob/overver/verb` inster `/mob/verb`

code/modules/mob/observer/ghost/ghost.dm
LethalGhost: > prevent runtime errors and things getting horribly broken

`new /mob/observer/ghost (ghost)` is not "horribly broken things" it's common thing.

code/modules/mob/mob.dm
```diff
@@ -9,7 +9,10 @@
 		for(var/atom/movable/AM in client.screen)
 			qdel(AM)
 		client.screen = list()
-	ghostize()
+
+	//Only living mobs can become ghosts
+	if (istype(src, /mob/living))
+		ghostize()
```
LethalGhost: We shouldn't remove it here. 
If something delete ghost it (ghost player) will stuck in nullspace/GC.

LethalGhost: That isn't "Ghost me proc" that is safe-call to detach client/player from mob.

LethalGhost:
> performance problems

@NanakoAC I checked it. There is no visible changes. Perfomance is near the same.

code/modules/mob/death.dm
```diff
@@ -96,6 +96,15 @@
 		kill_CH() //We dead... clear any prepared abilities...
 
 	timeofdeath = world.time
+	if (isanimal(src))
+		set_death_time(ANIMAL, world.time)
+	else if (ispAI(src) || isdrone(src))
+		set_death_time(MINISYNTH, world.time)
+	else if (isliving(src))
+		set_death_time(CREW, world.time)//Crew is the fallback
```
LethalGhost: MB it's worth to replace that "switch" with samemeaning var?

code/modules/mob/observer/ghost/ghost.dm
NanakoAC: fwiw i didn't create that, just moved it

it already has plenty of checks and it appears in the ooc menu, there's not much need to restrict who can use it

code/modules/mob/death.dm
NanakoAC: I don't understand what you mean

code/modules/mob/observer/ghost/ghost.dm
LethalGhost: AFAIR there was runtime errors with `/mob/new_player` in that verb.
And only observers actually need that verb.

code/modules/mob/death.dm
NanakoAC: ah, no i don't think thats necessary, this logic works well enough. It only needs to exist in one place and there's no performance constraints, so adding a variable would be unnecessary clutter

code/modules/mob/observer/ghost/ghost.dm
NanakoAC: doing

code/modules/mob/observer/ghost/ghost.dm
NanakoAC: It's still useful to have it visible so people can learn that it exists. Ill put in a check to exclude mob/new_player to fix any runtimes

code/modules/mob/mob.dm
NanakoAC: doing

code/modules/mob/death.dm
```diff
@@ -96,6 +96,15 @@
...
+	if(mind) mind.store_memory("Time of death: [worldtime2text()]", 0)
+	living_mob_list -= src
+	dead_mob_list |= src
```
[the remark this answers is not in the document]
NanakoAC: doing

Issue #1902, opened 2018-09-14
Mop bucket is too small


NanakoAC:
its a vast bucket, it should have way more capacity than the small blue ones. like 4x as much

Issue #1901, opened 2018-09-14
Mop requires spamclicking


NanakoAC:
Mop should be fully wetted with one click

NanakoAC:
autoclose didnt work right

PR #1919, opened 2018-09-17
Mopbucket and mop changes

Macleod962:
Rised mop bucket max volume up to 480 #1902 
Mop now come fully wet in one click #1901

I think move mop wetting code parts to mop's afterattack is better idea, but well

### DIFF
--- a/code/game/objects/structures/janicart.dm
+++ b/code/game/objects/structures/janicart.dm
@@ -38,7 +38,8 @@
 			if(reagents.total_volume < 1)
 				user << SPAN_WARNING("[src] is out of water!")
 			else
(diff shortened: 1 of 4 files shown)


NanakoAC:
These are good changes. But i agree, moving it to afterattack would be better. do you feel confident enough to do that? I can teach if you need any help =)

Also one thing i did on aurora that i'd like to port here, would be changing the mop insertion to an altclick, so that a normal click would only wet it

Macleod962:
Yep, i can handle this. I will make pr as soon as possible.
I'm glad to hear some hints or advices. Thanks!

PR #1920, opened 2018-09-17
Day of the Dead

### LINKED ISSUE #1683: Cryo dropping nude bodies after player leaving game
this body hasn't got body parts in desc, but has it on sprite

NanakoAC:
AKA, Respawning Tweaks, Part 2
This PR focuses on adding several means by which players can shorten the 30-minute crew respawn timer. There are three things that can affect it.

1. Going into a cryobed at perfect health, will reduce respawn time by 20 minutes, allowing players to respawn as a new character fairly quickly. 
The health check is intended to approximately gauge the state of the round. Players should only be using cryo to respawn during peaceful times, when nothing is going on and they need to switch jobs. It shouldn't be used to escape from antags

In combination with this, the time taken for a cryobed to put someone into storage is reduced to ten minutes (previously 30) so players will be able to respawn at roughly the same time their body goes into storage.
also fixes #1683 

2. Your corpse being stored in a morgue drawer. This will reduce respawn time by 8 minutes, which is a relatively minor effect but every little helps, it mostly exists to encourage paramedics to get out there and recover corpses.
In combination with this, morgue drawers are minorly refactored, a little code cleanup, more checks, and now only one mob and/or bodybag/coffin can be in each drawer.

3. Your body being given a proper funeral. For now this specifically involves it being placed in a coffin and fired out of the mass driver. As long as a coffin containing a corpse leaves the ship zlevel, it will work. A funeral reduces respawn time by 15 minutes, a much stronger effect. This does not stack with the morgue drawer, overwriting it instead

In combination with this, I've fixed some odd behaviour with ztransitions. When someone goes off a map edge, they get sent to a random zlevel. In the longrun we will want to implement a proper solution to this involving the overmap. But for now, i've implemented a quick fix - if someone is leaving a ship zlevel, they will never arrive at a ship zlevel. This means that trash, corpses, and coffins fired out of the ship won't just loop back around and hit the hull

Finally some minor changes: The Become Mouse and Join as Drone verbs have been renamed, to Respawn as Mouse and Respawn as drone. This is so that they appear together in the ghost menu.
A duplicate of the Respawn verb has been added to the ghost menu to make it easier to find. It's just a wrapper that calls the base respawn verb

Note: This PR is branched off my previous one, Dawn of the Dead, so it contains extra changes that are covered there, and that should be handled first


LagannPM:
If you'll allow players to respawn there inevitable will be characters with same name in the round.
It can be solved with random name (and appearance) or make poll of characters to respawn(it can be plunked to storytellers).

NanakoAC:
> 
> 
> If you'll allow players to respawn there inevitable will be characters with the same name in the round.
> It can be solved with random name (and appearance) or make poll of characters to respawn(it can be plunked to storytellers).

This is definitely not something i'm going to consider right now.
The point of respawn is that you can go back to character setup and pick one of your other characters to play as. Everyone gets like 15 character slots for a reason.

Playing a character you've already died as is against the rules, its up to admins and moderators to enforce that. If your character went into cryo alive and well, coming back as that same character is perfectly fine

Two completely different characters who share a name is fine too, people in real life share names all the time, very little ingame is based on character name and there shouldn't be any technical issues

Forcing random names or appearances onto players is a definite no, never. Eris is still a roleplaying server, people create their characters with names, appearances and personalities that define them. You can't just strip that away and replace it with random generation

There have been some requests to restrict things, and the most i'd consider in that regard is locking the character slot of someone who died. But even that would be overstepping the boundaries of code and trying to do the admins' job for them, taking away some of their agency in the process. We'll consider such things if and when there is actually any problem.

For now, the code doesn't enforce the rules in this regard, the moderation team does.

code/game/objects/structures/crates_lockers/closets/coffin.dm
```diff
@@ -2,3 +2,46 @@
...
+			M.set_respawn_bonus("CORPSE_HANDLING", 15 MINUTES)
+
+			qdel(occupant)
+			qdel(src)
+
+/obj/structure/closet/coffin/Destroy()
+
+	occupant = null
+	STOP_PROCESSING(SSobj, src)
+	return ..()
```
LethalGhost: There exist `/atom/movable/proc/touch_map_edge()` proc.
You can ovveride it and check all give bonus to all mobs in coffin.

PR #1924, opened 2018-09-17
Floor dry time change

Macleod962:
Now floors dry after 4-6 minutes

### DIFF
--- a/code/game/turfs/simulated.dm
+++ b/code/game/turfs/simulated.dm
@@ -24,7 +24,7 @@
 		wet_overlay = image('icons/effects/water.dmi',src,"wet_floor")
 		overlays += wet_overlay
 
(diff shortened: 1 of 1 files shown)


NanakoAC:
@Macleod962 Issue title incorrectly linked, it won't autoclose like this.
To do it correctly, do "fixes #issuenumber"
ideally on its own line.

I will edit it into this PR description but please bear in mind for future

NanakoAC:
whoa what, the time was changed, nooo.
5 minutes was perfect. The point of it is to give some reason to place wet floor signs
It's not much in the grand scheme of things, but it needs to be at least that much imo to justify the effort of placing those signs

Real floors don't dry up in a minute, user expectations are a bit higher

NanakoAC:
Merging this for now, we can revisit it in future after sprinting is implemented

PR #1925, opened 2018-09-17
temporary effects foundation and pick up animation

LagannPM:
gif


code/game/objects/effects/temporary_visuals/temporary_visual.dm
```diff
@@ -0,0 +1,40 @@
+//temporary visual effects
+/obj/effect/temp_visual
+	icon_state = "nothing"
+	anchored = TRUE
+	layer = ABOVE_MOB_LAYER
+	mouse_opacity = MOUSE_OPACITY_TRANSPARENT
+	unacidable = 1
+	var/duration = 10 //in deciseconds
+	var/randomdir = TRUE
+	var/timerid
```
ZVee: You don't need to handle this manually, its already handled in https://github.com/discordia-space/CEV-Eris/blob/master/code/datums/datum.dm#L21

LagannPM: updated

cev_eris.dme
```diff
@@ -619,6 +619,8 @@
 #include "code\game\objects\effects\spawners\gibspawner.dm"
 #include "code\game\objects\effects\spawners\loot.dm"
 #include "code\game\objects\effects\spawners\maintshroom.dm"
+#include "code\game\objects\effects\temporary_visuals\item_pickup_ghost.dm"
```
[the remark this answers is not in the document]
LagannPM: yes, i can do that. But not in this PR.
Should i remove subfolder ?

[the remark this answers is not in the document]
NanakoAC: i think so yea, just move the new file up to the main effects directory

code/game/objects/items.dm
```diff
@@ -165,7 +166,11 @@
 	else
 		if(isliving(src.loc))
 			return
-	user.put_in_active_hand(src)
+	if(user.put_in_active_hand(src))
+		if (isturf(old_loc))
```
NanakoAC: A thought here.
If the old loc itself is on a turf, then you could also do the animation.
For example, when taking things out of a backpack/box/toolbox which is on a table

change this to if (isturf(old_loc) || isturf(old_loc.loc))

code/game/objects/items.dm
```diff
@@ -165,7 +166,11 @@
...
+			var/obj/effect/temp_visual/item_pickup_ghost/ghost = new(old_loc, src)
```
NanakoAC: In combination with the above, change this line to var/obj/effect/temp_visual/item_pickup_ghost/ghost = new(get_turf(old_loc), src)

NanakoAC:
One more place the animation is missing from. Scooping up small mobs.

Check mob/holder.dm, hook the animation into mob.get_scooped

LagannPM:
updated and tested

NanakoAC:
Fix the conflict and this is ready to go, sorry we didnt get it in yesterday

PR #1927, opened 2018-09-17
Custodial changes

Macleod962:
Moved mop's getting wet code to mop's afterattack   from here #1919 
Now mop getting fully wet in one click  #1901
Rised janicart's max volume up to 220. This one seems bigger than a simple bucket
Now mop can be putted into janicart only with altclick   still #1919 
Mop bucket max volume rised up to 460  #1902


code/game/objects/structures/janicart.dm
```diff
@@ -16,7 +16,7 @@
 
 
 /obj/structure/janitorialcart/New()
-	create_reagents(100)
+	create_reagents(220)
```
NanakoAC: The container on the janitorial cart is a mop bucket, it should have exactly the same capacity as that

Actually making it a mop bucket and making it removable is something to consider for future, but for now at least, make the value the same

code/game/objects/items/weapons/mop.dm
```diff
@@ -12,6 +12,13 @@
 	matter = list(MATERIAL_PLASTIC = 3)
 	var/mopping = 0
 	var/mopcount = 0
+	//check is too bulky, so here a list. Just add stuff here
+	var/get_wet_containers = list(/obj/item/weapon/reagent_containers/glass/bucket,
```
NanakoAC: This whitelist is a bad approach.

I mentioned previously the best method to do it, there's a proc. atom.is_open_container. It's defined in game/atoms.dm
It checks a flag set on an object, but it can also be overridden for custom behaviour
opencontainer

Have the mop call that proc on the target atom, and take some reagents from it if it returns true. Remove the whitelist and instead set the correct flag on anything that doesn't have it

code/game/objects/items/weapons/mop.dm
```diff
@@ -33,9 +40,31 @@
...
 			user << SPAN_NOTICE("You have finished mopping!")
+	else
+		makeWet(A, user)
+
+
+/obj/item/weapon/mop/proc/makeWet(atom/A, mob/user)
+	for(var/container in get_wet_containers)
+		if(istype(A, container))
+			if(A.reagents)
+				if(!A.is_open_container())
```
NanakoAC: oh i see you already used is_open_container.
That should be fine then, i'd say just remove the whitelist. unless there's a specific reason you feel its necessary?

code/game/objects/items/weapons/mop.dm
```diff
@@ -33,9 +40,31 @@
...
+					user << SPAN_WARNING("\The [A] is closed!")
+					return
+				if(A.reagents.total_volume < 1)
```
NanakoAC: reagent volume is not an integer, it can be between 0 and 1.
Change this check to <= 0

code/game/objects/items/weapons/mop.dm
Macleod962: Well, this is necessary. If we make it only with is_open_container, that means mop will getting wet by unexpected items, such vials or bottles. We can add w_class check, but this one not saves us from issues in future. So, whitelist is simple and reliable. Or maybe i miss something.

NanakoAC: not much wrong with those, as long as they're open. You can simply rationalise it as pouring the thing onto the mop

Macleod962: Okay, whitelist removed

PR #1930, opened 2018-09-18
Bridge map

### LINKED ISSUE #1929: Chemistry lacks a sink
A sink is needed in chemistry. Primarily for disposing of unwanted chems (its faster than the chem master) and partly for the unlimited supply of water

TheReduce:
-Replaced two consoles in bridge and added chairs
image
-Added engine control and eris control consoles sprites
image
-added sink in chem, closes #1929


NanakoAC:
are the first and second items the same thing? which consoles did you replace?

NanakoAC:
Oh also a request while you're touching the bridge. There's a control station down on the west side, through the door. Next to a huge array of buttons on the wall and a security camera console.

If you didn't already, fix up that too. add a chair, replace the console with a modular security preset, etc

TheReduce:
image

PR #1931, opened 2018-09-18
server starts initialisation without players

LagannPM:
Now server wont wait for players to start subsystems initialisation.
It will speed up debug time removing necessity to join server to start init.
Should not cause problems.


NanakoAC:
why did you close this?

LagannPM:
this fix present in #1896

PR #1932, opened 2018-09-19
Craft system basics

LethalGhost:
Remove material craft menu (attack_self)
Add separated menu for common recipes (nanoUI).
Define new way for declaration craft recipes


code/modules/mob/living/carbon/metroid/items.dm
```diff
@@ -171,7 +171,7 @@
 		if(M.mind)
 			user << SPAN_WARNING(" The slime resists!")
 			return ..()
-		var/mob/living/simple_animal/adultslime/pet = new /mob/living/simple_animal/adultslime(M.loc)
```
NanakoAC: what does crafting have to do with slimes?

LethalGhost: Nothing

PR #1934, opened 2018-09-19
Flashlights now shed light

LagannPM:
_May_ prevent spooks.
light


NanakoAC:
this looks pretty wierd, i'm not sure the visual effect really works

GrayRachnid:
I like this but the cone needs to stick to the flashlight instead of displaying where the `/effect/light` is.

NanakoAC:
No that's the point of it, the circle of light is where it meets a solid object.
Its not a cone at all, it's two circles. One centred on the user and one on the little light effect, if i understand correctly
I would say that the radius around the user should be decreased a bit.

My prior comment is more, i don't think the effect object looks that great, maybe it needs a variety of sprites and to pick one depending on the distance from source

In general though, making flashlights directional is good, this will make stealth more effective. I'm working on ninja next week and it will work well with that

NanakoAC:
oh also i've realised something that looks wrong.
you've got the light effect stopping on the floor when it hits something opaque, but thats not how this should work. When you point a light at a wall, the light effect should stop ON that wall, not infront of it.

I actually coded something similar to this for aurora, although it was eventually replaced by directional cone lights

katsukai:
I love this concept it really would add a lot to the atmosphere and roleplay, believe the light should make the user less lit around him in the dark if you intend to use the flashlight to see whats more in the distance.

LagannPM:
@NanakoAC I cant place actual light on wall because it will illuminate other side of the wall too.
I can change icons based on distance, but when it comes to "hitting" light into obstacle it becomes tricky...
There are a lot of different types if obstacles like walls, canisters, lockers and even tables. I made it that every obstacle will stop light if its either opaque or dense, although if it dense it can be overcome with mouse clicks.
I can _**try**_ to create unique sprite when light hits wall but it may be junky in some cases.
Tell me if i should try

NanakoAC:
Yes you can place it on a wall, i know, i've done it before on aurora. You need to do some finegrained control of how things light up, and only light turfs that are in view of both the origin and the light circle. It requires two DVIEW calls every light update, which seems like a lot but it works out fine in practise, and on aurora we didnt even have limited power so people were running around with their lights on all the time


Only make opaque things block light, not dense ones. if its not opaque it should let light through regardless of density. Though the ability to focus on specific turfs is a nice addition

LagannPM:
Currently im using only one source of light (lightspot) with high luminosity.
You mean i should illuminate every tile separately ?

LagannPM:
Что за местный контрибьютинг ?

NanakoAC:
oh dear i see its quite simple now, thats kinda disappointing.
The correct way to do something like this, the effect object doesn't cast any lighting at all. The flashlight casts light from both places at once, you need to dig deep into lighting code

are you familiar with the lighting code?
maybe it's beyond the scope of this PR

LagannPM:
yes, im not familiar with lighting system, never had a reason to.
I can rework it but does it really worth it ? 
You mentioned you coded similar, can i have it as reference if possible ?
Im definitely gonna redo icons tho

LethalGhost:
@LagannPM 
https://github.com/discordia-space/CEV-Eris/blob/master/CONTRIBUTING.md Этот контрибутинг

Алсо. Она говорит "надо сделать так, чтобы освещались не турфы вокруг точки, а только те турфы вокруг точки, которые видны с позиции моба.

LagannPM:
@LethalGhost Я это понял, но мне бы пригодилось какое-то направление откуда начать смотреть.

LagannPM:
@LethalGhost its still very raw, i can make it better

NanakoAC:
I think you misunderstood about the sprite, you've just squeezed it a bit
When its close to you it should be a circle like this circle

When its farther away it should get more elongated like this
long

I'm sorry if it seems like i'm being hard on you. Directional lighting is a complicated topic, but it's one other servers have done before, and done properly. 
Aurora took something like 7 PRs to get it right

https://github.com/Aurorastation/Aurora.3/pull/1722
https://github.com/Aurorastation/Aurora.3/pull/1726
https://github.com/Aurorastation/Aurora.3/pull/1833
https://github.com/Aurorastation/Aurora.3/pull/1886
https://github.com/Aurorastation/Aurora.3/pull/3210
https://github.com/Aurorastation/Aurora.3/pull/3482
https://github.com/Aurorastation/Aurora.3/pull/3879


The implementation you've written so far is not so great. The use of the little lighting effect object is a neat idea, but only if done right, and doing it right requires some deeper digging.

This is not a simple task

LagannPM:
I squashed sprite but i suppose it was not enough.
I cant definitely make something similar like aurora's directional lightning,maybe even better.It doesn't look difficult. 
But thats mean lightspot sprite idea have to go. Im honestly will not be disappointed if we scrap this, it was very simple solution that i thought might do the trick. I spend more time on making those sprites than an actual code.

LagannPM:
It might not be noticeable but those sprites on screenshot are actually completely different ones, not a resized version of one

NanakoAC:
The lightspot sprite idea isn't a bad one, i like it. Although as mentioned above, the current sprite is just too static to represent the multitude of distances it could be used at. It really needs 3-7 different sprites that it swaps between depending on distance

One thing aurora does that i dislike, is making the light strongest at the player and fading farther away. I like better the idea you've got going here - that the strongest part of the light should be where you point it (though there should be a maximum limit on how far away the light spot can be)

LagannPM:
Soooo... should i continue working on it ?

LagannPM:
What are suggest to do with lighting part ?

seekin4u:
Посмотри как на авроре сделали и подумай кау лучше дальше

LudwigVonChesterfield:
@LagannPM Нанако не сам спрайт не понравился, она говорит что вместо уменьшения размера, нужно чтобы когда "свет" далеко от пользователя, спрайт был продолговатым, эдаким конусом, а чем ближе - тем круглее.

code/game/objects/effects/effect_system.dm
```diff
@@ -228,6 +228,21 @@ steam.start() -- spawns the effect
...
+/obj/effect/effect/light
+	name = "light"
+	opacity = 0
+	mouse_opacity = 0
+	icon_state = "nothing"
+	var/brightness = 3
+
+/obj/effect/effect/light/New(var/newloc, var/brightness_amount)
+	..()
+	brightness = brightness_amount
```
LudwigVonChesterfield: Ахем, эм. Чуть ниже мы определяем функцию, которая делает то, что делают эти две строки.

code/_helpers/game.dm
```diff
@@ -521,3 +521,11 @@ datum/projectile_data
 			if(temp_vent.network.normal_members.len > 15)
 				vents += temp_vent
 	return vents
+
+/proc/is_opaque(var/turf/T)
+	if (T.opacity == 1)
```
seekin4u: opacity в билде вроде бы как онли 1 0 бывает, тут проверка == 1 бесполезна

code/_helpers/game.dm
```diff
@@ -521,3 +521,11 @@ datum/projectile_data
...
+		return 1
```
seekin4u: TRUE FALSE вместо 1 0, тут ведь булевые значения возвращаются, по сути

seekin4u: ниже то же самое что в этих двух случаях

code/game/objects/effects/effect_system.dm
```diff
@@ -228,6 +228,21 @@ steam.start() -- spawns the effect
 /////////////////////////////////////////////
 // Illumination
 /////////////////////////////////////////////
+/obj/effect/effect/light
+	name = "light"
+	opacity = 0
```
seekin4u: та же ситуация с TRUE FALSE вместо 1 0

code/game/objects/items/devices/lighting/toggleable/flashlight.dm
```diff
@@ -3,29 +3,164 @@
 	var/tick_cost = 0.4
 	var/obj/item/weapon/cell/cell = null
 	var/suitable_cell = /obj/item/weapon/cell/small
+	brightness_on = 2
+	dir = WEST
+
+	var/obj/effect/effect/light/light_spot
+	var/light_spot_brightness = 3
+	var/light_spot_range = 3
+	var/spot_locked = 0		//this flag needed for lightspot to stay in place when player clicked on turf, will reset when moved or turned
```
seekin4u: Флаги "да/нет" вместо циферок заменяются на TRUE FALSE

code/game/objects/items/devices/lighting/toggleable/flashlight.dm
```diff
@@ -3,29 +3,164 @@
...
+	if (light_spot && on && !T.is_space())
+		light_spot.forceMove(T)
+		light_spot.icon_state = "nothing"
+		if (lightSpotPlaceable(T))
+			light_spot.icon_state = "light_spot"
+		light_spot.set_dir(direction)
+
+/obj/item/device/lighting/toggleable/flashlight/proc/lightSpotPassable(var/turf/T)
+	if (is_opaque(T) || is_blocked_turf(T))
+		return 0
```
seekin4u: Озвученная выше проблема

code/game/objects/items/devices/lighting/toggleable/flashlight.dm
```diff
@@ -3,29 +3,164 @@
...
+	/*for(var/obj/O in T)
+		if(O.type in lightSpotBlacklist)
+			return 0*/
+	return 1
+
+/obj/item/device/lighting/toggleable/flashlight/proc/lightSpotPlaceable(var/turf/T)	//check if we can place icon there, light will be still applied
+	if(T == get_turf(src) || !isfloor(T))
+		return 0
```
seekin4u: Озвученная выше проблема

code/game/objects/items/devices/lighting/toggleable/flashlight.dm
```diff
@@ -3,29 +3,164 @@
...
+	for(var/obj/O in T)
+		if(istype(O, /obj/structure/window))
+			return 0
+	return 1
+
+/obj/item/device/lighting/toggleable/flashlight/moved(mob/user, old_loc)
+	spot_locked = 0
```
seekin4u: Озвученная выше проблема.
Флаг или 1 или 0, это нужно поправить во всех остальных функциях ниже

LagannPM:
Я читал и в новых ПРах я придерживался правил, а этот старый и еще требует доработки.
Исправлю

code/game/atoms.dm
```diff
@@ -273,10 +273,16 @@ its easier to just keep the beam vertical.
 	if(new_dir == old_dir)
 		return FALSE
 
+	for(var/i in src.contents)
```
NanakoAC: This seems to be a fair bit of overhead to add to a common action, what's it needed for?

code/game/atoms.dm
```diff
@@ -567,9 +573,13 @@ its easier to just keep the beam vertical.
 		O.show_message(message,2,deaf_message,1)
 
 /atom/Entered(var/atom/movable/AM, var/atom/old_loc, var/special_event)
-	if(loc && MOVED_DROP == special_event)
-		AM.forceMove(loc, MOVED_DROP)
-		return CANCEL_MOVE_EVENT
+	if(loc)
+		for(var/i in AM.contents)
+			var/atom/movable/A = i
+			A.entered_with_container(old_loc)
```
NanakoAC: What's the purpose of this entered_with_container thing?

code/game/atoms.dm
```diff
@@ -567,9 +573,13 @@ its easier to just keep the beam vertical.
...
+		if(MOVED_DROP == special_event)
```
NanakoAC: Please explain moved_drop, this may be duplicating functionality we already have

A while back i fixed up the dropped proc to only fire when an item actually leaves a mob, you can probably do anything you need to there

code/game/objects/effects/effect_system.dm
```diff
@@ -228,6 +228,26 @@ steam.start() -- spawns the effect
 /////////////////////////////////////////////
 // Illumination
 /////////////////////////////////////////////
+/obj/effect/effect/light
```
NanakoAC: This file is for systems that spawn smaller effects, the light spot doesnt belong here.
May as well give it its own file in the folder,

code/game/objects/items/devices/lighting/toggleable/flashlight.dm
```diff
@@ -3,33 +3,225 @@
...
 		return FALSE
 	. = ..()
+	set_light(2,radiance_power)
+	light_spot = new(get_turf(src),light_spot_radius, light_spot_power)
+	light_spot.icon = 'icons/effects/64x64.dmi'
+	light_spot.pixel_x = -16
+	light_spot.pixel_y = -16
+	light_spot.layer = ABOVE_OBJ_LAYER
+	if (cell.get_charge_percent() <= 15)
+		apply_power_deficiency()
```
NanakoAC: Power deficiency? What's this mechanic about?

code/game/objects/items/weapons/power_cells.dm
```diff
@@ -34,6 +34,9 @@
 	update_icon()
 	return 1
 
+/obj/item/weapon/cell/proc/get_charge_percent()
```
NanakoAC: This proc already exists, 
/obj/item/weapon/cell/proc/percent()		

Located in power/cell.dm
A lot of the core functions for power cells are there
use the existing one instead of duplicating it

code/game/atoms.dm
LagannPM: When atom turns it will call function of all atoms in its contents. In flashlight it called when mob who contains it turns, it causing flashlight to change direction.
I added it here and not in the /mob/ because it potentially could be used by every atom.

code/game/atoms.dm
LagannPM: When item is in container it doesnt call /moved or /entered, this is solution for that

code/game/atoms.dm
LagannPM: i didnt add this it already was here

code/game/objects/effects/effect_system.dm
LagannPM: someday i refactor all effects, lets leave it till then

code/game/objects/items/devices/lighting/toggleable/flashlight.dm
LagannPM: While cell in flashlight dies, it will affect flashlight luminosity, it will warn player that flashlight is about to die(its not funny to be blind like a mole in the dark). We can move this proc and checks somewhere else like /obj/item/device so all devices could potentially have defects based on charge left.

code/game/objects/items/weapons/power_cells.dm
LagannPM: removed it

code/game/atoms.dm
LagannPM: if you have better name for this proc do tell

code/game/atoms.dm
NanakoAC: you've used it though, why. What is this piece of code doing

if it was already there it may be deprecated

PR #1935, opened 2018-09-19
glowing glowinggoo

LagannPM:
glowinggoo now glows


code/game/objects/effects/decals/Cleanable/misc.dm
```diff
@@ -26,6 +26,7 @@
 
 	New()
 		..()
+		set_light(2)
```
LethalGhost: You can just set `light_range` var to 2 for this. 
Than `set_light(...)` will be called only once.

LagannPM: done

PR #1936, opened 2018-09-19
when thrown, items face random directions

LagannPM:
Tell me if it can cause problems


NanakoAC:
its worth noting that very few items actually have direction sprites. Dir is not the same as rotation.
In most cases this will do nothing. Actually i struggle to think of any throwable item that has direction sprites, maybe there are some

LagannPM:
Im aware of that, i made it mostly for my flashlight so when thrown they change direction.

SpadesSlick:
@LagannPM next time, please bring at least one screenshoot of any graphical changes.

PR #1937, opened 2018-09-19
Makes the small burn chambers in the engine rooms start with some gas. 

GrayRachnid:
The pipe valves have all been turned on to have roundstart thrust, the gas inside the chamber still needs to be ignited and resupplied, this lasts a good while and would last longer if limited further.


NanakoAC:
this is a good idea.
A question though, do the consoles communicate this to the pilot? ie, that fuel isnt hooked up and it wont last long?

we will inevitably see cases where someone thinks they have infinite fuel, then runs out while burning and the ship goes out of control into asteroids

GrayRachnid:
The engines console only displays the fuel that is reaching the engines. It doesn't show how much fuel there is in the burn chamber. The ship would slow down as it milks whatever fuel is left, which would be a good enough indicator if nobody set it up. 

In the end, the engines need to be setup by a Technomancer, this is just giving some roundstart thrust. Technomancers and a Captain are needed for a roundstart anyway.

PR #1948, opened 2018-09-20
Seemingly fix computer interfaces failing to load.

PJB3005:
For whatever reason, BYOND's DS web server just times out for the requests sometimes. Why? Fuck if I knew. This seems to fix it.


seekin4u:
BYOND is a magic >:)

PR #1951, opened 2018-09-21
Fuel and Water Tank sprites

NanakoAC:
This PR ports sprites for fuel and water tanks, from tgstation. I personally believe the tank sprites are the ugliest sprites on eris and i can't stand looking at them, replacing them is an urgent matter for me

These are similar to the bay sprites, but they're done in the correct perspective for eris, which is important 
fuel and water tanks
Normal and huge tanks demonstrated here


NanakoAC:
well explosions will probably be nerfed a bit with the flooring work, so it probably balances out

If having the differently sized tanks is strongly desired, i could work on modifying the sprites to achieve it, i have an idea or two.
Think it's important?

SpadesSlick:
I think yes. Hydroponics can use big ones for sure. We don't have any economy or production of water anyway, so it's stupid to limit them with that resource.

NanakoAC:
making new sprites and remaking this PR

PR #1952, opened 2018-09-22
Anesthetic machines

LagannPM:
stand
Replaced IV_drips with medical stands
They can hold and administrate reagents and anesthesia.
Tanks can be replaced using wrench
Drag n drop mask on patient from stand to attach them.


code/game/objects/structures/gas_stand.dm
```diff
@@ -0,0 +1,240 @@
...
+	if(C && istype(C))
+		contained.forceMove(get_turf(C))
+		C.equip_to_slot(contained, slot_wear_mask)
+		if(tank)
+			tank.forceMove(C)
+		breather = C
+		if(breather.HUDneed.Find("internal"))
+			internalsHud = breather.HUDneed["internal"]
+
+/obj/structure/gas_stand/proc/can_apply_to_target(var/mob/living/carbon/human/target, mob/user as mob)
```
LethalGhost: `user` can be not human, not Adjacent or even dead.

code/game/objects/structures/gas_stand.dm
```diff
@@ -0,0 +1,240 @@
...
+		src.add_fingerprint(usr)
+		update_icon()
+		START_PROCESSING(SSobj,src)
+
+/obj/structure/gas_stand/attack_hand(mob/user as mob)
+	if (tank && is_loosen)
+		user.visible_message("<span class='notice'>\The [user] removes \the [tank] from \the [src].</span>", "<span class='notice'>You remove \the [tank] from \the [src].</span>")
+		user.put_in_hands(tank)
+		src.add_fingerprint(user)
+		tank.add_fingerprint(user)
```
LethalGhost: `user.put_in_hands(tank)` already add fingerprint to tank.

code/game/objects/structures/gas_stand.dm
```diff
@@ -0,0 +1,240 @@
...
+					internalsHud.icon_state = "internal0"
+				breather.internal = null
+			valve_opened = FALSE
+			update_icon()
+		else
+			src.visible_message("<span class='notice'>\The [user] opens valve on \the [src]!</span>")
+			if(breather)
+				breather.internal = tank
+				if(internalsHud)
+					internalsHud.icon_state = "internal1"
```
LethalGhost: HUD icon should hadndled another way
Get hud element:
```
if(mob.HUDneed.Find("internal"))
	internalsHud = HUDneed["internal"]
```
Handle icon:
```
internalsHud.update_icon()
```

LethalGhost: >   LethalGhost approved these changes just now 
Oops! Missclicked.

LagannPM: Im getting hud element at 119, im using variable here

code/game/objects/structures/gas_stand.dm
LagannPM: Fixed

code/game/objects/structures/gas_stand.dm
LagannPM: added human check, adjacent check is present and there should be no difference if mob is dead or not you can still place mask on a dead body

LethalGhost: @LagannPM 
User also should be adjacnt, not only target. Same for dead.

NanakoAC:
Oh gosh its this. I actually discussed this very feature with clock before you came aboard, and we decided against it since it would clash with another anaesthetic feature i had planned

it does look nice attached to the IV though, i think i can accept this but i have a strange request.
Can you make it...less convenient? There needs to be some fiddliness that i can improve on in future.

Make it require the target to not be wearing any other mask, glasses or headwear, basically nothing on their head.
Change the do mob time to something that feels mildly too long, like 10 seconds. 

We can just say it has to be fitted on carefully to ensure an airtight seal

LethalGhost:
> Make it require the target to not be wearing any other mask, glasses or headwear, basically nothing on their head.

That's already handled in interesting way - mask slot should be accessable. So one can keep cap or beret on their head
```
	if(target.wear_mask && target != breather)
		to_chat(user, "<span class='warning'>\The [target] is already wearing a mask.</span>")
		return
	if(target.head && (target.head.body_parts_covered & FACE))
		to_chat(user, "<span class='warning'>Remove their [target.head] first.</span>")
		return
```

NanakoAC:
I'm aware, but i'm requesting it as an extra requirement for the sake of making things slightly less convenient
The idea is to keep open an avenue of improvement for an eventual researchable thing

LagannPM:
@NanakoAC Its not IV stand, its unique structure, and yes i can make it shitty if that what you ask)

NanakoAC:
We had a talk about this, and it's being reworked to combine with the IV stand into a single object, it will be a good addition

LagannPM:
Updated

code/_onclick/drag_drop.dm
```diff
@@ -6,12 +6,14 @@
 	almost anything into a trash can.
 */
 /atom/MouseDrop(var/atom/over, src_location, over_location, src_control, over_control, params)
-	if(!usr || !over) return
-	if(!Adjacent(usr) || !over.Adjacent(usr)) return // should stop you from dragging through windows
+	if(!usr || !over)
```
LethalGhost: That seems `usr` can't be null here.

code/_onclick/drag_drop.dm
```diff
@@ -6,12 +6,14 @@
...
-	if(!Adjacent(usr) || !over.Adjacent(usr)) return // should stop you from dragging through windows
+	if(!usr || !over) 
+		return FALSE
+	if(!Adjacent(usr) || !over.Adjacent(usr)) // should stop you from dragging through windows
+		return FALSE
 
 	spawn(0)
 		over.MouseDrop_T(src, usr, src_location, over_location, src_control, over_control, params)
-	return
+	return TRUE
```
LethalGhost: This proc shouldn't return anything

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1488,3 +1488,28 @@ var/list/rank_prefix = list(\
...
+		affecting = organs_by_name[BP_GROIN]
+
+	if(affecting && (affecting.robotic >= ORGAN_ROBOT))
+		return FALSE
+	return (species && species.has_organ[organ_check])
+
+/mob/living/carbon/human/has_limb(var/limb_check)	//returns TRUE if found, 2 if limb is robotic, FALSE if not found and null if its chest or groin (dont pass those)
+
+	if (limb_check == BP_CHEST || limb_check == BP_GROIN)	//obviously doesnt work with them
+		return
```
LethalGhost: > //obviously doesnt work with them
And what about `BP_HEAD`?
That check looks unobvious.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1488,3 +1488,28 @@ var/list/rank_prefix = list(\
...
+
+	var/obj/item/organ/external/limb
+	limb = organs_by_name[limb_check]
+
+	if(limb && !limb.is_stump())
+		if(limb.robotic >= ORGAN_ROBOT)
+			return 2
```
LethalGhost: Why not use `return limb.robotic`?

code/modules/organs/blood.dm
```diff
@@ -238,3 +238,7 @@ proc/blood_splatter(var/target,var/datum/reagent/blood/source,var/large)
 	B.fluorescent  = 0
 	B.invisibility = 0
 	return B
+
+//Percentage of maximum blood volume.
+/mob/living/carbon/human/proc/get_blood_volume()
+	return round((vessel.get_reagent_amount(/datum/reagent/blood)/species.blood_volume)*100)
```
LethalGhost: reagent_holder procs work with ids not paths rigth now.

LagannPM: Then hold this PR untill i port medicine/chemistry stuff

code/modules/mob/living/carbon/human/human.dm
LagannPM: then how should i name this check then ? has_appendage ?

code/modules/mob/living/carbon/human/human.dm
LagannPM: will fix

code/modules/mob/living/carbon/human/human.dm
LethalGhost: What that proc should do?
Why `BP_HEAD` counted as limb but `BP_GROIN` not?

LagannPM: it should check if target has this body.
Arms, legs, head can be cut off

LethalGhost: @LagannPM groin can be cut off too

NanakoAC:
I'd really like to see this finished and merged today, it would be a good feature to show off and help build up excitement in the playerbase

LethalGhost:
> I'd really like to see this finished and merged today, it would be a good feature to show off and help build up excitement in the playerbase

It can't be merged with mentoined above errors.

PR #2024, opened 2018-09-27
Endure and Overcome

NanakoAC:
I've always wanted to have variable human health, and been told it couldn't be done. Well screw those guys, I've done it.
This PR implements full support for variable max health on human mobs. Defines that were previously flat values are now multipliers checked against max health. 
But that's all just plumbing work.

More interestingly, this PR reworks Toughness a bit
Now the Toughness stat increases your maximum health, by 1% per point
Humans have 200 health, so that's 2 health per point of toughness

Some old behaviour has been removed, it used to lower the threshold that you would collapse at. This is no longer necessary, as the increased maximum health encompasses this effect too

Also includes a bunch of minor fixes to typos and wierd names/descriptions.  Toughness and robustness hajve had their names set properly in code now, and the descriptions fixed. Robustness doesn't give health, it's used for offensive strength. Lifting, swinging, punching, pulling, pushing, etc

Max health bonuses are a lot more mechanically interesting, and this paves the way for more bonuses in future. Something else on the agenda is to add toughness bonuses to alcohol


LagannPM:
This will conflict with baymed port, close it or leave it as reference for me but add "do not merge" tag

NanakoAC:
Discussed this with lagann, its fine to merge for now. Baymed will take longer than expected, and he's just going to replace the entireity of life code then backport special things we did

code/datums/mob_stats.dm
```diff
@@ -1,11 +1,24 @@
 /datum/stat_holder
 	var/list/stat_list = list()
+	var/mob/living/host
 
-/datum/stat_holder/New()
+/datum/stat_holder/New(var/mob/living/L)
+	host = L
 	for(var/sttype in subtypesof(/datum/stat))
 		var/datum/stat/S = new sttype
+		S.holder = src
```
LethalGhost: It's better to pass holder with `new`

code/datums/mob_stats.dm
```diff
@@ -1,11 +1,24 @@
...
 		stat_list[S.name] = S
 
+/datum/stat_holder/Destroy()
+	for (var/a in stat_list)
+		qdel(a)
```
LethalGhost: `..()` is missed

code/datums/mob_stats.dm
```diff
@@ -20,6 +33,34 @@
...
+/mob/proc/changeStat(statName, Pure = null)
+	if (!istype(src, /mob/living))
+		return 0
+
+	var/mob/living/L = src
+	if (!L.stats)
+		return 0
+
+	return L.stats.changeStat(statName, Pure)
+
```
LethalGhost: Why that is `/mob` not `/mob/living` procs?

code/modules/mob/mob.dm
```diff
@@ -1173,7 +1173,14 @@ mob/proc/yank_out_object()
...
 	//This is called when the mob is let out of a holder
 	//Override for mob-specific functionality
+	return
+
+
+//This should be called whenever something is changed on a mob that may affect its max health
+//Override it to add anything that could concievably affect that mob
+//Overrides in living.dm, robot.dm and human.dm
+/mob/proc/update_max_health()
 	return
```
LethalGhost: Why that `/mob` level procs not `/mob/living`?

code/datums/mob_stats.dm
NanakoAC: Ease of use. This is a standard with almost all mob procs. It's how most of the inventory and damage stuff is handled

You define it at the /mob level but make it do nothing. Then you override it at the levels you care about and add features there.
It allows coding to be done with less typecasting

code/modules/mob/mob.dm
NanakoAC: as above

code/datums/mob_stats.dm
NanakoAC: Infact, this should probably be written into our coding standards. It's how these things are done because it's a good method.

Mob procs should almost always be defined at the /mob level and overridden later

seekin4u: Do we really need to give all this things to `/mob/` instead of `/mob/living/` ?
I mean if only humans and noone can use this - this should really be defined only for `/mob/living/` to not to rubbish already dirty mob's codebase.

NanakoAC: yes, we should, see above. This is established standard for all mob related procs.

NanakoAC:
Holder and missing parent call fixed

code/datums/mob_stats.dm
NanakoAC: but its been the standard since before eris was established, this is how bay designed all mob procs.
If you want to change it enmasse in future we can discuss that. but no reason for an exception right now

LethalGhost:
Cloning code should be updated to new health values too.

LethalGhost:
And things like Odysseus medhud

LethalGhost:
And helpshake/SPR code.

NanakoAC:
fuckit, too much effort to fix

PR #2033, opened 2018-09-27
Pickup, put down animations.

LudwigVonChesterfield:
Don't credit me, the idea is ported from Oracle station.

More animations! Now putting stuff on tables, closets, racks, disposal bins, is... Animated! Hooray!


code/modules/tables/interactions.dm
```diff
@@ -135,7 +135,7 @@
 		user << SPAN_WARNING("There's nothing to put \the [W] on! Try adding plating to \the [src] first.")
 		return
 
-	if (user.unEquip(W, src.loc))
-		set_pixel_click_offset(W, params)
+	set_pixel_click_offset(W, params)
+	user.unEquip(W, src.loc)
```
NanakoAC: nooop

this is a safety check. unEquip returns true if it succeeds, and THEN we move the object
these two should be restored to how they were

LudwigVonChesterfield: There are no cases in this very situation, when it will return FALSE. Prove me wrong.

LudwigVonChesterfield: (P.S.: And even if it does, the item will stay in hand, and there it doesn't matter what it's pixel_x and pixel_y is).

LudwigVonChesterfield:
I however see it as "beauty is in the eye of the beholder". And haven't been given any reason to not switch around the safety check.

PR #2035, opened 2018-09-28
Putdown animation safety fix

NanakoAC:
Restores safety checks
adds a user param to putdown animation
wraps putdown in spawn

This makes everything work perfectly without any need to rearrange safety checks

### DIFF
--- a/code/modules/mob/animations.dm
+++ b/code/modules/mob/animations.dm
@@ -318,31 +318,30 @@ note dizziness decrements automatically in the mob's Life() proc.
 	animate(I, pixel_x = to_x, pixel_y = to_y, time = 3, transform = matrix() * 0, easing = CUBIC_EASING)
 	sleep(3)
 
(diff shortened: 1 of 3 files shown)


code/modules/mob/animations.dm
```diff
@@ -318,31 +318,30 @@ note dizziness decrements automatically in the mob's Life() proc.
 	animate(I, pixel_x = to_x, pixel_y = to_y, time = 3, transform = matrix() * 0, easing = CUBIC_EASING)
 	sleep(3)
 
-/atom/movable/proc/do_putdown_animation(atom/target)
-	set waitfor = FALSE
```
ZVee: Instead of removing waitfor and wrapping code below in spawn, you could simply add `sleep()` with empty argument at the next line, and that would do the same. Since waitfor "magically" converts sleeps into spawns but with less overhead to be believed.
That way we also avoid building ladder with code blocks.

PR #2039, opened 2018-09-28
Inventory code clean up. WIP.

LethalGhost:
No time to explain check [`/code/modules/mob/inventory/_docs.dm`](https://github.com/discordia-space/CEV-Eris/blob/5daa1f85b344fb42b833a681ba3af82580b3697c/code/modules/mob/inventory/_docs.dm)


NanakoAC:
Yes time to explain. We're not merging a 243 file PR on a whim, and probably not this week. I'm going over this with a fine toothed comb

code/datums/uplink/uplink_items.dm
```diff
@@ -116,9 +116,8 @@ datum/uplink_item/dd_SortValue()
 		var/list/L = I
 		if(L.len) I = L[1]
 
-	if(istype(I) && ishuman(user))
-		var/mob/living/carbon/human/A = user
-		A.put_in_any_hand_if_possible(I)
```
NanakoAC: What have you done with the various put in hand functions?

We had many that governed exactly how it should function, like put in hand or delete

code/__defines/items_clothing.dm
```diff
@@ -77,7 +77,8 @@
 #define BLOCK_GAS_SMOKE_EFFECT     0x10 // Blocks the effect that chemical clouds would have on a mob -- glasses, mask and helmets ONLY! (NOTE: flag shared with ONESIZEFITSALL)
 #define FLEXIBLEMATERIAL           0x20 // At the moment, masks with this flag will not prevent eating even if they are covering your face.
 #define COVER_PREVENT_MANIPULATION 0x40 // Only clothing with this flag will prevent manipulation under it. Its for space suits and such, unlike from usual Bay12 rules of clothing manipulation.
-
+#define DRAG_N_DROP_UNEQUIP        0x80 // Allow you put intems in hands with drag and drop
```
NanakoAC: use DRAG_AND_DROP_UNEQUIP instead
it's not good to use colloquialisms or slang in code

code/__defines/items_clothing.dm
```diff
@@ -77,7 +77,8 @@
...
+#define LOUDLY_EQUIPEMENT          0x100// Play sound when equipped/unequipped
```
NanakoAC: This should be something like HAS_EQUIP_SOUND
Although is a flag even necessary? Couldn't we maybe just check whether an equipsound var is populated or not?

LethalGhost: We actually can. 
But it can be only TRUE or FALSE, what's the reason to make it variable instad flag?

code/datums/uplink/uplink_items.dm
LethalGhost: It was too weird proc used in 2 places.
```
/mob/proc/put_in_any_hand_if_possible(obj/item/W as obj, del_on_fail = 0, disable_warning = 1, redraw_mob = 1)
	if(equip_to_slot_if_possible(W, slot_l_hand, del_on_fail, disable_warning, redraw_mob))
		return TRUE
	else if(equip_to_slot_if_possible(W, slot_r_hand, del_on_fail, disable_warning, redraw_mob))
		return TRUE
	return FALSE
```
1) Priority - left hand - first, right - second. Active hand has no matter
2) If first `equip_to_slot_if_possible` fails item qdels, but after that other `equip..._if_possible` calls
and so on.

Furthemore, deleting item if you have no free hand but it still use your points - not good practice.
Also, returning value isn't check and deleted item (what still hold in `I`) is passing to caller.

FTW. It's easy, safety and more obvious to use `put_in_hands`

code/__defines/items_clothing.dm
LethalGhost: @LudwigVonChesterfield 
> What if we have different equip/unequip sounds?

We will add that var then.

code/modules/hydroponics/seed_storage.dm
```diff
@@ -224,12 +224,11 @@
 /obj/machinery/seed_storage/proc/add(var/obj/item/seeds/O as obj)
 	if (ismob(O.loc))
 		var/mob/user = O.loc
-		user.remove_from_mob(O)
+		user.drop_from_inventory(O, src)
 	else if(istype(O.loc,/obj/item/weapon/storage))
 		var/obj/item/weapon/storage/S = O.loc
 		S.remove_from_storage(O, src)
 
-	O.loc = src
```
LethalGhost: Replace with `O.forceMove(src)`

NanakoAC:
Ghost told me not to bother reviewing this, he's going to break it up into several smaller PRs. So closing for now

PR #2041, opened 2018-09-29
Inventory special cases

LethalGhost:
WebUI compare messed up (because of removed species)
All what changed - ‘Item.equipped’ proc call ‘update_icon’ now. Previously ‘Mob.equip_to_slot’ did that.


NanakoAC:
how does this relate to your big inventory overhaul PR?

PR #2079, opened 2018-09-30
Drone, Tank, Lattice, bugfixes

### LINKED ISSUE #2049: Tank storage unit draws over tanks
a tank storage unit should have a lower layer than the things it dispesnes

### LINKED ISSUE #2061: Drones are dense
Drones should not have density, they should be able to walk under peoples legs

### LINKED ISSUE #2062: Building on lattice doesn't work
goddamn i thought i coded this. I'm sure i did, i tested and it was working.

NanakoAC:
fixes #2049
fixes #2062
fixes #2061


code/modules/multiz/turf.dm
```diff
@@ -169,8 +169,13 @@ see multiz/movement.dm for some info.
 			ReplaceWithLattice()
 		return
 
-	if (istype(C, /obj/item/stack/material/steel))
+	if (istype(C, /obj/item/stack/material))
+		var/obj/item/stack/material/M = C
+		if (!istype(M.material, /material/steel))
```
LethalGhost: It's better to use `get_material` instead direct var check.

NanakoAC: sure, will do

Issue #2060, opened 2018-09-29
Robotic Storage Spam


NanakoAC:
When a robot is put into robotic storage, the storage unit starts spamming the common channel with an announcement about it, every second, forever

spaceplaceexplorer:
Cannot replicate locally. Need logs from related round. I suspect it was a telecomms issue.

NanakoAC:
It was solved by deleting the storage unit. It's pretty likely a problem with that object.
This is a critical priority bug witnessed and confirmed by many people, it's going to need to stay open for now, until we either fix it or can be sure its gone

spaceplaceexplorer:
The storage unit has an announcement radio inside of it.

PR #2083, opened 2018-09-30
Fix for spam caused by robots despawning in cryo

spaceplaceexplorer:
fixes https://github.com/discordia-space/CEV-Eris/issues/2060

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -276,7 +276,7 @@
 		qdel(I)
 	qdel(R.module)
 
-	return ..()
+	. = ..()
 
 // This function can not be undone; do not call this unless you are sure
 // Also make sure there is a valid control computer


code/game/machinery/cryopod.dm
```diff
@@ -276,7 +276,7 @@
 		qdel(I)
 	qdel(R.module)
 
-	return ..()
+	. = ..()
```
ZVee: Have you reproduced this bug, and tested if it fixed? Because to me this looks like you only copied solution from baystation's repo, which actually does the same in this scenario. Because if not, this PR not only close the issue of the bug which isn't fixed, but also does nothing.
But in the end this is byond.. so, it will be magic, if this resolved issue.

spaceplaceexplorer:
I don't actually understand dm.

Could not replicate spam with or without this change.

PR #2091, opened 2018-10-01
Brain's layer was changed.

### LINKED ISSUE #2051: Brain layer problem.
I'm under brain.
Я под мозгом.
sprite-0002

seekin4u:
Brain's layer now makes it to be ABOVE human, what is totally wrong
thing. ABOVE_MOB_LAYER was replased withh BELOW_MOB_LAYER.
Fix #2051.

### DIFF
--- a/code/modules/mob/living/carbon/brain/brain_item.dm
+++ b/code/modules/mob/living/carbon/brain/brain_item.dm
@@ -11,7 +11,7 @@
 	throwforce = 1.0
 	throw_speed = 3
 	throw_range = 5
-	layer = ABOVE_MOB_LAYER
+	layer = BELOW_MOB_LAYER
 	origin_tech = list(TECH_BIO = 3)
 	attack_verb = list("attacked", "slapped", "whacked")
 	var/mob/living/carbon/brain/brainmob = null


NanakoAC:
why is this wrong? Its intended so that you can see the brain ontop of the body if someone gets theirs taken out. Infact the same layering should be applied to all internal organs, it makes surgery simpler and less dumb

seekin4u:
I can't say that brain above the human is what i want to see whan i do surgery 🤔 
It's maybe simplier for surgery but not true imo.

NanakoAC:
The brain will only draw above the human once during surgery, when you remove it. This is so you can click on it to pick it up
If you make it draw under the mob, you'll have to rightclick the tile instead which is a pain, and just plain bad. Anything that comes out of something should draw ontop of the thing it came out of. Whether its beakers from a dispenser, food from a vendor, or organs from a human

~I don't see any reason for the change this PR makes

seekin4u:
Rightclick os not the only way to pick up items by the way, there is a much more simplier alt-clicking.
But okay, if you don't see any reason - close this PR and issue

NanakoAC:
Ah there is one thing you could do, if you could make it only draw ontop of lying mobs, that would work. Have mobs change their layer when laid on altars, surgical tables, etc

NanakoAC:
in any case this isn't the correct way to solve the problem, closing for now

PR #2097, opened 2018-10-01
Drying Rack rework

### LINKED ISSUE #1856: Drying Rack lacks icon.
image

Checked the icon folder, not there.

NanakoAC:
Drying Rack Proposed Position

This PR reworks they hydroponics drying rack, making it fully functional with a new icon ported from tg.

The rack dries all its contents in parallel, taking approximately 5-8 minutes to completely dry out a thing

Also generally optimises smartfridges and fixes little issues with them

fixes #1856 

IMPORTANT: This PR does not actually map in a drying rack. The above picture is my recommended location, please map one in there


code/game/machinery/kitchen/smartfridge.dm
```diff
@@ -149,26 +165,66 @@
...
+			continue
+		S.dryness += drying_power * (rand(0.85, 1.15))
+		if (S.dryness >= 1)
+			if(S.dried_type == S.type || !S.dried_type)
+				S.dry = 1
+				S.name = "dried [S.name]"
+				S.color = "#AAAAAA"
+			else
+				var/D = S.dried_type
+				D = new D(src)
```
LethalGhost: Why not just
```
var/D = new S.dried_type (src)
```
?

code/game/machinery/kitchen/smartfridge.dm
```diff
@@ -149,26 +165,66 @@
...
+
+
+
+
+
+
+
+
+
+/obj/machinery/smartfridge/New()
```
LethalGhost: Isn't it better to keep basic `/obj/machinery/smartfridge` together at file begining and subtypes of `/obj/machinery/smartfridge` and overrided procs at the end?

spaceplaceexplorer:
Could you add construction/deconstruction of drying racks in scope of this PR? Or perhaps a future one.

NanakoAC:
No because this exists
https://github.com/discordia-space/CEV-Eris/pull/2124

code/game/machinery/kitchen/smartfridge.dm
LethalGhost: But if you wanted to see how parent work you should be run up and down through the file to find different parts of code related to same type.
I think that way is cleaner:
```
- parent_type_defination
- parent_type_procs

- kind of space/separtator
- child_type_defintion
- child_type_procs
- repid last 3 steps if reqed
```

NanakoAC: I definitely agree that parent definition should be uptop, and all its vars, i just think child types should be below that, and all the main code and procs at the bottom.

For an example of why, I invite you to look at guns: https://i.imgur.com/uNBXf0z.png

All the different individual gun models either have their own file each, or are put into files with a couple other modules. These typically contain only variable definitions and occasionally a few special behaviours.
The core code that makes guns work is seperated too. People - especially new coders, often don't want to see that. They just come to these files to change the icon or description on their favourite gun.

Now we can't really do that for smartfridges, they're too niche to split over multiple files, it'd be a bit silly. But having the simple parts at the top is a good way to compromise. The complicated stuff is there if you want it. But you're less likely to want to look at that, than to look at name, icon, balancing values, etc

PR #2124, opened 2018-10-02
Crafting v1.1

### LINKED ISSUE #2064: Robots can't do crafting?
Recipe wants steel sheets. Have steel sheets in my active modules, crafting window cant find it

NanakoAC:
A major bugfixing and tweaking update for the crafting system. No new content, just fixes and improvements.

The primary goal is to get crafting working properly for robots:
fixes #2064 
Crafting now searches for materials and ingredients in a 3x3 radius around the tile infront of the user.
It does not search for tools there, you must still hold a tool
Searches for all types of items in the three active robot tool slots
obj/item/craft now supports drag and drop for ingredients, so that robots can make things without needing to be able to hold the ingredient
Added a more complex set of safety checks to prevent robots' modules being consumed in recipes, while still allowing them to use stacks and tools they have

In addition to consumed items and tools, a third recipe step type is added, materials. Material stacks are handled properly throughout and are based on the material of the stack, not its typepath

Names of most material datums are replaced with their defines. In general using defines is better

Support for recipe-level times are added, setting a per-step time overrides the time value on the recipe
setting step time works properly too

feedback improved in many places, fixed a bug where tools could be accidentally consumed

Buffed the drone's wood synth a bit so it can make wooden things

Fixed robots' ability to build plating floors too

Rebalanced crafting times across the board


code/datums/craft/step.dm
```diff
@@ -67,34 +79,43 @@
...
+			return
 
 	if(req_amount)
 		if(istype(I, /obj/item/stack))
 			var/obj/item/stack/S = I
 			if(!S.use(req_amount))
+				user << SPAN_WARNING("Not enough items in [S]. It has [S.get_amount()] units and we need [req_amount]")
 				return FALSE
-		else
+		else if (reqed_type) //No deleting tools
```
LethalGhost: You wanna make it impossible to consume items specified by type?

code/datums/craft/step.dm
```diff
@@ -117,3 +169,39 @@
...
+
+	//The item is on the user
+	if (user.canUnEquip(I))
+		//We test if they can remove it, this will return false for robot objects
+		return I
+
+
+	if (istype(I, /obj/item/stack))
+		//Robots are allowed to use stacks, since those will only deplete the amount but not destroy the item
+		return I
```
LethalGhost: That check should go befor `if (user.canUnEquip(I))` line 196

code/datums/craft/step.dm
```diff
@@ -117,3 +169,39 @@
...
+		//The item is held by someone else, can't use
+		return FALSE
+
+	//If we get here, the item is held by our user
+	if (I.loc != user)
+		//The item must be inside a container on their person, it's fine
+		return I
+
+	//The item is on the user
+	if (user.canUnEquip(I))
```
LethalGhost: That check not reqed if `req_amount == 0` item will not be unequipped or consumed.

code/datums/craft/step.dm
```diff
@@ -107,8 +128,39 @@
...
 	if(reqed_type)
-		return locate(reqed_type) in items
+		//Special handling for items that will be consumed
+		for(var/atom/movable/I in items)
+			//First we find the item
+			if (!istype(I, reqed_type))
+				//not the right type
+				continue
+			//Okay, so we found something that matches
+			if (is_valid_to_consume(I, user))
```
LethalGhost: That check not reqed if req_amount == 0 item will not be unequipped or consumed.

code/datums/craft/step.dm
NanakoAC: no, that check makes it ONLY possible to consume items specified by type.

Meaning that, if we got here with a reqed quality, it will be false, and execution won't get in there to the qdel call
so it makes it impossible to consume your tools

code/datums/craft/step.dm
NanakoAC: yes it will. With ingredients the code always assumes one unit, and unequips and consumes it appropriately

Also there is at least one case (tray shield) where someone has tried writing in a quantity for an ingredient. Which doesn't do anything right now, but may cause issues if i tried a check like that

is_valid_to_consume is a generic safety check. Crafting code is non performance critical, so it's better to be robust and safe. we're likely to see a lot of submissions of crafting recipes by inexperienced coders in future

code/datums/craft/step.dm
NanakoAC: see above

code/datums/craft/step.dm
NanakoAC: sure, that would be slightly more efficient, ill make that change

code/modules/mob/living/silicon/robot/robot_modules.dm
```diff
@@ -1040,8 +1040,8 @@ var/global/list/robot_modules = list(
 
 	var/datum/matter_synth/metal = new /datum/matter_synth/metal(25000)
 	var/datum/matter_synth/glass = new /datum/matter_synth/glass(25000)
-	var/datum/matter_synth/wood = new /datum/matter_synth/wood(4000)
-	var/datum/matter_synth/plastic = new /datum/matter_synth/plastic(2000)
+	var/datum/matter_synth/wood = new /datum/matter_synth/wood(10000)
```
NanakoAC: these changes are so drones can craft things with plastic and wood

code/datums/craft/step.dm
LethalGhost: But that already checked in `is_valid_to_consume(...)`.
For things like bots items specified with quality can be useable and must be able to consume.
And that unobvious behavior if you can describe a step requiring a given quality and mark it as comsumeable. Furthemore, description will say, what item will be consumed, but it will not.

NanakoAC: I don't understand what you're saying here. 
The system does not support consuming tools. Specified type items are consumed. Required quality tools are used but not consumed. You can't select and consume something based on its qualities.

There's no existing use for that behaviour and removing this check would be a very messy and half-assed way to add it. If you want that behaviour we can code it seperately. But for now, only items specified by required type should be possible to consume

NanakoAC:
Moved the stack check above can unequip check

code/datums/craft/step.dm
LethalGhost: You can define step as
`/obj/item/weapon/smth` or `QUALITY_SCREWDRIVING` - to "use item"
On the other hand you can define it with list:
`list(/obj/item/weapon/smth, amount)` or `list(QUALITY_SCREWDRIVING, amount)` to "consume item"

NanakoAC: Nooooo, the amount field for required qualities specifies the required level of that quality

We could add support for consuming tools in future, the crafting system is far from done. But it's beyond the scope of this PR. Afaik there is no current application for consuming your tools, and no reason to code that

LethalGhost: But qualities not working that way. That seems like they should, but right now they only affect times and probs, not restrict anything.
Any way that's not right plase for such check. You should catch it in step parsing and drop `consume amount` value then step will have correct description and so on.

NanakoAC: I feel like we're completely misunderstanding each other. Let me explain how i see this.
When a quality and an amount are specified in crafting, it means you need to use a tool with that amount of that qualitt. Eg QUALITY_BOLT_TURNING, 30 requires a wrench, normal tools have 30
The tool is used in the recipe but not consumed. It remains in your hand after that crafting step is done

The req_amount field is used for the required level of the quality. In the above example, that is 30.  It does not mean consume 30 wrenches, it means you need one wrench with a bolt turning quality of 30 or more
there is absolutely no support for eating tools. 

Tool qualities do not currently affect success rates or worktimes of crafting, thereare no provisions for that

For the second point;
It is a fine place for the check. If the check were removed, then tools would be consumed. IF thecks req amount - which is going to be true for tools because they have a numerical requirement for their quality
This works.

Tools will not be dropped, that's not what tools are for

Now closing arguments;

If you want to see crafting recipes which CONSUME the tools used in them, it is generally easy to do that by setting the tool as a required item instead of a required quality.

If you really want to see support for recipes which consume a tool selected by its quality, that is a completely seperate feature and it is beyond the scope of this PR

This is not the final crafting PR, it can and will be adjusted more in future. The purpose of this PR, as i said at the start, is to make crafting work for robots. This argument over tool consumption is pointless. If you want the ability to consume tools, add it. I am not opposing it. It's just not part of this work

NanakoAC:
checking the level of the tool quality after the use_tool call is intentional, not a bug. 
You try your crappy tool and learn that it's inadequate when you fail

PR #2131, opened 2018-10-02
Bunch of AI fixes part 1

### LINKED ISSUE #1827: AI can't examine things
title

### LINKED ISSUE #1829: AI can't understand other languages

### LINKED ISSUE #1981: AI can't access shield control subsystem
"AI can't access shield control subsystem"

### LINKED ISSUE #2112: The AI hears itself when it talks through a holopad.
image

Only the `transmitted` part should appear for the AI.

### LINKED ISSUE #2116: Cryo-ing as AI makes you an ANGEL instead of a ghost.
It also leaves you trapped in your AI core because angels can't go through walls.

Macleod962:
Fixes #2116 
AI would be ghost after core wiping

Fixes #2112 
Now AI hears holo-messages properly

Fixes #1981 
Shield control now usable for AI

Fixes #1829 
AI and pAI now understand another languages

Fixes #1827 
AI now can examine things with shift+click


code/modules/mob/say.dm
```diff
@@ -94,6 +94,9 @@
 		if(istype(other, src.type) || istype(src, other.type))
 			return TRUE
 		return FALSE
+	else	//AI and pAI check
+		if(isAI(src) || ispAI(src))
```
LethalGhost: Isn't that better to set `universal_speak` or `universal_understand` to `TRUE` for them?
Also what about pAI `universal translator` program?

Macleod962: Oh, right! Fixed

Issue #2009, opened 2018-09-25
Multiple hardsuit bugs


spaceplaceexplorer:
1. Hardsuit maintenance port cannot be crowbar open/closed despite being unlocked. Despite this, I had managed to open the hatch on a previous build, but was unable to close it again.

2. Hardsuit head/chest/glove/feet pieces can be unequipped by hand but not dropped, it should not be possible to unequip them by hand.

3. A fully deployed hardsuit control module (backpack slot) can be unequipped by hand, quickly retracting it but making it unable to be dropped until reequipped, fully deployed and then retracted.

4. Hardsuit modules do not function, f.e. meson visor despite being turned on, will not provide 'mesonic' green vision through walls.

NanakoAC:
thats a lot of issues, oh dear

NanakoAC:
removed point 5 as per comment on #2133

PR #2133, opened 2018-10-02
Fixes tool interaction based on tool quality for rig suits

spaceplaceexplorer:
fixes https://github.com/discordia-space/CEV-Eris/issues/2009

Only point 1. is related to rigs.
2. 3. are fixed by upcoming inventory rework.
4. modules function, except vision altering ones which still do not function
5. was not a bug to begin with, internals can be toggled by fully deploying the hardsuit and clicking the O2 icon on hud.

### DIFF
--- a/code/modules/clothing/spacesuits/rig/rig_attackby.dm
+++ b/code/modules/clothing/spacesuits/rig/rig_attackby.dm
@@ -122,7 +122,7 @@
 					user << SPAN_NOTICE("You [open ? "open" : "close"] the access panel.")
 					return
 			else
-				user << "\The [src] access panel is closed."
+				user << SPAN_DANGER("\The [src] access panel is locked.")
 				return
 
 		if(QUALITY_BOLT_TURNING)
--- a/code/modules/clothing/spacesuits/rig/rig_attackby.dm
+++ b/code/modules/clothing/spacesuits/rig/rig_attackby.dm
@@ -31,16 +31,9 @@
 		user << "You [locked ? "lock" : "unlock"] \the [src] access panel."
 		return
 
-	var/list/usable_qualities = list()
-	if(open)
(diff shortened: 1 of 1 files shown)


code/modules/clothing/spacesuits/rig/rig_attackby.dm
```diff
@@ -31,16 +31,9 @@
 		user << "You [locked ? "lock" : "unlock"] \the [src] access panel."
 		return
 
-	var/list/usable_qualities = list()
-	if(open)
-		usable_qualities.Add(QUALITY_WIRE_CUTTING, QUALITY_PULSING, QUALITY_CUTTING, QUALITY_BOLT_TURNING, QUALITY_SCREW_DRIVING)
-	if(!open && locked)
-		usable_qualities.Add(QUALITY_PRYING)
```
NanakoAC: Why change this section? 

I think the point of limiting the useable qualities list is to reduce cases where you're unable to do a normal interaction because one of its tool qualities take over.
And particularly to reduce cases where an item has multiple tool qualities and the one that gets first keeps superceding the one you want to use.
We should query as few tool qualities as possible

spaceplaceexplorer: In addition to correcting the wrong condition for when a crowbar can be applied, intent was to give the user feedback about his actions: despite using a tool that is not compatible with the current item state, it will tell them the problem. In this case, every tool except crowbar will tell the user that they need to open the access panel first, and applying the crowbar will tell them that they need to unlock it first. I would think it is preferable to receiving no feedback at all after using the wrong tool.

I will change line 125 because it should say "is locked" instead of "is closed"

NanakoAC: thats a good reason, okay approving

PR #2134, opened 2018-10-02
Inventory basics, step 1

LethalGhost:
Changes:
- Move drag_n_drop for unequip to `/obj/item` code, ability bound to flag `DRAG_AND_DROP_UNEQUIP`

Moving:
- Moved inventory related item procs/vars to `code/game/objects/items/_inventory.dm`

Docs:
- add item procs docs to `code/modules/mob/inventory/_docs.dm`
Current docs may not reflect the real procs! Procs/docs will be update with following PRs.

### DIFF
--- a/code/modules/mob/inventory/_docs.dm
+++ b/code/modules/mob/inventory/_docs.dm
@@ -1,0 +1,96 @@
+// ITEMS //
+/*
+Most procs from that categoy located in:
+ code/game/objects/items.dm
+*/
+
+
+/obj/item/proc/update_wear_icon(redraw_mob = TRUE)
+/*
+Replacement for outdated update_holding_icon()/update_clothing_icon()
+Found occupied slot and update it icon
+*/
+
+
+/obj/item/proc/pre_equip(var/mob/user, var/slot)
+/*
+Called just before an item is placed in an equipment slot.
+Use this to do any necessary preparations for equipping
+Immediately after this, the equipping will be handled and then equipped will be called.
+Returning a non-zero value will silently abort the equip operation
+Important notes:
+	- Can take few ticks if needed!
+	- Called before unequip (if holding by mob) and can still be located in another mob!
+By default here located equip sound calls.
+*/
+
+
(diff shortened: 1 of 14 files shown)


NanakoAC:
This seems to be just moving things around, right? Are there any actual changes to how procs work?

code/modules/mob/inventory/_docs.dm
```diff
@@ -0,0 +1,96 @@
...
+User is mob that equipped it
+Slot uses the slot_X defines found in items_clothing.dm
+*/
+
+/obj/item/proc/dropped(mob/Mob)
+/*
+Called whenever an item is dropped from INVENTORY SLOT.
+Important note:
+	- It is called after loc is set, so if placed in a container its loc will be that container.
+By default drop zoom if enabled.
```
NanakoAC: where's the drop zoom?

code/modules/mob/inventory/_docs.dm
```diff
@@ -0,0 +1,96 @@
...
+*/
+
+
+/obj/item/proc/can_be_equipped(mob/Mob, slot, disable_warning = FALSE)
```
NanakoAC: how do these relate to existing procs like mob/proc/mob_can_equip?

If they're intended as just an alternate code path, they should wrap the above

LethalGhost: As mentoined in description
> Current docs may not reflect the real procs! 

Because I would have to rewrite them with every PRs.
Welp, we can pass that part till last PRs, why not?

code/modules/mob/inventory/_docs.dm
LethalGhost: What do you mean?
[It's here actually](https://github.com/discordia-space/CEV-Eris/pull/2134/files#diff-068ab32518980ec97236e78fdac2425aR32)

NanakoAC: I'#m dumb, i didnt realise this file was unincluded documentation, thought i was looking at live code

disregard all comments on this file

Issue #2123, opened 2018-10-02
Computer Construction seems too simple


NanakoAC:
https://i.imgur.com/uPVi41H.png

I construct a modular console frame, and it comes pre-turned on with a logo on the screen.

It should be dead until more parts are installed. Like wiring, or circuits or console screen.
I'm not fully aware of the steps involved yet, but i'm sure it shouldn't _look_ finished after just the first step

LagannPM:
tell me if problem is present after #2188

PR #2137, opened 2018-10-03
Fixes modular computer overlay conditions

spaceplaceexplorer:
fixes https://github.com/discordia-space/CEV-Eris/issues/2123

### DIFF
--- a/code/modules/modular_computers/computers/machinery/modular_computer.dm
+++ b/code/modules/modular_computers/computers/machinery/modular_computer.dm
@@ -39,9 +39,12 @@
 	icon_state = icon_state_unpowered
 	overlays.Cut()
 
-	if(!cpu || !cpu.enabled)
-		if (!(stat & NOPOWER) || battery_powered)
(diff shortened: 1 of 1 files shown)


code/modules/modular_computers/computers/machinery/modular_computer.dm
```diff
@@ -39,9 +39,12 @@
 	icon_state = icon_state_unpowered
 	overlays.Cut()
 
-	if(!cpu || !cpu.enabled)
-		if (!(stat & NOPOWER) || battery_powered)
-			overlays.Add(screen_icon_screensaver)
```
NanakoAC: That is a lot of checks, is this many really necessary?

I notice it has a if !cpu check there. Shouldn't that be enough? A newly built frame just made of steel, shouldn't have a cpu in it yet. maybe figure out why that check isnt doing the job

am i misunderstanding something?

NanakoAC:
unsuitable fix, fixing it properly is too risky until after modular computer upstream update. Closing

PR #2144, opened 2018-10-03
Fixes runtime in putdown animations

spaceplaceexplorer:
Runtime fixes


code/modules/mob/animations.dm
```diff
@@ -327,9 +327,13 @@ note dizziness decrements automatically in the mob's Life() proc.
 
 /atom/movable/proc/do_putdown_animation(atom/target, mob/user)
 	spawn()
+		if (!istype(target))
```
LudwigVonChesterfield: Want to point out, how this doesn’t fix the runtime itself, barely masking it.

NanakoAC: how do you figure that?

LudwigVonChesterfield: How do I figure that simply “checking” whether passed argument isn’t null(istype()) does not solve the problem of argument ever becoming null? Shouldn’t the proc not even be called if there is no target? I say spawn has something to do with this.

NanakoAC: because there are a variety of situations where target can be null legitimately and that causes a runtime error

Saying "spawn has something to do with this" is a strangely accusatory way to say "the target atom might have been deleted last frame". That's what a safety check is for

NanakoAC:
add a safety check for user too. and make sure checks are in both pickup and putdown

NanakoAC:
oh and as well as checking type, check QDELETED for user and target

code/modules/mob/animations.dm
```diff
@@ -327,9 +333,17 @@ note dizziness decrements automatically in the mob's Life() proc.
...
+		if (!istype(src) && !QDELETED(src))
+			return
+		if (!istype(target) && !QDELETED(target))
+			return
+		if (!istype(user) && !QDELETED(user))
+			return
 		var/old_invisibility = invisibility // I don't know, it may be used.
 		invisibility = 100
 		var/turf/old_turf = get_turf(user)
+		if (!istype(old_turf) && !QDELETED(user))
```
LethalGhost: You already checked `QDELETED(user)`.

code/modules/mob/animations.dm
```diff
@@ -327,9 +333,17 @@ note dizziness decrements automatically in the mob's Life() proc.
 
 /atom/movable/proc/do_putdown_animation(atom/target, mob/user)
 	spawn()
+		if (!istype(src) && !QDELETED(src))
+			return
+		if (!istype(target) && !QDELETED(target))
+			return
+		if (!istype(user) && !QDELETED(user))
+			return
```
LethalGhost: Shouldn't 
`if (!istype(user) && !QDELETED(user))`
 be 
`if (!istype(user) || QDELETED(user))`
?

spaceplaceexplorer: Yes

PR #2146, opened 2018-10-03
Pickweight for Obj/Randoms

NanakoAC:
Converts most of the obj/random spawners to use pickweight instead of the old prob method

I'm not sure if its faster, but it's a bit more flexible, easier to understand and a loot easier to write when adding new things.  On aurora we switched over to pickweight a long time ago

basic pick is retained for the nonweighted lists, and also for the material stacks, they didnt work with pickweight due to their parameters

Also converts obj/random to run on initialize rather than new.
Updates lockers to support this so that they'll still contain things spawned on them
Optimises and extends the spawners a bit, post spawn functionality added i'll use shortly


code/_helpers/lists.dm
```diff
@@ -110,7 +110,11 @@ proc/listclearnulls(list/list)
 		result = first ^ second
 	return result
 
-//Pretends to pick an element based on its weight but really just seems to pick a random element.
```
NanakoAC: I wonder what silly person wrote this comment.

There was a minor error in here, the starting value for total, but otherwise it was fine, it's a tried and tested method

code/game/objects/random/materials.dm
```diff
@@ -2,6 +2,7 @@
 	name = "random building material"
 	icon_state = "material-grey"
 
+//This stuff can't be easily converted to pickweight because of these amount fields
```
GrayRachnid: For consistency's sake, can't you just make a new obj that's a child of these materials that just has the amount changed on it?

NanakoAC: i'll do something about it, DNM for now

code/game/objects/items/stacks/stack.dm
```diff
@@ -22,13 +22,21 @@
...
 	var/list/datum/matter_synth/synths = null
 
-/obj/item/stack/New(var/loc, var/amount=null)
-	..()
+	//If either of these two are set to nonzero values, the stack will have randomised quantity on spawn
+	//Used for the /random subtypes of material stacks. any stack works
+	var/rand_min = 0
+	var/rand_max = 0
+
+/obj/item/stack/Initialize(var/loc, var/amount=null)
```
LethalGhost: Can `amount` argument be accessed?

code/game/objects/structures/crates_lockers/closets.dm
```diff
@@ -46,7 +46,12 @@
 /obj/structure/closet/can_prevent_fall()
 	return TRUE
 
-/obj/structure/closet/Initialize()
+/obj/structure/closet/Initialize(mapload)
+	..()
+	populate_contents()
+	return mapload ? INITIALIZE_HINT_LATELOAD : INITIALIZE_HINT_NORMAL
```
LethalGhost: `populate_contents()` will be called twice if `mapload` is not null.

NanakoAC: welp, will fix

code/game/objects/items/stacks/stack.dm
NanakoAC: derp, i didnt think of this
will fix

code/game/objects/structures/crates_lockers/closets.dm
NanakoAC: done

NanakoAC:
This should be ready to go

PR #2149, opened 2018-10-03
Fixes rig visors not working with glasses overlay

spaceplaceexplorer:
Adds an override into process_glasses to bypass newly added cell dependency of various glasses.
Adds a check into glasses_overlay to apply active rig visor overlay.

### DIFF
--- a/code/_onclick/hud/screen_objects/base_screen_objects.dm
+++ b/code/_onclick/hud/screen_objects/base_screen_objects.dm
@@ -1090,6 +1090,12 @@ obj/screen/fire/DEADelize()
 		if (G.active && G.overlay)//check here need if someone want call this func directly
 			overlays |= G.overlay
 
+	if(istype(H.back,/obj/item/weapon/rig))
+		var/obj/item/weapon/rig/O = H.back
+		if(O.visor && O.visor.active && O.visor.vision && O.visor.vision.glasses && (!O.helmet || (H.head && O.helmet == H.head)))
+			if (O.visor.vision.glasses.overlay)
+				overlays |= O.visor.vision.glasses.overlay
+
 
 /*	if(owner.gun_move_icon)
 		if(!(target_permissions & TARGET_CAN_MOVE))
--- a/code/_onclick/hud/screen_objects/base_screen_objects.dm
+++ b/code/_onclick/hud/screen_objects/base_screen_objects.dm
@@ -1090,8 +1090,8 @@ obj/screen/fire/DEADelize()
 		if (G.active && G.overlay)//check here need if someone want call this func directly
 			overlays |= G.overlay
 
-	if(istype(H.back,/obj/item/weapon/rig))
-		var/obj/item/weapon/rig/O = H.back
+	if(istype(H.wearing_rig,/obj/item/weapon/rig))
+		var/obj/item/weapon/rig/O = H.wearing_rig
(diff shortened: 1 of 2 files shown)


code/_onclick/hud/screen_objects/base_screen_objects.dm
```diff
@@ -1090,6 +1090,12 @@ obj/screen/fire/DEADelize()
 		if (G.active && G.overlay)//check here need if someone want call this func directly
 			overlays |= G.overlay
 
+	if(istype(H.back,/obj/item/weapon/rig))
+		var/obj/item/weapon/rig/O = H.back
+		if(O.visor && O.visor.active && O.visor.vision && O.visor.vision.glasses && (!O.helmet || (H.head && O.helmet == H.head)))
```
NanakoAC: jesus fuck thats a lot of checks.

code/_onclick/hud/screen_objects/base_screen_objects.dm
```diff
@@ -1090,6 +1090,12 @@ obj/screen/fire/DEADelize()
 		if (G.active && G.overlay)//check here need if someone want call this func directly
 			overlays |= G.overlay
 
+	if(istype(H.back,/obj/item/weapon/rig))
```
NanakoAC: A request;
Mobs, or at least humans, have a wearing_rig var.
It currently doesn't work, its not getting populated when you wear one

Find out why, fix it, and then check it instead of this check here

Reason being, the design spec for rigs doesn't require them to be on the back. There's a future possibility of rigs that come in a belt, or a shirt, or a tie or whatever. Lets try to avoid checking a specific slot

code/_onclick/hud/screen_objects/base_screen_objects.dm
NanakoAC: There's a lot of things i don't understand here. The first one is, why is this snippet necessary at all. Since there's a call to process_glasses in process_rig

code/modules/mob/living/carbon/human/human_helpers.dm
```diff
@@ -59,8 +59,8 @@
 	if(istype(back,/obj/item/weapon/rig))
 		process_rig(back)
 
-/mob/living/carbon/human/proc/process_glasses(var/obj/item/clothing/glasses/G)
-	if(G && G.active)
+/mob/living/carbon/human/proc/process_glasses(var/obj/item/clothing/glasses/G, var/forceActive)
+	if(G && (G.active || forceActive))
```
NanakoAC: What's the need for this forceactive flag? To override cell dependancy right?

But even hardsuit vision modules have cell dependancy, the suit has its own internal power source
Checking the active flag should be enough, and the suit should set and control that active flag internally when appropriate

spaceplaceexplorer: Modules have their own power drain, vision modules use an instance of their glasses internally to pass along to vision processors for the proper effects. Glasses keep an internal cell that they use while active in Process(). Neither the glasses cell nor the glasses instance can be accesses by players. The active flag for glasses does not come into play here as it depends on the glasses cell.

Rig could charge/discharge the internal cell of the glasses, or the glasses could check for being inside a module and utilize rig power, but that is complicating things for no gain since module power drain functionality is already in place.

Regardless, balancing is not the scope of this PR, merely restoring lost functionality. A future PR can deal with balancing the module power drain should that be desired.

Issue #2122, opened 2018-10-02
Nanoui Breaks Before Roundstart


NanakoAC:
Attempting to access the UI of an object before the round has started, breaks that object, permanantly. You have to shut down the server and restart to make it work.

This also includes the crafting menu

This is a huge inconvenience for debugging

PR #2164, opened 2018-10-05
Move asset loading into pregame

spaceplaceexplorer:
fixes https://github.com/discordia-space/CEV-Eris/issues/2122

I have seen no side effects from this.


NanakoAC:
Does this entirely prevent the bug? Is there any possibility of getting in before assets are sent if i spam the observe button? Cuz i do that

PR #2169, opened 2018-10-06
Debugging: WIP Storyteller

NanakoAC:
Don't even think about merging this, it's just here to provide a diff so people can help me debug.
This PR will be deleted in time

Somewhere i've fucked up something so bad that the server won't launch, the error handler won't even work. I had to stick in some code of my own to prevent the error handler from throwing errors itself

List of errors
https://pastebin.com/UwCBrZJ4


NanakoAC:
problem solved, shutting down

PR #2170, opened 2018-10-06
Fixes vents and scrubbers randomly hibernating

spaceplaceexplorer:
Vents and scrubbers hibernated for 10 to 20 seconds whenever they thought they were idle. In case of scrubbers, there was nothing to wake them up, causing situations where they took extraordinarily long to do their job due to hibernating even when they were supposed to be working, because for one process() call everything seemed ok.

I've moved a few conditions around to compensate for whatever minor performance hit this causes, the checks in place were already pretty sane except for scrubbers, which I'm assuming is what prompted whoever decided to 'fix' it to use random based hibernation. Originally they were updating their network regardless of work done.


NanakoAC:
A thought comes to mind.

Instead of removing the hibernating feature completely, what about just having a wake up proc. Which would be called by changing anything on the local air alarm.

Since hibernating is based on a var, you could just set it to zero from the air alarm to immediately wake up and start doing things

NanakoAC:
discussed, its fine for now, approving

PR #2171, opened 2018-10-06
Inventory slots

LethalGhost:
Inclusion of slots datum.
Replace outdated `update_clothing_icon` and `update_held_icon` procs with `update_wear_icon`
Most slots vars currently unused. They added for future updates of equip/unequip code.


code/controllers/subsystems/inventory.dm
```diff
@@ -0,0 +1,24 @@
+SUBSYSTEM_DEF(inventory)
```
NanakoAC: What the heck do we need an inventory subsystem for?

LethalGhost: It's not fire or smth like that.
I'm not sure is it really reqed but that seems like proper way to initialize slot datums.
And that's easier to debug SubSystem when global list of slots.

PR #2172, opened 2018-10-06
Rig Cutting

NanakoAC:
Inspired by a comment on #2154

There is currently no way to forcibly open a hardsuit module which is locked. The lock is an impenetrable fortress. I find this unrealistic and limiting.

To fix it, this PR adds the ability to cut them open with a welder.
But not easily, oh no. 
It has the longest worktime and the second highest difficulty that we've defined, it is a very slow process if you lack mechanical skill

Cutting open the panel permanantly destroys the lock, it can never be locked again, at least for now.
Maybe in future we'll add a way to repair that, but its not so important


code/modules/clothing/spacesuits/rig/rig_attackby.dm
```diff
@@ -7,9 +7,7 @@
 		if(shock(user)) //Handles removing charge from the cell, as well. No need to do that here.
 			return
 
-	// Pass repair items on to the chestpiece.
-	if(chest && (istype(I,/obj/item/stack/material) || QUALITY_WELDING in I.tool_qualities))
-		return chest.attackby(I,user)
+
```
LethalGhost: You can't repair it with material stacks from now?

NanakoAC:
updated, cutting now destroys the panel lock

code/modules/clothing/spacesuits/rig/rig.dm
```diff
@@ -558,7 +558,8 @@
 		ai_override_enabled = !ai_override_enabled
 		notify_ai("Synthetic suit control has been [ai_override_enabled ? "enabled" : "disabled"].")
 	else if(href_list["toggle_suit_lock"])
-		locked = !locked
+		if (locked != -1)
```
spaceplaceexplorer: User could use with some feedback why the button does nothing. Beeps in error etc.

code/modules/clothing/spacesuits/rig/rig_attackby.dm
```diff
@@ -116,7 +118,7 @@
 				return
 
 		if(QUALITY_PRYING)
-			if(!locked)
+			if(locked != TRUE)
```
spaceplaceexplorer: Locked is a ternary value, true/false has no meaning and is misleading. Should be 1 or a defined constant.

code/modules/clothing/spacesuits/rig/rig_attackby.dm
```diff
@@ -142,9 +144,27 @@
 				user << "\The [src] access panel is closed."
 				return
 
+		if(QUALITY_WELDING)
+			//Cutting through the cover lock. This allows access to the wires inside so you can disable access requirements
+			//Ridiculously difficult to do, hijacking a rig will take a long time if you don't have good mechanical training
+			if(locked == TRUE)
```
spaceplaceexplorer: Same here.

spaceplaceexplorer:
Rest checks out.

Issue #2016, opened 2018-09-26
Shower does not update character sprite


spaceplaceexplorer:
Taking a shower and removing blood from your items will not update your sprite, items have to be unequipped and put back on.

NanakoAC:
Human item updating code is a mess. 
On aurora i wrote a major project to fix it all. Never finished it. I worked on it on and off throughout a  year, never quite got it done. 

I hope to do it for eris, but it'll be a fair bit of work

PR #2173, opened 2018-10-06
Fixes showers not updating player sprite

spaceplaceexplorer:
fixes https://github.com/discordia-space/CEV-Eris/issues/2016


code/game/objects/structures/watercloset.dm
```diff
@@ -297,6 +297,7 @@
 				if(H.belt.clean_blood())
 					H.update_inv_belt(0)
 			H.clean_blood(washshoes)
+			H.update_icons()
```
LethalGhost: OMG that proc above... just creazy

PR #2187, opened 2018-10-08
Removing datum pooling.

ZVee:
Before i dive into numerous `New` procs, i wan't to get rid off pooling first (which i just noticed).


ZVee:
No, it even decreases it.

NanakoAC:
I read the links provided and understand completely. It seems pooling is no good. oh well

NanakoAC:
This doesn'ty seem worth the risk to merge so close to launch. High potential to break stuff, minimal benefit to end users if it doesn't

PR #2191, opened 2018-10-08
An Ardent Orison

NanakoAC:
Adds an advanced welding tool 

welder

It has slightly higher stats than the base welder in all categories, but otherwise functions the same.
It's been added to the appropriate random pools

In addition, this PR fixes inhand welders not updating their sprite when turned on/off. Now they do, and its clearly visible when someone has an active welder in their hand

In addition, the fuel capacity of the basic welder is nerfed slightly.
This will encourage use of welding backpacks, or tanks


code/game/objects/random/tools.dm
```diff
@@ -3,54 +3,55 @@
 	icon_state = "tool-grey"
 
 /obj/random/tool/item_to_spawn()
-	return pickweight(list(/obj/random/rare = 1,\
```
NanakoAC: I removed the trailing slashes from this file

but actual changes are just adding the advanced welder to this and the advanced list

code/modules/mob/update_icons.dm
```diff
@@ -4,6 +4,9 @@
 /mob/proc/regenerate_icons()		//TODO: phase this out completely if possible
 	return
 
+/mob/proc/update_hands()
+	return
+
```
LethalGhost: Are we really requed that proc?
It seems like part of code what direct access `l_hand`/`r_hand` vars.

NanakoAC: it's a simple wrapper for a few common parts, to save on copypasting 2-4 lines everywhere

LethalGhost: But why can we need call `update both hands`?
Equip/uneqip handle slot icon updates, `Item.update_wear_icon()` update current slot if icon was changed.

NanakoAC: You really can't see a use for updating all the icons related to someone's hands??

What if you're not sure what was changed? It's an obviously useful subset of features which is more efficient than calling regenerate icons

it sounds like your idea is better in this case, but i still believe the proc will be useful

NanakoAC:
DNM for now, going to make a better sprite and rework the code a little

Issue #1994, opened 2018-09-23
Bug with cells of stungloves


Walfanger:
"Ironhammer Stun Gloves does not accept power cells different then default one. You can't make them use better or worser cells if needed"

PR #2196, opened 2018-10-09
Fixes stungloves cell compatibility

spaceplaceexplorer:
fixes https://github.com/discordia-space/CEV-Eris/issues/1994

Fixes some 4am coding ideas.


code/modules/power/cell.dm
```diff
@@ -61,6 +60,7 @@
 	return use(cell_amt) / CELLRATE
 
 /obj/item/weapon/cell/update_icon()
+	var/charge_status
```
NanakoAC: Is this part of the idea you were discussing last night for optimising cells inside apcs and chargers?

how does it work?

PR #2215, opened 2018-10-10
Fixes payday runtime

spaceplaceexplorer:


code/modules/events/payday.dm
```diff
@@ -4,16 +4,20 @@
 /datum/event/payday/start()
 	var/datum/transaction/T = new(0, "Central Payment Account", "Payday", "NTGalaxyNet Terminal #[rand(111,1111)]")
 
-	for(var/record in data_core.general)
-		var/datum/data/record/R = record
```
NanakoAC: Revert lines 7 and 8, this is actually an optimisation

By doing an untyped iteration through a list, and then explicitly typecasting each one, you save on running an istype check against each element in the list

PR #2217, opened 2018-10-10
Fixes beeping during shrapnel removal

spaceplaceexplorer:
Now passes severity 1 into malfunction, which handles what the implant should do.


code/modules/surgery/implant.dm
```diff
@@ -228,8 +228,7 @@
 			var/fail_prob = 10
 			fail_prob += 100 - tool_quality(tool)
 			if (prob(fail_prob))
+				user.visible_message("\red You scrape something inside [target]'s [affected.name]." )
 				var/obj/item/weapon/implant/imp = affected.implants[1]
```
LethalGhost: Also, `/obj/item/weapon/implant/imp` can have another type. 
Isn't it better to use `istype(imp)` and move beeping instead removing?

spaceplaceexplorer: All implants have their malfunction proc which handles any sound they make when they malfunction. Surgery does not handle what implants do when they malfunction and merely tells them they have malfunctioned and how much.

LethalGhost: Main point here - it can runtime without `istype(imp)`

PR #2223, opened 2018-10-11
Job Slots Expansion

NanakoAC:
Eris is big, and its made for a lot of people. We're still pinning down exactly what player numbers we want to target, but it's going to be a lot. We need a lot of job slots.

This PR expands the number of slots for almost every non-singleton job. Some are affected more than others. 
Generally at the player numbers we're seeing now, running out of slots should be a complete non issue, and nobody should be unable to get the job they want

The biggest and most necessary change is to janitors. We desperately need more of those.

Increasing ironhammer will be the most controversial, but they have a tendancy to all die and leave the station unprotected

feedback welcome


GrayRachnid:
1) We don't have the playerbase to supply all these numbers.
2) This PR doesn't include any map edits to add more equipment for these open slots.
3) 8 Sec officer slots will just mean more people respawning into a sec officer after they get killed by an antag.
4) 6 Janitors is insane. Two janitors can do enough work for the ship.
5) The hulk doesn't have space for 5 miners.
6) There's no space for a 3rd chemist in chemistry.
7) Imo 3 cyborg slots is too much, even 2 is too much. Give the roboticist something to do.
8) 5 doctors and 3 paramedics? Irl there's more paramedics than doctors stationed in most clinics. Plus, these five doctors are all useless until defibs are added because everyone dies in a few shots.

I'm fine with the gardener change. The rest I mostly oppose however.

NanakoAC:
Firstly, this is mostly future facing
People die, a lot. And often can't be replaced because of slot limits, thats an issue.
Sharing equipment is a thing, not everyone needs their own set.  Almost every job can find some spare equipment scattered around maintenace, or order it from cargo, more things will be added in future, but mapping PRs get messy

Our numbers are growing and we will need the extra slots soon. Right now though, it will allow stacking up many people into a role that's needed, like mining or cleaning

The hulk is a massive shuttle that has space for over 40 people to do the thing it's supposed to do - take them from one place to another. Then the miners get their asses outdoors and start digging. No shortage of space on an asteroid

6 janitors is not insane, this is the biggest ship i've ever seen. Even 20 janitors wouldn't be enough to clean it up


It adds nothing to gameplay to have people forced to go assistant because they can't do the job that they want - and that is needed, due to the slots already having been consumed by a succession of people who joined and died, or went SSD, or ended up doing something different. 
With the exception of singleton roles, there's no point to it, it's not fun for the station to lack a vital role, or for people who want to be in that role to be barred from it

And about people respawning as sec; the respawn timer is there for a reason. There's no reason someone can't do that if they want to.  Metagaming knowledge about the round is another matter, but thats nothing specific to security, and i don't believe it's a significant problem

Voldirs:
I think slot numbers should be scaleable rather than hardcoded. It's just too much.

LagannPM:
agree with above, it also draws job assigning by heads obsolete.
My modular computer PR gives ability for every head to grant permanent access to their department.
It should be dynamic to prevent overpopulating some jobs like security and force people to fill shortage of other jobs

NanakoAC:
Job assigning with heads has some merit
I don't agree with the overpopulating jobs though. We've already recently added a feature that allows someone to go to cryo and respawn fairly quickly, they can use this if they find themselves at a loose end

In addition, the crew manifest is displayed to people before joining, they can easily see what roles exist - and are inactive

The critical problem here that this is aiming to solve though, is people being unable to join a needed role because all of the slots are occupied by dead or SSD players, and the job isnt actually getting done

Voldirs:
@NanakoAC 5 miners is too much, there is not so much content for them also there are only 3 voidsuits for them, 3 chemists is the same, there is only one chem dispenser 2 chemists is ideal for such workplace, 3 robots is too much because there it leaves not so much job for roboticists and 6 janitors is  too much just because, maybe 2 or 3 is fine if they all get mops and buckets.

NanakoAC:
Numbers revised a bit, based on feedback

Multiple janitors may struggle for equipment for now, but cleaning gear is quite easy to source. I'll add some features in the near future to make buckets, mops and janicarts craftable

GrayRachnid:
The numbers are better. I disagree with having 2 med specialists however, there's already 2 inspectors. 

This also still doesn't solve the problem of where you're going to put them. Where will a 2nd bartender work? Where will a 4th miner get his gear from? Where will a 2nd med specialist work? 

And in regards to the medbay numbers, I believe there should be more paramedics than doctors.

NanakoAC:
where will they work? That's up to them to figure out.
In most cases the extra slot will be used to replace a dead or SSD crewmember

I have a plan to expand the forensics lab too, everything is done in steps

LagannPM:
Can we make checks in player logout/login to open those extra slots ? i know that this can be abused but whatever

NanakoAC:
nah that's a bad way to do things.
I will definitely code some dynamic features to open slots, and to reopen them for dead/SSD players in some form. But thats in the future

This isn't a temporary measure though, it's a first step. Setting sane baseline values that include wiggle room. Dynamic slot opening features will build on it in future because there's no way we can ever support a massive population with just what this PR adds

NanakoAC:
@SpadesSlick This has your requested changes, its done, merge it

PR #2225, opened 2018-10-11
Merged plant_controller into MC's subsystem.

ZVee:


code/modules/hydroponics/grown_inedible.dm
```diff
@@ -8,9 +8,8 @@
 	var/plantname
 	var/potency = 1
 
-/obj/item/weapon/grown/New(newloc,planttype)
-
-	..()
+/obj/item/weapon/grown/Initialize(mapload, planttype)
+	. = ..()
```
LethalGhost: Will `planttype` argument be accessable?

ZVee: I see no problem with access, unless someone messes with new and skips loc as first argument writing something like `new type(planttype)` instead of `new type(loc, planttype)` or even using named arguments which are actually dead with Initialization as i know.

PR #2231, opened 2018-10-11
Mouse trap damage nerf

LagannPM:


NanakoAC:
I literally just buffed them.

They're intended to provide a mild threat when wandering maintenance, They do a maximum of 15 damage

Voldirs:
If I remember correctly, one can place unlimited number of mousetraps at one tile... Looks like possible insta-kill trap of mousetraps. Sounds too hilarious.

NanakoAC:
an infinite number will still instakill even if the damage values are nerfed. This PR both failed to solve that problem and reverted meaningful content.

To fix that, maybe a limit of one active trap per tile, or one trap triggered per entry

PR #2238, opened 2018-10-13
Pain Indication(Clouds)

LudwigVonChesterfield:
The gif speaks for itself.


NanakoAC:
As a longterm feature, I'm not really enthused about this.
It may be overwritten or made obsolete in future, as @spaceplaceexplorer is planning to port actual scream audio, and at some point I plan to port my larger icons for security huds, which will cover up a lot of that speechbubble space.
In addition to that it just looks kind of silly, 

However, it does help to visually communicate important information. On that basis we'll accept it, at least temporarily. I can't promise that this feature will stay though

PR #2240, opened 2018-10-13
Pouches.

LudwigVonChesterfield:
<s>Just like on that server where you shoot bugs with rifles or shotguns!</s>

Sprites are made by WatlerJe, so thats a big thanks to him!

image

All pouches have a "sliding" mechanic, which can be enabled/disabled on right click --> "Toggle Slide". The mechanic allows you to instantly get the item from the pouch, if it is the only one left inside.

<b>The current pouches in existence are:</b>
    + <i>Small Generic Pouch</i>
        One slot. Basically does nothing, and is just there.
    + <i>Medium Generic Pouch</i>
        Two slots. Can store any SMALL or smaller item in.
    + <i>Engineering Tools Pouch</i>
        Three slots. Can store any tools, and all the analyzers.
    + <i>Engineering Supply Pouch</i>
        Two slots. Can store tools, analyzers, cells, circuitboards, and sheets of materials!
    + <i>Medical Supply Pouch</i>
        Three slots. Stores medical items.
    + <i>Flare Pouch</i>
        Five slots, all are up for flares.
    + <i>Ammo Pouch</i>
        Three slots. Stores magazines, or ammo casings.
    + <i>Pistol Holster</i>
        Stores one pistol. Has sliding behavior activated by default.
    + <i>Baton Holster</i>
        Stores the stunbaton(Actually any melee, or even a crowbar). Has sliding behavior activated by default.


NanakoAC:
I'm a bit confused about the purpose of these pouches. What makes them different from other storage items?

How are you supposed to use them? Can they be worn on any specific slots, or attached to clothing?

This "sliding" behaviour sounds useful, if terribly named. But i'm not liking the restriction of it only working when there's a single item. Wouldn't it be more useful to have a single click always remove the last item that was placed in?

LudwigVonChesterfield:
They are put into your pockets.

Faryaa:
> I'm a bit confused about the purpose of these pouches. What makes them different from other storage items?
> 
> How are you supposed to use them? Can they be worn on any specific slots, or attached to clothing?
> 
> This "sliding" behaviour sounds useful, if terribly named. But i'm not liking the restriction of it only working when there's a single item. Wouldn't it be more useful to have a single click always remove the last item that was placed in?

They occupy your pocket slots effectively replacing your not very useful small pockets with more useful specialised ones. It always bugged me how less your pants can hold in terms of stuff. Like a single pill occupying a whole pocket.

SpadesSlick:
CM need to store sheets of materials in those poaches for balance needs, but we don't. Tools should be enough.

Faryaa:
> CM need to store sheets of materials in those poaches for balance needs, but we don't. Tools should be enough.

What about power cells and boards? Those are needed by engineers for repairs but occupy as much space as full sheet of any resource occupy. Having a pouch to carry a spare power cell and APC board would be handy for electricians.

SpadesSlick:
Oh wait, there is two of them. Well, maybe if we made those poaches rare.

NanakoAC:
Also, switch the two generic pouches to use the normal inventory system instead, based on w_class of items.
with a very small capacity, like 6-12
That'll make them actually useful when combined with the sliding feature

LudwigVonChesterfield:
All the requested changes are done, but I also felt like populating some closets with the pouches.

Hos, warden, officers get baton holster and pistol holster.

Hop, captain get a pistol holster.

Technomancer Exultant gets an engineering supply pouch.

CMO gets his own medical supply pouch.

And I gave Janitor the medium generic one. I just always felt like janitors lack storage space.

Faryaa:
I was just thinking - what if flare pouch would also accept glowsticks, syringes and autoinjectors? Those have similar shape and pouch would be more useful. Just ask chemist to provide you with filled injectors or syringes for quick emergency use.

code/game/objects/items/weapons/storage/pouches.dm
```diff
@@ -0,0 +1,212 @@
...
+	else
+		..()
+
+/obj/item/weapon/storage/pouch/small_generic
+	name = "small generic pouch"
+	desc = "Can hold anything in it, but only about once."
+	icon_state = "small_generic"
+	item_state = "small_generic"
+
+	max_storage_space = ITEM_SIZE_TINY * 2
```
NanakoAC: Set the storage space of small pouch to ITEM_SIZE_SMALL * 2
and set the storage space of medium pouch to ITEM_SIZE_SMALL * 3

code/game/objects/items/weapons/storage/pouches.dm
```diff
@@ -0,0 +1,212 @@
...
+	storage_slots = 3
+	w_class = ITEM_SIZE_NORMAL
+	max_w_class = ITEM_SIZE_NORMAL
+
+	can_hold = list(
+		/obj/item/ammo_magazine,
+		/obj/item/ammo_casing
+		)
+
+/obj/item/weapon/storage/pouch/flare
```
NanakoAC: Since it's very specific and niche purpose, make the flare pouch start pre-filled

In its Initialize proc, create five flares inside it.
Do it like...
```/obj/item/weapon/storage/pouch/flare/Initialize(var/roundstart)
if (roundstart)//make the flares```

code/game/objects/items/weapons/storage/pouches.dm
LudwigVonChesterfield: That would be slightly too big. And having medium pouch contain 3 small items basically defeats the entire purpose of specialized pouches. I think just how it is is already good.

code/game/objects/items/weapons/storage/pouches.dm
LudwigVonChesterfield: There was also a suggestion on allowing glow sticks and syringes be kept in these. Should I?

code/game/objects/items/weapons/storage/pouches.dm
NanakoAC: The medium pouch is okay, but the small one is useless, it has almost no benefit. 
Making them double and triple the number of things a pocket can hold, respectively, would be pretty good

I can see that it would cause some issues with the specialised pouches, there's a problem there
Any alternate ideas to make the small pouch more useful?

I'm thinking making crowbars bigger, so that they no longer fit in pockets, and require a pouch or toolbelt to store

code/game/objects/items/weapons/storage/pouches.dm
NanakoAC: yes this is actually an excellent idea
Syringes, autoinjectors, glowsticks, hypospray, pens/crayons and vials! this will give it a significant medical use

In that case scrap the prespawned flare idea, and rename it something more generic like Cylinder pouch

code/game/objects/items/weapons/storage/pouches.dm
LudwigVonChesterfield: It is more useful than just a pocket already. It fits two pens instead of one!

NanakoAC: But nobody uses pens.
Pockets are variably used for...
Flashlight, crowbar, health analyzer, welding tool, syringe, ammo clip

Basically things that you urgently need onhand right now.
Even a small pouch should offer some meaningful advantage by allowing you to store two of these things

NanakoAC: Also give the janitor a small pouch instead of a medium one, and make the medium pouches generally rarer

PR #2245, opened 2018-10-16
Adds cyberlimbs for slim bodytype

SubberTheFabulous:
All bodytypes used the default bodytype's cyberlimbs. This leaves untouched the cyberlimbs for fat bodytype, though.
Please, don't make me do screenshots.


SubberTheFabulous:
According to Lethal Ghost, there is. It just uses this: 
`icon_state = "[organ_tag]_[gender][body]"`
It uses the same thing BlueBay used, iirc, so, it should be okay regardless.

Issue #1967, opened 2018-09-22
You can't make some more meat spikes


MrFoxik:
So you can't make your own kitchen, and make more meat of humans/monkeys/etc

NanakoAC:
Sounds like something for a crafting recipe

spaceplaceexplorer:
Deconstruction for meatspikes too if it doesn't already exist.

PR #2247, opened 2018-10-17
Adds meat spike construction/deconstruction

ghost:
fixes https://github.com/discordia-space/CEV-Eris/issues/1967

Cost is 3 metal sheets.

### DIFF
--- a/code/datums/craft/recipe.dm
+++ b/code/datums/craft/recipe.dm
@@ -5,7 +5,7 @@
 	var/result
 
 	var/list/steps
-	var/flags = CRAFT_ONE_PER_TURF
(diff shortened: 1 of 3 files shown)


spaceplaceexplorer:
1 metal sheet is already equal 1 metal rod. Even a flimsy table frame is made out metal sheets. There's no point in microcrafting.

NanakoAC:
suire there is, crafting is about immersion and busywork, we should use what makes the most sense.
The crafting system is designed to replace the old method of magically turning metal sheets into finished objects, by simulating the intermediate steps (and using tools)

Recipes that just use sheets should be cut down

PR #2249, opened 2018-10-17
Repeating Tool Sounds

NanakoAC:
I always found it odd that using tools would play a sound only once, no matter how long the worktime is. And for the rest of the time, which could be up to a minute, you just stand there in silence. This is especially jarring for anyone else who might be observing you, as without the sounds they have no way to tell what you're doing

To remedy this, i've added and utilised a new feature. Repeating sound support.
Note that this is not at all the same as looping sounds. 

Repeating sound plays a sound file with given settings at regular intervals, with random variance optionally applied to that interval.
It's open for anyone and anything to use, and it makes use of addtimer and callbacks

Secondly, this PR implements it in tools. Now tools will make their tool sound roughly every 2.5 seconds while a tool use operation is going on.
Naturally, the sound is cancelled if the tool action is aborted, and stops when the action is finished

Each instance of the sound applies frequency variance if set, so it will sound slightly different. It gives an organic, immersive feel to tools


code/game/objects/items.dm
```diff
@@ -490,22 +490,54 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 			fail_chance += round(H.shock_stage/120 * 40)
 			base_time += round(H.shock_stage/120 * 40)
 
+	//Precalculate worktime here
+	var/time_to_finish = 0
+	if (base_time)
+		time_to_finish = base_time - get_tool_quality(required_quality) - user.stats.getStat(required_stat)
```
spaceplaceexplorer: Can't this result in negative time?

NanakoAC: you are correct, will fix

code/game/objects/items.dm
```diff
@@ -490,22 +490,54 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
...
+				toolsound.stop()
+				toolsound = null
 			return TOOL_USE_CANCEL
 		else
 			target.used_now = FALSE
 
+	//Safe cleanup
+	if (toolsound)
+		toolsound.stop()
+	toolsound = null
```
LudwigVonChesterfield: Why do we null it even if it's already null?

NanakoAC: incase it's not already null
to ensure it can GC cleanly

LudwigVonChesterfield: Don't we... Uh. Check if it's null one line above?

PR #2250, opened 2018-10-17
GC: Includes coordinates and small fixes

LagannPM:
Does not fix rig's runtimes


code/controllers/subsystems/garbage.dm
```diff
@@ -166,6 +166,8 @@ SUBSYSTEM_DEF(garbage)
 				var/datum/qdel_item/I = items[type]
 				if(!I.failures)
 					log_to_dd("GC: -- \ref[D] | [type] was unable to be GC'd --")
+					if(I.coords)
+						log_to_dd("GC: -- \ref[D] | location is X=[I.coords.x_pos], Y=[I.coords.y_pos], Z=[I.coords.z_pos] --")
```
spaceplaceexplorer: This will output 2 lines instead of 1 for no reason. They should be merged.

code/modules/clothing/clothing.dm
```diff
@@ -12,6 +12,12 @@
 	var/list/valid_accessory_slots
 	var/list/restricted_accessory_slots
 
+/obj/item/clothing/Destroy()
+	for(var/obj/item/clothing/accessory/A in accessories)
+		qdel(A)
+	accessories = null
```
spaceplaceexplorer: Should this not be accessories.Cut() if the intent is to free up references?

code/game/objects/items.dm
```diff
@@ -59,9 +59,13 @@
 	var/list/item_icons = list()
 
 /obj/item/Destroy()
+	qdel(hidden_uplink)
+	hidden_uplink = null
 	if(ismob(loc))
 		var/mob/m = loc
 		m.drop_from_inventory(src)
+		m.update_inv_r_hand()
+		m.update_inv_l_hand()
```
LethalGhost: That update icon calls shouldn't be here.

code/modules/clothing/clothing.dm
LudwigVonChesterfield: We null the reference to the list itself. And the "Cut" part happens a row above.

LudwigVonChesterfield: Though I think there is a define like QDEL_LIST()

PR #2251, opened 2018-10-17
Wall icon runtime fix

### LINKED ISSUE #2145: Wall icon runtime
Happened while loading round, not good.
[16:34:33] Runtime in wall_icon.dm,53: list index out of bounds
  proc name: update icon (/turf/simulated/wall/update_icon)
  src:  (103,67,1) (/turf/simulated/wall/r_wall)
  call stack:
   (103,67,1) (/turf/simulated/wall/r_wall): update icon()
  the low wall (/obj/structure/low_wall): update connections(1, 0)

LagannPM:
fixes  #2145

### DIFF
--- a/code/game/turfs/simulated/wall_icon.dm
+++ b/code/game/turfs/simulated/wall_icon.dm
@@ -44,6 +44,8 @@
 	update_material()
 
 /turf/simulated/wall/update_icon()
+	..()
+	
 	if(!material)
 		return
 


NanakoAC:
can you explain why the bug happened and why a parent call fixes it?

LagannPM:
To be frank, i looked at baystation code and that's was the only difference, but after adding that there was error no more.
Maybe i've mistaken and there was no enough testing

PR #2255, opened 2018-10-21
Fix iffy code.

PJB3005:
DM allows some seriously dumb things to pass like defining ..() as a proc. SpacemanDMM (DM code parser) doesn't allow this for good reasons so this commit fixes that nonsense.

**I haven't tested this, because I'm not behind a Windows PC**


code/__marcos/subsystems.dm
```diff
@@ -4,9 +4,9 @@ Moved it here from machines subsystem, if to leave there, the macros won't be ac
 
 #define START_PROCESSING_IN_LIST(Datum, List) \
 if (Datum.is_processing) {\
-	if(Datum.is_processing != #Processor)\
```
NanakoAC: This line is pretty perplexing. What was #Processor before?

Replacing it with a hardcoded string is pretty suspect. It seems even wierder that the string specifically references machines. Things aside from machines can process

code/controllers/subsystems/machines.dm
```diff
@@ -3,35 +3,6 @@
 #define SSMACHINES_POWERNETS     3
 #define SSMACHINES_POWER_OBJECTS 4
 
-#define START_PROCESSING_IN_LIST(Datum, List) \
```
NanakoAC: Was this a duplicate define?

code/controllers/subsystems/machines.dm
```diff
@@ -3,35 +3,6 @@
...
-#define STOP_PROCESSING_IN_LIST(Datum, List) \
-if(Datum.is_processing) {\
-	if(SSmachines.List.Remove(Datum)) {\
-		Datum.is_processing = null;\
-	} else {\
-		crash_with("Failed to stop processing. [log_info_line(Datum)] is being processed by [is_processing] and not found in SSmachines.[#List]"); \
-	}\
-}
-
-#define START_PROCESSING_PIPENET(Datum) START_PROCESSING_IN_LIST(Datum, pipenets)
```
NanakoAC: I see these were nested defines, is that bad?

code/modules/shield_generators/shield_generator.dm
```diff
@@ -14,7 +14,7 @@
 	var/datum/wires/shield_generator/wires
 	var/list/field_segments = list()	// List of all shield segments owned by this generator.
 	var/list/damaged_segments = list()	// List of shield segments that have failed and are currently regenerating.
-	var/list/event_log() = list()				// List of relevant events for this shield
```
NanakoAC: oh hey this is my code. Those brackets were a typo, not sure why i never noticed it.

code/__marcos/subsystems.dm
PJB3005: @NanakoAC if you look at `MC.dm` there's `START_PROCESSING` defines that are similar to this one we're in right now. Except this one is specifically for specific sub-lists of `SSmachines`.

This line now matches the line below which is what it's supposed to be. I believe `#Processor` was copy pasted from `START_PROCESSING` (where it does make sense) and not replaced.

code/controllers/subsystems/machines.dm
PJB3005: Sure was.

code/controllers/subsystems/machines.dm
PJB3005: No problem with that.

code/modules/shield_generators/shield_generator.dm
PJB3005: You didn't notice because DM somehow has no problem parsing this mess.

code/controllers/subsystems/machines.dm
PJB3005: These are also duplicates.

PR #2257, opened 2018-10-22
Spider/roach refactoring part 1

spaceplaceexplorer:
Fixes multiple bugs with spider behavior.
Adds a new list helper proc for getting nearest objects.
Refactors roaches and spiders.

All webs and cocoons now have 5 health, all large cocoons have 20. Only 1 web and 1 cocoon allowed per tile. Spiders making a new cocoon on a tile now instead add objects into the old one.

Spiders no longer apply their toxins twice under certain circumstances, toxin type and amount is now properly applied.

Spiders sucking out fluids out of cocooned mobs now apply 100 brute to symbolize what is happening. Even without the 100 brute, they are very much dead already.

/obj/machinery/bot (mulebot) is now no longer a valid target as it was handled poorly and not even taking damage from their attacks.


code/modules/mob/living/superior_animal/giant_spider/ai.dm
```diff
@@ -1,161 +1,2 @@
-/mob/living/superior_animal/giant_spider/proc/FindTarget()
```
NanakoAC: All of this has been centralised into the main superior animal life, right?

code/modules/mob/living/superior_animal/giant_spider/giant_spider.dm
```diff
@@ -58,23 +57,22 @@
 	get_light_and_color(parent)
 	..()
 
-/mob/living/superior_animal/giant_spider/AttackingTarget()
+/mob/living/superior_animal/giant_spider/attemptAttackOnTarget()
 	var/target = ..()
+	if (target)
+		playsound(src, 'sound/weapons/spiderlunge.ogg', 30, 1, -3)
```
NanakoAC: stick a prob call around this, because hearing that sound every time is going to get pretty stale and lose the novelty

code/modules/mob/living/superior_animal/giant_spider/giant_spider.dm
```diff
@@ -98,66 +96,67 @@
 		if(busy == MOVING_TO_TARGET)
 			if(cocoon_target == C && get_dist(src,cocoon_target) > 1)
 				cocoon_target = null
-			busy = 0
-			stop_automated_movement = 0
+				busy = 0
+				stop_automated_movement = 0
 
 /mob/living/superior_animal/giant_spider/nurse/Life()
 	..()
```
NanakoAC: all of this stuff should really be in the AI file rather than life code

code/modules/mob/living/superior_animal/roach/life.dm
```diff
@@ -1,26 +0,0 @@
-/mob/living/superior_animal/roach/Life()
```
NanakoAC: Put this file back and leave it empty, it needs to be there as a place to put special life code

Also there was some, at least one roach regenerates chemicals over time. This file is the place for that, where is it now?

code/modules/mob/living/superior_animal/giant_spider/giant_spider.dm
```diff
@@ -168,32 +167,40 @@
...
+									O.loc = C
+
+								for(var/mob/living/M in targetTurf)
+									if((M.stat == CONSCIOUS) || istype(M, /mob/living/superior_animal/giant_spider))
 										continue
 									large_cocoon = 1
-									fed++
-									src.visible_message(SPAN_WARNING("\The [src] sticks a proboscis into \the [cocoon_target] and sucks a viscous substance out."))
+
+									if ((M.getBruteLoss() < 150) && istype(M, /mob/living/carbon))
```
NanakoAC: I am really not keen on this. The hardcoded check for 150, the direct application of 100 damage, it smells like bad code

Rather than directly applying damage, it would make more sense to drain blood, That's at least something more flexible and gracefully handled.

The way this is implemented doesn't sit right with me at all

code/modules/mob/living/superior_animal/giant_spider/ai.dm
spaceplaceexplorer: Yes.

code/modules/mob/living/superior_animal/roach/life.dm
spaceplaceexplorer: It is where it always was, in the file for the special roach. Since it is a property of only that one roach.

PR #2260, opened 2018-10-24
Spider/roach refactoring part 2

spaceplaceexplorer:
Spiders and roaches are now carbon and have everything that comes along with it, i.e. damage from environment / unsuitable air gas / body temperature. Everything is documented in superior_animal.dm

They can now gib or turn to dust under certain conditions, if damage dealt is sufficient or enough heat is applied. These are missing custom animations.

Fire burning overlay needs better icons, I've used the ones borgs have. They work good enough as placeholders.


NanakoAC:
Do chems work properly on them now?
Can they be poisoned
can they be melted with acid
can they be doused in fuel and set aflame?

Players keep asking for these features, they really want it

spaceplaceexplorer:
Yes to all of the above.

code/modules/mob/living/carbon/superior_animal/defense.dm
```diff
@@ -0,0 +1,286 @@
...
+			LAssailant = M
+
+			M.do_attack_animation(src)
+			playsound(loc, 'sound/weapons/thudswoosh.ogg', 50, 1, -1)
+			visible_message(SPAN_WARNING("[M] has grabbed [src] passively!"))
+
+			return 1
+
+		if (I_DISARM)
+			playsound(loc, 'sound/weapons/punchmiss.ogg', 25, 1, -1)
```
NanakoAC: Is it not possible for disarming to actually accomplish anything, like pushing the target away?

spaceplaceexplorer: I'm open to suggestions but pushing away will not accomplish anything due to how their current AI works, they would just immediately return to their previous tile. I can refine this in a future PR with some chance to make them bugger off for a bit so people can run away.

code/modules/mob/living/carbon/superior_animal/defense.dm
```diff
@@ -0,0 +1,286 @@
...
+		for(var/g in environment.gas)
+			if(gas_data.flags[g] & XGM_GAS_CONTAMINANT && environment.gas[g] > gas_data.overlay_limit[g] + 1)
+				pl_effects()
+				break
+
+	if(istype(get_turf(src), /turf/space))
+		if (bodytemperature > 1)
+			bodytemperature = max(1,bodytemperature - 10*(1-get_cold_protection(0)))
+
+		if (min_air_pressure > 0)
```
NanakoAC: ah a thought here

give roaches and spiders a fairly low minimum pressure.
They should still die in a vacuum,  but they should survive just fine in thin atmospheres that would suffocate a human

code/modules/mob/living/carbon/superior_animal/defense.dm
```diff
@@ -0,0 +1,286 @@
...
+	if (overkill_dust && (getFireLoss() >= maxHealth*2))
+		if (bodytemperature > max_bodytemperature*2)
+			dust()
+			return
+
+	if ((bodytemperature > max_bodytemperature) || (bodytemperature < min_bodytemperature))
+		adjustFireLoss(5)
+		updatehealth()
+
+/mob/living/carbon/superior_animal/handle_breath(datum/gas_mixture/breath)
```
NanakoAC: breathing code, oh boy

Are there any noticeable performance impacts from this stuff?
Perhaps as an optimisation, you could give nonhuman mobs a longer interval between breath checks. IIRC humans check every life ticks

code/modules/mob/living/carbon/superior_animal/giant_spider/attack.dm
```diff
@@ -0,0 +1,7 @@
+/mob/living/carbon/superior_animal/giant_spider/UnarmedAttack(var/atom/A, var/proximity)
```
NanakoAC: Is this proc called on humans? This wouldn't work for them (or maybe all carbon mobs) since they have multiple reagent holders. Chems need to be injected into bloodstr

code/modules/mob/living/carbon/superior_animal/life.dm
```diff
@@ -0,0 +1,86 @@
+/mob/living/carbon/superior_animal/Life()
+	. = ..()
+
+	objectsInView = null
+
+	if(client)
+		return .
```
NanakoAC: There's never any need to do "return ."

. contains the return value and will always be returned unless you specify "return somethingelse"

code/modules/mob/living/carbon/superior_animal/life.dm
```diff
@@ -0,0 +1,86 @@
...
+					var/moving_to = pick(cardinal)
+					set_dir(moving_to)
+					Move(get_step(src,moving_to))
+					turns_since_move = 0
+
+	//Speaking
+	if(speak_chance && prob(speak_chance))
+		visible_emote(emote_see)
+
+/mob/living/carbon/superior_animal/handle_chemicals_in_body()
```
NanakoAC: Did you test out chems in practise? Do all the medicines heal appropriately, the poisons damage, the acids burn, etc?

code/modules/mob/living/carbon/superior_animal/roach/blattedin.dm
```diff
@@ -0,0 +1,10 @@
+/datum/reagent/toxin/blattedin/touch_mob(var/mob/living/carbon/superior_animal/roach/bug, amount)
```
NanakoAC: Does this reviving actually work now?

code/modules/mob/living/carbon/superior_animal/roach/blattedin.dm
```diff
@@ -0,0 +1,10 @@
...
+	if(istype(bug))
+		if(bug.stat == DEAD)
+			if((bug.blattedin_revives_left >= 0) || prob(70))//Roaches sometimes can come back to life from healing vapors
```
NanakoAC: Nobody really got to test it before since it was broken, but these values are pretty silly as is

Change that || to && so that it's hard capped

code/modules/mob/living/carbon/superior_animal/defense.dm
spaceplaceexplorer: I've tried a hundred nurse spiders and checked profiler, it was much less demanding than the constant icon update cells in apcs were doing which nobody noticed.

code/modules/mob/living/carbon/superior_animal/giant_spider/attack.dm
spaceplaceexplorer: This is a proc that use used for unarmed attacks by player controlled mobs afaik. What I did was move all special attack info into this and had the AI utilize it for attacks instead.

code/modules/mob/living/carbon/superior_animal/life.dm
spaceplaceexplorer: Yes to all of that. Everything that works with carbon works with these. They do not have blood yet but I plan to add that in the future.

code/modules/mob/living/carbon/superior_animal/roach/blattedin.dm
spaceplaceexplorer: It should apply the same as acid spraying which I test so yes. Though I did not actually test the revival.

code/modules/mob/living/carbon/superior_animal/roach/blattedin.dm
spaceplaceexplorer: I didn't even notice that, will fix.

code/modules/mob/living/carbon/superior_animal/defense.dm
spaceplaceexplorer: Done

code/modules/mob/living/carbon/superior_animal/life.dm
spaceplaceexplorer: After further testing this is an issue with chem smoke ignoring mobs.

code/modules/mob/living/carbon/superior_animal/giant_spider/attack.dm
NanakoAC: but it contains a call just a few lines down ```L.reagents.add_reagent(poison_type, poison_per_bite)```

Who is that intended for? I don't see how it could function correctly in any circumstance

PR #2261, opened 2018-10-25
Fixes splashes calling touch procs incorrectly

spaceplaceexplorer:
Splash was calling touch procs twice for turfs and objects and it was not properly respecting the amount to be splashed.

This fixes issues such as fuel not applying fire stacks properly because it was always spilled on the ground and none remained to apply to the actual target.


NanakoAC:
I fixed this exact bug on aurora too

This is getting merged right now, and if baymed doesn't include a fix, we'll ensure this is preserved

PR #2262, opened 2018-10-25
Adds a new proc to objects - is_hot()

spaceplaceexplorer:
Objects now have a new proc `is_hot()` and var heat to determine whether they should ignite a mob upon hitting it.

This is checked when getting hit by a projectile in `bullet_act()`, melee in `apply_hit_effect()`, and thrown object in `hitby()`.

Added a new define for `HEAT_MOBIGNITE_THRESHOLD 400`, which is the minimum amount of heat needed to call `IgniteMob()` when hit. By itself, this does not cause fires, only if the target already has fire_stacks present.

All burn damage energy and beam projectiles cause such heat ignition, so do many items such as cigs, candles, matches, flares and welders.

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -472,10 +472,11 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 			return TRUE
 
 //Use this proc if you want to handle all types of failure yourself. It used in surgery, for example, to deal damage to patient.
-/obj/item/proc/use_tool_extended(var/mob/living/user, var/atom/target, base_time, required_quality, fail_chance, required_stat = null, instant_finish_tier = 110, forced_sound = null, var/sound_repeat = 2.5 SECONDS)	if (is_hot() >= HEAT_MOBIGNITE_THRESHOLD)
-	if (isliving(target))
-		var/mob/living/L = target
-		L.IgniteMob()
+/obj/item/proc/use_tool_extended(var/mob/living/user, var/atom/target, base_time, required_quality, fail_chance, required_stat = null, instant_finish_tier = 110, forced_sound = null, var/sound_repeat = 2.5 SECONDS)
+	if (is_hot() >= HEAT_MOBIGNITE_THRESHOLD)
+		if (isliving(target))
+			var/mob/living/L = target
+			L.IgniteMob()
 
 	if(target.used_now)
 		user << SPAN_WARNING("[target.name] is used by someone. Wait for them to finish.")
@@ -493,21 +494,53 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 			fail_chance += round(H.shock_stage/120 * 40)
 			base_time += round(H.shock_stage/120 * 40)
 
-	if(forced_sound != NO_WORKSOUND)
-		if(forced_sound)
-			playsound(src.loc, forced_sound, 100, 1)
-		else
-			playsound(src.loc, src.worksound, 100, 1)
+	//Precalculate worktime here
+	var/time_to_finish = 0
+	if (base_time)
+		time_to_finish = base_time - get_tool_quality(required_quality) - user.stats.getStat(required_stat)
 
-	if(base_time && (instant_finish_tier >= get_tool_quality(required_quality)))
+	if((instant_finish_tier < get_tool_quality(required_quality)) || time_to_finish < 0)
+		time_to_finish = 0
+
+
+	//Repeating sound code!
+	//A datum/repeating_sound is a little object we can use to make a sound repeat a few times
+	var/datum/repeating_sound/toolsound = null
+	if(forced_sound != NO_WORKSOUND)
+
+		var/soundfile
+		if(forced_sound)
+			soundfile = forced_sound
+		else
+			soundfile = worksound
+
+		if (sound_repeat && time_to_finish)
+			//It will repeat roughly every 2.5 seconds until our tool finishes
+			toolsound = new/datum/repeating_sound(sound_repeat,time_to_finish,0.15, src, soundfile, 80, 1)
+		else
+			playsound(src.loc, soundfile, 100, 1)
+
+	//The we handle the doafter for the tool usage
+	if(time_to_finish)
 		target.used_now = TRUE
-		var/time_to_finish = base_time - get_tool_quality(required_quality) - user.stats.getStat(required_stat)
+
 		if(!do_after(user, time_to_finish, user))
+			//If the doafter fails
 			user << SPAN_WARNING("You need to stand still to finish the task properly!")
 			target.used_now = FALSE
+			if (toolsound)
+				//We abort the repeating sound.
+				//Stop function will delete itself too
+				toolsound.stop()
+				toolsound = null
 			return TOOL_USE_CANCEL
 		else
 			target.used_now = FALSE
+
+	//Safe cleanup
+	if (toolsound)
+		toolsound.stop()
+		toolsound = null
 
 	var/stat_modifer = 0
 	if(required_stat)
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -461,7 +461,7 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 //Simple form ideal for basic use. That proc will return TRUE only when everything was done right, and FALSE if something went wrong, ot user was unlucky.
 //Editionaly, handle_failure proc will be called for a critical failure roll.
 /obj/item/proc/use_tool(var/mob/living/user, var/atom/target, base_time, required_quality, fail_chance, required_stat = null, instant_finish_tier = 110, forced_sound = null)
-	if (required_quality == QUALITY_WELDING)
+	if (ignites_with_touch)
 		if (isliving(target))
 			var/mob/living/L = target
 			L.IgniteMob()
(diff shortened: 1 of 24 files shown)


code/game/objects/items.dm
```diff
@@ -461,6 +461,11 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
 //Simple form ideal for basic use. That proc will return TRUE only when everything was done right, and FALSE if something went wrong, ot user was unlucky.
 //Editionaly, handle_failure proc will be called for a critical failure roll.
 /obj/item/proc/use_tool(var/mob/living/user, var/atom/target, base_time, required_quality, fail_chance, required_stat = null, instant_finish_tier = 110, forced_sound = null)
+	if (required_quality == QUALITY_WELDING)
```
NanakoAC: use your new variable here instead of qualities.

For example, a laser scalpel or an electric drill mitght also ignite

code/game/objects/items/weapons/tools/_tools.dm
```diff
@@ -217,31 +217,29 @@
 				spawn(100)
 					H.disabilities &= ~NEARSIGHTED
 
-//Prosthetic repair
-/obj/item/weapon/tool/attack(var/mob/living/carbon/human/H, var/mob/living/user)
 
-	if(get_tool_type(user, list(QUALITY_WELDING)))
-		if(ishuman(H))
-			var/obj/item/organ/external/S = H.organs_by_name[user.targeted_organ]
+/obj/item/weapon/tool/attack(mob/living/M, mob/living/user, var/target_zone)
```
NanakoAC: this section seems a bit out of scope, what's this all about?

code/game/objects/items/weapons/tools/weldingtools.dm
```diff
@@ -31,6 +31,7 @@
 		..()
 		damtype = BURN
 		force = force_ignited
+		ignites_with_touch = 1
```
NanakoAC: use TRUE/FALSE instead of 1/0

code/game/objects/objs.dm
```diff
@@ -12,6 +12,7 @@
 	var/damtype = "brute"
 	var/armor_penetration = 0
 	var/corporation = null
+	var/ignites_with_touch //if 1, ignites things when it hits them, does not add fire stacks by itself
```
NanakoAC: Give it an initial value of FALSE

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -290,6 +290,9 @@ meteor_act
 		var/obj/item/organ/external/affecting = get_organ(zone)
 		var/hit_area = affecting.name
 
+		if (O.ignites_with_touch)
+			IgniteMob()
```
NanakoAC: does this proc check the fire stacks?

code/game/objects/objs.dm
spaceplaceexplorer: Why?

code/game/objects/items/weapons/tools/weldingtools.dm
spaceplaceexplorer: The variable can be potentially used to represent temperature for ignition purposes. I use 1 and 0 everywhere for clarity since there is no boolean type in byond and it is misleading.

code/modules/mob/living/carbon/human/human_defense.dm
spaceplaceexplorer: IgniteMob only ignites when there are fire_stacks present.

code/game/objects/items/weapons/tools/_tools.dm
spaceplaceexplorer: The prosthetic repair was not resolving the attack properly and under certain circumstances just returning without feedback so I fixed the conditions and returns.

NanakoAC:
and flares

LudwigVonChesterfield:
Why not have a “can_ignite()” proc instead of a var you need to keep of all the time, toggle when lit/extinguished, and etc.?

spaceplaceexplorer:
How would you imagine such a proc to work? I'm not saying it's a bad idea.

spaceplaceexplorer:
Changed the var into a proc

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -16,17 +16,23 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 /obj/item/weapon/flame
 	var/lit = 0
 
+/obj/item/weapon/flame/ignites_with_touch()
+	if (lit)
```
LudwigVonChesterfield: Could write return lit instead.

code/game/objects/items/weapons/tools/weldingtools.dm
```diff
@@ -55,3 +55,7 @@
 	switched_on_qualities = list(QUALITY_WELDING = 40, QUALITY_CAUTERIZING = 15, QUALITY_WIRE_CUTTING = 15)
 	max_fuel = 40
 	force_ignited = WEAPON_FORCE_PAINFULL*1.15 //Slightly more powerful, not much more so
+
+/obj/item/weapon/tool/weldingtool/ignites_with_touch()
+	if (damtype == BURN)
```
LudwigVonChesterfield: Doesn’t it have a lit var?

spaceplaceexplorer: It does not.

spaceplaceexplorer:
All requested changes now included.

code/game/objects/items/weapons/cigs_lighters.dm
spaceplaceexplorer: It now returns heat.

LethalGhost:
Proc called `is_hot` should return TRUE/FALSE.
`get_temperature` is more suitable name

code/game/objects/objs.dm
```diff
@@ -12,8 +12,10 @@
 	var/damtype = "brute"
 	var/armor_penetration = 0
 	var/corporation = null
+	var/heat = 0
 
-
+/obj/proc/is_hot()
+	return heat
```
LethalGhost: That really should be `/obj` level proc/var?

spaceplaceexplorer: All objects can collide with mobs. Therefor all objects should have such property.

code/game/objects/items/devices/lighting/flare.dm
```diff
@@ -10,6 +10,7 @@
 	var/on_damage = 7
 	var/produce_heat = 1500
 	turn_on_sound = 'sound/effects/Custom_flare.ogg'
+	heat = 1000
```
NanakoAC: set to 1873, flares burn reallly hot

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -15,6 +15,11 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 //For anything that can light stuff on fire
 /obj/item/weapon/flame
 	var/lit = 0
+	heat = 1000
```
NanakoAC: 1670, temperature of a burning candle

code/game/objects/items/weapons/tools/weldingtools.dm
```diff
@@ -23,6 +23,8 @@
 	create_hot_spot = TRUE
 	glow_color = COLOR_ORANGE
 
+	heat = 3800
```
NanakoAC: 2250 for normal welder, 3773 for advanced one

code/modules/mob/living/living_defense.dm
```diff
@@ -50,6 +50,8 @@
 
 
 /mob/living/bullet_act(var/obj/item/projectile/P, var/def_zone)
+	if (P.is_hot() >= HEAT_MOBIGNITE_THRESHOLD)
```
NanakoAC: do projectiles work ? IIRC you said they didn't.

Can you ignite someone by dousing them in flame and firing a laser at them?

code/modules/overmap/ships/computers/sensors.dm
```diff
@@ -36,7 +36,7 @@
 		data["range"] = sensors.range
 		data["health"] = sensors.health
 		data["max_health"] = sensors.max_health
-		data["heat"] = sensors.heat
+		data["heat"] = sensors.current_heat
```
NanakoAC: what's the changes to this file all about?

code/modules/projectiles/projectile.dm
```diff
@@ -72,6 +72,10 @@
 	var/matrix/effect_transform			// matrix to rotate and scale projectile effects - putting it here so it doesn't
 										//  have to be recreated multiple times
 
+/obj/item/projectile/is_hot()
+	if (damage_type == BURN)
+		return damage*heat
```
NanakoAC: this will produce completely insane values, best not. Just go with 3800 as a flat value for lasers

code/modules/projectiles/projectile/beams.dm
```diff
@@ -14,6 +14,8 @@
 	tracer_type = /obj/effect/projectile/laser/tracer
 	impact_type = /obj/effect/projectile/laser/impact
 
+	heat = 200
+
 /obj/item/projectile/beam/practice
```
NanakoAC: zero out the heat for practise beams. they're basically just harmless holograms

code/modules/mob/living/living_defense.dm
spaceplaceexplorer: They work now and yes you can.

code/modules/overmap/ships/computers/sensors.dm
spaceplaceexplorer: That object was already using the heat war so I renamed it to current_heat instead. It was only used locally by that object.

code/modules/projectiles/projectile/beams.dm
spaceplaceexplorer: It is already 0 if they deal 0 damage.

PR #2271, opened 2018-10-29
Optimizes machines and processing subsystems

spaceplaceexplorer:
Instead of copying the entire processing list, machines and processing subsystems now use last position in their list to resume paused runs. A wrapper now moves the last position in list as needed when an object stops being processed.

This drastically reduces overhead for processing lists for these subsystems.

### DIFF
--- a/code/__defines/MC.dm
+++ b/code/__defines/MC.dm
@@ -32,6 +32,7 @@ if (Datum.is_processing) {\
 
 #define STOP_PROCESSING(Processor, Datum) \
 if(Datum.is_processing) {\
+	Processor.stopProcessingWrapper(Datum);\
 	if(Processor.processing.Remove(Datum)) {\
 		Datum.is_processing = null;\
 	} else {\
--- a/code/controllers/subsystems/machines.dm
+++ b/code/controllers/subsystems/machines.dm
@@ -23,7 +23,7 @@ SUBSYSTEM_DEF(machines)
 	var/list/power_objects = list()
 
 	var/list/processing
-	var/list/current_run = list()
+	var/nextProcessingListPosition = 0
 
 /datum/controller/subsystem/machines/PreInit()
 	 processing = machinery
@@ -33,6 +33,14 @@ SUBSYSTEM_DEF(machines)
 	setup_atmos_machinery(machinery)
 	fire()
 	..()
+
+/datum/controller/subsystem/machines/stopProcessingWrapper(var/D, var/list/L = processing) //called before a thing stops being processed
+	if (!nextProcessingListPosition) //0 position means subsystem is not paused
+		return
+	var/position = L.Find(D)
+	if (position)
+		if (position < nextProcessingListPosition) //queue moved, position moves with it
+			nextProcessingListPosition -= 1
 
 #define INTERNAL_PROCESS_STEP(this_step, check_resumed, proc_to_call, cost_var, next_step)\
 if(current_step == this_step || (check_resumed && !resumed)) {\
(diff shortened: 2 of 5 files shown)


NanakoAC:
this needs a whole lot more comments in your code. Make it simpler for future users please

code/__defines/MC.dm
```diff
@@ -32,6 +32,7 @@ if (Datum.is_processing) {\
 
 #define STOP_PROCESSING(Processor, Datum) \
 if(Datum.is_processing) {\
+	Processor.stopProcessingWrapper(Datum);\
```
NanakoAC: Add in a comment here too, explaining that this wrapper exists to update the pivot when things stop processing

code/controllers/subsystems/machines.dm
```diff
@@ -110,67 +118,111 @@ datum/controller/subsystem/machines/proc/setup_atmos_machinery(list/machines)
...
-			if(M.use_power)
-				M.auto_use_power()
+		nextProcessingListPosition = 1 //fresh start, otherwise from saved posisition
+
+	//localizations
+	var/list/local_list = machinery
+	var/obj/machinery/thing
+	var/wait = src.wait
+
+	var/tickCheckPeriod = round(local_list.len/16+1) //pause process every 1/16th length of list
```
NanakoAC: Recalculating this every time you go to process doesn't seem right. What if the list has changed since the last time it was calculated

code/controllers/subsystems/machines.dm
```diff
@@ -110,67 +118,111 @@ datum/controller/subsystem/machines/proc/setup_atmos_machinery(list/machines)
...
 		else
-			powernets.Remove(PN)
-			PN.is_processing = null
-		if(MC_TICK_CHECK)
+			local_list.Remove(thing)
+			thing.is_processing = null
+			nextProcessingListPosition-- //removing processed thing from list moves the queue to the left, adjust accordingly
+
+		if(!(nextProcessingListPosition%tickCheckPeriod)) //pauses every tickCheckPeriod-th processed thing
+			pause()
```
NanakoAC: Are you sure this pausing is beneficial? What's the goal of it?

code/controllers/subsystems/processing/processing.dm
```diff
@@ -7,33 +7,50 @@ SUBSYSTEM_DEF(processing)
...
 	var/process_proc = /datum/proc/Process
 
 	var/debug_last_thing
 	var/debug_original_process_proc // initial() does not work with procs
+	var/nextProcessingListPosition = 0 //position of next thing to be processed, inside of currently processed list
+
+/datum/controller/subsystem/processing/stopProcessingWrapper(var/D) //called before a thing stops being processed
+	if (!nextProcessingListPosition) //0 position means currently processed list is not paused or running, no point in adjusting last position due to removals from list
+		return
+	var/position = processing.Find(D) //find exact position in list
```
NanakoAC: Does this have potential be a net loss in performance? For example, if a big explosion causes many items to be deleted at once?

code/controllers/subsystems/machines.dm
spaceplaceexplorer: Sacrifices performance due to rescheduling for multiple smaller process calls, which makes for smaller perceived lag spikes.

code/controllers/subsystems/processing/processing.dm
spaceplaceexplorer: The deletions would have to happen during a process call or between rescheduling. Otherwise the proc returns at start due to no process running. It is unlikely to affect performance.

code/controllers/subsystems/machines.dm
spaceplaceexplorer: The tick period can be an arbitrary number and it would not produce anomalous behavior. This is merely a guide for how many times it's supposed to pause based on current length of list.

code/__defines/MC.dm
NanakoAC: Comment above the macro
This is the main entrypoint through which people will find this. a comment here is important

code/controllers/subsystems/machines.dm
ZVee: What about those 4 flags https://github.com/discordia-space/CEV-Eris/blob/master/code/__defines/MC.dm#L56 , is it necessary to add more methods to "pause" than adjusting flags?

spaceplaceexplorer: Idea was not to have it run a check on 8000 machines, and instead have it run only at most 16 times. CHECK_TICK might be just a relatively simple condition check, but it adds up into considerable overhead if it has to be done 8000 times.

ZVee: Okay.
Btw, we process every machines mostly not to process them, but for use_power proc to calculate power consumption on the maps (have you checked if there is any issues introduced with those pauses?).
If someone rework power consumtion calculations to be added and removed only when there is power change or dis/connections of machinery while normally we only work with cached value, then we don't need to add every machine into process list by default (it will become something similar as objects, when we only add things into process, where we actually need to process something), which should cut off needless processing by a lot i guess.
Also someone finally started to rework this on https://github.com/Baystation12/Baystation12/pull/23643 for areas, then it will be nice to do so for cables i think later.

spaceplaceexplorer: Reworking how power usage works is a good idea but not in scope of what I am doing in this PR, which is removing overhead from the subsystem itself rather than changing when and how things are processed.

Ultimately machines should be split into two categories, ones that are 'dumb' machines and only drain power with no other processing, and machines that actually require a process call periodically. Which is again outside of scope for this.

PR #2272, opened 2018-10-31
Multi z drawing rework

spaceplaceexplorer:
Added a new var to turfs `isTransparent`. This is used for multi z visibility to indicate a turf is see-through. Currently it is hardcoded to open and space turfs only.

Reworked how multi z drawing works. You can now see all the way down to the bottom from the top of the ship, with proper darkening applied. This works in space too. Mobs are now visible through space z levels.

Fixes some issues with spiders not calling updates for open spaces.
Ported toggle_move_stars from baystation12 which fixed ship transit changing only 1 z level.

spacecarp


code/modules/multiz/turf_icon.dm
```diff
@@ -1,102 +1,98 @@
...
+/turf/space/update_icon(var/updateAboveTurf = TRUE)
+	if (SSticker.current_state < GAME_STATE_PLAYING)
 		return
 
 	overlays.Cut()
 	var/turf/below = GetBelow(src)
-	if(below)
-		if(below.icon == 'icons/turf/space.dmi')
+	if (below)
+		if (below.is_space())
```
NanakoAC: it may be good to check if below.is_hole ?

It's set true for space and openspace tiles, providing an easy way to identify them. I don't see any usage of it here

NanakoAC:
what's the core of this optimisation? why is it faster, and any idea by how much?

spaceplaceexplorer:
It's about 10% faster. Previously it was needlessly creating the same image for each update, questionably populating `overlays` and ignored pixel positions/transformations. It also had duplicated code for space turfs and wasn't propagating updates upwards or drawing any objects floating inside open turfs below other open turfs.

code/modules/multiz/turf_icon.dm
NanakoAC: oh that is odd, you'd expect is_space to only return true for space tiles specifically. And if that's not what it does, then it seems you've actually changed functionality here, yes?
Can you have a little look around and see if there would be any consequences of making it check only for turf/space?

I implemented is_hole, ported from bay. It should generally be used in these performance-intensive situations, since checking a variable is faster than running a function. 
It's an expectation that any circumstance that would allow a turf to become a hole, the turf will update that var appropriately

PR #2273, opened 2018-10-31
removed logging of missing coords

LagannPM:
If container of another atoms is to be deleted they appear coordinateless.


NanakoAC:
Couldn't the whole function be better optimised too?

For example, by calling get_turf(src) and if that returns anything, then grab the xyz of the resulting turf

PR #2277, opened 2018-11-03
Storage HUD rework

spaceplaceexplorer:
Introduces a new HUD framework.

HUD framework is per client instead of per item with nested object oriented HUD elements, supports basic alignment functions, relative positioning and resizing, few features such as scaling to client view screen size or alignment relative to client view are still missing.

Storage HUD looks the same as before, but now utilizes this framework. Clicking on a storage item slot now clicks the item itself, making clicking small items easier.

Fixes numerous storage item bugs and refactors some internals for storage.dm
Storage items now have a new proc, `refresh_all` to refresh hud of all mobs that see the contents.


code/_onclick/hud/HUD_element/HUD_element.dm
```diff
@@ -0,0 +1,95 @@
...
+
+	if (_data)
+		_data.Cut()
+
+	vis_contents.Cut()
+
+	var/list/HUD_element/elements = getElements()
+	for(var/HUD_element/E in elements)
+		elements -= E
+		if (!QDELETED(E))
```
LudwigVonChesterfield: I don't think qdel() will attempt to delete if we're being qdeleted, so the check is slightl... Worhtless?

LudwigVonChesterfield: Okay, I am absolutely sure it won't attempt to qdel() what's already being qdeleted, but the check itself is faster than what happens in qdel() before checking for QDELETED. Sorry.

spaceplaceexplorer: You are correct, this was a leftover from when I was debugging an issue with destroy call being called twice. Removed it now.

PR #2278, opened 2018-11-04
Chem smoke now calls touch_mob on mobs

spaceplaceexplorer:
Fixes support roaches being unable to revive dead roaches.


code/game/objects/effects/chem/chemsmoke.dm
```diff
@@ -158,8 +158,10 @@
 		for(var/turf/T in targetTurfs)
 			chemholder.reagents.touch_turf(T)
 			for(var/atom/A in T.contents)
-				if(istype(A, /obj/effect/effect/smoke/chem) || ismob(A))
+				if(istype(A, /obj/effect/effect/smoke/chem))
 					continue
+				else if (ismob(A))
```
NanakoAC: wouldnt it be better to use trans_to instead of these checking mob and object types? IIRC that handles these checks

PR #2279, opened 2018-11-04
Fixes faulty memoization runtime in wall_icon

spaceplaceexplorer:
Fixes https://github.com/discordia-space/CEV-Eris/issues/2145

The only other cause for this runtime could be that `overlays.Cut()` is somehow responsible, but I cannot imagine how.

### DIFF
--- a/code/game/turfs/simulated/wall_icon.dm
+++ b/code/game/turfs/simulated/wall_icon.dm
@@ -47,8 +47,7 @@
 	if(!material)
 		return
 
-	if(!damage_overlays[1]) //list hasn't been populated
-		generate_overlays()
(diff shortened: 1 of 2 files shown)


NanakoAC:
This is not a good fix.
The wall overlays is a global list, it's supposed to be generated exactly once

The old approach had a check for it in every update icon as a hack to cover it not being properly set before. It is inefficient and needed fixed

Your fix makes the situation worse though. Now instead of a simple check of a list element, you've factored in a function call and a list copy (by returning damage overlays from that function) with every update

This is performance critical code and these changes make it less efficient. Even if it fixes a bug, i can't approve this solution

PR #2281, opened 2018-11-04
Modular Computers V2

NanakoAC:
The modular computers PR, now with several more passes of fixing and tweaking.

Only one remaining issue, cyrillic filenames don't work properly. Ghost is supposed to fix this but it doesn't seem to be getting done, and its not important enough to block merging


spaceplaceexplorer:
Cannot pick occupation in preferences. Only assistant button works.

PR #2290, opened 2018-11-07
The Engineer's Friend

NanakoAC:
This PR is the first step towards implementing makeshift tools. 
This first step is Duct Tape!

This PR implements tape as a fully featured tool, with QUALITY_ADHESIVE and QUALITY_SEALING. It has a new use type, stock, which is currently a non-refillable resource

Quality adhesive is barely used right now, but i plan to make extensive use of it in makeshift tool crafting recipes in a followup PR
Quality sealing is used for several repairing methods. Currently tape can be used to repair:
Punctured voidsuits
Damaged Floors
Damaged windows
Supermatter crystals

The last one is not a recommended application, but it might work under extreme circumstances
Duct tape is by far the best item for repairing punctured space suits, and it is advised for powergaming engineers to carry some around in case their suit takes a few holes. It's the only thing that can repair the suit while you're still wearing it, keeping some in your pocket might save your life

Also included are two variants of tape
Fiber tape is an advanced tool version, better all around. Found as rare loot in maintenance
Very related: https://www.youtube.com/watch?v=haPvuhznuyI

Web tape is a makeshift tool version, it's literally made from a bandage and a spiderweb. Bandages are made from any piece of clothing, spiderwebs can now occasionally be found in maintenance even when spiders aren't around. It needs to be constructed on the floor, as you have to drag and drop the web. but it works quite nicely

Duct tape itself, the base item, is now a common tool loot item, and it can also be printed with plastic at an autolathe. Emergency toolboxes also include some

Lastly, this also includes some tweaks to core tool usage. The consumption of power/fuel/stock is no longer a flat amount per operation, but is instead based on the actual time spent performing a tool operation. There is a small minimum value to this, but largely it means that the less time you spend working, the less resources will be consumed. Thus providing more value to technical skills and advanced tools

A small mapping change was necessary to fix mapped-in taperoll objects which existed, since they were retyped to a subtype of tool


code/game/objects/items/weapons/tools/_tools.dm
```diff
@@ -17,13 +17,66 @@
...
+	if (use_stock_cost)
+		stock = max_stock
+
+	update_icon()
+	return
+
+//For killing processes like hot spots
+/obj/item/weapon/tool/Destroy()
+	if (src in SSobj.processing)
+		STOP_PROCESSING(SSobj, src)
```
LudwigVonChesterfield: STOP_PROCESSING already checks if src is being processed.

code/game/objects/items/weapons/tools/_tools.dm
```diff
@@ -17,13 +17,66 @@
...
+	return ..()
+
+//Ignite plasma around, if we need it
+/obj/item/weapon/tool/Process()
+	if(switched_on)
+		if(create_hot_spot)
+			var/turf/location = src.loc
```
spaceplaceexplorer: Just use `get_turf(src)` for this whole section, you don't need any intricate location checking.
`#define get_turf(atom) get_step(atom, 0)`

code/game/objects/items/weapons/tools/_tools.dm
```diff
@@ -44,11 +97,244 @@
...
+
+	//Safe cleanup
+	if (toolsound)
+		toolsound.stop()
+		toolsound = null
+
+	var/stat_modifer = 0
+	if(required_stat)
+		stat_modifer = user.stats.getStat(required_stat)
+	fail_chance = fail_chance - get_tool_quality(required_quality) - stat_modifer
```
spaceplaceexplorer: Can't this result in negative `fail_chance`?

code/game/objects/items/weapons/tools/_tools.dm
```diff
@@ -44,11 +97,244 @@
...
+
+
+
+
+/******************************
+	/* Tool Failure */
+*******************************/
+
+//Critical failure rolls. If you use use_tool_extended, you might want to call that proc as well.
+/obj/item/proc/handle_failure(var/mob/living/user, var/atom/target, required_stat = null, required_quality)
```
spaceplaceexplorer: `required_stat = null` there is no need to assign null in arguments, they are null by default.

code/game/objects/items/weapons/tools/_tools.dm
```diff
@@ -44,11 +97,244 @@
...
+
+	var/crit_fail_chance = 25
+	if(required_stat)
+		crit_fail_chance = crit_fail_chance - user.stats.getStat(required_stat)
```
spaceplaceexplorer: Can this be negative?

PR #2292, opened 2018-11-07
Cyberplant colors

spaceplaceexplorer:
Cyberplants now pick a random color from a predetermined list when spawned or when walked over.
Clicking cyberplants with empty hand randomizes what type of plant is being displayed.

Fixes cyberplants not emitting any light. Light is now based on plant color.


code/game/objects/structures/cyberplants.dm
```diff
@@ -16,44 +17,136 @@
...
+	plant = prepare_icon(state)
+
+/obj/structure/cyberplant/proc/change_color(var/color)
+	if (!plant)
+		return
+
+	if(!color)
+		color = pick(possible_colors)
+
+	light_color = color
```
PJB3005: Editing `light_color` directly is bad. You should use `set_light(l_color=color)` instead.

spaceplaceexplorer: Done.

PR #2294, opened 2018-11-07
 Movement delay & smoothness refactor. 

PJB3005:
1: kills tick compensation. It was a broken system that could best be summed up as "old coders threw random formulas at movement code until they got a movement speed they were happy with".
2: Movement now uses glide size to be SMOOTH AS HELL.
3: client FPS had to be disabled because of https://secure.byond.com/forum/?post=2241289
4: rework client movement input code to use a server-side move loop. This improves responsiveness significantly.

Credits to /vg/'s coders for figuring out how to do all this. Most of this is ported from there.


code/__defines/misc.dm
```diff
@@ -208,3 +208,5 @@
 #define EMAIL_BROADCAST "broadcast@internal-services.net"
 
 #define LEGACY_RECORD_STRUCTURE(X, Y) GLOBAL_LIST_EMPTY(##X);/datum/computer_file/data/##Y/var/list/fields[0];/datum/computer_file/data/##Y/New(){..();GLOB.##X.Add(src);}/datum/computer_file/data/##Y/Destroy(){. = ..();GLOB.##X.Remove(src);}
+
+#define WORLD_ICON_SIZE 32
```
spaceplaceexplorer: This can be accessed with `world.icon_size`, why make it a define?

code/game/atoms_movable.dm
```diff
@@ -296,3 +299,69 @@
...
+		AM.set_glide_size(glide_size, min, max)
+
+
+//This proc should never be overridden elsewhere at /atom/movable to keep directions sane.
+// Spoiler alert: it is, in moved.dm
+/atom/movable/Move(NewLoc, Dir = 0, step_x = 0, step_y = 0, var/glide_size_override = 0)
+	if (glide_size_override > 0)
+		set_glide_size(glide_size_override)
+	
+	if (Dir & (Dir - 1))
```
spaceplaceexplorer: Please comment in code what this whole condition body is doing.

code/game/gamemodes/events/meteors.dm
```diff
@@ -268,7 +268,7 @@
 	z_original = z
 
 
-/obj/effect/meteor/Move()
+/obj/effect/meteor/Move(NewLoc, Dir = 0, step_x = 0, step_y = 0, var/glide_size_override = 0)
```
spaceplaceexplorer: Why set default values of arguments for all inherited `Move` procs? Shouldn't they utilize default value of parent when present for consistency?

code/modules/recycling/conveyor2.dm
```diff
@@ -73,11 +73,15 @@
...
 				if(A.loc == src.loc) // prevents the object from being affected if it's not currently here.
-					step(A,movedir)
+					if (debug_wait)
+						world << "[wait], [world.time], [DELAY2GLIDESIZE(wait)]"
+					step_glide(A, movedir, DELAY2GLIDESIZE(wait))
 					items_moved++
 			if(items_moved >= 10)
 				break
 
+/obj/machinery/conveyor/var/debug_wait = FALSE
```
spaceplaceexplorer: Leftover debug.

code/__defines/misc.dm
PJB3005: On /vg/ it's a define because of constant expressions. To be honest I flat forgot about `world.icon_size` when writing this so I'll replace this with that.

code/game/atoms_movable.dm
PJB3005: I didn't write this, I simply moved it from `mob_movement.dm` to here because it had no reason to be in `mob_movement.dm`.

I can still comment it if you want though.

code/game/gamemodes/events/meteors.dm
PJB3005: If you don't re-specify the default value of the parameter in the child, it'll be null inside that child (if not provided), while it'll be correct in the parent only. They're not inherited so yes they have to be re-specified.

rect_screenshot_temp

PJB3005:
Alright this should be good to go if nobody spots anything in review.

code/_helpers/delays.dm
```diff
@@ -0,0 +1,32 @@
+//////////////////////////////////
+// /vg/ MODULARIZED DELAYS - by N3X15
+//////////////////////////////////
```
NanakoAC: what is a delay controller? What's it for?

code/datums/observation/moved.dm
```diff
@@ -39,13 +39,13 @@ GLOBAL_DATUM_INIT(moved_event, /decl/observ/moved, new)
...
 // Entered() typically lifts the moved event, but in the case of null-space we'll have to handle it.
-/atom/movable/Move()
+/atom/movable/Move(NewLoc, Dir = 0, step_x = 0, step_y = 0, var/glide_size_override = 0)
 	var/old_loc = loc
 	. = ..()
 	if(. && !loc)
 		GLOB.moved_event.raise_event(src, old_loc, null)
 
-/atom/movable/forceMove(atom/destination)
+/atom/movable/forceMove(atom/destination, var/special_event, glide_size_override=0)
```
NanakoAC: special event?

code/game/atoms.dm
```diff
@@ -1,7 +1,7 @@
 /atom
 	layer = TURF_LAYER
 	plane = GAME_PLANE
-	appearance_flags = TILE_BOUND|PIXEL_SCALE
+	appearance_flags = TILE_BOUND|PIXEL_SCALE|LONG_GLIDE
```
NanakoAC: what does this long glide flag do?

code/_helpers/delays.dm
PJB3005: Basically just a thin wrapper to remove duplication with `world.time` based cooldowns. You can set a delay and then check later whether the controller is still blocked.

code/datums/observation/moved.dm
PJB3005: It's defined at the root level `forceMove` so I copied it over. Don't really know what it's for.

code/game/atoms.dm
PJB3005: `glide_size` controls the speed of gliding animations when gliding between two turfs happens. Problem is that the API is dumb and it's actually represented in pixels/frame, instead of something sane like the time the glide takes.

What this flag does is make diagonal glides take `sqrt(2)` times longer so that glide sizes work implicitly diagonally or cardinally.

code/modules/client/edge_sliding/move_loop.dm
```diff
@@ -0,0 +1,72 @@
+var/list/opposite_dirs = list(SOUTH,NORTH,null,WEST,null,null,null,EAST)
```
NanakoAC: a list already exists for this, reverse_dir i think

code/modules/mob/mob_movement.dm
```diff
@@ -180,28 +123,20 @@
...
-	//This doesn't entirely remove the problem, but it keeps travel times accurate to within 0.1 seconds
-	//Over an infinite distance, and prevents the inaccuracy from compounding. Thus making it basically a non-issue
-	var/leftover = world.time - move_delay
-	if (leftover > 1)
-		leftover = 0
+	if(mob.incorporeal_move && isobserver(mob))
+		Process_Incorpmove(Dir)
+		return
 
 	if(locate(/obj/effect/stop/, mob.loc))
```
NanakoAC: is this ever used anywhere? Could probably be removed

PR #2296, opened 2018-11-07
Tiger Tiger, Burning Bright

NanakoAC:
Buffs flamethrowers, after testing them out against plants in practise, they are found to be lacking
Now they have a gas mult which increases the amount of gas thrown (theoretically it fabricates gas out of nothing, but this is a common thing)

In addition, flamethrower range increased to 2 tiles, giving them much greater effectiveness against plant infestations. Perhaps even making them a viable weapon against other creatures

and removed some world debug messages that were left in before


code/game/objects/items/weapons/flamethrower.dm
```diff
@@ -167,12 +168,12 @@
 
 
 /obj/item/weapon/flamethrower/proc/ignite_turf(turf/target)
-	world << "Igniting turf [target.x] [target.y]"
 	//TODO: DEFERRED Consider checking to make sure tank pressure is high enough before doing this...
 	//Transfer 5% of current tank air contents to turf
 	var/datum/gas_mixture/air_transfer = ptank.air_contents.remove_ratio(0.02*(throw_amount/100))
-	//air_transfer.toxins = air_transfer.toxins * 5 // This is me not comprehending the air system. I realize this is retarded and I could probably make it work without fucking it up like this, but there you have it. -- TLE
+	air_transfer.multiply(gas_mult)
 	new/obj/effect/decal/cleanable/liquid_fuel/flamethrower_fuel(target,air_transfer.gas["plasma"],get_dir(loc,target))
```
spaceplaceexplorer: Move the multiplier here like this `air_transfer.gas["plasma"]*gas_mult`, no more actual gas fabrication.

NanakoAC:
theres some bug with it depositiing unignited fuel pools, investigating

code/game/objects/effects/decals/Cleanable/fuel.dm
```diff
@@ -6,66 +6,74 @@
 	anchored = 1
 	var/amount = 1
 
-	New(turf/newLoc,amt=1,nologs=0)
-		if(usr && usr.client && !nologs)
-			message_admins("Liquid fuel has spilled in [newLoc.loc.name] ([newLoc.x],[newLoc.y],[newLoc.z]) (<A HREF='?_src_=holder;adminplayerobservecoodjump=1;X=[newLoc.x];Y=[newLoc.y];Z=[newLoc.z]'>JMP</a>)")
-			log_game("Liquid fuel has spilled in [newLoc.loc.name] ([newLoc.x],[newLoc.y],[newLoc.z])")
-		src.amount = amt
+/obj/effect/decal/cleanable/liquid_fuel/New(turf/newLoc,amt=1,nologs=0)
```
NanakoAC: This is just me implementing coding standards all over this file. Proper indentation and full paths

code/game/objects/effects/decals/Cleanable/fuel.dm
```diff
@@ -6,66 +6,74 @@
...
+					other_fuel.Spread(exclude)
+			else
+				new/obj/effect/decal/cleanable/liquid_fuel(target, amount*0.25,1)
+			amount *= 0.75
+
+
+/obj/effect/decal/cleanable/liquid_fuel/flamethrower_fuel
+	icon_state = "mustard"
+	anchored = 0
+	var/turf/origin
```
NanakoAC: Actual functional changes start here

code/game/objects/effects/decals/Cleanable/fuel.dm
```diff
@@ -6,66 +6,74 @@
...
+	New(newLoc, amt = 1, d = 0, var/turf/_origin)
```
LudwigVonChesterfield: >This is just me implementing coding standards all over this file. Proper indentation and full paths

New proc declared in relative path.

code/game/objects/effects/decals/Cleanable/fuel.dm
```diff
@@ -6,66 +6,74 @@
...
-	flamethrower_fuel
-		icon_state = "mustard"
-		anchored = 0
-		New(newLoc, amt = 1, d = 0)
-			set_dir(d) //Setting this direction means you won't get torched by your own flamethrower.
-			. = ..()
+	for(var/d in list(turn(dir,90),turn(dir,-90), dir))
+		var/turf/simulated/O = get_step(S,d)
+		if (O == origin)
+			continue //No torching the user
```
LudwigVonChesterfield: If we have this check, why do we need to set dir in New, not in Initialize?

code/ZAS/Fire.dm
```diff
@@ -7,6 +7,7 @@ If it gains pressure too slowly, it may leak or just rupture instead of explodin
 */
 
 //#define FIREDBG
+#define MINIMUM_FUEL_VOLUME 0.0005 //Used to prevent leaving patches with astronomically tiny amounts of fuel
```
spaceplaceexplorer: This should be defined per reagent, in reagent define itself, so it can be extended to things like water.

code/game/objects/effects/decals/Cleanable/fuel.dm
spaceplaceexplorer: `amt` is not a good argument name and goes against guidelines for naming vars. Since you want to clean up the code, might as well change that.

code/game/objects/items/weapons/flamethrower.dm
```diff
@@ -167,18 +168,16 @@
...
 	//Transfer 5% of current tank air contents to turf
 	var/datum/gas_mixture/air_transfer = ptank.air_contents.remove_ratio(0.02*(throw_amount/100))
-	//air_transfer.toxins = air_transfer.toxins * 5 // This is me not comprehending the air system. I realize this is retarded and I could probably make it work without fucking it up like this, but there you have it. -- TLE
-	new/obj/effect/decal/cleanable/liquid_fuel/flamethrower_fuel(target,air_transfer.gas["plasma"],get_dir(loc,target))
+
+	new/obj/effect/decal/cleanable/liquid_fuel/flamethrower_fuel(target,air_transfer.gas["plasma"]*gas_mult,get_dir(loc,target), get_turf(src))
+
 	air_transfer.gas["plasma"] = 0
-	target.assume_air(air_transfer)
+	//target.assume_air(air_transfer)
```
spaceplaceexplorer: Not that I disagree with making flamethrowers not actually spew gas into air, but this makes `air_transfer` obsolete and should call `qdel` on it, no need to even remove plasma from it either.

code/game/objects/effects/decals/Cleanable/fuel.dm
NanakoAC: dir is for the direction it travels, it moves away from the user.
but this check is to prevent fuel from spreading back onto the user's tile

code/ZAS/Fire.dm
NanakoAC: out of scope as discussed

PR #2297, opened 2018-11-07
Puts an upper cap on fire_stacks burn temperature

spaceplaceexplorer:


code/modules/mob/living/living_defense.dm
```diff
@@ -302,7 +302,7 @@
 
 	//Scale quadratically so that single digit numbers of fire stacks don't burn ridiculously hot.
 	//lower limit of 700 K, same as matches and roughly the temperature of a cool flame.
-	return max(2.25*round(FIRESUIT_MAX_HEAT_PROTECTION_TEMPERATURE*(fire_stacks/FIRE_MAX_FIRESUIT_STACKS)**2), 700)
```
NanakoAC: You've put a cap on it, but i don't see any changes to the logic of how temperature is calculated.

Doesn't this just mean it will quickly reach a point of max temperature, and then fire stacks will continue increasing without any effect? 

That doesn't seem ideal behaviour. Wouldn't it be better to rescale things so that the new maximum temperature is reached at the same number of max stacks as before?

PR #2299, opened 2018-11-08
Fixes redundancy during QDELETED checks

spaceplaceexplorer:
QDELETED already does these checks.

`#define QDELETED(X) (!X || QDELING(X))`

In cases where istype(X) was used it actually hid runtimes instead of solving them. If any such occur we will now know something wrong is happening.


NanakoAC:
istype is used to simultaneously check for null, and check that the thing is of the correct type.  How was it hiding anything? Why are you removing it, this is not redundant. qdeleted doesn't do any verification of the object's type

code/_helpers/text.dm
```diff
@@ -483,7 +483,7 @@ proc/TextPreview(var/string, var/len=40)
 //Generates a clickable link which will jump the camera/ghost to the target atom
 //Useful for admin procs
 /proc/jumplink(var/atom/target)
-	if (!istype(target) || QDELETED(target))
```
NanakoAC: this is not redundant, it checks to ensure an atom (and not a simple datum) is passed

code/modules/mob/animations.dm
```diff
@@ -315,16 +315,16 @@ note dizziness decrements automatically in the mob's Life() proc.
 
 /atom/movable/proc/do_putdown_animation(atom/target, mob/user)
 	spawn()
-		if (!istype(src) || QDELETED(src))
+		if (QDELETED(src))
 			return
-		if (!istype(target) || QDELETED(target))
```
NanakoAC: same here, the istype exists for literally checking the type.

If you don't specify a type in the proc call, it implicitly uses the type the variable is declared with

istype(src) is bullshit though

code/_helpers/text.dm
spaceplaceexplorer: If you are passing it a simple datum, then the problem which needs to be addressed is not in this proc but in the proc calling it, and this proc will merely silence the issue.

If procs silence bad usage then it will never get fixed.

code/modules/mob/animations.dm
spaceplaceexplorer: Same as above.

spaceplaceexplorer:
I'm going to paste here original explanation:

If something returns false with `istype` then there are two options: it is null or it is in a place where it should not be.
There is no case where you want to silence an error with `istype`,
Because it hints at a greater problem with types being in places where you do not want them. Used as a silencer for incorrect proc usage does not fix problems, just hides them.

PR #2300, opened 2018-11-08
Minimal interface for tax program

LagannPM:
TaxQuickly 2559 is an uplink program, it doesn't have an interface only pop up input windows for input.
Ferociously clicking on it can cause pda interface to break.
Since TaxQuickly  is added randomly to every pda i added minimal interface for logging in.
Its still doesn't serve any purpose for non traitor.


NanakoAC:
PR description please

PR #2176, opened 2018-10-07
Inventory equip

LethalGhost:
Update `equip` part of inventory code. Don't confuse with `hands` part (put_in_hands, put_in_acvive_hand, drop_l_hand).
Basic procs:
- `equip_in_slot` - UNSAFE make all "equip works"
- `equip_in_slot_if_possible` - basic proc if you wanna propertly equip item
- `equip_in_slot_or_del` - equip_in_slot_if_possible + del if can't be equipped
- `equip_to_appropriate_slot` - equip_in_slot_if_possible + autoselect slot

BTW, touch hands, helpers parts.
All common changes and ideas described in [`code/modules/mob/inventory/_docs.dm`](https://github.com/discordia-space/CEV-Eris/blob/35a7ba3f6741812cf00044042b2879adb1141a5c/code/modules/mob/inventory/_docs.dm)
Check it (and it's changes if you wanna know what changes better).

Important change:
Remove `/obj/item/proc/mob_can_equip(Mob...)`
It code was splited into `mob/can_equip(Item, slot...)`, `Item/can_be_equipped(Mob, slot...)` and slot datums `/can_equip(Mob, Item...)`

Side changes:
move equip sound from `equip_to_slot_if_possible` to `pre_equip`
remove `can_pickup` proc, all functionality kind of included in `equip_to_slot_if_possible`

PR #2301, opened 2018-11-08
Inventory unequip

LethalGhost:
Analogue PR for ["Inventory equip"](https://github.com/discordia-space/CEV-Eris/pull/2176) but related to opposite act.

Most changes reflect one made there.

Updated inventory behavior:
- `u_equip` - analogue for `equip_to_slot`, make all "unequip works" (drop item layer, plane, screen_loc, remove item from user screen)
- `drop_from_inventory` - wrapped arond `u_equip` and handle item and world ineraction - set new loc for item, etc.
- `unEquip` - checks if item really can be unequipped and calls `drop_from_inventory`. Return FALSE if item can't be unequipped.

All procs can be called for any kind of mob without special checks. 
Trying to `drop_from_inventory` robot modules will cause runtime error.
But you can `drop_from_inventory`/`unEquip` items from robot grabber.

Related changes:
- `/obj/item/proc/mob_can_unequip` replaced with `/proc/mob_can_unequip` (analogue to `/proc/mob_can_equip`) which checked `Mob/canUnequip(Item,...)` and `Item.can_be_unequipped(Mob,...)`


NanakoAC:
we really need this, but is it actually ready for review? I see there's a conflict

code/modules/mob/living/carbon/human/inventory.dm
```diff
@@ -62,12 +62,27 @@ This saves us from having to call add_fingerprint() any time something is put in
...
 
-/mob/living/carbon/human/u_equip(obj/item/W as obj)
+/mob/living/carbon/human/u_equip(obj/item/Item)
+	if(src.client)
+		src.client.screen -= Item
+	Item.layer = initial(Item.layer)
+	Item.plane = initial(Item.plane)
+	Item.screen_loc = null
+
+	legacy_u_equip(Item)
```
NanakoAC: This is silly, why would we have a function that calls legacy_(itself)

LethalGhost: But that's legacy `u_equip` it will be removed lately, but that can't be done now.

NanakoAC: sure it can, just put that code back where it came from

Infact i've already done it, its in master now, so you can just completely erase your changes to that function and leave u_equip as is

code/game/gamemodes/endgame/supermatter_cascade/blob.dm
```diff
@@ -79,15 +79,18 @@
 
 	Consume(user)
 
-/turf/unsimulated/wall/supermatter/attackby(obj/item/weapon/W as obj, mob/living/user as mob)
```
NanakoAC: what on earth is this? A wall made of supermatter?

code/datums/craft/step.dm
```diff
@@ -231,8 +231,7 @@
 		return I
 
 	//The item is on the user
-	if (user.canUnEquip(I))
```
NanakoAC: Why switch this from a mob function to a generic one? What's the gain here, how does this help?

code/game/machinery/autolathe.dm
```diff
@@ -327,36 +327,27 @@
...
-		user << SPAN_NOTICE("You put \the [eating] into the autolathe.")
-		SSnano.update_uis(src)
+		if(user.unEquip(eating, src))
+			container = eating
+			user << SPAN_NOTICE("You put \the [eating] into the autolathe.")
+			SSnano.update_uis(src)
 
 /obj/machinery/autolathe/proc/eat(var/mob/living/user)
-	if(!istype(user))
+	if(!istype(user) || stat)
```
NanakoAC: Don't check stat directly, use incapacitated()

code/game/machinery/excelsior/implantmaker.dm
```diff
@@ -29,15 +29,15 @@
 		return
 
 	if(istype(I,/obj/item/weapon/implant) || istype(I,/obj/item/robot_parts) || istype(I,/obj/item/prosthesis))
-		user.remove_from_mob(I)
-		qdel(I)
+		if(user.unEquip(I))
+			qdel(I)
```
NanakoAC: This appears to be double indented? Does it compile?

code/modules/holodeck/HolodeckControl.dm
```diff
@@ -214,8 +214,7 @@
 	if(isobj(obj))
 		var/mob/M = obj.loc
 		if(ismob(M))
-			M.remove_from_mob(obj)
-			M.update_icons()	//so their overlays update
+			M.drop_from_inventory(obj)
```
NanakoAC: SAVEPOINT 1

code/datums/craft/step.dm
LethalGhost: Mob proc checks only mob stats but generic one checks both `mob.canUnequip()` and `item.canBeUnequipped()`

code/game/gamemodes/endgame/supermatter_cascade/blob.dm
LethalGhost: Funny. Can't unswer that question.

code/game/machinery/autolathe.dm
LethalGhost: That's machinery stat, not mob one.

LethalGhost: Well, may be I shouldn't put them together. That's what cause confuse.

code/modules/mob/living/carbon/human/inventory.dm
LethalGhost: But there is only half of changes in your PR, not icon handling, for example. 
I think there is no real reason to mess item variable changing with all that slots related code.
Code will be cleaner if we keep both procs for now. It's even not break compability with anything.

code/modules/mob/inventory/_docs.dm
```diff
@@ -57,6 +57,53 @@ Ovverided for human proc check "slot_back" first.
 
 
 
+
+// UNEQUIP //
+
+/mob/proc/u_equip(obj/item/Item)
+/*
+This is an UNSAFE proc.
+Inequip analogue for equip_to_slot(...)
```
NanakoAC: if its an opposite for another function, why not name them appropriately

code/modules/mob/inventory/_docs.dm
```diff
@@ -57,6 +57,53 @@ Ovverided for human proc check "slot_back" first.
...
+
+It calls:
+	- Mob.u_equip(Item)
+	- Item.forceMove(Target)
+*/
+
+
+/mob/proc/unEquip(obj/item/Item, atom/Target)
+/*
+This differs from drop_from_inventory(...) in that it checks if the item can be unequipped first.
```
NanakoAC: IF only this one checks, then how can drop_from_inventory be a safe proc?

And if this is the one that checks, and u_equip is the one that does the heavy lifting, why do we need the middleman of drop_from_inventory at all?

code/modules/mob/inventory/_docs.dm
```diff
@@ -86,7 +133,7 @@ Don't confuse with Item.can_be_equipped(Mob, slot, disable_warning)
 */
 
 
-/proc/mob_can_equip(mob/living/L, obj/item/Item, slot, disable_warning = FALSE)
+/proc/mob_can_equip(mob/living/Mob, obj/item/Item, slot, disable_warning)
```
NanakoAC: Removing the default value of false here is confusing, should revert that

code/modules/mob/inventory/_docs.dm
```diff
@@ -96,6 +143,34 @@ It calls:
...
 
 
+/mob/proc/can_unequip(obj/item/Item, slot, disable_warning = FALSE)
+/*
+Return TRUE if Mob can unequip Item
+Don't confuse with Item.can_be_unequipped(Mob, slot, disable_warning)
+*/
+
+
+/proc/mob_can_unequip(mob/living/Mob, obj/item/Item, disable_warning)
```
NanakoAC: mob_can_unequip and mob.can_unequip ? 

that's hugely confusing

code/modules/mob/inventory/hands.dm
```diff
@@ -0,0 +1,8 @@
+/mob/proc/get_all_hands()
+	return list()
+
+/mob/living/carbon/human/get_all_hands()
```
NanakoAC: is there a robot version for this which returns their module slots?

code/modules/mob/inventory/unequip.dm
```diff
@@ -0,0 +1,25 @@
...
+
+/mob/proc/drop_from_inventory(var/obj/item/Item, var/atom/Target = null, drop_flag = null)
+	src.u_equip(Item)
+
+	if(!Item || !Item.loc)
+		return TRUE // self destroying objects (tk, grabs)
+
+	if(!Target)
+		Target = src.loc
+		drop_flag = MOVED_DROP
```
NanakoAC: what is drop flag?
what possible value for it could be passed into this function?

code/modules/mob/living/carbon/human/inventory.dm
```diff
@@ -178,32 +164,31 @@ This saves us from having to call add_fingerprint() any time something is put in
 			return BP_R_ARM
 
 
-/mob/living/carbon/human/equip_to_slot(obj/item/W, slot, redraw_mob = 1)
+/mob/living/carbon/human/equip_to_slot(obj/item/Item, slot, redraw_mob = 1)
```
NanakoAC: why use Item here?

It's commonplace everywhere to use an uppercase I for a local variable like this. It prevents potential namespace pollution of more common words that are often used for object variables

code/modules/mob/living/silicon/robot/inventory.dm
```diff
@@ -290,9 +290,14 @@
...
-		return TRUE
-	if (istype(I.loc, /obj/item/weapon/gripper)) //Robots are allowed to drop the things in their gripper
-		return TRUE
-	return ..(I) //This will be false for things directly equipped
\ No newline at end of file
+/mob/living/silicon/robot/can_unequip(obj/item/I)
+	//Robots are allowed to drop the things in their gripper
+	if(I && istype(module_active, /obj/item/weapon/gripper))
+		var/obj/item/weapon/gripper/G = module_active
+		return I == G || G.wrapped == I
```
NanakoAC: I == G ?
That would allow the robot to drop the gripper itself. That is bad.

Also I used checking loc of gripper to account for possible buggy situations where something would cease to be in the gripped variable but still be held inside the gripper

code/datums/craft/step.dm
LethalGhost: No. I replaced one function with one complex which made of two.
`item.MobCanUnequip(...)` with `mob_can_unequip(Item, Mob,...)` which calls `Mob.canUnequip(Item, ...)` and `Item.canBeUnequipped(Mob,..)`

PR #2303, opened 2018-11-08
New sprites for crates.

SpadesSlick:
image
image
Made by Toriate and edited by Silentium.


NanakoAC:
After poking around with them a bit, i'm worried these crates are too garish
The brightness of the light parts seems like a bit much, as does the weight of their darker outlines. But i think they also look kind of flat at the front, they should maybe have a little shading lower down

The only one i think is quite right is the grey and red one in the lower picture, that white stripe looks like specularity
Has Wrotler had a look at these yet?

There's nothing fundamentally wrong with them, but i think a few more tweaks are needed

SpadesSlick:
I see. We can try something different a bit later then.

SpadesSlick:
PR will be redone.

PR #2320, opened 2018-11-09
Blob fix

NanakoAC:


NanakoAC:
oh fuck i accidentally closed this instead of merging it

PR #2322, opened 2018-11-09
Computers light

LagannPM:
Computers, PDAs, tablets etc. are now slightly glow
default
default


LethalGhost:
All OK, but can't light update using screen color of app not machinery itself?

PJB3005:
Those screenshots are way too bright. You should lower the light power, preferably a lot.

LagannPM:
@NanakoAC LED and screen is two completely different things, this light is cast by screen and it independent of LED. LED is more like i flashlight on back side of PDA

NanakoAC:
> 
> 
> @NanakoAC LED and screen is two completely different things, this light is cast by screen and it independent of LED. LED is more like i flashlight on back side of PDA

Theoretically they're different objects, but technically they both cast their light from the same source - the PDA object itself. 

Afaik, an object can only cast one light at a time

PJB3005:
> They are mounted on walls and half of this light is lost because of that

Uh, that's not how it works is it? They still emit that much light just on "one side" of the wall, which still isn't ideal.

LagannPM:
Ah, and also i wont see telescreens and consoles lights when power is out.

PR #2323, opened 2018-11-10
Fixes pre_equip delaying spawns

spaceplaceexplorer:


LethalGhost:
That's not seems right.

PR #2324, opened 2018-11-10
Space Equipping Fix

NanakoAC:
Turns out the isworn/isheld procs were flawed. 
By setting them to check for a zero equip slot, all problems are solved. The equip slot is always zero until after the first time an item is successfully equipped

### DIFF
--- a/code/game/objects/items/_inventory.dm
+++ b/code/game/objects/items/_inventory.dm
@@ -63,7 +63,7 @@
 
 /obj/item/proc/is_worn()
 	//If equip_slot is zero then it has never been equipped
-	if (!equip_slot)
+	if (equip_slot == slot_none)
 		return FALSE
 
 	if (ismob(loc))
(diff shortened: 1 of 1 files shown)


code/game/objects/items/_inventory.dm
```diff
@@ -62,11 +62,19 @@
 
 
 /obj/item/proc/is_worn()
+	//If equip_slot is zero then it has never been equipped
+	if (!equip_slot)
```
spaceplaceexplorer: `#define slot_none 0` already exists, use it.

code/game/objects/items/_inventory.dm
```diff
@@ -62,11 +62,19 @@
...
+		return FALSE
+
 	if (ismob(loc))
 		return !(equip_slot in unworn_slots)
 
 
 /obj/item/proc/is_held()
+	//If equip_slot is zero then it has never been equipped
+	if (!equip_slot)
```
spaceplaceexplorer: Same here.

PR #2331, opened 2018-11-10
Makes extinguisher stop making wet floors

spaceplaceexplorer:


LagannPM:
Is this really okay? Extinguishers were wetting floors since ever and its a part of it mechanic. You see, extinguishers are filled with water in ss13 instead of a gas and because of it they wet floors.
I tried to do this on other server and got a lot of hate for that.

NanakoAC:
i'm thinking a better approach would be to make slimes take damage from moving over wet floors

this is mostly about slime damage isn't it?

speaking of which, have you actually checked that water works on them? I think you broke it with your recent superior mob carbonisation

LagannPM:
Extinguishers should not be used agains slimes, there should be a chem weapon or firehose. Its dumb that you have to "extinguish" them

spaceplaceexplorer:
I did not touch slimes during any commit I have made in any PR, they are unrelated to superior_animals.

I do not care how other servers balance their extinguishers. Wetting entire hallway with one click, repeatedly, with a common tool is bad design and makes for terrible firefighting. The way slips work now they deny a whole hallway with no effort.

PR #2332, opened 2018-11-10
Inventory Deletion Fix

NanakoAC:
Fixes deleted items not being removed from your UI

### DIFF
--- a/code/modules/mob/living/carbon/human/inventory.dm
+++ b/code/modules/mob/living/carbon/human/inventory.dm
@@ -63,6 +63,13 @@ This saves us from having to call add_fingerprint() any time something is put in
 	return (O && !O.is_stump() && (!check_usablility || O.is_usable()))
 
 /mob/living/carbon/human/u_equip(obj/item/W as obj)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/inventory.dm
```diff
@@ -63,6 +63,13 @@ This saves us from having to call add_fingerprint() any time something is put in
 	return (O && !O.is_stump() && (!check_usablility || O.is_usable()))
 
 /mob/living/carbon/human/u_equip(obj/item/W as obj)
+	if(src.client)
+		src.client.screen -= W
+	W.layer = initial(W.layer)
+	W.plane = initial(W.plane)
+	W.screen_loc = null
```
LethalGhost: Values will be dropped twice. Here and in `mob/drop_from_inventory(...)`

NanakoAC: its fine, it's a temporary fix

Your PR will sort it properly

PR #2342, opened 2018-11-11
Flamethrower fixes and autolathe recipes

LagannPM:
Added recipes:
- Pneumatic crowbar
- Spraybottle
- Flamethrower
- Pill bottle

Removed obsolete circuits
Some minor fixes and tweaks to flamethrower
Added recipe for handcrafted flamethrower
Fixed uppercase letters in crafting messages


code/game/objects/items/weapons/flamethrower.dm
```diff
@@ -176,16 +169,15 @@
 	qdel(air_transfer)
 	//Burn it based on transfered gas
 	target.hotspot_expose((ptank.air_contents.temperature*2) + 380,500) // -- More of my "how do I shot fire?" dickery. -- TLE
+	return
 
-
-
+/obj/item/weapon/flamethrower/no_tank/New(var/loc)
```
NanakoAC: we don't need a no_tank version, the basic one already has no tank. That's why there's a /full version

LagannPM: Ahh, yes. I forgot about it. Before that there was also an igniter

PR #2347, opened 2018-11-11
PDA Label Fix

NanakoAC:
PDAs now properly name themselves with the owner's name and job when an ID is inserted.

This actually applies to ALL modular computers, including consoles and laptops. So you can see at a glance if a PC has an ID inserted, and whose


code/modules/modular_computers/computers/modular_computer/core.dm
```diff
@@ -236,6 +236,16 @@
 		update_icon()
 	return TRUE
 
+
+/obj/item/modular_computer/proc/update_label()
+	if (card_slot)
+		var/obj/item/weapon/card/id/I = card_slot.stored_card
```
spaceplaceexplorer: `/obj/item/modular_computer/GetIdCard()` is used instead of direct var access.

PR #2350, opened 2018-11-11
Event Threat Health balancing

NanakoAC:
This PR slightly reduces the power of three main event threats. Generally numerical adjustments are in the range of 15-25%, small steps. With one exception, the nurse spiders.

Generally, all three of these threats have their own gimmicks, of which being tanky is not an intrinsic part. These changes make them a bit more manageable without compromising their unique points

Spiders:
Max health reduced on the two tankier classes

Nurse spider toxin per bite greatly, since this was never really balanced before. 
Sleep toxin causes the victim to fall asleep with a dose of 5+ units. Nurses now inject 2 per bite, meaning they need three attacks to knock someone out. Fewer hits will just cause slowness and drowsines

Slimes:
Reduced health regen (including from feeeding) and max health of baby and adult classes

Blobs: 
Reduced health, resistance and regen of the normal blob parts. Larger Shield blobs are unchanged

### DIFF
--- a/code/game/gamemodes/events/blob.dm
+++ b/code/game/gamemodes/events/blob.dm
@@ -68,9 +68,10 @@
 	anchored = 1
 	mouse_opacity = 2
 
-	var/maxHealth = 30
+	var/maxHealth = 25
 	var/health = 1
-	var/brute_resist = 1.3
+	var/health_regen = 1.8
+	var/brute_resist = 1.25
 	var/fire_resist = 0.75
 	var/expandType = /obj/effect/blob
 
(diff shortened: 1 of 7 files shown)


code/game/gamemodes/events/blob.dm
```diff
@@ -68,9 +68,10 @@
 	anchored = 1
 	mouse_opacity = 2
 
-	var/maxHealth = 30
+	var/maxHealth = 25
 	var/health = 1
-	var/brute_resist = 1.3
+	var/health_regen = 1.8
```
NanakoAC: This was a hardcoded 2 before, now its a variable

PR #2353, opened 2018-11-12
Manifest Activity Tweaks

NanakoAC:
Primary change in this PR is to make the crew manifest update player status appropriately.

If a player has been gone (dead, afk or disconnected) for 15 minutes, they will be marked as Inactive on the manifest.
Note that this does not override any manually set status through medical records. If a status other than Active has been set (SSD, disabled, deceased, etc) that will show instead of inactive

If a player has been gone for more than 60 minutes, they are removed from the manifest entirely. This overrides everything. 
This is intended to stop the manifest getting clogged up with tons of long-gone players during a long round

These two times are just a starting point, and open to feedback

Other changes:
Mind datums now have a var that tracks last activity time for that character. It is set by the Inactivity subsystem as long as they have a connected client and are alive. It stops updating on death

Inactivity subsystem now runs regardless of AFK kick config settings, because of the above (it still wont kick people unless that is set though.

Some time defines were moved to a proper file in the __defines folder. They were previously defined in _helpers which was causing issues using them with earlier files in the helpers folder.

Entirely removed OOC manifest behaviour. Now players in the lobby see the same manifest as everyone else.
This behaviour only served to marginalize an already little-used feature, namely setting statuses through crew records

Anyways, i also plan to use some of this functionality in future for recycling job slots from players who are gone


NanakoAC:
Line 21, inactivity.dm

PR #2355, opened 2018-11-12
Tool Tweaks

NanakoAC:
Moderate reworking of the tool degradation feature in response to user feedback. Generally with the aim of making tool degradation more understandable, and less like your tools are cheap junk

Passive degradation (on normal use) majorly nerfed. Generally 50-70% less than before.  More advanced tools take less damage
Tools now get an instant hit of damage on any failure, this is in addition to the possiblity of crit failure that causes major damage.
Randomly spawned tools (in maintenance) can now start off in a mildly damaged state.  They have been abandoned and left to rust
This won't affect tools in toolboxes, vendors, created ones, starting gear, or those ordered at cargo.

Damaged tools now grant less materials when recycled at an autolathe
Autolathes can now store decimal quantities of material sheets, but will still only output whole sheets

Fixed a bug where you could make multiple taped copies of an item using duct tape


code/game/machinery/autolathe.dm
```diff
@@ -369,15 +369,16 @@
 	var/mass_per_sheet = 0 // Amount of material constituting one sheet.
 
 	for(var/obj/O in eating.GetAllContents(includeSelf = TRUE))
-		if(O.matter && O.matter.len)
-			for(var/material in O.matter)
+		var/list/_matter = O.get_matter()
+		if(_matter && _matter)
```
spaceplaceexplorer: This is a duplicate check. Do not check for `.len` before calling `for` either, it checks automatically.

code/game/objects/objs.dm
```diff
@@ -216,3 +216,9 @@
 	playsound(src.loc, 'sound/weapons/guns/interact/pistol_magout.ogg', 75, 1)
 	M << SPAN_NOTICE("You insert [I] into [src].")
 	return TRUE
+
+
+//Returns the list of matter in this object
+//You can override it to customise exactly what is returned.
+/obj/proc/get_matter()
```
spaceplaceexplorer: Many objects had matter set, this does not take it into account at all.

PR #2367, opened 2018-11-13
AFK Kick Fix

NanakoAC:
fixes players getting kicked for no reason


code/modules/client/client procs.dm
```diff
@@ -308,8 +308,9 @@
 //checks if a client is afk
 //3000 frames = 5 minutes
 /client/proc/is_afk(duration=3000)
-	if(inactivity > duration)
-		return inactivity
+	if (duration)
```
PJB3005: Isn't the problem here that something is calling `is_afk` with a duration of zero, instead of `is_afk` doing this check?

PR #2369, opened 2018-11-13
Floor Icon Refactor

NanakoAC:
This is the groundwork for a flooring project clock has me working on. 

What this PR does:

Deprecates several flags related to floor smoothing, instead replacing them with a more nuanced system. This system will be expanded in a couple days to include whitelists and blacklists

Sets up floors to only update once during roundstart, instead of the 9 times most floors would previously do. This will save on a vast quanttity of roundstart processing.

Seperates the concept of inner and outer corners on floors into two flags

Generally makes floor icon updating more efficient and cleaner

Switches maintenance floors to use a flat base icon for now. They will be worked on properly soon

Some other little things i forgot


code/__defines/turfs.dm
```diff
@@ -5,11 +5,16 @@
...
+#define TURF_IS_FRAGILE         1024
+#define TURF_ACID_IMMUNE        2048
+#define TURF_HIDES_THINGS		4096
+
+
+
+//Used for floor/wall smoothing
+#define SMOOTH_NONE 0	//Smooth only with itself
+#define SMOOTH_ALL 1	//Smooth with all of type
+#define SMOOTH_WHITELIST 2	//Smooth with a whitelist of subtypes
```
NanakoAC: implementing this in another PR shortly. This is one of three PRs ill be submitting this week about floors

code/game/objects/structures/low_wall.dm
```diff
@@ -39,18 +39,32 @@
...
+
 /obj/structure/low_wall/Destroy()
 	for (var/obj/structure/window/W in loc)
 		if (!QDELETED(W))
 			W.shatter()
 
+	//If we're on a floor, make it no longer be counted as a wall
+	var/turf/simulated/floor/T = loc
+	if (istype(T))
+		T.is_wall = FALSE
```
spaceplaceexplorer: This does not handle a case where there is something else on the turf that set `is_wall` to `TRUE`. To handle such a case, `is_wall` should be a numeric value and each atom that sets `is_wall` to `TRUE` would instead add `1` to it, and on destroy subtract `1` from it.

code/game/turfs/simulated/floor_icon.dm
```diff
@@ -21,107 +24,88 @@ var/list/flooring_cache = list()
...
 
 
-				if ((flooring.flags & TURF_USE0ICON) && has_border)
-					icon_state = flooring.icon_base+"0"
+			//We can only have inner corners if we're smoothed with something
+			if (has_smooth)
+				if(flooring.flags & TURF_HAS_INNER_CORNERS)
+
+					//Quick way to check if we're smoothed with both north and east
+					if((has_smooth & NORTHEAST) == NORTHEAST)
```
spaceplaceexplorer: `== NORTHEAST` is not needed. See above `if(flooring.flags & TURF_HAS_INNER_CORNERS)` for example on how to use binary ops to work with flags. This applies to rest of these conditions too.

code/game/objects/structures/low_wall.dm
NanakoAC: there are no such cases nor are any planned

code/game/turfs/simulated/floor_icon.dm
NanakoAC: it is needed, otherwise it returns true for either north or east. 
This check ensures its only true for both

code/game/objects/structures/low_wall.dm
spaceplaceexplorer: Is there anything preventing two low walls from being on a single turf?

NanakoAC: Now there is, yes. i pushed a fix to prevent multiple dense objects being crafted on the same tile

PR #2372, opened 2018-11-14
Ritual Refactor

NanakoAC:
Neotheology rituals are a horrible mess codewise.

How things currently work:
Just about any interaction with them generates a pile of new datums.
The cruciform stores a list of typepaths for the rituals it can cast. Every time you open a book, it creates a new copy of each of those rituals to take some values from it.
Every time you speak a ritual, it creates a copy of each of them again to test it. This is especially galling with group rituals

I have changed things.

Now, all the ritual datums are held in a single global list. Only one copy of each ritual exists, ever. The list is associative using the name as the key. Probably we should add some kind of ID instead but this works for now
The cruciform - and cruciform modules - each hold a list of ritual names, and get rituals by retrieving one from the all_rituals list using the name as the key.
Modules hold a list of typepaths for initialisation, and they generate their list of names by looking through the global list and finding subtypes of their whitelist.

This PR is by no means a final solution, but it's a giant step forward towards making this code be not so awful.
This may or may not fix the crashes with ritual books, but it will at least eliminate one possible cause

Working alone, i've not been able to test out group rituals yet. I'd l;ike to do so

### DIFF
--- a/code/_helpers/global_lists.dm
+++ b/code/_helpers/global_lists.dm
@@ -46,8 +46,8 @@ var/list/obj/item/device/uplink/world_uplinks = list()
 
 
 //Neotheology
-var/global/list/all_rituals = list() //List of all rituals
-var/list/global_ritual_cooldowns = list() // internal lists. Use ritual's cooldown_category
+GLOBAL_LIST_EMPTY(all_rituals)//List of all rituals
+GLOBAL_LIST_EMPTY(global_ritual_cooldowns) // internal lists. Use ritual's cooldown_category
 
 //Preferences stuff
 	//Bodybuilds
@@ -227,7 +227,7 @@ var/global/list/unworn_slots = list(slot_l_hand,slot_r_hand, slot_l_store, slot_
 
 		//Rituals which are just categories for subclasses will have a null phrase
 		if (R.phrase)
-			all_rituals[R.name] = R
+			GLOB.all_rituals[R.name] = R
 
 
 	return 1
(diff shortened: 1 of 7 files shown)


code/_helpers/global_lists.dm
```diff
@@ -44,6 +44,11 @@ var/global/list/poster_designs = list()
 // Uplinks
 var/list/obj/item/device/uplink/world_uplinks = list()
 
+
+//Neotheology
+var/global/list/all_rituals = list() //List of all rituals
+var/list/global_ritual_cooldowns = list() // internal lists. Use ritual's cooldown_category
```
seekin4u: "global ritual cooldown" list which is not global :ohmy:
Is it supposed to be non global?

NanakoAC: its declared in global scope, works fine,ehh

spaceplaceexplorer: Put this inside `/datum/ritual/` define as `var/global/list/global_ritual_cooldowns = list()`

NanakoAC: I just took it FROM there
Its a global list, global_lists.dm is the best place for it. it's the defacto location for global lists

spaceplaceexplorer: If you want it to be a global accessible to everything, why are you not using `GLOB.`?

NanakoAC: I actually don't know how to use glob, i wanted to

will look into it, ive also asked cups for help since he ported it

PR #2373, opened 2018-11-14
Removes observer respawn penalty

ghost:
Observer respawn penalty punishes only players that choose to observe at start of the round, or newly joined players that decide to observe. There is no reason to specifically target newly joined players or prevent people from playing for 30 minutes, just because they decided not to stare at a lobby screen.

If someone has concerns that this will lead to metagaming, it did not prevent metagaming in the first place, only delayed players from joining. Argument can be made that it discouraged metagaming, but again this is moot since it only applied to a very small subsection of players: newly joined and round start observers.

### DIFF
--- a/code/modules/mob/new_player/new_player.dm
+++ b/code/modules/mob/new_player/new_player.dm
@@ -106,8 +106,8 @@
 		new_player_panel_proc()
 
 	if(href_list["observe"])
-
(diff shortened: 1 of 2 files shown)


NanakoAC:
despite the creator of this PR being gone, i think it was a good idea. Is it still possible to merge it?

SpadesSlick:
Let's see how it will work in game. In best case we get even more people playing.

PR #2377, opened 2018-11-16
Material Sharding and Precision

NanakoAC:
Result of the discussion from two days ago, this PR does several things:

Autolathes now properly track materials down to 2 decimal places. Amounts smaller than 0.01 sheets are still chopped off, but this is 100x the precision we had before. Recipe costs are likewise properly represented.

Autolathes can now eject any amount of a material smaller than 1 unit, in the form of a material shard. 

Material shards now rotate randomly and scale visually based on the amount of material they have. It makes broken windows look far more interesting

In addition, bonus feature: Adds a Stack Split verb for stacks, allowing you to take any number from a stack and place it in your hands as a new stack. 
Technomancers can finally share and not just hog whole stacks each!


LethalGhost:
Category not set for new verb

code/game/objects/items/stacks/stack.dm
```diff
@@ -341,6 +344,28 @@
...
+
+
+	if (!Adjacent(usr))
+		usr << SPAN_WARNING("You need to be in arm's reach for that!")
+		return
+
+	var/quantity = input(usr,
+	"This stack contains [amount]/[max_amount]. How many would you like to split off into a new stack?\n\
+	The new stack will be put into your hands if possible", "Split Stack", round(amount * 0.5)) as num
+
```
LethalGhost: Safety checks missed after quantity input.

code/game/objects/items/stacks/stack.dm
```diff
@@ -341,6 +344,28 @@
 	else
 		return ..()
 
+//Verb to split stacks
+/obj/item/stack/verb/split_verb()
+	set src in view()
+	set name = "Split"
```
LethalGhost: Category set missed

code/game/objects/items/stacks/stack.dm
```diff
@@ -341,6 +344,28 @@
...
+	var/quantity = input(usr,
+	"This stack contains [amount]/[max_amount]. How many would you like to split off into a new stack?\n\
+	The new stack will be put into your hands if possible", "Split Stack", round(amount * 0.5)) as num
+
+	var/obj/item/stack/S = split(round(quantity, 1))
+	if (istype(S))
+		//Try to put the new stack into the user's hands
+		if (!(usr.put_in_hands(S)))
+			//If that fails, leave it beside the original stack
+			S.forceMove(loc)
```
LethalGhost: 1) `put_in_hands()` already drop item on floor under mob if it fails to equip item in hands
2) If mob use that verb while holding stack in hands new stack can stuck in him.

code/game/machinery/autolathe.dm
```diff
@@ -628,13 +630,19 @@
...
-		return
-	var/obj/item/stack/material/S = new M.stack_type(loc)
-	S.amount = eject
-	stored_material[material] -= eject
+	if (whole_amount)
+		var/eject = stored_material[material]
+		eject = amount == -1 ? eject : min(eject, whole_amount)
+		//We eject a number of sheets equal to the whole amount
+		var/obj/item/stack/material/S = new M.stack_type(loc)
+		S.amount = whole_amount
```
LethalGhost: `amount` can be passed as argument. 
It's better than direct var access.

code/game/machinery/autolathe.dm
```diff
@@ -615,11 +615,13 @@
...
 
 
+//Autolathes can eject decimal quantities of material as a shard
 /obj/machinery/autolathe/proc/eject(var/material, var/amount)
 	if(!(material in stored_material))
 		return
 
-	amount = round(amount) //The autolathe can contain less than a full sheet, but only ejects whole sheets
+	var/whole_amount = round(amount)
+	var/remainder = amount - whole_amount
```
LethalGhost: It's better to start actions after all check.
Especially "do actions with `amount`" before "check `amount` value"

code/game/machinery/autolathe.dm
```diff
@@ -1,5 +1,5 @@
 // makes sure that discounted prices from upgraded lathe no less than 1 unit.
-#define SANITIZE_LATHE_COST(n) max(1, n) // helps to fix prices where "* mat_efficiency" is used.
+
```
LethalGhost: 1) Why it was removed? 
2) Isn't it worth to add `round(..)` call here instead of copy-pasting values all over that file?

LethalGhost: I mean yeah values can now be less than 1, but that's not used as restriction from negative values too?

LethalGhost:
All in all. 
Why we add not-int values for material if we (not long ago) up all values to use only int for material interacitons?
There also exist few weak points:
- one can make shards from sheets but can't pick up shards with sheets
- one can make sheet (not shard) with less than 1 unit of material
- one can make 0.01 shards from sheets and convert them in 1 unit sheets (material fabric!)
- deconstruction of machineries return rounded amount of material
- why that affect only autolathe?

I can't see any usefull points.

code/game/objects/items/stacks/stack.dm
NanakoAC: safety checks are in the split function iirc. 
What kind of checks would you like therE?

code/game/machinery/autolathe.dm
NanakoAC: I don't see any way negative values could possibly happen. but i could put the round in there, sure

NanakoAC:
Why this PR?
The rounding based approach has a number of problems:
1. Our printing costs are based on the matter of objects * 1.25. This creates a lot of loss with rounding and in many cases it was the same price to print something as to recycle it
2. Material efficiency can be upgraded on autolathes with new parts. And this would not be reflected because all of your gains from efficiency would be lost to rounding errors.
3. I added a new feature recently where recycled tools would be worth less matter if they were damaged. This wasn't working correctly with the rounding.

We had a major and heated debate about this three days ago. A lot of people had a lot of ideas, but in the end this design is the only one clock signed off on. Using shards to represent <1 unit of matter gives players a tangible representation of small amounts, and makes it understandable to them


Yes you can't pickup shards with sheets that's intended. because then you'd end up with a sheet stack containing a decimal amount. Shards are pretty tiny, gather them up and take them to an autolathe for recycling

How do you make a sheet from a shard? That's not intended, i'll look into it. I definitely don't want to create material duplication bugs

Will fix the deconstruction too

What else should it affect aside from autolathe?

LethalGhost:
@NanakoAC 
> How do you make a sheet from a shard?

With welding tool.
[Here](https://github.com/discordia-space/CEV-Eris/blob/9c8b447514315e68ff9d21a3b131edae305c4111/code/game/objects/items/weapons/material/shards.dm#L81)

>What else should it affect aside from autolathe?

RnD machinery and mecha fabricators.

code/game/machinery/autolathe.dm
LethalGhost: mat_efficency count is thin place. There was situation ... in short if someone add more part to autolathe it mat_efficiency can become negative. It's count that way:
`mat_efficiency = 1.1 - man_rating * 0.1`

But you right. That's not that important point, if values go below 0 we'll notice it immediately.

NanakoAC:
Reworked shard welding. Now you can weld together all the shards in a tile to make as many sheets as possible. No materials are wasted or created from nothing, preservation of matter is observed

NanakoAC:
all requested changes are done, ready for review

code/game/objects/items/stacks/stack.dm
```diff
@@ -341,6 +344,34 @@
...
+	if (!usr.IsAdvancedToolUser())
+		return
+
+	if (!Adjacent(usr))
+		usr << SPAN_WARNING("You need to be in arm's reach for that!")
+		return
+
+	var/quantity = input(usr,
+	"This stack contains [amount]/[max_amount]. How many would you like to split off into a new stack?\n\
+	The new stack will be put into your hands if possible", "Split Stack", round(amount * 0.5)) as num
```
LethalGhost:
```suggestion
	The new stack will be put into your hands if possible", "Split Stack", round(amount * 0.5)) as null|num
```
That will add `cancel` button to input dialogue

NanakoAC:
fixed a safety check incase the stack gets moved
made sure destruction ejects things properly, and accounts for there being enough sheets for multiple stacjks

code/game/machinery/autolathe.dm
```diff
@@ -742,4 +755,3 @@
 #undef ERR_NOREAGENT
 #undef ERR_NOLICENSE
 
-#undef SANITIZE_LATHE_COST
```
LethalGhost: That line shouldn't be removed

PR #2378, opened 2018-11-16
Fixes storage numbered display

spaceplaceexplorer:
Storage bags, such as plant bags and trash bags, once again utilize numbered item display.

numbers

### DIFF
--- a/code/game/objects/items/weapons/storage/bags.dm
+++ b/code/game/objects/items/weapons/storage/bags.dm
@@ -18,7 +18,7 @@
 /obj/item/weapon/storage/bag
 	allow_quick_gather = 1
 	allow_quick_empty = 1
-	display_contents_with_number = 0 // UNStABLE AS FuCK, turn on when it stops crashing clients
+	display_contents_with_number = 1
 	use_to_pickup = 1
 	slot_flags = SLOT_BELT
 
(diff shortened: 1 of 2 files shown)


NanakoAC:
Can you make the size of tiles relative to the total size of each stack? That would be useful

spaceplaceexplorer:
No because it would require multiple rows for display and then one item type would end up displayed through two or more rows, having to split the number displayed, making it both ugly and unclear.

spaceplaceexplorer:
It cannot work in single row for the same reason storage space based inventory cannot work in single row past certain amount, in case of these bags the amount of items can reach a lot higher number than regular bags. It would extend the inventory way past screen limits if there were sufficient item types present, to prevent this they would have to be displayed clumped up, making numbers overlapped and unreadable.

In short, no, you are wrong.

NanakoAC:
relative, not absolute, everything can be rescaled to fit the space available
Having many types  of items with only 1 of each in the stack is of course a problem, but it'll be a problem here too. How does this system handle a bag containing 1 each of 50 different items?

NanakoAC:
Size relative storage would still be desireable, but it seems to be beyond scope for now. This PR is a net improvement and does fix some bugs with bag display, so we will merge it

PR #2379, opened 2018-11-18
Floor Icons 2: The Return

NanakoAC:
Second stage of requested floor icon work
maintfloor
This PR fully implements support for whitelists, blacklists, and the greylist (which is just using both lists at once).
This allows specific floor types to be used as exceptions to the smoothing rule

But more importantly it implements atom based smoothing
This has a flexible, if rather complex system, which matches atoms by type, and also by exact values of variables, as well as a priority value to govern how values in opposite lists override each other

Included is an example, implemented on plating, to make a floor not smooth with dense anchored objects, except airlocks


code/game/turfs/flooring/flooring.dm
```diff
@@ -55,16 +55,60 @@ var/list/flooring_types
...
+	//SMOOTH_NONE: Ignore all of type
+	//SMOOTH_ALL: Smooth with all of type
+	//SMOOTH_WHITELIST: Ignore all except types on this list
+	//SMOOTH_BLACKLIST: Smooth with all except types on this list
+	//SMOOTH_GREYLIST: Objects only: Use both lists
 
 	//How we smooth with other flooring
 	var/floor_smooth = SMOOTH_ALL
+	var/list/flooring_whitelist = list() //Smooth with nothing except the contents of this list
+	var/list/flooring_blacklist = list() //Smooth with nothing except the contents of this list
```
LethalGhost: Wrong comment

code/game/turfs/simulated/floor_icon.dm
```diff
@@ -132,28 +140,171 @@ var/list/flooring_cache = list()
...
+						if (!AM.is_between_turfs(origin, T))
+							continue
+
+						//From here on out, we do dangerous stuff that may runtime if the coder screwed up
+
+
+						var/match = TRUE
+						for (var/d_var in d_vars)
+							//For each variable we want to check
+							if (AM.vars[d_var] != d_vars[d_var])
```
LethalGhost: `AM.vars[d_var]` will runtime if `d_var` not in `AM.vars`

code/game/turfs/flooring/flooring.dm
```diff
@@ -55,16 +55,60 @@ var/list/flooring_types
...
+			This can be used, for example, to check that objects are dense and anchored
+			there are no safety checks on this, it will probably throw runtimes if you make typos
+
+	Common example:
+	Don't smooth with dense anchored objects except airlocks
+
+	smooth_movable_atom = SMOOTH_GREYLIST
+	movable_atom_blacklist = list(
+		list(/obj, list("density" = TRUE, "anchored" = TRUE), 1)
+		)
```
LethalGhost: Wouldn't it be easier if items order will define priority?

code/game/turfs/simulated/floor_icon.dm
```diff
@@ -132,28 +140,171 @@ var/list/flooring_cache = list()
...
+						if (match)
+							//If we've successfully found an atom which matches a list entry
+							best_priority = d_priority //This one is king until a higher priority overrides it
+
+							//And this is a whitelist, so this match forces is_linked to true
+							is_linked = TRUE
+
+
+		if (smooth_movable_atom == SMOOTH_BLACKLIST || smooth_movable_atom == SMOOTH_GREYLIST)
+			//All of this blacklist code is copypasted from above, with only minor name changes
```
LethalGhost: >code is copypasted from above

Ugh

code/game/turfs/simulated/floor_icon.dm
NanakoAC: I'm aware, the comment indicates as such.
But it will be in there, because the type is already checked beforehand.

code/game/turfs/flooring/flooring.dm
NanakoAC: sadly not, because the priority is primarily to determine prominence across both lists, not just within a single list

code/game/turfs/simulated/floor_icon.dm
NanakoAC: got an alternative idea?

code/game/turfs/simulated/floor_icon.dm
```diff
@@ -132,28 +140,171 @@ var/list/flooring_cache = list()
...
+				else if(floor_smooth == SMOOTH_BLACKLIST)
+					is_linked = TRUE //Default to true for the blacklist, then make it false if a match comes up
+					for (var/v in flooring_whitelist)
+						if (istype(t.flooring, v))
+							//Found a match on the list
+							is_linked = FALSE
+							break
+	else
+		//If this is a countercheck, we skip all of the above, start off with true, and go straight to the atom lists
+		is_linked = TRUE
```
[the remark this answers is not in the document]
NanakoAC: why do that? 
it doesn't really make much difference either way

NanakoAC:
New Commit up, fixes several requested changes and conflicts, rebases to master. 
Also fixed several visual bugs with grass and simplified external edge code. Now grass turfs have external corners, and that bug with lines on dirt is gone

code/game/turfs/simulated/floor_icon.dm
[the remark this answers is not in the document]
LethalGhost: Yeah but that structure will be easy to read and update. 
Structures like that is not good style and here it can be easily avoided:
```
if(condition 1)
	if(condition 2)
		if(condition 3)
			if(condition 4)
				if(condition 5)
					if(condition 6)
						if(condition 7)
							if(condition 8)
								return
							else
								expression 8
						else
							expression 7
					else
						expression 6
				else
					expression 5
			else
				expression 4
		else
			expression 3
	else
		expression 2
else
	expression 1

```

code/game/turfs/simulated/floor_icon.dm
```diff
@@ -132,28 +135,174 @@ var/list/flooring_cache = list()
...
-		var/turf/simulated/floor/t = T
-		if (flooring.floor_smooth || t.flooring.name == flooring.name)
-			is_linked = TRUE
+							//And this is a whitelist, so this match forces is_linked to true
+							is_linked = TRUE
+
+
+		if (smooth_movable_atom == SMOOTH_BLACKLIST || smooth_movable_atom == SMOOTH_GREYLIST)
+			//All of this blacklist code is copypasted from above, with only minor name changes
+			if (smooth_movable_atom == SMOOTH_BLACKLIST || smooth_movable_atom == SMOOTH_GREYLIST)
```
LethalGhost: That check is redundand.
default

code/game/turfs/flooring/flooring.dm
LethalGhost: But can't we merge lists then?
Remove most `SMOOTH_X` defines and update list strucure that way:
```
list_name = list(
	list(/obj, list(vars), BLACKLIST / WHITELIST)
)
```

Priority will be handled with oreder.
That also help remove copypasted code below.

code/game/turfs/flooring/flooring.dm
LethalGhost: But it still same as comment for whitelist. That's not right.

code/game/turfs/simulated/floor_icon.dm
LethalGhost: That's only part of problem.
That specified type can be change but changer will have no clue that it vars uses here.

code/game/turfs/flooring/flooring.dm
NanakoAC: ehh, i guess. maybe it'd be better that way
I really just want to get this merged and done with though

PR #2382, opened 2018-11-23
Serbian language for future antag use and some language tweaks.

SpadesSlick:
image

Nuke ops will use this later.


code/modules/mob/language/station.dm
```diff
@@ -97,4 +98,3 @@
 					 "portal", "systema", "electronika", "nigilizm", "anarhizm", "communistu", "tupuye", "sopla", "obshivka", "obtekaemost", \
```
NanakoAC: A lot of these syllables don't seem like syllables at all, but whole words

They're supposed to be little bits of words, pieced together to make a whole word.  Usually 2-4 letters
Six letters is kind of pushing it for a syllable, yet some of these have 11

i think they're too long, and so are the words that result from them

LudwigVonChesterfield: I think it's intended to be just words. If the spaces inbetween them would be 100% chance it would be fine. (Most are Russian words in transleet).

NanakoAC:
I thought i left a comment here, can't find it o.o

To repeat, i'm concerned about the syllable length. A syllable is supposed to be a piece of a word, typically 2-5 letters, 6 is pushing it. But this language has syllables of up to 11 letters, most of them seem to be whole words.

I don't think this is how the language system is supposed to be used. It's creating overly long compound words

I'm not exactly familiar with the serbian language irl, but from a bit of googling i found sites like http://www.novosti.rs/
It seems to employ a lot of shorter syllables too, lots of 1-4 letter words i'm not seeing there

In short, i think the syllables, especially the longer ones, should be cut up into several smaller parts

SpadesSlick:
Syllables were created with Chinese-like languages in mind, they won't work well with Europe ones. To make it work with Russian, you need to implement whole system with not one, but 5-6 independent lists of that. 

But we can counter that. Creators of original system include ``space_chance`` var that can define how often they are merged. It's that for 85% for those languages, and as you can see on the screen, it works fine. 

The purpose of this PR is to create new language for antags, not to create whole new system. We can have have nice results with current one, it's looks like funny gibberish, just like it should be. I don't see a reason to waste time on more complex system, you cannot get more realistic results without that, and it's generally out of the scope of PR.

PR #2387, opened 2018-11-27
Jetpacks and Z Travel

### LINKED ISSUE #1031: RIG jetpack doesn't let you use the "Move-Down" and "Move-Up" verbs to change Z-Levels
Literally that

NanakoAC:
This PR is a complete rewrite of everything related to transitioning between Z levels. Largely to make it not broken. Key features:
********************************************************
Jetpacks now work. Most of them do anyways, the following ones work perfectly:

- Normal back-mounted jetpacks
- Robot internal jetpacks
- Hardsuit maneuvring thrusters

Unfortunately, mecha thrusters do not work. I can and will fix them soon, but this PR has taken far too long and it needs to get merged
fixes #1031 
********************************************************
Robot jetpacks have been mechanically reworked. They are now a full fledged component (not installed by default) instead of an upgrade. Once installed they are used through verbs in the silicon commands menu, no more taking up an equip slot.
There are now no restrictions on installing it, a jetpack can be installed into any chassis. In addition the cost of it has been significantly reworked. it's still expensive but you can find all the materials on ship

The robot jetpack now has a built in air compressor. Whenever the user is in a pressurised environment (ie, inside the ship) it will start sucking air from the environment to refill itself.  This is fairly slow and has a power cost
********************************************************
New Jetpack FX, salvaged and reworked from existing sprites
JetFX

These properly represent the fact that our jetpacks work by emitting pressurised gas
The sprites are fully directional, appropriately offset, and thoroughly worked to ensure a thrust visual effect only happens when your jetpack actually uses up gas. Every thrust you see, some gas was used

In addition it includes a stabilize effect. When you stop moving, the jetpack does a little reverse thrust to stop your inertia, complete with a bit of recoil
********************************************************
Players can now transition up and down between zlevels through the use of four new FULLY ANIMATED methods:
Jetpacks
Jetpacks

Climbing:
Climbing
Can currently only be done in 0g, climbing requires you to find a wall that runs between both zlevels.
It requires no equipment but isn't very safe, you'll often slip out of control

Magboot Climbing
Magboots Magboots
A safer alternative to climbing, requires the same things (wall, 0g) and also requires you to have activated magboots. Faster and zero slipping

Jumping
jumping jumping
The least safe of all options but also the fastest. Just requires a floor under you, or a ceiling above you to go downwards. You need something to push off of
Only works in 0g for now. But in future i plan to add ways for certain special people to leap between floors, using hardsuit modules, cybernetics, antag powers, etc

********************************************************
Includes improved jetpack sprites, by shazbot
Jetsprites

********************************************************
Catwalks now properly act like floors in space and allow you to walk on them without (much) slipping
Jetpack gas costs rebalanced
Jetpacks added to random loot
Some tweaks to bluespacetech equipment

And probably a ton of other little things i've forgotten about


SpadesSlick:
Do we have any ways to craft or print jetpacks?

SpadesSlick:
Also, please give PR a normal name, I need to put something on the master commit name.

mattroks101:
Transitioning z-levels via jetpack seems incredibly slow.

NanakoAC:
> 
> 
> Transitioning z-levels via jetpack seems incredibly slow.

Moving upwards in an environment that has gravity, is the slowest possible transition and that's what's pictured there.

It goes much faster when moving downwards, or when the gravity is off

NanakoAC:
conflicts resolved, ready for review/approval

PR #2390, opened 2018-12-02
Rods now open craft menu on attack_self()

### LINKED ISSUE #2376: Rods and Grilles need to use crafting
When you attack_self on a stack of rods, it places a grille in your tile.

What it should do is open the crafting menu. And Grille should be added as a crafting option using rods

seekin4u:
Constructing "grille" is now happens with craft_menu, self_attack no
more make lattice under your legs.
Resolves #2376 .


seekin4u:
img

PR #2393, opened 2018-12-02
Rigs fix

### LINKED ISSUE #2386: Installed rig modules not showing visually
IHS hardsuit doesnt show the graphics for maneuvring thrusters, probably others are broken

seekin4u:
Overlays are now draw properly. You may add onmob or item (to rig in
inventar slot) overlay to any rig (if it accepts `rig_module` you want to install and "animate" with overlay).

See `/obj/item/rig_module` icon's definitions for HOWTO. Do not forget to add all sprites which you specify in `rig_module`, this is highly impoortant.

Also fixes #2386


code/modules/clothing/spacesuits/rig/modules/modules.dm
```diff
@@ -54,13 +55,37 @@
 
 	var/list/stat_rig_module/stat_modules = new()
 
+/obj/item/rig_module/New()
```
LethalGhost: It's better to call that in `Initialize()` instead of `New()`

code/modules/clothing/spacesuits/rig/rig.dm
```diff
@@ -517,27 +519,47 @@
...
+	add_item_overlay(chest)
 
 	if(wearer)
 		wearer.update_inv_shoes()
 		wearer.update_inv_gloves()
 		wearer.update_inv_head()
+		wearer.update_inv_wear_mask()
 		wearer.update_inv_wear_suit()
+		wearer.update_inv_w_uniform()
 		wearer.update_inv_back()
```
LethalGhost: Why so much overlays should be updated?

seekin4u: This is left after first iteration, will correct

code/modules/clothing/spacesuits/rig/modules/modules.dm
seekin4u: This is just bay's outdated proc, moved up becouse i like `New()`/`Destroy()` going right after object definition.

code/modules/admin/bluespacetech.dm
```diff
@@ -15,20 +15,26 @@ ADMIN_VERB_ADD(/client/proc/cmd_dev_bst, R_ADMIN|R_DEBUG, TRUE)
...
 
 
-	if(!check_rights(R_ADMIN|R_DEBUG))
-		return
+	if(!check_rights(R_ADMIN|R_DEBUG))	return
+
+
 
+
+	//I couldn't get the normal way to work so this works.
```
NanakoAC: what have you done to BSTs and why, this doesnt seem related to hardsuits

code/modules/admin/bluespacetech.dm
```diff
@@ -61,9 +69,9 @@ ADMIN_VERB_ADD(/client/proc/cmd_dev_bst, R_ADMIN|R_DEBUG, TRUE)
 	bst.regenerate_icons()
 
 	//Add the rest of the languages
+	//Because universal speak doesn't work right.
 	bst.add_language(LANGUAGE_COMMON)
 	bst.add_language(LANGUAGE_CYRILLIC)
-	bst.add_language(LANGUAGE_SERBIAN)
```
NanakoAC: this was just added

seekin4u:
I will revert it, just fucked up when pressed "update from master", probably i will have finished it by this evening.

seekin4u:
This is fine now, stealthy merge diff parts are gone, there are only mine changes now.

code/modules/clothing/spacesuits/rig/rig.dm
```diff
@@ -538,6 +534,30 @@
...
 	return
 
+//this makes your modules' overlays be drawn upon your rig's chest if it exists
+/obj/item/weapon/rig/proc/add_item_overlay(obj/item/I)
+	if(istype(I, chest))//e.g. we have parts which are supposed to be on our helmet, so we check
```
NanakoAC: Why does this check only chest and then stop?

Aren't there modules which add overlays to the helmet, as the comment indicates? Those wouldn't work with this code

code/modules/clothing/spacesuits/rig/rig.dm
```diff
@@ -538,6 +534,30 @@
...
+			for(var/obj/item/rig_module/module in installed_modules)
+				if(module.suit_overlay)
+					I.overlays += image("icon" = item_effect_icon, "icon_state" = module.suit_overlay_item, "dir" = SOUTH)
+
+/obj/item/weapon/rig/proc/get_mob_overlay(slot)
+	if(slot != slot_wear_suit_str || !active)//this means we should not show our modules till reg isn't engaged and charged
+		return
+
+	var/mob/living/carbon/human/H = wearer
+	if(H.body_build.index)//slim, fat, but not standart. Standart is already set in icons.
```
NanakoAC: standard*

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -703,23 +702,28 @@ var/global/list/damage_icon_parts = list()
...
-		// Accessories - copied from uniform, BOILERPLATE because fuck this system.
 		var/obj/item/clothing/suit/suit = wear_suit
 		if(istype(suit) && suit.accessories.len)
 			for(var/obj/item/clothing/accessory/A in suit.accessories)
 				standing.overlays |= A.get_mob_overlay()
 
-		overlays_standing[SUIT_LAYER]	= standing
+		if(back && istype(back, /obj/item/weapon/rig))//and the main part also needed
+			var/obj/item/weapon/rig/R = back//this is safecheck, again
+			if(R && R.active)//we are fully engaged and charged
```
NanakoAC: don't some modules have overlays when not active too?

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -767,22 +765,22 @@ var/global/list/damage_icon_parts = list()
 		if(back.item_state_slots && back.item_state_slots[slot_back_str])
 			overlay_state = back.item_state_slots[slot_back_str]
 
-
 		//apply color
 		var/image/standing = image(icon = overlay_icon, icon_state = overlay_state)
 		standing.color = back.color
 
-		//create the image
+		if(back)//rig's sprite by itself, then moduls
```
NanakoAC: modules*

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -767,22 +765,22 @@ var/global/list/damage_icon_parts = list()
...
+			if(istype(back, /obj/item/weapon/rig))
```
NanakoAC: This is bad. In future i'm planning to add rigs that are worn on slots other than the back. This is why rigs have a var to indicate which slot they should be worn in.

Specific case code like this in update inv procs is also generally bad.
This should call a generic proc on the worn item, and that item should make its own overlays as needed

code/modules/mob/living/carbon/human/update_icons.dm
LethalGhost: BTW.
That's check rig activation not modules.

code/modules/mob/living/carbon/human/update_icons.dm
seekin4u: There was the same code before, if you know how to evade checks  like this, and make this proc clearer - tell me, i'm already bored with all this trashy and old code which i wanted to "refactor" but discarded all changes becouse firstly i want this PR.

NanakoAC: Call a generic function

if (back)
back.apply_overlays(standing)

This icon updating code is generic, it needs to not have special cases like this in it

seekin4u: In this case i rerally don't like to make the only one function for one-time use, as for me it is clearer to take one string in this place. Btw `if(back)` is really unnecessary here, will fix soon.

code/modules/clothing/spacesuits/rig/rig.dm
seekin4u: Okay i got it.
It is better now to left this part. The reason of it - each `rig_module` only has `var/obj/item/weapon/rig/holder` as a reference to it's "owner", it isn't attached to a specific part - a hand or a helmet, so i can't fugire out how to correlate rig_module (e.g. flashlight) to rig's helmet.
Well, no, i can always add some little code to EACH rig_module to ONLY know where it should be drawn and make it more complicated, but this is just a liiitle unnecessarry feature which makes me to do extra changes.

NanakoAC: it would be a good feature, please do it

a spike coming from the wrist should be an overlay on the gloves, not on the suit

PR #2394, opened 2018-12-03
Bullet hole effect

mattroks101:
Adds a bullet hole effect. Ported from InterBay. Different strength bullets and lasers will leave different bullet holes. May be hard to see depending on the walls.
image
image

The reason I did it this way instead of as an overlay is walls continuously clear and re-apply their overlay because of their damage system. Without somehow changing that this was the only way I could think to add a bullet hole effect. Maybe someone else here can think of a better way of doing it.


SpadesSlick:
Can you remove them while repairing walls? Also, rubber ammo commonly used on our ship by civilians and Ironhammer personal. I don't think rubber ammo should left those kind of holes.

mattroks101:
Rubber bullets leave the really light mark, the dent one, because they deal way less damage than regular bullets.
image

Also yeah I'll make them clear when you repair walls too.

code/game/turfs/simulated/walls.dm
```diff
@@ -209,14 +214,43 @@
...
 
+	if(bulletholes >= 30)
+		clear_bulletholes()
+
+	var/obj/effect/overlay/bmark/BM = new(src)
+
+	BM.pixel_x = p_x
+	BM.pixel_y = p_y
+
+	if(decaltype == 1)
```
LethalGhost: It's better to use defines instead of magic numbers

NanakoAC:
this is a great idea

LethalGhost:
BTW, is it possible to handle hole icon with hole code? If so it will be possible to reuse that holes with other objects.

code/game/turfs/simulated/walls.dm
mattroks101: Got rid of it altogether.

mattroks101:
I don't know. Maybe? What else needs bullet holes?

NanakoAC:
low walls should have hole support too.
And tables, maybe even floors
Barricades

none of these are high priority though. Is it difficult to extend bullet hole support to other objects?

mattroks101:
Hmmm... perhaps it could be just made into a more generic structure proc or something?

mattroks101:
I don't think I understand what you're saying. Do you want me to move all of this stuff up to like the atom level so that it can be used on turfs or structures?

LethalGhost:
@mattroks101 
I suggested to move that lines:

into `/obj/effect/overlay/bmark` code.
To keep icon update with the rest of b-holes mechanics.

code/game/turfs/simulated/walls.dm
```diff
@@ -209,14 +213,34 @@
 			projectile_reflection(Proj)
 			return PROJECTILE_CONTINUE // complete projectile permutation
 
-	//cut some projectile damage here and not in projectile.dm, becouse we need not to all things what are using get_str_dam() becomes thin and weak.
+	//cut some projectile damage here and not in projectile.dm, because we need not to all things what are using get_str_dam() becomes thin and weak.
```
NanakoAC: this entire comment makes no sense. As long as you're correcting it, maybe rewrite it

code/game/turfs/simulated/walls.dm
```diff
@@ -209,14 +213,34 @@
...
+	//Actually adding the bullet hole effect now.
+	var/obj/effect/overlay/bmark/BM = new(src)
+
+	BM.pixel_x = p_x
+	BM.pixel_y = p_y
+	// offset correction
+	BM.pixel_x--
+	BM.pixel_y--
+
+	if(Proj.damage >= 20)//If it does a lot of damage it makes a nice big black hole.
```
NanakoAC: use a define for this value. like WEAPON_FORCE_PAINFULL

code/game/turfs/simulated/walls.dm
```diff
@@ -209,14 +213,34 @@
...
-	//cut some projectile damage here and not in projectile.dm, becouse we need not to all things what are using get_str_dam() becomes thin and weak.
+	//cut some projectile damage here and not in projectile.dm, because we need not to all things what are using get_str_dam() becomes thin and weak.
 	//in general, bullets have 35-95 damage, and they are plased in ~30 bullets magazines, so 50*30 = 150, but plasteel walls have only 400 hp =|
 	//but you may also increase materials thickness or etc.
 	proj_damage = round(Proj.get_structure_damage() / 3)//Yo may replace 3 to 5-6 to make walls fucking stronk as a Poland
 
 	//cap the amount of damage, so that things like emitters can't destroy walls in one hit.
 	var/damage = min(proj_damage, 100)
 
+	if(bulletholes >= 30)
```
NanakoAC: Why clear holes if they're over 30?
Wouldn't it make more sense only to clear the single oldest one, or to not add new ones, or even just to let them keep building up infinitely ?

code/game/turfs/simulated/walls.dm
mattroks101: I have no idea what it even means I just force of habit corrected the one misspelling I noticed. Someone who actually knows what this is trying to say should rewrite it.

code/game/turfs/simulated/walls.dm
mattroks101: Do we have defines for this sort of thing already?

code/game/turfs/simulated/walls.dm
mattroks101: Well it makes a new object for every hole so I figured an infinite amount of objects would probably lag the game. It's a pretty arbitrary number but I have never been able to even see it get to thirty without the wall breaking.

code/game/turfs/simulated/walls.dm
NanakoAC: We do indeed. Defines/weapons.dm

NanakoAC:
Whats your last commit about, RE refactoring?

Is this PR ready for farther review?

mattroks101:
I stuck it on the atom level as requested.

RandomColonist:
If possible, as a suggestion, maybe add forensic interaction for it to be used by detective? For example, you will be able to identify caliber/gather residue from bullet hole or identify laser model looking on laser hole.

SpadesSlick:
Adding holes to more structures is out of scope for this PR, and should be done on different one. Everything else seems fine.

PR #2398, opened 2018-12-04
reduced power consumption of the drill

Slepoi:


NanakoAC:
Balance changes should provide some justification. And generally change by up to 15% at a time unless you can provide a case for majorly restructuring a mechanic

NanakoAC:
farthermore, its worth noting that many of our devices consume power faster than on other servers, as we care more about battery scarcity and encouraging seeking of higher quality cells

Slepoi:
I tested it, everything turned out fine.
The player used a small battery and it worked for 15 seconds so he thought it was a mistake.

NanakoAC:
Ah well there is an issue to fix there then. Those massive drills are made for large power cells, they shouldn't accept a small or medium sized one. Can you make a PR to fix that?

PR #2401, opened 2018-12-05
Terahertz Scanners v2.0

NanakoAC:
This PR completely rewrites the T-Ray scanner basically from scratch.
Developed in response to player complaints about how useless this scanner was, and in preparation for its upcoming new functionality as the Exterminator's Best Friend for locating roach burrows. (They're not in this PR, coming soon)

For those not aware, TRay is the scanner that allows you to see underfloor pipes and wires. And burrows soon.

New Features:
Smooth updating and no flickering
Smooth
The overlay now updates smoothly as you move around, instead of flickering on and off. The flicker effect is completely gone.

Customiseable range and power costs
https://images-ext-1.discordapp.net/external/mYR9EMUm92s8jmLULl1YyVdvUpvmygQ5X_ANQr1mURQ/https/i.imgur.com/eaSli5Y.mp4
The scanner now has a selectable range, it can have a radius from 1 to 7 tiles
The power costs scale up modestly with the range. Yes power costs are now a thing, important balancing component. It takes any small cell like most devices

Ignore the power valuies in that video, not representative of final product
A normal cell provides enough power for 3-11 minutes of constant use, depending on power settings. You can of course replace it with a better cell if you want more

Nano interface
UI
Attackself now opens a shiny new Nano UI window that allows you to adjust range, view remaining power, and also a clever little estimate of remaining lifetime based on current power setting

Note that it can still be toggled with a single click, altclicking the scanner will do that without opening any UI windows

Performance improvements:
This scanner was huuuugely laggy. I've improved it in so many ways. Quite probably it will still have a significant performance cost, but it should be less than before. Performance related features:
-Drawing the overlays now makes use of mutable appearances which should cut down on the cost of creating images. This also solves all display related bugs like only seeing some parts of a pipe.
-The scanner uses a Moved observation registry and only updates when the user moves or changes a setting. No more updating overlays every second
-Limitations: The scanner now only operates when held in a hand. It will still remain turned on in any other location (and consume power) but it will not waste any script time drawing overlays when its in a pocket, container, or on the ground etc. 
Combined with the new power cost, this means no more just turning it on and leaving it in your pocket running forever. Use it when you need it


code/game/objects/items/devices/t_scanner.dm
```diff
@@ -10,53 +16,71 @@
...
+	var/suitable_cell = /obj/item/weapon/cell/small
+	var/active_power_usage = 25 //Watts
+
+	var/turn_on_sound = 'sound/effects/Custom_flashlight.ogg'
+
+	/*Enabled and active are seperate things.
+	Enabled determines the power status. Is the scanner turned on or not?
+	The scanner is enabled as long as it has power, and the power switch is turned on. While enabled it will use power
+	*/
+	var/enabled = 0
```
seekin4u: This is the only thing not according to our contributing, 1/0 instead of TRUE/FALSE

code/game/objects/items/devices/t_scanner.dm
```diff
@@ -65,72 +89,315 @@
...
-		I.plane = ABOVE_HUD_PLANE
 
-		//Pipes are special
-		if(istype(scanned, /obj/machinery/atmospherics/pipe))
-			var/obj/machinery/atmospherics/pipe/P = scanned
-			I.color = P.pipe_color
-			I.overlays += P.overlays
-
-		I.alpha = 128
-		I.mouse_opacity = 0
```
LethalGhost: That line shouldn't be omitted. 
Without it holopipes will catch mouse click and prevent interactions with items on turf.

NanakoAC: its not working, tried adding it back but they still intercept clicks, see latest commit

I don't think its a big deal, pipes have safety checks so you still need to remove the floor to interact with them. You can turn off the scanner to do stuff, or put it in your pocket

SpadesSlick: It's a big deal, because you can use it in combat to protect yourself. Mobs should be above those pipes anyway.

LethalGhost: Checked it.
That task is out of PR scope.

PR #2408, opened 2018-12-09
Burrows

NanakoAC:
Burrows are here
What are burrows?

Burrows are holes which lead into an abstract tunnel network which is located inside the walls of eris.
They provide secret routes around the ship for roaches, spreading plants, and various other mobs

Where are burrows?
Burrows are on floors against a wall. But they are generally hidden under floor tiles, and can only be seen with the use of a T-ray scanner (or by tearing up the tiles to reveal them)

Burrows spawn all over the ship, both in and out of maintenance. The ones in maintenance tend to be visible due to lack of floorings to cover them. Burrows can spawn anywhere randomly, but in addition a burrow will always have a guaranteed spawn near any group of roaches, spiders, or maintshrooms that spawn at roundstart

What do burrows do?

Burrows serve a large number of purposes:
1. Migration. 
Every 10 minutes or so, every burrow that has mobs near it, has a chance to send some of its mobs elsewhere. This means that over time roaches and spiders will spread out among the ship, instead of just staying in predetermined locations. You can never know for sure where you'll find them
When migrating, mobs will almost always pick a destination in maintenance. There is only a small chance (1%) that they will invade inhabited areas of the ship. 

2. Reinforcements
When a fuhrer roach is in danger, it can send out a distress call, which will summon reinforcements to all nearby burrows in earshot. This will typically cause many roaches to flood into an area, and escape is recommended. These reinforcements are always summoned from another burrow somewhere on the ship, no roaches are magically created

3. Evacuation
When an area becomes breached, or a fuhrer roach is in danger and has already used up their reinforcements, nearby creatures will attempt to evacuate an area by fleeing into a burrow. When they do

4. Plants
Maintshooms and spacevines can use burrows to spread around. In this case they will always spread to burrows near the source. This could include on different decks.

How do we deal with burrows?

Burrows are intended to provide an obstacle that players can deal with. They can be destroyed to make departments safe.
If you're serious about dealing with them, you'll want to gather up the Tools of The Exterminator:

- A T-Ray scanner	
		these allow you to see burrows hidden under floors. Vital if you want to find the darned things.
Speaking of which, the code from the  T Ray scanner PR was merged into this branch. so that one should be merged first
TRAY		

- A digging tool
		You need some kind of excavating equipment to collapse a burrow. The best thing would be a diamond drill, heavily modded to improve workspeed and precision. But any digging tools will do really. Drill, jackhammer, pickaxe, shovel, even a crowbar can do the job. Collapsing burrows is an extremely difficult task though,and doing it with a crowbar will take a looong time.

In addition to digging tools, collapsing burrows uses a compound stat check. The sum total of the user's mechanical and robustness skill is used
		
- Traps:
		Always useful tools, a mechanical trap will instakill many types of roach. Even mousetraps can do a fair amount of damage. They can be placed directly ontop of burrows and will hit anything emerging from there
		

Just look for them!
When mobs are coming through a burrow, it shakes and vibrates, while making a fair bit of noise. If its under a floor, the floor tiles will shake too
shaking

Other Changes 
--------------------------------------------------------

Roaches:
-------
Roach numbers increased. The spawning values are tweaked so there will now be more spawns, and more in each spawn
All Roaches move more slowly now. Slight decrease for most of them, much heavier speed decrease for panzer and fuhrer
Panzer roach health buffed heavily. It's much more of a tank now, as befits its role
Blattedin toxin from the support roaches fixed and buffed a bit. It was previously poisoning roaches, now it heals them gradually.

Fuhrer Roach majorly redesigned:
1. The fuhrer now has a ton of health. 200 to be precise, making him equal to a human player. Significant damage increase too.
2. Fuhrer rarity majorly decreased, they are now a rare sight among roach groups, there's probably less than 10 on the whole ship
3. Fuhrer speed halved, he's very slow and can be outrun easily, will always be on the back lines in combat.
4. Fuhrer gains a powerful new ability, distress call.
His Distress var is increased by a bit whenever he personally takes damage, or whenever any roach dies near him. This value is used for the chance of doing a distress call.
The distress call has two effects, and each of these effects can only be triggered once, ever (per each fuhrer roach)

The first time it's used, distress call summons reinforcements. All burrows in a wide radius will pull reinforcements from elsewhere on the ship, resulting in a major influx of new roaches in the general vicinity. You get 20 seconds to run before they arrive

The second time its used, distress call will sound a retreat. The fuhrer and all other roaches will flee into the nearest burrow, resurfacing elsewhere on the ship. Unless you can kill him quickly, he will escape.

Overall, the fuhrer is now intended to serve as a sort of miniboss.
When you find one, a good strategy would be to run away, and focus first on securing the surrounding area, collapsing all burrows in nearby rooms to limit his ability to call reinforcements.

When you do want to fight him, go in hard. If you kill him fast enough you can prevent his escape. Roaches should be fully vulnerable to stunning methods like flashbangs, and luring him into a trap should prevent movement entirely

Other mobs:
Spiders and slimes have had their health nerfed a bit. They're still pretty tanky, but less absurdly so
The infestation event now spawns its creatures through burrows, prioritising those in crew areas.
Successfully destroying a burrow while creatures are emerging from it will cancel that emergence and refund half of the points to the storyteller

Plants:
Maintshrooms can no longer spread through airlocks. They spread through burrows instead.
Spacevines still can spread through airlocks, but not if the airlock is welded
Reduced maximum range and growing speed of spreading plants generally

Jobs:
Miners given some mechanical skill. The intention is to make them the best at collapsing burrows. Destroying burrows is a job miners can do on-ship while waiting to get to an asteroid

Burrows come with a big pile of new sound effects, courtesy of user Scampsie on freesound.org, and specifically this lengthy sample: https://freesound.org/people/scampsie/sounds/417091/
Licensed under CC BY 3.0 https://creativecommons.org/licenses/by/3.0/


code/_helpers/maths.dm
```diff
@@ -138,3 +138,17 @@
 
 /proc/rand_between(var/lower, var/upper)
 	return (rand() * (upper - lower)) + lower
+
+
+/proc/dist3D(var/atom/A, var/atom/B)
```
NanakoAC: This finds distances accurately across Zlevels. At some point soon i'll get around to writing 3D range procs too

code/_helpers/matrices.dm
```diff
@@ -120,6 +120,18 @@
 
 	return list(R + x,R,R, G,G + x,G, B,B,B + x)
 
+
+//Changes our pixel offset by offset pixels towards the target atom
+/atom/proc/offset_to(var/atom/target, var/offset = 1)
```
NanakoAC: Copied over from the jetpacks PR

code/datums/craft/step.dm
```diff
@@ -121,7 +121,7 @@
 			return
 		if(target)
 			announce_action(start_msg, user, I, target)
-		if(!I.use_tool(user, target || user, time, reqed_quality))
```
NanakoAC: Crafting now uses the sum of cog and mec stats

We don't have variable difficulty for craft recipes yet, but its a step forward

code/datums/mob_stats.dm
```diff
@@ -14,11 +14,56 @@
 	var/datum/stat/S = stat_list[statName]
 	S.changeValue(Value)
 
-/datum/stat_holder/proc/getStat(statName, Pure = null)
-	var/datum/stat/S = stat_list[statName]
-	return S ? S.getValue(Pure) : 0
-
 
+/datum/stat_holder/proc/getStat(statName, Pure = null)
```
NanakoAC: Extended this proc to support compound stat checks

code/game/area/Space Station 13 areas.dm
```diff
@@ -28,7 +28,7 @@ NOTE: there are two lists of areas in the end of this file: centcom and station
 	var/lightswitch = 1
 
 	var/eject = null
-
+	var/is_maintenance = FALSE
```
NanakoAC: this new var simplifies a lot of maintenance checking logic, and allows any area to be designated maint

code/__defines/turfs.dm
```diff
@@ -19,4 +19,8 @@
 #define SMOOTH_ALL 1	//Smooth with all of type
 #define SMOOTH_WHITELIST 2	//Smooth with a whitelist of subtypes
 #define SMOOTH_BLACKLIST 3 //Smooth with all but a blacklist of subtypes
-#define SMOOTH_GREYLIST 4 // Use a whitelist and a blacklist at the same time. atom smoothing only
\ No newline at end of file
+#define SMOOTH_GREYLIST 4 // Use a whitelist and a blacklist at the same time. atom smoothing only
+
+
+#define DECK_HEIGHT 3	//3 metres in vertical height between decks
+//Note that the height of a turf's airspace is defined elsewhere as 2.5 metres, this adds extra to account for floor thickness
```
LethalGhost: Can't that be relative to original 2.5 meters?
Like `#define DECK_HEIGH (ATMOS_TILE_HEIGHT + 0.5)`.

code/datums/mob_stats.dm
LethalGhost: May be that should be separated proc-wraper?

code/game/objects/effects/effect_system.dm
```diff
@@ -13,6 +15,19 @@ would spawn and follow the beaker, even if it is carried or thrown.
 	unacidable = 1//So effect are not targeted by alien acid.
 	pass_flags = PASSTABLE | PASSGRILLE
 
+
+/obj/effect/New()
+	.=..()
+	if (random_rotation)
+		var/matrix/M = transform
+		if (random_rotation == 1)
```
LethalGhost: Magic numbers!

code/__defines/turfs.dm
NanakoAC: good idea, will do

NanakoAC: good idea, will do

code/game/objects/effects/effect_system.dm
NanakoAC: oops, should be TRUE, will fix

LethalGhost: Probably shouldn't. Exist case where `random_rotation == 2`

SpadesSlick:
PR should be named in more informative way.

code/datums/mob_stats.dm
NanakoAC: No real need, this isn't performance critical code. This method is quite flexible

PR #2409, opened 2018-12-09
Adds new RCD Sprites

Michiyamenotehifunana:
dreamseeker_2018-12-09_19-22-30

This PR will replace the old "classic" RCD sprites with new ones, made by yours truly. I'm not sure how the licensing stuff works here, so I'll just declare these sprites as CC-BY-SA until further notice.

dreamseeker_2018-12-09_19-06-10

I have also added code that provides an "ammo" counter for the RCD and flashes a yellow light when there is insufficient ammo.

Fun fact: This PR was very much inspired by https://github.com/tgstation/tgstation/pull/36283, also made by me. The basic shape of the sprite is actually still the same, but I have added more detail to it.


code/game/objects/items/weapons/RCD.dm
```diff
@@ -151,9 +155,19 @@
 	else
 		qdel(T)
 
+	update_icon()	//Updates the ammo counter if ammo is succesfully used
```
LethalGhost: Why not move `update_icon()` in `useResource(...)` proc?

Michiyamenotehifunana: I'll try that and see if it works.

LethalGhost: Thank. Will wait for result.

PR #2411, opened 2018-12-10
Adds the RXD, a ghetto crossbow made from RCDs

Michiyamenotehifunana:
dreamseeker_2018-12-10_18-17-02

The RXD is a RCD with crossbow parts welded onto it. It flashforges bolts using matter units when its string is drawn back. Each bolt costs 10 matter units to forge, and flashforged bolts can be reclaimed for 5 matter units each. It cannot be manually loaded with bolts, nor can its bolts be removed except when firing.

crossbowpreview

I am unsure as to how to properly balance this, so the stats could use some tweaking in the future.


NanakoAC:
this is great

the arrows are the same as a normal crossbow, right?

Michiyamenotehifunana:
They use a subvariant of the normal crossbow bolts named "Flashforged Bolt". They have half the throwforce of plain crossbow bolts.

PR #2412, opened 2018-12-10
Scrap, and related machinery.

LudwigVonChesterfield:
This PR adds a new way of getting loot - scrap. Scrap is a bunch of random junk tossed in together.

Each scrap pile has a predetermined loot in it, and when you hit it with a shovel you update the visible loot.

Scrap beacon is a beacon that attracts scrap all around it.

* [x] Scrap.
* [x] Random reagents in scrap lumps.
* [x] Scrap stacker machine.
* [x] Scrap pile ripper machine.
* [x] Generator based on scrap(?).
* [x] Scrap beacon.


NanakoAC:
Rather big project, and interesting looking. I'm generally in favor of what i see, though i may have some concerns about details

I am concerned about adding too many different loot lists, it becomes harder to maintain and to keep adding content in several places
I think these things should draw heavily from existing lists, like the junk list

Where exactly will scrap be found? I'm a bit perplexed about exactly how its implemented. A machine to attract it seems wierd for tight knit maintenance areas

code/game/objects/random/food.dm
```diff
@@ -1,8 +1,26 @@
...
+				/obj/item/weapon/reagent_containers/food/snacks/liquidfood = 2,
+				/obj/item/weapon/reagent_containers/food/snacks/pie = 1,
+				/obj/item/weapon/reagent_containers/food/snacks/sandwich = 1))
+
+/obj/random/junkfood/rotten
 	name = "random spoiled food"
 	icon_state = "food-red"
 
-/obj/random/rotfood/item_to_spawn()
+/obj/random/junkfood/rotten/item_to_spawn()
```
NanakoAC: rotten food doesn't really need its own seperate list, does it? 

Wouldn't it be better to use the junkfood list and then adjust things in postspawn to add toxins or somesuch?

code/game/objects/random/food.dm
```diff
@@ -16,15 +34,17 @@
...
 				/obj/item/weapon/reagent_containers/food/snacks/sandwich = 1))
 
-/obj/random/rotfood/low_chance
+/obj/random/junkfood/rotten/low_chance
 	name = "low chance spoiled food"
 	icon_state = "food-red-low"
 	spawn_nothing_percentage = 60
 
-/obj/random/rotfood/spawn_item()
+/obj/random/junkfood/rotten/spawn_item()
```
NanakoAC: There's a post spawn function for this, use it. 
Overriding spawn item generally shouldn't be done

code/game/objects/random/random.dm
```diff
@@ -53,3 +53,13 @@
 
 /obj/random/single/item_to_spawn()
 	return ispath(spawn_object) ? spawn_object : text2path(spawn_object)
+
+/obj/randomcatcher
```
NanakoAC: What is this catcher thing all about?

code/modules/scrap/oldificator.dm
```diff
@@ -0,0 +1,250 @@
+/obj/proc/make_old()
+	color = pick("#996633", "#663300", "#666666")
+	light_color = color
+	name = pick("old ", "expired ", "dirty ") + initial(name)
+	desc += pick(" Warranty has expired.", " The inscriptions on this thing were erased by time.", " Looks completely wasted.")
```
NanakoAC: You may want to put a linebreak before these things

code/modules/scrap/oldificator.dm
```diff
@@ -0,0 +1,250 @@
+/obj/proc/make_old()
+	color = pick("#996633", "#663300", "#666666")
+	light_color = color
+	name = pick("old ", "expired ", "dirty ") + initial(name)
```
NanakoAC: More adjectives!

Rusty
ancient
weathered
beaten
broken
tarnished
frayed
etc, get a thesaurus out, make this a big list

code/modules/scrap/oldificator.dm
```diff
@@ -0,0 +1,250 @@
...
+	desc += pick(" Warranty has expired.", " The inscriptions on this thing were erased by time.", " Looks completely wasted.")
+
+	germ_level = pick(80,110,160)
+	if(prob(40))
+		if(prob(70))
+			light_power = light_power / pick(1.5, 2, 2.5)
```
NanakoAC: Afaik, altering these will almost certainly accomplish nothing, since these vars are temporary and their values are set by things when they turn on/off

code/modules/scrap/oldificator.dm
```diff
@@ -0,0 +1,250 @@
...
+	unreliability += rand(1, 5)
+
+/obj/item/weapon/storage/make_old()
+	var/del_count = rand(0,contents.len)
+	for(var/i = 1 to del_count)
+		var/removed_item = pick(contents)
+		contents -= removed_item
+		qdel(removed_item)
+
+	if(prob(75))
```
NanakoAC: Add a check for whether it has storage slots at all before doing this.

In general, apply that principle everywhere. Potentially turning zero values into nonzero ones (even negative) can have unintended consequences

code/modules/scrap/oldificator.dm
```diff
@@ -0,0 +1,250 @@
...
+		storage_slots = max(contents.len, max(0, storage_slots - pick(2, 2, 2, 3, 3, 4)))
+	if(prob(75))
+		max_storage_space = max_storage_space / 2
```
NanakoAC: as above, only do this if any storage space is set

code/modules/scrap/oldificator.dm
```diff
@@ -0,0 +1,250 @@
...
+
+	..()
+
+/obj/item/weapon/reagent_containers/make_old()
+	for(var/datum/reagent/R in reagents.reagent_list)
+		R.volume = rand(0, R.volume)
+	reagents.add_reagent("toxin", rand(0,10))
```
NanakoAC: Instead of a flat value, make this based on the volume of reagents.
It'd be silly to have a pill with 1u of medicine and 10u added toxins

Also, this should probably only be done if there were any reagents at all to begin with
and maybe wrap it in a prob call so toxins aren't guaranteed

code/modules/scrap/oldificator.dm
```diff
@@ -0,0 +1,250 @@
...
+/obj/item/ammo_magazine/make_old()
+	var/del_count = rand(0,contents.len)
+	for(var/i = 1 to del_count)
+		var/removed_item = pick(stored_ammo)
+		stored_ammo -= removed_item
+		qdel(removed_item)
+	..()
+
+/obj/item/weapon/cell/make_old()
+	charge = max(0, rand(0, maxcharge * 2) - maxcharge)
```
NanakoAC: use rand_between for this. It does this annoying math for you, and allows a value between any two arbitrary numbers, including negatives and decimals

Do something like rand_between (maxcharge*0.75, maxcharge)

code/modules/scrap/oldificator.dm
```diff
@@ -0,0 +1,250 @@
...
+	if(prob(40))
```
NanakoAC: 40 is a very high chance for something to explode, drop this down a bit

code/modules/scrap/oldificator.dm
```diff
@@ -0,0 +1,250 @@
...
+		rigged = TRUE
+		if(prob(80))
+			charge = maxcharge  //make it BOOM hard
+	..()
+
+/obj/item/weapon/stock_parts/make_old()
+	var/degrade = pick(0,1,1,1,2)
```
NanakoAC: Do stock parts support non integer ratings?

code/modules/scrap/oldificator.dm
```diff
@@ -0,0 +1,250 @@
...
+
+/obj/item/stack/material/make_old()
+	return
+
+/obj/item/stack/rods/make_old()
+	return
+
+/obj/item/weapon/tank/make_old()
+	var/new_vol = pick(0.2,0.4,0,6,0,8)
+	air_contents.gas["oxygen"] *= new_vol
```
NanakoAC: I'm very sure there's a better way to do this than looping through all the possible gasses
check xgm_gas_mixture

code/modules/scrap/oldificator.dm
```diff
@@ -0,0 +1,250 @@
...
+	air_contents.gas["nitrogen"] *= new_vol
+	air_contents.gas["sleeping_agent"] *= new_vol
+	air_contents.temperature = 293
+	volume *= new_vol
+	air_contents.update_values()
+	..()
+
+/obj/item/weapon/circuitboard/make_old()
+	if(prob(75))
+		build_path = pick(/obj/machinery/washing_machine, /obj/machinery/broken, /obj/machinery/shower)
```
NanakoAC: does building a broken machine really make sense?

In any case this is a good idea, but it should have a bigger possibilities list
also will this automatically update the name of the circuit board? Probably not, make sure that is done

code/modules/scrap/oldificator.dm
```diff
@@ -0,0 +1,250 @@
...
+		var/obj/item/weapon/aiModule/brokenmodule = new /obj/item/weapon/aiModule/broken
+		brokenmodule.name = src.name
+		brokenmodule.desc = src.desc
+		brokenmodule.make_old()
+		qdel(src)
+	..()
+
+/obj/item/clothing/suit/space/make_old()
+	if(prob(75))
+		var/datum/breach/B = new()
```
NanakoAC: don't do this, there are functions to handle it. Look at random/voidsuit, i think i used it in there for spawning pre-damaged voidsuits

code/modules/scrap/oldificator.dm
```diff
@@ -0,0 +1,250 @@
...
+		B.class = pick(25,50,75)
+		B.damtype = pick(BRUTE, BURN)
+		B.update_descriptor()
+		B.holder = src
+	..()
+
+/obj/item/clothing/make_old()
+	if(prob(50))
+		slowdown += pick(0.5, 0.5, 1, 1.5)
+	if(prob(75))
```
NanakoAC: halving all the values at 75% is pretty severe, nerf this a bit, and maybe randomise exactly how much they're reduced

code/modules/scrap/oldificator.dm
```diff
@@ -0,0 +1,250 @@
...
+		var/datum/breach/B = new()
+		breaches += B
+		B.class = pick(25,50,75)
+		B.damtype = pick(BRUTE, BURN)
+		B.update_descriptor()
+		B.holder = src
+	..()
+
+/obj/item/clothing/make_old()
+	if(prob(50))
```
NanakoAC: this is a very high slowdown chance, it'll make found clothing not very useful, drop this to 15% imo

code/modules/scrap/oldificator.dm
```diff
@@ -0,0 +1,250 @@
...
+		seconds_electrified = -1
+	if(prob(60))
+		shut_up = 0
+	if(prob(60))
+		shoot_inventory = 1
+	if(prob(75))
+		var/del_count = rand(0,product_records.len)
+		for(var/i = 1 to del_count)
+			var/removed_item = pick(contents)
+			contents -= removed_item
```
NanakoAC: i've heavily modified vendors, this part should be reviewed

they don't contain items anymore until those items are vended

code/game/objects/random/random.dm
LudwigVonChesterfield: Sometimes random types spawn a random type which then spawns an item. The catcher is used in these cases to get the proper item instead of getting the result of the spawn which would be a random type(which is intended to yet spawn another item)

code/modules/scrap/oldificator.dm
LudwigVonChesterfield: it really depends, and as a general case this has rights to live.

code/modules/scrap/oldificator.dm
LudwigVonChesterfield: 1) there is already a 10% chance of no toxins (rand can give out 0)
2) it would make sense for a reagent container, even empty pill to get toxic after lying in a pile of scrap, full of waste and toxic material for... who knows how long

code/modules/scrap/oldificator.dm
LudwigVonChesterfield: there are no non-integer rating. it's written as such so a chance of 1 would be thrice as higher as of 0 and 2.

code/modules/scrap/oldificator.dm
LudwigVonChesterfield: the point of it being a fake circuit board, the player thinks it's a, per say AI laws console, but in the end it appears to be circutry for a washing machine. bamboozled player is bamboozled

code/modules/scrap/oldificator.dm
LudwigVonChesterfield: well there would be a lot of scrap to search though

code/modules/scrap/oldificator.dm
```diff
@@ -0,0 +1,250 @@
...
+		equip_delay += rand(0, 2 SECONDS)
+	..()
+
+/obj/item/weapon/aiModule/broken
+	name = "\improper broken core AI module"
+	desc = "broken Core AI Module: 'Reconfigures the AI's core laws.'"
+
+/obj/machinery/broken/Initialize()
+	..()
+	explosion(loc, 1, 2, 3, 3)
```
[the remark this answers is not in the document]
NanakoAC: If the intent is to have them explode when created from an old circuitboard, that should be handled in the Created proc

For those spawned at roundstart, they should explode when interacted with. Bumped, or attackhand or somesuch

code/modules/scrap/oldificator.dm
NanakoAC: i dont think it does, there's vanishingly few cases where it'd matter. Glowing slime cores maybe?

this won't work as intended as is, it should be done with specific overrides for items like flashlights

code/modules/scrap/oldificator.dm
NanakoAC: i was asking because using nonintegers could be interesting here

code/modules/scrap/oldificator.dm
NanakoAC: ah, well in that case it should be given a generic but enticing name and description

Unknown Circuit Board
What will it construct? Try it out and see

code/modules/scrap/oldificator.dm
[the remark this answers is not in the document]
LudwigVonChesterfield: the intent is to have them explode when they are initialized, per say, sometimes when becoming old a machinery can become "broken", and should make an explosion.

code/modules/scrap/oldificator.dm
LudwigVonChesterfield: I will drop it to 30%, because I think you misunderstand the entire concept of roaming through a ton of trash just to find something at least barely useful. Trash shouldn't be a lootbag full of goodies, rather a pile of... Trash, with something good inside.

code/modules/scrap/oldificator.dm
LudwigVonChesterfield: the list of adjectives isn't as broad, because they are used for any object, so for say "rusty plastic spoon" would make no sense.

code/modules/scrap/oldificator.dm
LudwigVonChesterfield: I did prettify this part, though the "better way" does exactly that, but it is conviniently wrapped inside a proc.

LudwigVonChesterfield:
Is updated to fix what has been stated by Nanako, still far from complete.

LudwigVonChesterfield:
The PR is complete and ready for review.

PR #2413, opened 2018-12-11
Fungo D'Artigliera v1.1

NanakoAC:
A significant update to maintshrooms, this PR does several things.

1. Colors!
Colors of the wind
Instead of being a static blue, the cap of the mushroom is now tinted according to the chemicals contained in it. This gives a little visual aid to guess what might be in there

It tweaks the saturation and lightness of the code to avoid too much wierdness involved with tinting, generally going for pastel shades

2. Clipping
Secondly, this fixes an issue that has annoyed me for too long
Maintshrooms are supposed to grow on walls, and as such they overlap the wall a bit. This is great when the wall is to the west, east or north. But overlapping walls to the south just looked awful and buggy. It may seem minor but i wanted to fix it, the fix came at great cost of effort but the result is worthwhile
Clipped

This employs some black magic to cut off the bottom of the sprite, generating a cutoff cached icon at runtime and reusing it for all the shrooms. Also generally refactors wall hugging code, adding a new seed trait for it

3. Chems
Adds more chems to the list of possibilities, self explanatory. Some are chosen for colors, some for interesting effects.
Polyacid has a special effect that causes it to damage floors, which is pretty fun. In past times, this would have been a problem due to lots of breaches, but our floors are triple layered bulwarks these days


Voldirs:
And still one problem remains, shrooms can have really vast variety of chems, but most of them do nothing for human when sprayed. It kinda defeats their purpose.

NanakoAC:
> And still one problem remains, shrooms can have really vast variety of chems, but most of them do nothing for human when sprayed. It kinda defeats their purpose.
> And in the same time chems are too robust if injected, such as cyanide, slime jelly and mutation toxins

A correction to that
Most of them do nothing most of the time. But all chems are situational

If you find some mushrooms that spray bicaridine, you can use it as a base, fight roaches and retreat there to heal.
If you know what's in them, they can be useful. Thats part of the point of this PR

speaking of which, i wonder if a reagent scanner can detect their contents...

PR #2419, opened 2018-12-17
T Ray Scanner availability

NanakoAC:
This PR makes T-ray scanners more available to the people who need them:
A scanner is added to each miner equipment closet (also a single free toolmod)
Scanners are added to several robot modules, including drone

In addition this fixes a couple niggling issues:
Engineering grippers can now pickup floor tiles, so they can repair floors
Fixed a runtime error with BST backpacks

And one significant fix, i decided to implement a quick hack for the longstanding issue of robot items running out of power and breaking. For now, the charge of cells in their tools (which are unremovable and can't be recharged) is set to a really high value on initialize so it shouldn't run out.

This is a temporary fix until we (probably cups) gets around to coding a proper fix to wire them to the robot's central power cell

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_modules.dm
+++ b/code/modules/mob/living/silicon/robot/robot_modules.dm
@@ -112,6 +112,12 @@ var/global/list/robot_modules = list(
 		T.degradation = 0 //We don't want robot tools breaking
 
 
+	for (var/obj/item/I in modules)
+		for (var/obj/item/weapon/cell/C in I)
+			C.charge = 999999999 //A quick hack to stop robot modules running out of power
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/silicon/robot/robot_modules.dm
```diff
@@ -112,6 +112,12 @@ var/global/list/robot_modules = list(
 		T.degradation = 0 //We don't want robot tools breaking
 
 
+	for (var/obj/item/I in modules)
+		for (var/obj/item/weapon/cell/C in I)
+			C.charge = 999999999 //A quick hack to stop robot modules running out of power
```
mattroks101: I don't know of a better way to do this but let's just make sure we don't forget this exists later.

PR #2420, opened 2018-12-17
Auxiliary Fueltank Fix

NanakoAC:
Fixes the expanded fuel tank toolmod not being able to be filled

### DIFF
--- a/code/game/objects/items/weapons/tools/_tools.dm
+++ b/code/game/objects/items/weapons/tools/_tools.dm
@@ -633,6 +633,8 @@
 	precision = initial(precision)
 	suitable_cell = initial(suitable_cell)
 	max_fuel = initial(max_fuel)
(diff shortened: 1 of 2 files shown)


code/game/objects/items/weapons/tools/_tools.dm
```diff
@@ -633,6 +633,8 @@
 	precision = initial(precision)
 	suitable_cell = initial(suitable_cell)
 	max_fuel = initial(max_fuel)
+	if (reagents)
+		reagents.maximum_volume = max_fuel
```
LethalGhost: I think it's better to update reagents at the end of this proc to proper handle upgrades and simplify it code.

NanakoAC: done

PR #2421, opened 2018-12-17
Cruciform fix

LagannPM:


code/controllers/subsystems/jobs.dm
```diff
@@ -384,6 +384,12 @@ SUBSYSTEM_DEF(job)
 		if(equipped != 1)
 			var/obj/item/clothing/glasses/G = H.glasses
 			G.prescription = 1
+	
+	if(H.religion == "Neotheology" && !locate(/obj/item/weapon/implant/core_implant/cruciform, H))
```
NanakoAC: The chaplain spawns with a special version of the cruciform which has more powers, does that work correctly?

LagannPM: yes it is

PR #2429, opened 2018-12-23
Hardsuit Storage

NanakoAC:
This PR Adds two main new features

1. the hardsuit Internal Storage Module
I checked other codebases, or at least bay and aurora, surprisingly they didnt make this yet. Now i have.

Internal storage gives a RIG suit 16 units of storage space, which is 2/3rds of what a satchel gives. It allows the RIG to be seamlessly used as a storage container, both worn and on the ground. The storage module itself can also be used as a container while its out of the suit, the contents are stored in the module.

There is no security, anyone who has the rig can use the internal storage. We could add secure versions in future that check the user has access rights to activate the suit. I definitely plan to add expanded versions in future, as a fancy science thing

The internal storage module comes pre-installed on the Technomancer Exultant hardsuit, and only that one. Anyone else who wants one needs to go scavenging, or buy from someone else

2. Random Rig Modules
Every possible module in the game is added to a new random list, and distributed among maint loot. The really fancy mounted weapons are very rare, finding the storage module is very common

Also adds a Rare modules list which contains all the good stuff and no weighting, that's found in the rare loot pool

Also mildly refactored internal storage code a tiny bit


code/modules/clothing/suits/storage.dm
```diff
@@ -15,12 +15,14 @@
 	. = ..()
 
 /obj/item/clothing/suit/storage/attack_hand(mob/user)
-	if (pockets.handle_attack_hand(user))
-		..(user)
+	if ((is_worn() || is_held()) && pockets.handle_attack_hand(user))
```
NanakoAC: i think this logic might be wrong, will investigate tomorrow

Voldirs:
>/obj/item/rig_module/stealth_field

Don't include this module, it needs to be reworked badly or plainly removed. it's almost impossible to counter invisibility in combat.

Also power costs for modules should be checked too.

NanakoAC:
> > /obj/item/rig_module/stealth_field
> 
> Don't include this module, it needs to be reworked badly or plainly removed. it's almost impossible to counter invisibility in combat.
> 
> Also power costs for modules should be checked too.

It'd make a good story though! Most of the rare modules are just that, pretty darned rare. the odds of someone finding them are quite low

Stealth field can surely do with a rework, i plan to work on it while revamping ninja. But i don't think having it here for now would be so bad

code/modules/clothing/suits/storage.dm
NanakoAC: fixed

code/game/objects/random/rig.dm
```diff
@@ -41,4 +41,99 @@
...
-		module.sabotage_tank()
\ No newline at end of file
+		module.sabotage_tank()
+
+
+
+//Installable modules!
+
+/obj/random/rig_module
+
```
mattroks101: So does this just not get a name? Or did you forget to put something here?

NanakoAC: will fix

NanakoAC: fixed

PR #2435, opened 2018-12-29
New "Gladstone" shotgun

SpadesSlick:
image
Sprites by @Michiyamenotehifunana.

Adds new shotgun with extended ammo capacity and low recoil. Later we will create more stats for weapons to play around, so even if new guns don't have a lot of differences now, this will change.
image


code/modules/projectiles/guns/projectile/shotgun/bull.dm
```diff
@@ -15,7 +15,7 @@
 	origin_tech = list(TECH_COMBAT = 4, TECH_MATERIAL = 4)
 	burst_delay = null
 	fire_delay = null
-	bulletinsert_sound 	= 'sound/weapons/guns/interact/shotgun_instert.ogg'
+	bulletinsert_sound = 'sound/weapons/guns/interact/shotgun_instert.ogg'
```
NanakoAC: instert should be insert

PR #2440, opened 2018-12-31
Folding sheets

LagannPM:
default
Bedsheets can be folded and rolled
Added bedsheet/resting interactions


NanakoAC:
These are the annoying sheets that get picked up when you click on them right?

Can you make folding the default action instead of alt click, and implement drag and drop to pick them up.
(also maybe possibly have them be placed on a bed when you use them on it, or a mob lying in the bed)

LagannPM:
Had to rename pickup() proc

code/game/objects/items.dm
```diff
@@ -127,15 +127,23 @@
 	return ..(user, distance, "", message)
 
 /obj/item/attack_hand(mob/user as mob)
-	var/atom/old_loc = src.loc
-
-	src.pickup(user)
+	src.before_pickup(user)
```
NanakoAC: This doesn't fit with established design patterns, look at pre_equip for an example

It should be called pre_pickup, and it should return a value which dictates whether or not the pickup action completes.

something like
if (!pre_pickup(user))
return

LagannPM:
done

PR #2444, opened 2019-01-02
Add printable card broadcasters

greggbot:
They exist, so let's give them a design.
All this does is allow the printing of the read-only ID broadcasters used in modular PDAs. It's nice if you want to assemble one from scratch, or need to replace one.


SpadesSlick:
Maybe it's better to add this to autolathe discs? That's not looks like advanced spare part.

SpadesSlick:
Indeed. Looks like we need a new disc for modular computers spare parts. But this could be done in separate PR.

PR #2446, opened 2019-01-03
Makes maintshroom traits splicable

greggbot:
So Xenobotany can have a bit more fun with Eris' unique flora.
This does NOT:
Make them mutate randomly.
Add them to random seeds.
It ONLY makes them splicable. I sort of like the idea of traits being "locked" behind maintshroom samples, so I've kept it that way, but it could easily be changed.
TRAIT_WALL_HUGGER is under GENE_VIGOUR, the same gene as spreading, which I figured made sense.
TRAIT_CHEM_SPRAYER is under GENE_FRUIT, which is where traits like juicy/thorny and the like are.


NanakoAC:
If i understand correctly, this means you could take a sample of a shroom that sprays fuel, splice it into a plant of your own, and create plants that spray fuel?

this sounds good

PR #2449, opened 2019-01-04
Full graphical floors rework

SpadesSlick:
This PR do one simple and huge thing - reworks all of the ship floors to new sprites. While work is not finished, and few new floors should be added as well, this is enough to provide a base for smaller PRs.

New floors use new logic for merging, allowing to create floor borders without placing floor decals on map directly. This should improve the speed of the mapping procces, and now all of the floors can be changed by players in-game. When no decals are involved, for players this should be simple, they only need to craft proper floor tile and place them where they need it.

While the overall graphics quality may be a bit lower now, those floors are made for use with new palette, and should be used with colored lights, meaning that this whole set got huge potential for future use.

Because of the nature of this PR - task replacing almost every floor tile on a ship, map merge was not attempted, and looks like it's not needed here. Change was too big anyway.

image
image
image


code/game/turfs/flooring/flooring_effects.dm
```diff
@@ -15,8 +21,6 @@
...
 
-/obj/effect/damagedfloor/New(loc)
-	var/turf/simulated/floor/F = loc
-	if(istype(F))
-		F.break_tile(1)
-	qdel(src)
\ No newline at end of file
+/obj/effect/damagedfloor/rust/New(loc)
+	..()
+	icon_state = "rust[rand(1,9)]"
```
LethalGhost: `src` will be deleted in 25.
Set icon before update `icon_state`

code/game/turfs/flooring/flooring_effects.dm
```diff
@@ -6,6 +6,12 @@
 /obj/effect/damagedfloor
 	icon_state = "scorched1"
 
+/obj/effect/damagedfloor/New(loc)
+	var/turf/simulated/floor/F = loc
+	if(istype(F))
+		F.break_tile(1)
+	qdel(src)
```
LethalGhost: It's better to use `Initialize` instead `New` for this.

PR #2453, opened 2019-01-05
New cryo sprites.

SpadesSlick:
image

Those sprites should work better with floors from #2449. They are the part of the rework, and they should be merged after that PR.


NanakoAC:
Didn't we discuss adding extra cabling to the rear part? I wanted to use the existing rear part and you mentioned that instead.

These beds aren't meant to be storage units, they're just entrances to storage, People in them are moved to some hidden place deep within the ship for longterm cryostorage. The rear part needs to look big enough to suck people through it and into the floor/walls

SpadesSlick:
If we want it to be reworked the way you are suggesting, we need to sprite whole cryo storage. And maybe we will a bit later. But not now. For now we need something that will work with our new floor, because old cryo is looking like shit on them.

Cables were added actually, well, only one cable. I don't think we need more.

SpadesSlick:
It looks like shit that way, I told you. We can't mash those thing together.

PR #2455, opened 2019-01-05
Interface rework and new AI interface

LagannPM:
Start of rework of interface system
Currently applied only to AI
https://cdn.discordapp.com/attachments/475698003827752962/530986674541297675/unknown.png
default
supports mouse interactions and colorizing
default
Supports autoaligning using layouts

P.S.
This is not final layout of AI it will change later, currently adding it to make AI playable
Proper documentation will be later
Also added two buttons to ghost to navigate between levels


NanakoAC:
Potential conflict with #2429, that should be merged first

code/_onclick/hud/HUD_element/_HUD_defines.dm
```diff
@@ -0,0 +1,24 @@
+//  values for setAlignment(var/horizontal, var/vertical) and alignElements(var/horizontal, var/vertical, var/list/HUD_element/targets) 
+#define HUD_NO_ALIGNMENT 0 						 //no alignment
+#define HUD_CENTER_ALIGNMENT 1					 //central alignment
+
+#define HUD_HORIZONTAL_WEST_OUTSIDE_ALIGNMENT 2	 //bordering west side of parent from outside
```
NanakoAC: shouldn't these be bitfields? What if you want to anchor to a corner?

LagannPM: no, there are two types of alignment vertical and horizontal
proc that align uses two of them as arguments
if you want corner you align both horizontal and vertical to one side
Atleast it how it was created by captcha

code/game/machinery/camera/presets.dm
```diff
@@ -28,49 +28,49 @@ var/global/list/engineering_networks = list(
 	network = list(NETWORK_CRESCENT)
 
 /obj/machinery/camera/network/fist_section
-	network = list(NETWORK_FIRST_SECTION)
+	network = list(NETWORK_AI, NETWORK_FIRST_SECTION)
```
NanakoAC: why was it necessary to insert NETWORK_AI everywhere here? 

It would be a lot better to give the AI a list of networks it can use, surely?

code/modules/mob/observer/observer_UI.dm
```diff
@@ -0,0 +1,36 @@
+/datum/interface/ghost_Eris
```
NanakoAC: you added an interface for ghosts too?

NanakoAC:
I'm testing things out. These move up/down buttons for ghosts are a great idea, but their positioning is a pain. Move them to the lower right corner please.  Being in the middle of the screen is too distracting

NanakoAC:
Same thing with the buttons on either side of the AI screen. Move them both down to their lower corners.
They can still be columns, but the lowest button should sit in the lower right corner

Also, is there any possibility of scaling these buttons? 32x32 seems excessive and they do take up a not insignificant amount of screenspace, which may obscure mobs underneath.

Is there any option for players to adjust their transparency, or to hide or disable some of these?

NanakoAC:
Remove the Call Evacuation button. Entirely, just remove it.

We do not want that to be easy to do. A clearly visible button to do it is not needed

PR #2461, opened 2019-01-05
Serbian Mercenaries: Phase 1

NanakoAC:
Ready to go.

Future features that didn't make it in this time:

Team uplink
Mission time counter
Better objectives
Command points


SpadesSlick:
This should a last time we are not using map merger.

PR #2491, opened 2019-01-06
Adminjump fix

NanakoAC:
There were two verbs named jump to mob and they were interfering with each other

tweaked names and paths slightly to fix this


SpadesSlick:
God, I fucking love Bay.

PR #2497, opened 2019-01-07
fixes spawn bugs

### LINKED ISSUE #2510: Equipment overload issues
BUG: If your bag is overloaded because of starting equipment your loadout stuff may never spawn in

reported in junkyard, needs testing

LagannPM:
fixes missing pda/id when using custom jumpsuit, disappearing items when cant store them, and might fix spawn in space
fixes #2510

### DIFF
--- a/code/controllers/subsystems/ticker.dm
+++ b/code/controllers/subsystems/ticker.dm
@@ -176,8 +176,8 @@ SUBSYSTEM_DEF(ticker)
 	Master.SetRunLevel(RUNLEVEL_GAME)
 	create_characters() //Create player characters and transfer them
 	collect_minds()
-	equip_characters()
 	move_characters_to_spawnpoints()
+	equip_characters()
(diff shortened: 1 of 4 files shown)


code/controllers/subsystems/ticker.dm
```diff
@@ -176,8 +176,8 @@ SUBSYSTEM_DEF(ticker)
 	Master.SetRunLevel(RUNLEVEL_GAME)
 	create_characters() //Create player characters and transfer them
 	collect_minds()
-	equip_characters()
 	move_characters_to_spawnpoints()
+	equip_characters()
```
NanakoAC: i don't understand the logic behind moving people to spawn after equipping them. Won't this result in people spawning as naked bald people for a second before their clothes appear?

LagannPM: no, there shouldnt be any ticks between that and this is intended way to do it
otherwise loadout items are dropped in null space

PR #2498, opened 2019-01-07
Fix emergency pods

Camper12345:
* No more crash on pods init
* Pod doors are now can open


code/controllers/evacuation/evacuation_pods.dm
```diff
@@ -28,6 +28,10 @@
 			if (pod.arming_controller)
 				pod.arming_controller.arm()
 
+			if (istype(pod.active_docking_controller, /datum/computer/file/embedded_program/docking/simple/escape_pod))
```
NanakoAC: this is a hack, not the way to do things

The pods need a bit of refactoring because I reworked shuttle code and didn't get around to them

code/modules/surgery/other.dm
```diff
@@ -1,4 +1,4 @@
-//Procedures in this file: Inernal wound patching, Implant removal.
+//Procedures in this file: Inernal wound patching, Implant removal, Hardsuit removal.
```
NanakoAC: unrelated? I don't see any changes to surgery here

maps/CEVEris/_CEV_Eris.dmm
```diff
@@ -1040,7 +1040,7 @@
...
 "aua" = (/obj/structure/cable/green{d1 = 1; d2 = 8; icon_state = "1-8"},/obj/machinery/hologram/holopad,/turf/simulated/floor/tiled,/area/eris/eschangarb)
 "aub" = (/turf/simulated/shuttle/wall/escpod{tag = "icon-escpodwall16"; icon_state = "escpodwall16"},/area/shuttle/escape_pod1/station)
-"auc" = (/obj/machinery/embedded_controller/radio/simple_docking_controller/escape_pod{frequency = 1380; id_tag = "escape_pod_1"; pixel_x = -25; pixel_y = 0; tag_door = "escape_pod_1_hatch"},/obj/structure/bed/chair/shuttle,/turf/simulated/shuttle/floor{icon_state = "cargofloor1"},/area/shuttle/escape_pod1/station)
+"auc" = (/obj/machinery/embedded_controller/radio/simple_docking_controller/escape_pod{frequency = 1380; id_tag = "escape_pod_1"; pixel_x = -25; pixel_y = 0; tag_door = "escape_pod_1_hatch"; tag = "escape_pod_1_controller"},/obj/structure/bed/chair/shuttle,/turf/simulated/shuttle/floor{icon_state = "cargofloor1"},/area/shuttle/escape_pod1/station)
```
NanakoAC: what have you done here and why?

Camper12345: I've added tag to the shuttle docking controller, so pod could link itself with its controller. This was the cause of crash on initialization

code/modules/surgery/other.dm
Camper12345: You're right

maps/CEVEris/shuttles-eris.dm
```diff
@@ -8,7 +8,7 @@
 /datum/shuttle/autodock/ferry/escape_pod/erispod/New()
 	name = "Escape Pod [number]"
 	default_docking_controller = "escape_pod_[number]"
```
NanakoAC: default docking controller should be the same value as arming controller, indicating the panel inside the pod

Camper12345: I've spend some time trying to figure ut what does it mean.
escape_pods.dm, line 9

Camper12345: Even more, pod controller and ship-side controller aren't the same controller

Camper12345:
It works. But yes, it really needs refactoring.

Camper12345:
Tested, it works

PR #2540, opened 2019-01-08
Switches the logging to the correct one on bluespace techs.

GrayRachnid:
the debug log doesn't display it in realtime.


GrayRachnid:
I didn't notice it in any other abuse-able verbs. This is the biggest one that I could find.

PR #2549, opened 2019-01-08
Closets and crates now can be hacked.

### LINKED ISSUE #2513: Locked crates can't be hacked
Code exists to hack locked crates but it gets blocked

SpadesSlick:
Closes #2513.

Not only they are now can be hacked, but also hack is now silent with aural dumper.
image


code/game/objects/structures/crates_lockers/closets.dm
```diff
@@ -376,14 +376,19 @@
...
 		)
 		if(I.use_tool(user, src, WORKTIME_LONG, QUALITY_PULSING, FAILCHANCE_HARD, required_stat = STAT_MEC))
 			if(hack_stage < hack_require)
-				playsound(loc, 'sound/items/glitch.ogg', 60, 1, -3)
+				var/obj/item/weapon/tool/T = I
+				if (istype(T))
+					if(!(T.silenced))
+						playsound(loc, 'sound/items/glitch.ogg', 60, 1, -3)
+				else
+					playsound(loc, 'sound/items/glitch.ogg', 60, 1, -3)
```
NanakoAC: Looks like the same audio settings for both silenced and unsilenced
Give the silenced version a volume of 3 and an extrarange of -5

PR #2564, opened 2019-01-09
Autolathe improvements and buckle sound tweak

LagannPM:


code/game/machinery/autolathe.dm
```diff
@@ -201,14 +201,19 @@
...
 
-
 /obj/machinery/autolathe/attackby(var/obj/item/I, var/mob/user)
 	if(default_deconstruction(I, user))
 		return
 
 	if(default_part_replacement(I, user))
 		return
 
+	if(istype(I, /obj/item/weapon/disk/autolathe_disk))
```
NanakoAC: Recycling disks should be allowed in the case that
1. The disk has/had limited uses
and
2. Its uses have been depleted to zero

code/game/machinery/autolathe.dm
```diff
@@ -361,8 +379,14 @@
 		return FALSE
 
 	if(!eating.matter || !eating.matter.len)
-		user << SPAN_NOTICE("\The [eating] does not contain significant amounts of useful materials and cannot be accepted.")
-		return
+		user << SPAN_NOTICE("\The [src] refuses to accept \the [eating] as it has non-null license.")
```
NanakoAC: This error message makes no sense, why change it

code/game/machinery/autolathe.dm
```diff
@@ -361,8 +379,14 @@
...
+		return FALSE
+
+	if(istype(eating, /obj/item/weapon/disk/autolathe_disk))
+		var/obj/item/weapon/disk/autolathe_disk/disk = eating
+		if(disk.license)
+			user << SPAN_NOTICE("\The [eating] does not contain significant amounts of useful materials and cannot be accepted.")
```
NanakoAC: aha now i see, you've got it swapped around

code/game/objects/buckling.dm
```diff
@@ -81,7 +81,7 @@
 	
 	if(do_after(user, buckleTime, src))
 		if(buckle_mob(M))
-			playsound(src.loc, 'sound/machines/Custom_closetopen.ogg', 15, 1, -3)
+			playsound(src.loc, 'sound/machines/Custom_closetopen.ogg', 65, 1, -3)
```
NanakoAC: does this need to be so loud? what's the reasoning behind this

loud sounds are rather intrusive,

code/game/objects/buckling.dm
```diff
@@ -106,6 +106,7 @@
 	if(do_after(user, unbuckleTime, src))
 		var/mob/living/M = unbuckle_mob()
 		if(M)
+			playsound(src.loc, 'sound/machines/Custom_closetopen.ogg', 65, 1, -3)
```
NanakoAC: same here, i think quieter was fine

code/game/machinery/autolathe.dm
LagannPM: those cases are taken into account

code/game/objects/buckling.dm
LagannPM: they are not, they barely noticable, before they couldnt be heard at all

code/game/objects/buckling.dm
LagannPM: same as above

PR #2569, opened 2019-01-09
Borg spawn fixes

LagannPM:
Spawn fixes


code/controllers/subsystems/jobs.dm
```diff
@@ -509,7 +509,9 @@ proc/EquipCustomLoadout(var/mob/living/carbon/human/H, var/datum/job/job)
 
 	var/datum/spawnpoint/SP
 	if(late)
-		if(!pref_spawn)
+		if(rank == "Robot")
```
NanakoAC: magical special cases like this are hacky. Can you find a better approach than sticking in checks for robots?

For example, possibly making this a mob proc instead of a generic one, so robots can override it and write their own

LagannPM: there is no need to create it exclusively for borgs, as it only exception i can think of.
But i can force borgs to use roundstart spawnpoints instead.
In reality borgs shouldnt have options in pref spawnpoints. Humans cant use those spawnpoints and borgs cant use human ones.

LethalGhost: There exists that vars in spawnpoints code:
```
	var/restrict_job = null
	var/disallow_job = null
```

NanakoAC: can you use those vars?

LagannPM: yes, i forgot about them

PR #2580, opened 2019-01-10
Mopping Tweaks

NanakoAC:
A small whimsical feature to improve job mechanics for the janitor. I noticed a lot of janitors arent really doing their job and the ship is always messy.

The main feature of this PR is a new alternate usage mode for the mop, attackself to toggle it into sweep mode. This cleans a 3-tile arc infront of you, and it does so faster than cleaning a single tile. Comes with a neat visual effect Sweeping!
 However it has drawbacks:
-It only does a light clean, removing one stain or layer of dirt. So particularly soiled floors may require 3-5 applications
-It needs space to use. If any dense objects or tiles are included in your sweep arc, you hit it and get stunned briefly.
-Any mobs caught in the way will be whacked with the mop. Could annoy passing crewmen and get you in trouble if not careful

This gives the sweep a little bit of skill to using it. Keeping your cleaning rate optimal requires awareness of your surroundings
also buffed the singletile cleaning a little, and removed some textspam from it

Lastly, increases the availability of mops a little, by adding a spare one to the custodial closet and several spares to the orderable custodial crate


code/_helpers/atoms.dm
```diff
@@ -18,3 +18,5 @@
 
 	return null//If we get here, the holder must be buried many layers deep in nested containers, or else is somehow contained in nullspace
 
+
```
mattroks101: Was this intentional?

NanakoAC: I started writing a function to get pixel offsets, then ended up using forcemove instead and deleted the thing i wrote, guess i missed a space.

PR #2582, opened 2019-01-10
Return Sanitze to name input in preferences

LethalGhost:


code/modules/client/preference_setup/general/01_basic.dm
```diff
@@ -65,10 +65,15 @@ datum/preferences
 	var/datum/species/S = all_species[pref.species]
 
 	if(href_list["rename"])
-		var/raw_name = input(user, "Choose your character's name:", "Character Name")  as text|null
+		var/raw_name = input(user, "Choose your character's name:", "Character Name", pref.real_name)  as text|null
 		if (!isnull(raw_name) && CanUseTopic(user))
-			pref.real_name = raw_name
-			return TOPIC_REFRESH
+			var/new_name = sanitize_name(raw_name, pref.species)
```
NanakoAC: Will this properly handle cyrillic characters and such?

PR #2584, opened 2019-01-10
Medical vendors and some minor access changes

Voldirs:
1)Medical equipment room got appropriate access requirment
2)NanoMed Plus got medical equipment access
3)IHS med spec got medical equipment access
Non-medical department roles that have this access include: captain, first officer, MEO, IHS med spec
That is to prevent literally everyone from using that vendor to heal themselves ignoring medical staff.


NanakoAC:
Hmm, i'm unsure about this. 

Ironhammer certainly shouldn't be stealing medical supplies, but sometimes they need to, especially when theres no doctors. I think this might be likely to just cause them to smash windows to get into the equipment room

could do with input

Kurgis:
Could compromise, by making the vendor in the Examination Room have no access requirements, while the one in the Medbay has them. One issue I noticed is that you just vend them, and it just gives you the items. This doesn't really make sense for non-Moebius staff, since they are just giving free stuff out. Would it be possible to create a new vendor for the entire crew, except the medicine must be purchased? It would put extra pressure for crew to get money.

Additionally, aren't the NanoMeds intended to provide emergency supplies to crew, or am I incorrect?

Voldirs:
No, no, no, IHS should not get meds from moebius for free. Only if they buy them.
And yes, if they are no medics on the ship and you desperately need those supplies, go on, break a window, it's not a problem considering the situation.
Wall mounted vendors are not changed, they have no access requirement.

Wesreidau:
I mentioned this issue to Nanako and I suggested wall-mounted vendors could be actual vending machines, charging credits for gauze, ointment, autoinjectors (5u tricord anyone?) and maybe even pills. These vendors can go in the main halls and accident-prone places to provide emergency care at a cost. This keeps a lot of people from swamping the medical lobby with minor injuries and furthers the medical-for-money narrative.

Voldirs:
Nah, basic medicine, like bandages, ointment should be free, it's essential stuff. But advanced trauma/burn kits, and drugs should have a cost. Anyway, it's irrelevant for now because all vendors are owned by guild.

PR #2585, opened 2019-01-10
AI Law Update

### LINKED ISSUE #1301: Fix AI's lawset

NanakoAC:
Asimov is really not working out in practise, it was always meant to be temporary.

This PR reworks the nanotrasen lawset and sets it back to the default. Naturally all mentions of NT are eris now.
Although these laws aren't explicitly hierarchical, the obey law is nevertheless moved to the top to pointedly illuminate the servile status of AIs. They shouldn't be overrulling command staff

fixes #1301 
also fixes a leftover global debug message in mind.dm


code/datums/ai_law_sets.dm
```diff
@@ -11,22 +11,22 @@
 	..()
 
 /******************** Nanotrasen/Malf ********************/
-/datum/ai_laws/nanotrasen
+/datum/ai_laws/eris
 	name = "NT Default"
```
Kurgis: Shouldn't this be renamed to "Eris Default"?

code/game/objects/items/weapons/AI_modules.dm
```diff
@@ -284,7 +284,7 @@ AI MODULES
 	name = "'NT Default' Core AI Module"
```
Kurgis: Shouldn't this be renamed to "Eris Default"?

code/datums/ai_law_sets.dm
NanakoAC: GOOD POINT
will fix

code/game/objects/items/weapons/AI_modules.dm
LethalGhost: May be `/obj/item/weapon/aiModule` should rebuild its name in `Initialize`

PR #2587, opened 2019-01-10
Blob, Gravity and Power event tweaks

NanakoAC:
Some iterative rebalances based on player observation:

Blob nerfed a bit. Specifically:
The core no longer regenerates health, so heroic sacrifices to kill it are more likely to succeed.  And will at least make it easier to kill later.
Blobs no longer do a special powerful attack when attempting to grow into someone. Instead they will now do their normal attack and growing will not fail. This generally makes them do less damage, and using acid resistant clothing is more effective. However they will grow under your feet and keep doing damage every second while standing on them
Slightly reduced blob health values generally.

Gravity Event: 
Majorly reduced weight, this one is a pain and it doesnt need to be so common

Grid Check/Power loss event:
Majorly reduced weight, but roughly doubled duration. It's more serious when it happens, and players are encouraged to find lights and adapt to it for a while rather than just waiting on it to pass.
Added in critical area checks to prevent this event from causing the AI to run out of power.


code/game/gamemodes/events/apc_damage.dm
```diff
@@ -44,3 +44,6 @@
 /proc/is_valid_apc(var/obj/machinery/power/apc/apc)
 	var/area/A = get_area(apc)
 	return !(A && (A.flags & AREA_FLAG_CRITICAL)) && !apc.emagged && isOnShipLevel(apc)
+
```
mattroks101: More blank lines. These aren't needed are they?

NanakoAC: is there a quick and easy way to remove them?

mattroks101: Undoing what you did in the file to create them? I guess it doesn't hurt anything it just creates clutter in the diff.

NanakoAC:
blankspace fixed

PR #2596, opened 2019-01-11
Make splash occur only on harm intent

### LINKED ISSUE #2546: Harm Intent to Splash Beakers
Gray: SUGG: Make it so you need to be on harm intent to splash beakers on stuff.

hitsedesen:
closes #2546


LagannPM:
What about sink interaction ?

hitsedesen:
image

LagannPM:
it should be "into the sink" instead.

hitsedesen:
Why so?

LagannPM:
and reagents should be just removed if you dispose them into sink, as they are disposed safely.
Otherwise you just splash content of the beaker all over the sink.

code/modules/reagents/reagent_containers/glass.dm
```diff
@@ -108,6 +97,17 @@
 		else
 			name = "[base_name] ([label_text])"
 
+/obj/item/weapon/reagent_containers/glass/beaker/pre_attack(atom/a, mob/user, params)
+	if(user.a_intent == I_HURT)
+		if(standard_splash_mob(user, a))
+			return 1
```
seekin4u: Code convention says to use TRUE/FALSE instead of 1/0

NanakoAC:
Please clarify the state of sink interaction now

hitsedesen:
`splash onto`, as above

EvenInDeathIStillServe:
It should be "You pour the solution down the sink."
Because that's what you do, you don't splash water on a sink: it's just going to get the floor wet. You deliberately pour it into the sink, and it goes down the drain.

NanakoAC:
change it to pouring chemicals down the sink please

PR #2601, opened 2019-01-11
QOL Door interactions, apc lights, inhand ID door interaction

LagannPM:
Clicking on turf where is door located with empty hand will toogle it
Struggle no longer while trying to click on closed door icon


code/_onclick/ai.dm
```diff
@@ -249,4 +253,38 @@
 /obj/machinery/telecomms/AICtrlClick()
 	Topic(src, list("input"="toggle"))
 	return 1
-*/
\ No newline at end of file
+*/
+
+//QOL feature, clicking on turf can toogle doors
+/turf/AICtrlClick()
```
NanakoAC: Needs user here and other places in this file

PR #2605, opened 2019-01-12
Fix bug when items stucks in mob

### LINKED ISSUE #2511: Held items bugging out
Issue persisting from months ago that ghost was meant to fix, items in the hand can enter a bugged state where they are located somewhere else but the hand still holds a reference to them, thus preventing that hand from picking up anything else

Ghost was supposed to fix this

LethalGhost:
Fix #2511


NanakoAC:
I've seen this bug ocurring on a lot more than embedding. It often occurs when placing things on a table, or into a machine. This will likely not fix it. Seems good anyways though

PR #2607, opened 2019-01-12
Openspace vis_contents

Camper12345:
* Openspaces now use vis_contents as a displaying method
* Floor plane system. Object on different floors now have different planes.
      - It seems like nothing is broken.


maps/CEVEris/_map_data_Eris.dm
```diff
@@ -203,28 +206,27 @@ ADMIN_VERB_ADD(/client/proc/test_MD, R_DEBUG, null)
 	var/is_accessable_level = TRUE // Prob modifier for random access (space travelling)
 	var/generate_asteroid= FALSE
 	var/tmp/z_level
+	var/height = -1	///< The number of Z-Levels in the map.
```
LethalGhost: Why default height is -1?

NanakoAC:
This is going to need a lot more explanation

what exactly is vis contents, why make this change, what are the benefits to us?

maps/CEVEris/_map_data_Eris.dm
Camper12345: This is for no one forgot to set the height ont he map. It's asserted at the New()

LethalGhost:
@NanakoAC 
`vis_contents` is new feature what was introdused in byond 512.
It's allow objects to be dispayed on another objects. Like overlays but respect dispayed objects changes (transform matrix, icon, dir changes, etc)

That PR replace original open space code (based on overlays for objects and "shadows" for mobs).
Key benefits is:
- Native support instead custom overlays based code
- Smoth movement for objects visible through open space
- Respect to all changes for objects visible through open space
- It's allow extended multi-z interations for players (overlays-system bounds objects to turf, `vis_contents` one - not)


I'm also agree. That PR required careful checks it can couse some bugs with multi-z interactions with objects and mobs.

LethalGhost:
@NanakoAC 
There was lots of disscussion about that in Shipyard last few days.
But.. well... probably mostly in russian.

code/__defines/_planes+layers.dm
```diff
@@ -136,16 +136,16 @@ What is the naming convention for planes or layers?
...
 //HUD layer defines
 
-#define FULLSCREEN_PLANE 18
+#define FULLSCREEN_PLANE 32000
 #define FLASH_LAYER 18
 #define FULLSCREEN_LAYER 18.1
 #define UI_DAMAGE_LAYER 18.2
 
-#define HUD_PLANE 19
+#define HUD_PLANE 32680
```
NanakoAC: why are these numbers so high? Not that it matters, but seems rather arbitrary

code/game/objects/items/weapons/tools/tape.dm
```diff
@@ -126,24 +126,36 @@
...
 /obj/item/weapon/ducttape/proc/attach(var/obj/item/weapon/W)
 	stuck = W
 	W.forceMove(src)
-	if (istype(W, /obj/item/weapon/paper))
-		icon_state = W.icon_state + "_taped"
-		overlays = W.overlays
+	update_icon()
+	name = W.name + " (taped)"
+
+/obj/item/weapon/ducttape/update_icon()
```
NanakoAC: why is duct tape modified here?

code/game/turfs/simulated/floor_icon.dm
```diff
@@ -308,7 +308,7 @@ var/list/flooring_cache = list()
 	if(!flooring_cache[cache_key])
 		var/image/I = image(icon = 'icons/turf/damage_overlays.dmi', icon_state = icon_base)
 
-		I.plane = FLOOR_PLANE
```
NanakoAC: when would a floor not be on the floor plane?

code/modules/lighting/lighting_overlay.dm
```diff
@@ -17,6 +17,16 @@
 
 	var/needs_update = FALSE
 
+/*/atom/movable/lighting_overlay/init_plane()
```
NanakoAC: some commented out code that seems to have been added here?

code/modules/mob/inventory.dm
```diff
@@ -149,7 +149,7 @@
 	if (src.client)
 		src.client.screen -= I
 	I.layer = initial(I.layer)
-	I.plane = initial(I.plane)
+	I.set_plane(initial(I.plane))
```
NanakoAC: set_plane seems to only be used in a few cases, while many examples of directly setting the plane remain.
Why so?

maps/CEVEris/_map_data_Eris.dm
NanakoAC: Are you sure? It looks like it was asserted at new, but you've deleted this ? https://i.imgur.com/cjpEcXw.png

code/__defines/_planes+layers.dm
Camper12345: It's close to 32767 which is maximum plane value. No z-level planes should be higher than hud, so this is good way to ensure it

code/game/objects/items/weapons/tools/tape.dm
Camper12345: Duct tape allows you to tape items to the walls and uses some plane magic for this

code/game/turfs/simulated/floor_icon.dm
Camper12345: Floor plane is no longer constant, but decal still needs to be be on the same plane as the floor

code/modules/mob/inventory.dm
Camper12345: There is not so many places where plane is assigned, excluding object definitions. I changed almost all of them.

PR #2616, opened 2019-01-13
Examine Tcomms machinery to see damage

EvenInDeathIStillServe:
TComms machinery can now be examined to know if they are damaged or just unpowered. Different message depending on severity of damage.

### DIFF
--- a/code/game/machinery/telecomms/machine_interactions.dm
+++ b/code/game/machinery/telecomms/machine_interactions.dm
@@ -14,6 +14,23 @@
 	var/temp = "" // output message
 	var/construct_op = 0
 
+/obj/machinery/telecomms/examine(mob/user)
+	..(user)
+	switch(integrity)
+		if(100)
+			user << SPAN_NOTICE("It's in perfect condition.")
+		if(85 to 99)
+			user << "It has some minor faults."
+		if(65 to 85)
+			user << "It's slightly damaged."
+		if(45 to 65)
+			user << SPAN_WARNING("It's badly damaged.")
+		if(25 to 45)
+			user << SPAN_WARNING("It's heavily damaged.")
+		else
+			user << SPAN_WARNING("It's falling apart.")
+	if(stat & BROKEN)
+		user << SPAN_DANGER("It's broken.")
 
 /obj/machinery/telecomms/attackby(obj/item/I, mob/user)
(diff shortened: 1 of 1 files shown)


code/game/machinery/telecomms/machine_interactions.dm
```diff
@@ -14,6 +14,23 @@
...
+		if(65 to 85)
+			user << "It's slightly damaged."
+		if(45 to 65)
+			user << SPAN_WARNING("It's badly damaged.")
+		if(25 to 45)
+			user << SPAN_WARNING("It's heavily damaged.")
+		else
+			user << SPAN_WARNING("It's falling apart.")
+	if(stat & BROKEN)
+		user << SPAN_DANGER("It's broken.")
```
LethalGhost: Isn't it better to check `stat` first and if it's not `BROKEN` check `integrity`?
That seems right now you can get message like _"This is telecomms machinry. It's in perfect condition It's broken."_

LagannPM:
Isn't it better to move this code to code/game/machinery instead ?

EvenInDeathIStillServe:
No, because integrity is a variable specific to Telecomms machinery. Autolathes, gibbers, etc do not get damaged.

code/game/machinery/telecomms/machine_interactions.dm
EvenInDeathIStillServe: Tcomms machinery has two separate damage conditions: integrity, which is caused by overheating, and the BROKEN stat, which only happens when you remove cables from it while disassembling it.

I have changed the broken message to better reflect its status.

LagannPM:
We actually need a destroyable autolathes, gibbers, etc. , if this not a problem for you can you move integrity variable, examine code, harm intent interactions and repair code to /obj/machinery instead ?

EvenInDeathIStillServe:
That is acceptable. I'll see what I can do.

NanakoAC:
@ITCups @EvenInDeathIStillServe Please give me a status report on this PR

EvenInDeathIStillServe:
Haven't done anything other than my initial commit. I can try to get it done in about a week.

NanakoAC:
requested closed by author, hope he'll have time to work on it again soon

PR #2625, opened 2019-01-14
Welding sounds & messages

EvenInDeathIStillServe:
This PR does 3 things:

- Adds a message and sound effect (from /tg/) when toggling welding tools.
- Adds 4 new sound effects for welding.
~~Makes tool worksounds lists, so more than 1 sound effect can be added, and will be randomly played.~~ _Reverted because there was a much cleaner way to do this._
- Adds a new list for welding sounds, the worksound is randomly picked from that list. Also works for repeating sounds.


EvenInDeathIStillServe:
> This isn't making the most of the new sounds. It needs to randomise the sound file for each time it plays, not just once at the start
> 
> Look at zmethods for how to do this

Done. 

I reverted the change I made that turned the worksounds into lists, since it wasn't necessary. Also, made it so that welding tools spawn sparks throughout the whole task instead of just the start.

code/game/sound.dm
```diff
@@ -95,6 +95,13 @@ var/list/climb_sound = list(
 	'sound/effects/ladder4.ogg'
 )
 
+var/list/weld_sound = list(
+	'sound/items/welding1.ogg',
```
NanakoAC: More variety is good but there's nothing wrong with the old welder sound, add it to this list

EvenInDeathIStillServe: Have you listened to the sounds? They are very different from the old sound, it would seem out of place.

NanakoAC: Yes i have, I disagree. Requested change stands. please add it back in that list

EvenInDeathIStillServe:
Added the old sound to the list.

PR #2629, opened 2019-01-14
Effect system fixes

EvenInDeathIStillServe:
I started this PR to fix a bug which made sparks invisible: Sparks were defined as /obj/effect/sparks, but the icon for effect objects is defined in /obj/effect/effect.

After being disgusted by this redundancy, I proceeded to replace all instances of /obj/effect/effect with the much more acceptable /obj/effect.

I then went on to fix a bug which made plasma gas invisible: its icon_state was "plasma", while the .dmi containing gas icons had "phoron" and "plasma-purple" and no "plasma". I replaced the "phoron" with "plasma-orange" and adjusted the icon_state to that.

Then I redefined sparks to be /obj/effect/temp_visual/sparks.

Also I made it so that when sparks are spawned, the sparks don't move all at once anymore: each one has a random move delay.


NanakoAC:
the redundancy was dumb, but it may have had a reason. This is broad ranging and high risk.

I'll need to ask you to test and confirm a lot of things before we can merge this, including:
Jetpack trails
Smoke and chemsmoke
Chempuffs (from pepperspray and spraybottles)
Fire
Blood
Rubble from burrows

And pretty much any other kind of effect you can think of

LethalGhost:
>  test and confirm a lot of things 

Nothing of that was changed. 90% of PR - replacing `/obj/effect/effect` with new path `/obj/effect`.
That's can't affect mechanics.

EvenInDeathIStillServe:
> > test and confirm a lot of things
> 
> Nothing of that was changed. 90% of PR - replacing `/obj/effect/effect` with new path `/obj/effect`.
> That's can't affect mechanics.

While that may seem true, it appears to be quite the contrary. All the effects work properly, yes, but this change had the side effect of making all effects unclickable, including blood on the ground. Seems like this had a purpose after all. I will rectify by reverting the removal of the extra /effect and instead replace it with the more accurate /visual.

NanakoAC:
Unclickability is caused by the mouse_opacity var. 
Setting it to true will allow things to be clicked. 
Setting it false makes clicks pass through the object, and makes it not show up in the rightclick menu

Theres no need for base effects to have mouse opacity, but decal/cleanable should have them. The proposed effect/visual subtype seems unnecessary

EvenInDeathIStillServe:
Closing this and opening a new PR without messing with /obj/effect/effect.

PR #2638, opened 2019-01-15
Mopping Tweaks III

NanakoAC:
Third and hopefully final (for now) tweaks to mopping.  After spending almost a whole round doing it

General aim here is to make floors stay clean longer once cleaned, so that mopping isn't an eternal chore of re-cleaning the same spots over and over in high traffic areas

Changes:
Maximum dirt is now a flooring var, the floorings used in main hallways and high traffic areas have a higher maximum capacity, and will take 2-2.5x as many crossings to reach maximum dirt level

When sweep mode cleans a tile to below half of max dirt (no longer visible) the dirt is set to 0

Also moved some cleaning related procs and vars into a new file, instead of having them scattered over several turf files


SpadesSlick:
Is it possible to rework moping with new quality - "cleaning"? There are a lot of ways to clean things and floor, with soap for example. Use of janitor mop for cleaning weapons from blood could be interesting this.

NanakoAC:
i'm not sure
not for this PR at least, but i think so. certainly, cleaning things could be tools and have a quality, most of the heavy lifting might still be in the turf code i think

ill think on it

code/game/turfs/flooring/flooring_premade.dm
```diff
@@ -4,6 +4,7 @@
 	icon_state = "carpet"
 	initial_flooring = /decl/flooring/carpet
 
+
```
mattroks101: Unrelated blankspace change.

NanakoAC:
fixed

PR #2653, opened 2019-01-18
Economy Phase 2: Payday

NanakoAC:
This PR reworks, reintroduces, and properly integrates crew wages. 

Crew are now paid once per hour, with an email being sent to notify them of this, as well as the shipwide announcement of payday.

The exact values of payments are a new set of defines:

Dumb Labour: 450. Paid to janitor, actor and guild tech
Normal Labour: 600, paid to everyone not specifically mentioned
Hazard Pay: 750: Paid to a few dangerous jobs like miner and IH operative
Professional: 900: Paid to the educated professionals with fancy degrees. Doctors and scientists and inspectors
Command: 1200: Paid to command staff of each department

The following roles are not paid at all:
Chef and Bartender: These two are expected to make their own money.
Assistant: They can go find a job
Captain and Guild Merchant: They each control a wealthy account and can draw whatever salary they like from it.

Departments themselves are also paid, each department gets an hourly budget infusion to their departmental funds, drawn from the same source as their wages

Payments are properly sourced, all the money comes from somewhere, is transferred into the local department account and then paid from there to individual crewmembers. Every department has funding sources specified.

Retainers:
Ironhammer, Technomancers, Civilian.
These are paid from the ship account

Moebius:
Medical, research. These are paid by external funding from moebius corp.

Church is paid by external funding from the Church of Neotheology

Guild gets no funding, the merchant is expected to make enough money to pay the local guild employees. Miners and techs will go unpaid if he's not selling things.

Other changes:
The Ship Account now starts with three million credits. This is unlikely to run out during a normal round, and is used to pay other departments as outlined above. Captain is encouraged to roleplay with the money and not just make it rain from the catwalks

This is a placeholder, and will eventually be replaced with ways for the captain to earn largescale credits by sending crewmembers on profitable expeditions

Extended crew records a bit with new fields
Some refactoring of accounts and transfers


Voldirs:
>Dumb Labour: actor 

https://cdn3.whatculture.com/images/2014/07/Tobey-Maguire-Spider-Man-Crying_edited-1.jpg

Boredone:
While I might of just missed it, this seems to make it so the Chef, Bartender, and Assistants also start completely broke? Since if I read it right, your wage determines how much you start with, and they get none for various reasons.
Captain and Merchant too of course, but that's probably intentional and fine since they can use their department money instead.
Personally feel like they should at-least start with some money, if not much.

NanakoAC:
> 
> 
> While I might of just missed it, this seems to make it so the Chef, Bartender, and Assistants also start completely broke? Since if I read it right, your wage determines how much you start with, and they get none for various reasons.
> Captain and Merchant too of course, but that's probably intentional and fine since they can use their department money instead.
> Personally feel like they should at-least start with some money, if not much.

good point, will fix

NanakoAC:
Fixed the starting balance for those three mentioned roles

PR #2657, opened 2019-01-18
Ammo loading/unloading fixes and tweaks

LagannPM:
Loading magazines and ammoboxes is a little different now
You can load ammo in magazine by clicking on it with ammo in your hand
You can unload full stack from it by clicking on it with empty hand
You can unload one casing from it by alt clicking with ammo or empty hand on it
***
Stacking ammo from turf is easier now
***
Ammobox quick unload on selfclick is removed, click on turf for that
***
Magazines and ammo can be unloaded from guns by dragging gun in empty hand(just like removing battery)
***
Fixed 32 caliber not fitting in ammo box and a typo ".38"

### DIFF
--- a/code/modules/projectiles/ammunition.dm
+++ b/code/modules/projectiles/ammunition.dm
@@ -68,9 +68,30 @@
 			user << "\blue You inscribe \"[label_text]\" into \the [initial(BB.name)]."
 			BB.name = "[initial(BB.name)] (\"[label_text]\")"
 		return TRUE
-	else if(istype(I, /obj/item/ammo_casing) && (src.amount != src.maxamount) && (src.desc == I.desc))
+	else if(istype(I, /obj/item/ammo_casing))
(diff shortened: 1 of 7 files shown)


Voldirs:
I'd suggest another implementation. Alt-click is too unobvious. If one clicks the mag or box with empty hand he gets full stack of ammo which is 5, or less if there were less ammo left.
If one clicks the mag with stack of ammo, he loads one ammo from that stack.

LagannPM:
I'll leave alt click interaction but ill implement your suggestion

EvenInDeathIStillServe:
If you're going to add alt-click interaction, the game should tell you when you examine the ammo box.

NanakoAC:
I agree with @LagannPM here.
If altclick isn't obvious, its because players haven't learned to use it yet. They will in time. And the more things we add with this control scheme, the more players will grow accustomed to it

PR #2665, opened 2019-01-19
Protolathe Cell Exploit Fix

### LINKED ISSUE #2652: BUG: Silver resource duping possible with protolathe.
Within the Protolathe menu is the option for both advanced reagent scanners and spectrometers.
brp1er
After printing, they come immediately with small batteries pre-installed despite costing no silver for printing.
brp2er
These can be recycled for silver after being removed, after which, you can recycle the devices themselves for a partial return on resources spent on the silver.
brp3er

NanakoAC:
fixes #2652 
Fixes missing Created() procs causing certain protolathed devices to spawn with cells in them, allowing endless duplication of silver.

The solution is thorough, but a little hacky. It'll do until we make centralised power supply datums, and standardise cell behaviour across all objects


cev_eris.dme
```diff
@@ -2143,6 +2144,7 @@
 #include "code\modules\recycling\disposal.dm"
 #include "code\modules\recycling\sortingmachinery.dm"
 #include "code\modules\research\circuitprinter.dm"
+#include "code\modules\research\debug.dm"
```
mattroks101: debug still here though

NanakoAC: oh no that's deliberate.

This is a debug object that maxes out research levels when fed into the destructive analyzer, there's no way to obtain it ingame. It has potential usefulness for admin shenanigans, and it will be useful to have active as a debugging tool in future

mattroks101: If you say so.

PR #2671, opened 2019-01-19
Adds looping sounds, ambience for supermatter & TEG

EvenInDeathIStillServe:
Ports looping sound framework from /TG/.

Supermatter and the TEG now have ambient sounds. No more completely silent engine rooms.

TEG sound is licensed under CC0. https://freesound.org/people/qubodup/sounds/189896/
Supermatter sound was taken from /TG/ with no license cited.

Supermatter volume maxes out at 2000 power, TEG volume maxes out at 500KW.
My hearing is absolutely terrible, so testing is welcome to see if the max volume needs to be toned down.


NanakoAC:
First thoughts:
Theres a constant background sound at the start, when nothing is running. This should not be.

Before the engine is started, when everything is cold and dead, the engine room should be entirely silent. The ambient sound should only come when things are started up

As for the teg sounds, i'm hearing those and they're quite loud, but the output is only this teg sounds

Definitely doesn't feel right, it needs to start up a lot quieter, and scale only gradually

The range on the teg sounds is also odd. Having a limited range is good, but it feels a bit too small. I walk a couple paces away and it cuts out bizarrely. Maybe a bit more range is in order

It would perhaps be good to have tiered volume, with the range and volume increasing in discrete, noticeable jumps as the teg output hits certain targets. I reckon that would be more atmospheric, and provide good feedback of your progress in starting up

Also for reasons i can't quite discern, theres an odd ringing noise in the lower level. I'm not sure if this is intended. Maybe its the supermatter itself, but closing the blast doors doesn't seem to baffle it

With extended listening, the noise becomes rather repetitive. Have you tried experimenting with pitch variance? It sounds like it set to have no variance right now

Overall though, good feature, good idea, implementation needs some work. Lets refine and test it in stages

Wesreidau:
The diesel generator sound is nothing like what a TEG would sound like, if those make any real sound at all. Here's a recording of a "power generator" at CERN.
https://freesound.org/people/tcharollais/sounds/221858/

And an artificial "nuclear power plant" sound created for ambience. I think it suits the SM nicely.
https://freesound.org/people/DarkMime/sounds/418983/

NanakoAC:
Wesreidau is Nikov on discord, he's a regular engineering player and has applicable real life experience in his field. Can you two discuss this feature between you and let me know what you come up with?

EvenInDeathIStillServe:
I've switched the sounds with the ones provided by Nikov. Supermatter noise is no longer a constant ringing, but now ominous waves with sinusoidal volume.
Sounds are now overall a lot quieter, as they should be.
I've added functionality for giving a looping sound extra range and varying frequency and gave it to the SM and TEG.
Sounds should no longer be annoying, and can now be considered proper ambiance.

code/game/sound.dm
```diff
@@ -340,6 +340,14 @@ var/const/FALLOFF_SOUNDS = 0.5
 
 	src << S
 
+#define CHANNEL_HIGHEST_AVAILABLE 1014
+
+/proc/open_sound_channel()
```
hitsedesen: `sound_channels`

NanakoAC:
requested closed by author, hope he'll have time to work on it again soon

PR #2683, opened 2019-01-21
Swaps Suits to use Primarily Extra_Allowed

Boredone:
Swaps majority of /clothing/suits to use Extra_Allowed instead of allowed, letting a large majority of suits carry things like guns, knives, and emergency oxygen tanks.
Rigsuits and Spacesuits also gained access to weapon/material stuff, like Fire Axes.


NanakoAC:
For quick reference, the base list contains:

	allowed = list(
	/obj/item/weapon/storage/pouch/,
	/obj/item/weapon/gun,
	/obj/item/weapon/melee,
	/obj/item/weapon/material,
	/obj/item/ammo_magazine,
	/obj/item/ammo_casing,
	/obj/item/weapon/handcuffs,
	/obj/item/weapon/tank,
	/obj/item/device/suit_cooling_unit,
	/obj/item/weapon/cell)

It's pretty comprehensive and should cover most cases

Boredone:
Ah, that's going to be a much larger buff then I planned for. Going to make some adjustments to that list though, since for example Suit Cooling doesn't really do much if you don't got pressure resistance, and just weapon/material lets you fit stuff like Fire Axes on your jacket, which feels iffy.

NanakoAC:
fire axes in the suit slot is quite intended, it represents another slot for something strapped around your torso, and should fit any kind of weapon at all, especially the largest ones.

code/modules/clothing/clothing.dm
```diff
@@ -438,24 +438,24 @@ BLIND     // can't see anything
...
+		/obj/item/weapon/melee,
+		/obj/item/weapon/material/kitchen,
+		/obj/item/weapon/material/knife,
+		/obj/item/weapon/material/butterfly,
+		/obj/item/ammo_magazine,
+		/obj/item/ammo_casing,
+		/obj/item/weapon/handcuffs,
+		/obj/item/weapon/tank/emergency_oxygen,
+		/obj/item/weapon/cell,
+		/obj/item/device/lighting/toggleable/flashlight)
```
NanakoAC: This is just reverting a lot of my recent work, no please.

Suit storage is a very large slot for something strapped over your torso, it should hold almost anything, weapon/material and weapon/tank were fine

revert changes to this list please

NanakoAC:
The mentality here should be for a longterm fix. We don't need a ton of very specific items like matchboxes or flashlights in these hardcoded allowed lists. The suit storage should be given the lowest level of common type which makes sense. Generally i want to remove a lot of the frustration, wierdness and special cases with trying to figure out what a suit holds. most things should fit most suits, without needing to think about it too much

Extra allowed should be used sparingly, for a few things that you might find special holders for in certain suits. It should not have a big bunch of things like you've done here

code/modules/clothing/clothing.dm
Boredone: Ah, whoops. Was wondering why there was two lists.

code/modules/clothing/clothing.dm
```diff
@@ -438,24 +438,27 @@ BLIND     // can't see anything
...
+		/obj/item/weapon/material,
+		/obj/item/ammo_magazine,
+		/obj/item/ammo_casing,
+		/obj/item/weapon/handcuffs,
+		/obj/item/weapon/tank,
+		/obj/item/device/suit_cooling_unit,
+		/obj/item/weapon/cell,
+		/obj/item/weapon/storage/fancy/cigarettes,
+		/obj/item/weapon/flame/lighter,
+		/obj/item/device/lighting/toggleable/flashlight,
```
NanakoAC: I'm thinking these three entries could be reduced to 
/obj/item/weapon/storage/fancy
/obj/item/weapon/flame
/obj/item/device/lighting

Any thoughts on wierdness that might result? anything in those types that you think shouldn't fit there?

code/modules/clothing/spacesuits/rig/rig.dm
```diff
@@ -97,16 +97,19 @@
 	allowed = list(
```
NanakoAC: this list duplicates most of the stuff in the main list now, so delete it

code/modules/clothing/suits/armor.dm
```diff
@@ -1,6 +1,6 @@
 
 /obj/item/clothing/suit/armor
-	allowed = list(/obj/item/weapon/gun/energy,/obj/item/device/radio,/obj/item/weapon/reagent_containers/spray/pepper,/obj/item/weapon/gun/projectile,/obj/item/ammo_magazine,/obj/item/ammo_casing,/obj/item/weapon/melee/baton,/obj/item/weapon/handcuffs)
+	extra_allowed = list(/obj/item/device/radio,/obj/item/weapon/reagent_containers/spray/pepper)
```
NanakoAC: instead of that last entry, add
 /obj/item/weapon/reagent_containers/spray

to the base allowed list

code/modules/clothing/suits/armor.dm
```diff
@@ -214,8 +214,8 @@
 	desc = "A synthetic armor vest. This one has added webbing and ballistic plates."
 	icon_state = "webvest"
 	armor = list(melee = 50, bullet = 45, laser = 50, energy = 25, bomb = 30, bio = 0, rad = 0)
-	allowed = list(/obj/item/weapon/gun/energy,/obj/item/device/radio,/obj/item/weapon/reagent_containers/spray/pepper,/obj/item/weapon/gun/projectile,/obj/item/ammo_magazine,/obj/item/ammo_casing,/obj/item/weapon/melee/baton,/obj/item/weapon/handcuffs)
-
+	extra_allowed = list(/obj/item/device/radio,/obj/item/weapon/reagent_containers/spray/pepper)
```
NanakoAC: ditto here

code/modules/clothing/suits/jobs.dm
```diff
@@ -63,7 +40,10 @@
 	item_state = "apron"
 	blood_overlay_type = "armor"
 	body_parts_covered = 0
-	allowed = list (/obj/item/weapon/reagent_containers/spray/plantbgone,/obj/item/device/scanner/analyzer/plant_analyzer,/obj/item/seeds,/obj/item/weapon/reagent_containers/glass/fertilizer,/obj/item/weapon/material/minihoe)
+	extra_allowed = list (\
+		/obj/item/weapon/reagent_containers/spray/plantbgone,/obj/item/device/scanner/analyzer/plant_analyzer,
```
NanakoAC: instead of the plant analyzer, add /obj/item/device/scanner to the base allowed list

code/modules/clothing/suits/utility.dm
```diff
@@ -107,10 +104,7 @@
 	gas_transfer_coefficient = 0.90
 	permeability_coefficient = 0.50
 	body_parts_covered = UPPER_TORSO|LOWER_TORSO|LEGS|ARMS
-	allowed = list(
-		/obj/item/device/lighting/toggleable/flashlight,/obj/item/weapon/tank/emergency_oxygen,
-		/obj/item/clothing/head/radiation,/obj/item/clothing/mask/gas
-	)
+	extra_allowed = list(/obj/item/clothing/head/radiation,/obj/item/clothing/mask/gas)
```
NanakoAC: add /obj/item/clothing/mask to the base allowed list

Issue #2575, opened 2019-01-09
Bug - In-Stomach Cooking Recipies


Wesreidau:
Gray: BUG: Eating meat and flour makes you shit meatballs.

Nikov: Old bug I've seen on a few stations, caused by reagent combinations inside the mob. You hear the bubbling chemistry sound too.

PR #2685, opened 2019-01-21
No more shitting meatballs upon consuming protein and flour.

LudwigVonChesterfield:
I have taken a very simple approach to solve #2575, is just outright ban these recipes inside the mob.

Perhaps a better solution would be to force a mob to ingest whatever has been created as a result of such recipe. That I leave for a judgement of somebody else.

### DIFF
--- a/code/modules/reagents/recipes.dm
+++ b/code/modules/reagents/recipes.dm
@@ -52,21 +52,30 @@
 	var/mix_message = "The solution begins to bubble."
 	var/reaction_sound = 'sound/effects/bubbles.ogg'
 
+	var/list/require_containers = list() // This reaction will only occure in these containers(Or their subtypes).
+	var/list/blacklist_containers = list() // This reaction will not occure in these containers(Or their subtypes).
+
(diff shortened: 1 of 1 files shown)


GrayRachnid:
does this stop mixing inaprov and dylo in the body?

LudwigVonChesterfield:
No, should it?

GrayRachnid:
Nop

NanakoAC:
hmmm, i'm not 100% convinced that this is the ideal solution, but it will do. and the feature does seem potentially useful

code/modules/reagents/recipes.dm
```diff
@@ -52,21 +52,30 @@
...
 	if(!holder.has_all_reagents(catalysts))
-		return 0
+		return FALSE
 
 	//check that none of the inhibitors are present in the required amount
 	if(holder.has_any_reagent(inhibitors))
-		return 0
+		return FALSE
+
+	if(require_containers.len && !is_type_in_list(holder.my_atom, require_containers))
```
NanakoAC: does this run istype on all the things in this list?

LudwigVonChesterfield: Yes, exactly.

PR #2692, opened 2019-01-22
Design Materials Refactor

NanakoAC:
I hated doing this, but it needed to be done after some recent changes exposed flaws

This PR refactors all of the science designs to calculate their material costs the same way autolathes do - by reading them directly from the matter of the target object. And thus ensuring that a portion of the matter can be reclaimed by recycling it. 

With this, there should now be absolutely nothing in the game that can be printed, which can't also be recycled
A couple costs were manually changed, circuitboards now broadly require silicon (which, naturally, can be obtained by recycling other circuit boards)

Most of the costs weren't changed, but simply moved out of the design datum and onto the matter of the object instead. Hence the vast number of files changed.
Setting costs on the design datum is still allowed, and these are added to the matter costs. Costs on the design datum are now intended for use as "extra" costs associated with manufacture, but which don't show up in the finished product. The prime example being sulphuric acid

Finally, some small tweaks to cargo. Greatly reduced the cost of material crates, standardised prices of specific materials, and added a bulk metal crate to the guild ordering. will likely be used by robotics

### DIFF
--- a/code/modules/cargo/packs.dm
+++ b/code/modules/cargo/packs.dm
@@ -473,29 +473,43 @@ var/list/all_supply_groups = list("Operations","Security","Hospitality","Enginee
 	crate_name = "Replacement lights"
 	group = "Engineering"
 
-/datum/supply_pack/metal50
-	name = "50 metal sheets"
+/datum/supply_pack/metal120
+	name = "120 metal sheets"
 	contains = list(/obj/item/stack/material/steel)
-	amount = 50
-	cost = 1000
+	amount = 120
+	cost = 360
 	containertype = /obj/structure/closet/crate/secure
 	crate_name = "Metal sheets crate"
 	group = "Engineering"
 
-/datum/supply_pack/glass50
-	name = "50 glass sheets"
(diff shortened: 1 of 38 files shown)


code/modules/cargo/packs.dm
```diff
@@ -473,29 +473,43 @@ var/list/all_supply_groups = list("Operations","Security","Hospitality","Enginee
...
 	group = "Engineering"
 
 /datum/supply_pack/wood50
-	name = "50 wooden planks"
+	name = "120 wooden planks"
 	contains = list(/obj/item/stack/material/wood)
-	amount = 50
-	cost = 1000
+	amount = 120
+	cost = 1200
```
seekin4u: You may use defines for defining one sheet price and than make `cost = ammount * WOOD_SHEET_PRICE`, i think this would be better than multiplying these numbers with your hands

NanakoAC: I will do this later in future economy work, since there are other places it will need to be written in too

PR #2700, opened 2019-01-23
Wedging tools into airlocks to keep 'em open.

### LINKED ISSUE #2617: Sugg - Crowbar Wedging Door Open
So I play a lot of engineering. Right now, you can cut power and pry open a door with a crowbar. This is fine, although the failure rate and subsequent hunt for backup power wires can get annoying. What isn't fine is how the doors will close on engineers while they are working on it. Sometimes you need to keep a door open when you can't drop the bolts, or to allow passage of multiple people,  or to resist an AI chomping on someone. So I'm thinking of the following.

With the crowbar in hand, pry an unpowered door or firelock open. That door is now "wedged" open on the crowbar. So long as you are one tile from the door and still holding the crowbar, the door remains wedged open. When you take the crowbar out of your hands or step away from the door, it closes. You can get screwdrivers etc into your other hand and work on the door.

Likewise, with any powered door that is already open a crowbar can be jammed into it, wedging it open. Useful when you don't trust the door to stay open, when you need to keep a door open without hacking the bolts down. Also logically consistent.

This is of particular use with firelocks, as the current code has them closing too fast for the door to be worked on under it, say if I need to hack into a secure section which is also breached to space (common engineering task). What I have seen other servers do is have the firelocks stay forced open if they are opened with a crowbar. However once someone escaping a breach has cleared the lock, it will stay open. By having the escapee wedge it open with the crowbar only so long as they are next to it, it will be open only long enough for they and their friends to escape, or for them to succumb to a plasma fire. After which the door closes and the firelock resumes controlling the damage.

The AI, by the way, should not be able to force the door with the crowbar jammed in it. Trying to move one heavy piece of steel in a track down another heavy piece of steel with a steel wedge in the way is more likely to destroy the airlock, a complex machine, than the piece of carbon steel.

LudwigVonChesterfield:
closes #2617

The title sums it all up pretty neatly.

image

* It may look derpy with tools that are not crowbar.

You wedge a tool in by holding it in hand, and in right-click dropdown menu of airlock clicking "Wedge item", or by putting it on the same tile as the airlock.

You take the tool out in one of three ways: dropdown menu "Take out wedged item", grab intent-click, mousedroping the airlock onto you.

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -1,3 +1,5 @@
+GLOBAL_LIST_EMPTY(wedge_icon_cache)
+
 /obj/machinery/door/airlock
 	name = "Airlock"
 	icon = 'icons/obj/doors/Doorint.dmi'
@@ -40,6 +42,7 @@
 	var/open_sound_unpowered = 'sound/machines/airlock_creaking.ogg'
 	var/_wifi_id
 	var/datum/wifi/receiver/button/door/wifi_receiver
+	var/obj/item/wedged_item
 
 /obj/machinery/door/airlock/attack_generic(var/mob/user, var/damage)
 	if(stat & (BROKEN|NOPOWER))
@@ -417,7 +420,7 @@ There are 9 wires.
 
 
 
-/obj/machinery/door/airlock/bumpopen(mob/living/user as mob) //Airlocks now zap you when you 'bump' them open when they're electrified. --NeoFite
+/obj/machinery/door/airlock/bumpopen(mob/living/user) //Airlocks now zap you when you 'bump' them open when they're electrified. --NeoFite
 	if(!issilicon(usr))
 		if(src.isElectrified())
 			if(!src.justzap)
@@ -433,10 +436,7 @@ There are 9 wires.
 			user.halloss += 10
 			user.stunned += 10
 			return
-	..(user)
-
-/obj/machinery/door/airlock/bumpopen(mob/living/simple_animal/user as mob)
-	..(user)
+	..()
 
 /obj/machinery/door/airlock/proc/isElectrified()
 	if(src.electrified_until != 0)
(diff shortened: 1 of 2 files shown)


code/game/machinery/doors/airlock.dm
```diff
@@ -1,3 +1,5 @@
+var/global/list/wedge_icon_cache = list()
```
seekin4u: У нас GLOB есть, в общем то

ITCups:
Verbs are very bad way of doing interactions, you might want another method.
Maybe just dropping tool on ground might be better solution.

LudwigVonChesterfield:
Verbs in context of things that are one pixel wide(open airlock) are great, however I will take your recommendation and add it as another way

code/game/machinery/doors/airlock.dm
LudwigVonChesterfield: GLOB был ошибкой когда есть world переменные

ITCups: Glob уменьшает зону видимости переменной
Когда ипользуешь глоб ты точно знаешь что она глобальная и ктонибудь другой случайно не достучится до нее не используя GLOB.

NanakoAC:
I'm testing now, so far its lacking audio polish. The feature is working, the tool is taking damage, but the only way i can see that is by var viewing.

The door needs to audibly attempt to close, make at least the buzzer sound, maybe some new machinery grinding sound too. Maybe even play its closing animation halfway then revert it, i believe this could be done with flick. And definitely make the door do a Shakeanimation, that proc is located in helpers/matrices

code/game/machinery/doors/airlock.dm
```diff
@@ -570,10 +570,77 @@ There are 9 wires.
...
+	var/obj/item/weapon/tool/T = usr.get_active_hand()
+	if(istype(T) && T.w_class >= ITEM_SIZE_NORMAL) // We do the checks before proc call, because see "proc overhead".
+		if(!density)
+			force_wedge_item(T)
+			to_chat(usr, SPAN_NOTICE("You wedge [T] into [src]."))
+		else
+			to_chat(usr, SPAN_NOTICE("[T] can't be wedged into [src], while [src] is open."))
+
+/obj/machinery/door/airlock/proc/take_out_wedged_item(mob/living/user)
+	set name = "Take wedged item out"
```
NanakoAC: "Take wedged item out" is a somewhat overlong and clumsy name for a verb, please revise

Possible alternative, Remove Blockage

code/game/machinery/doors/airlock.dm
```diff
@@ -1056,14 +1134,19 @@ There are 9 wires.
...
 /obj/structure/closet/airlock_crush(var/crush_damage)
 	..()
 	damage(crush_damage)
 	for(var/atom/movable/AM in src)
 		AM.airlock_crush()
 	return 1
 
+/obj/item/weapon/tool/airlock_crush(crush_damage)
+	. = ..() // Perhaps some function to this was planned, however currently this proc's return is not used anywhere, how peculiar. ~Luduk
+	unreliability += crush_damage * degradation * (1 - get_tool_quality(QUALITY_PRYING) * 0.01)
```
NanakoAC: In practise, the rate of damage seems way too high, throw in a *0.4 multiplier in here ontop of the other calculations

NanakoAC:
Final comment, add the ability to wedge a tool in using altclick on the door with it in active hand. There should be functionality existing to do this

Wesreidau:
As the original suggestor:

Could this be caused by using the harm or maybe grab or disarm intent on an open airlock with the tool in question? The door closing on a crowbar naturally and getting jammed open seems like a bad approach. Stray tools fly around and this should be a deliberate action, not an accidental one.

LudwigVonChesterfield:
You can do it deliberately, and I did describe how so in the PR title, however I purposefully do not allow door-jamming while it is open.

code/game/machinery/doors/airlock.dm
Wesreidau: "Unjam Airlock" "Unwedge Airlock"

Wesreidau:
Yes, but can't someone grab / disarm intent-click the crowbar onto the open airlock to wedge it? And you say you purposefully don't allow door-jamming while it is open. This confuses me, since the point of door wedging is keeping an open door open.

NanakoAC:
This is almost ready, however there's currently no indication for the user that the door is repeatedly attempting to close. 
When i said about doing a shake and sfx, i meant for each individual attempt at closing, not only the first time.

Probably those noises would get annoying and repetitive, so shake only will be fine. Make it shake every time the tool gets crushed and takes some damage

PR #2713, opened 2019-01-24
Contribution Guidelines updated

NanakoAC:
A first draft of new contribution guidelines. Much of what was already there was fine, and is kept, but quite a few things added

This should by no means be considered the final word on the subject, but more of a step foward


CONTRIBUTING.md
```diff
@@ -128,45 +159,11 @@ del(src)
 ```
 ***
 
-Do not return unused values from functions. Do not use return if there's no any actions in function after it.
```
mattroks101: This is telling you to not return values arbitrarily, this is a good practice. If you do plan on using that value at a later date than it's fine. This is not a helpful thing to delete.

CONTRIBUTING.md
```diff
@@ -128,45 +159,11 @@ del(src)
...
-```
-***Bad:***
-```
-proc/mutate_count(var/obj, var/value)
-	obj.count = value
-	return 1
-```
-***
-
-Instance variables should be used with only with source object denotation to avoid namespace conflicts.
```
mattroks101: This is telling you not to use `src` arbitrarily and only use it when you have two conflicting variable names. Ideally you would probably name it something else, but this is also sound advice and tells people the only acceptable instance of `src`. Unless you're going to re-write this to offer similar advice about using `src`, I do not think this should be deleted.

CONTRIBUTING.md
```diff
@@ -128,45 +159,11 @@ del(src)
...
-	src.name = name
-```
-***Bad:***
-```
-/obj/item/
-	var/name = ""
 
-/obj/item/proc/set_name(name)
-	name = name
-	# You will probably get a runtime error.
```
mattroks101: The `#` here is weird though.

CONTRIBUTING.md
```diff
@@ -86,7 +117,7 @@ Boolean variables and return values should use TRUE and FALSE constans instead o
 ```
 ***
 
-Using colon operator (:) for object property and procs access is deprecated.
+Using colon operator (:) for object property and procs access is generally inadviseable
```
mattroks101: Inadvisable. Also it is deprecated I'm not sure why you changed that to "generally inadvisable". You shouldn't be using it at all.

CONTRIBUTING.md
```diff
@@ -179,17 +176,7 @@ var/count = 0
 var/c = 1
 ```
 ***
-Avoid digits in names.
```
mattroks101: Digits **should** be avoided in names. As having a millions things names var/text1, var/text2, var/text3 makes code borderline unreadable. This is also good sound coding advise that you are removing. Please do not encourage bad coding styles and practices.

CONTRIBUTING.md
NanakoAC: Its not a sensical rule, this is like telling people to brush their teeth before bed, and not to set their house on fire.  It's not worth telling people not to do something so harmless and afaik uncommon

CONTRIBUTING.md
NanakoAC: It doesn't say that at all. Infact i'm pretty sure the above line is telling you to always use src.thing

I don't see where your assertion comes from, what do you think instance variables means?

CONTRIBUTING.md
LethalGhost: That mistake appears too often actually.

CONTRIBUTING.md
NanakoAC: Deprecated by byond, are you sure? 

I recently found a perfectly good use for it in unifying behaviour over two distinct types, after a safety check to remove all possibility of runtime errors.

Used properly, it can be a useful tool

CONTRIBUTING.md
NanakoAC: This doesn't encourage using numbers everywhere, its still adviseable to have descriptive variable names. But there are many cases where you have multiple objects of an identical purpose, and calling them thing1 and thing2 is not functionally different from thingA and thingB, or thing_one and thing_two

There already is a rule, elsewhere in this file, about having properly descriptive names.  That fulfills what you're asking this rule to.

This rule was just specifically and arbitrarily excluding digits from being used in situations where they were perfectly valid

CONTRIBUTING.md
mattroks101: > Instance variables should be used with only with source object denotation to avoid namespace conflicts.

> only with source object denotation to avoid namespace conflicts.

> only with

>  **only**

NanakoAC: Yes, but what does instance variables mean?
Function parameters are not instance variables. 

An instance is an object or datum, a thing that exists in the world, a practical application of code.
It's saying all variables belonging to this instance should use src. IE, use src.thing every time you refer to something on yourself

It doesn't say to do this in the case of a conflict, it's telling you to do it always to prevent any possibility of a conflict from occurring

There's no need for that, its inefficient and wasteful

NanakoAC: but moreover, theres no need for it even with function inputs. As i just told sahqo, if you use src in a function you have to keep using it for the rest of that context, its adding pointless labour to the user and introducing points of failure if you slip up. 

If you name your input parameters properly, thats exactly one point where you have to fix things. It's more foolproof

NanakoAC:
I need to add:
Rule about cargo cult coding: Don't do things you don't understand
Rule about proper proc parameter naming: use _var to prevent conflicts

NanakoAC:
also to add: Don't design unmaintainable systems that put a great burden on future coders

NanakoAC:
updated

NanakoAC:
requested changes all done, added a section about unnecessary return calls

CONTRIBUTING.md
```diff
@@ -1,15 +1,53 @@
 # General
-
+* Where possible, large projects should be broken up into several smaller pull requests, and/or done in phases over time
```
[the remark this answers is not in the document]
mattroks101: Period here is still missing.

CONTRIBUTING.md
```diff
@@ -1,15 +1,53 @@
...
 * Pull requests should not contain commented code except TODOs and explanation comments.
-* Pull requests should not contain any debug output, variables or procs.
+* Pull requests should not contain any debug output, variables or procs, unless these are of value to admins/coders for live debugging
```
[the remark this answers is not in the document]
mattroks101: And here.

CONTRIBUTING.md
```diff
@@ -66,8 +104,40 @@ Spaces are needed between function agruments (declaration and definition). Space
...
+	do_thing
+	. = result_of_doing_thing
+	return
+	
+/proc/do_thing()
+	do_thing
+	do_other_thing
+	return
+```
+***
```
LethalGhost: Style does not match other parts of guidlines.

NanakoAC: in what manner?

NanakoAC:
fixed periods
i don't know what ghost wants

CONTRIBUTING.md
LethalGhost: Sry that was problem with displaying indents in WebUI

ACCount12:
Seems good enough to merge as it is to me.

PR #2715, opened 2019-01-24
Fix dramatic lights turning

hitsedesen:
The bug happened because of the assumption that `Process()` is called every tick


LudwigVonChesterfield:
#2729

PR #2718, opened 2019-01-25
Adds three new randomized crates

greggbot:
More stuff for the merchant to sell is good. Hardsuit modules and pouches in particular will see a fair bit of demand, and I imagine the rarer aster powercells might too.


Voldirs:
Add random pouch crate maybe? That's what is going to be sold really good considering how rare they are to people who don't want to get to maints

code/game/objects/random/powercells.dm
```diff
@@ -3,35 +3,74 @@
...
-	name = "low chance random tool"
+	name = "low chance random powercell"
 	icon_state = "battery-green-low"
 	spawn_nothing_percentage = 60
+
+/obj/random/powercell/asters
+	name = "random asters powercell"
+	icon_state = "battery-orange"
+
+/obj/random/powercell/asters/item_to_spawn()
```
NanakoAC: This asters powercell list is still an unfortunate method.

Please implement as described-  one single random powercells list as a variable, and have this variant modify that list before picking from it

PR #2723, opened 2019-01-25
Fixes a lot of stuff, makes a lot of changes.

### LINKED ISSUE #1662: Ironhammer Holobadges
Right now security holobadges are stored inside a box in IH Sergeant locker and therefore never issued to operatives: sergeant is either non-existant or simply does not care/not even aware of their existance.
It would be better to move the box into the table inside Ironhammer barracks, give a single badge to every operative/inspector on spawn or put a single badge into every operative/inspector locker just like every equipment. That way badges would be used much more often for some police RP.

### LINKED ISSUE #1727: Virology ejector doesn't send things into distance
https://i.imgur.com/AuNCopY.png

When you drop a monkey into a disposal tube which ejects into space, it should go flying into the horizon and never be seen again.

Instead, for some reason, these monkeys get fired about 3-4 tiles away and stop. Then your chat gets spammed as they suffocate and die while floating there. The spam continues as their bones break post mortem

I'd say anything that forcefully ejects into space should fly away and vanish, not linger near the hull

Not sure if this is due to external hull flooring providing friction (in which case it's a mapping problem) or disposal chutes not taking gravity into account (a coding problem)

### LINKED ISSUE #2432: Air ports outside engineering have wrong pressure
There are air ports outside of engineering intended for filling up canisters and mechas with normal mixed air
airport

They are apparently outputting at a pressure of 2285 kpa, though. And not the  ~4500kpa that these tanks normally work at

Docking your mech here for a refill is actually quite likely to reduce the gas in your tank, essentially setting you to half of what you'd normally expect
These probably need some pumps to control their output pressure

### LINKED ISSUE #2591: Bug - Ironhammer Medical Window
Window missing under the privacy shutters, aft side of the ironhammer medical office.

bugironhammer

### LINKED ISSUE #2678: Missing roof near engine room
https://i.imgur.com/hi3N0TB.png

just under my ghost the top of a wall can be seen, needs a hull floor tile here

### LINKED ISSUE #2721: More ATMs needed
At the very least we need:

One ATM inside the guild lobby. On the north wall. Something will have to be moved to make room

One ATM inside the guild itself, in the restricted space. Perhaps in the cargo bay or the merchant office. Merchant needs a way to manage money without leaving his department

One ATM in each of the four ship sections, not including the above two

One ATM somewhere in or near the diner area, for paying for food and drink

GrayRachnid:
Fixes #2591
Fixes #2432 
Fixes #1662
Fixes #2678
Fixes #2721
Fixes #1727

Also adds EFTpos scanners to each head of staff's office.


NanakoAC:
I edited the description to make the autoclosing work correctly

PR #2725, opened 2019-01-25
Protolathe now accepts reagents

### LINKED ISSUE #2719: BUG: Chemicals cannot be inserted into the Protolathe, despite recipes requiring chemical reagents.
~~Protolathe requires admin-intervention to refill with silicon, for component parts.~~ [The relevant code only mentions the word "beaker" once.](https://github.com/discordia-space/CEV-Eris/blob/master/code/modules/research/protolathe.dm) There doesn't appear to be any way to fill it.

EDIT: Admin-intervention is not possible, the Protolathe cannot store reagents. #2720

seekin4u:
It now accepts any reagents from any reagent_containers. The magic is in
`attackby` proc.
Fix #2719


code/modules/research/protolathe.dm
```diff
@@ -121,21 +121,20 @@
 	if(default_part_replacement(I, user))
 		return
 	if(I.is_open_container())
-		return 1
```
seekin4u: This is the only true change in this PR - if proc returns 1 in this case (e.g. when you trying to click on protolathe with beaker) nothing will happen. 0 allow reagents transfering.

PR #2727, opened 2019-01-25
Protolathe and curcuit implanter's markdown fix

seekin4u:
Now DISPOSE ALL REAGENTS pops only if there are any reagents, not for everyone in
list as it was before.


seekin4u:
image

Issue #2704, opened 2019-01-23
Gravity generator lobby runtime


NanakoAC:
Idled in lobby for a while, power died naturally, gravity generator tries to play its shutdown sound

[20:21:47] Runtime in sound.dm,294: Cannot execute null.return air().
  proc name: playsound local (/mob/proc/playsound_local)
  src: NanakoAC (/mob/new_player)
  src.loc: null
  call stack:
  NanakoAC (/mob/new_player): playsound local(the floor (117,63,4) (/turf/simulated/floor/tiled/dark/gray_perforated), 'sound/effects/alert.ogg', 100, 1, 0.5, null, null, null, 1)
  the gravitational generator (/obj/machinery/gravity_generator/main/station): shake everyone()
  the gravitational generator (/obj/machinery/gravity_generator/main/station): grav off()
  the gravitational generator (/obj/machinery/gravity_generator/main/station): set state(0)
  the gravitational generator (/obj/machinery/gravity_generator/main/station): Process(20)
  Machines (/datum/controller/subsystem/machines): process machinery(1)
  Machines (/datum/controller/subsystem/machines): fire(1)
  Machines (/datum/controller/subsystem/machines): ignite(1)
  Master (/datum/controller/master): RunQueue()
  Master (/datum/controller/master): Loop()
  Master (/datum/controller/master): StartProcessing(0)

NanakoAC:
fixed

PR #2728, opened 2019-01-25
Fix gravity generator sound's runtime

seekin4u:
Occured when there are players in lobby and generator calls `playsound` and goes for every mob in `SSmob.mob_list` which contains also `/mob/new_player/` from lobby. Runtime log is in #2704


code/game/sound.dm
```diff
@@ -286,39 +286,41 @@ var/const/FALLOFF_SOUNDS = 0.5
...
-		if (S.volume <= 0)
-			return	//no volume means no sound
-
-		var/dx = turf_source.x - T.x // Hearing from the right/left
-		S.x = dx
-		var/dz = turf_source.y - T.y // Hearing from infront/behind
-		S.z = dz
+		//
+		//world << "isturf(turf_source):[isturf(turf_source)]:(turf_source)[turf_source]_X:[turf_source.x]_Y:[turf_source.y]:(distance[distance])"
+		if(T)//this means mob is aboard or somewhere near
```
seekin4u: This is the only added place, as most of problems was that T is turf without any air in lobby(or where ever) and distance, hearer_env and source_env are used below few times, and i can't just remove or replace them with constant value.

code/game/sound.dm
```diff
@@ -342,7 +344,7 @@ var/const/FALLOFF_SOUNDS = 0.5
 
 		else if (pressure_factor < 0.5)
 			S.environment = SPACE
-		else
+		else if(!istype(src, /mob/new_player/))//again we may be in lobby
```
seekin4u: so this is duct tape again, code below seems to be copy-pasted, as it exists in "for humans" part, but i'm not sure.

PR #2741, opened 2019-01-27
Fix for round delay

### LINKED ISSUE #2696: Delay Round End button doesn't actually work
please

this is my lifeline

seekin4u:
Now round delay works as it should work. Pressing "delay round" button
delays lobby count if you are in lobby/pre_game and delays round end if
you are in game already.
Fixes #2696 .


code/modules/admin/admin.dm
```diff
@@ -709,11 +709,11 @@ ADMIN_VERB_ADD(/datum/admins/proc/delay, R_SERVER, FALSE)
 
 	if(!check_rights(R_SERVER))
 		return
-	if (SSticker.current_state != GAME_STATE_PREGAME)
+	if (SSticker.current_state != GAME_STATE_PREGAME && SSticker.current_state != GAME_STATE_STARTUP)
```
NanakoAC: this looks like its delaying the round start, not the end? Are you sure this delays the end of the round?

seekin4u: This button might be pressed before full lobby initialization (before `STATE_STARTUP`) but in that case it will delay round end before game actually starts (but not lobby delay as it should be).

PR #2743, opened 2019-01-27
Scrap, and related machinery.

LudwigVonChesterfield:
This PR adds a new way of getting loot - scrap. Scrap is a bunch of random junk tossed in together.

Each scrap pile has a predetermined loot in it, and when you hit it with a shovel you update the visible loot.

Scrap beacon is a beacon that attracts scrap all around it.

* [x] Scrap.
* [x] Random reagents in scrap lumps.
* [x] Scrap stacker machine.
* [x] Scrap pile ripper machine.
* [x] Generator based on scrap(?).
* [x] Scrap beacon.

I am mighty sorry for the making of a seperate PR for this, but the previous one was slightly screwed up.


mattroks101:
Do you want to closed the old PR then?

LudwigVonChesterfield:
Of course! Silly me

code/game/objects/random/food.dm
```diff
@@ -11,32 +11,39 @@
...
-	return food
+/obj/random/junkfood/rotten/post_spawn(list/spawns)
+	for(var/obj/item/weapon/reagent_containers/food in spawns)
+		if(!food.reagents)
+			return
+		if(prob(80))
+			food.reagents.add_reagent("toxin", 25)
+		if(prob(30)) // So sometimes the rot is visible.
+			food.make_old()
+	return spawns // In the random.dm it is said that the function should return the list of items, yet the return of the function is never used. ~Luduk
```
NanakoAC: the comment was a mistake, it can be deleted

NanakoAC:
spent 15 minutes wandering around maintenance, so far not found any scrap piles. These things seem remarkably rare, if they're even implemented at all, i'm not sure they are.

Scrap piles should definitely be spawning in the random junk list, and maybe lowkeyrandom too

NanakoAC:
the art for scrap piles is pretty nice, i'm enthusiastic about how this will look.

It will seem a bit odd in certain places though, so i'd advise doing something like i did with burrow spawning - making them appear on clear floor tiles against a wall

NanakoAC:
Reviewwing!

This food should not expire
https://i.imgur.com/LKrjyIk.png

These rations should be sealed and permanantly edible. Short circuit their make_old proc so theyre always fresh

NanakoAC:
Some useability concerns

Its not at all explained to the user that they should use a shovel on it, or that the loot list will change when they do

While digging through a clothing pile, i seem to have exhausted everything in it, but it still lets me dig and search it, the contents are empty though.  When exhausted, the pile should disappear

What happens if people insert things into a scrap pile then use a shovel on it?

NanakoAC:
Also, balance concern. To make searching these piles worth the effort, could you make each one have some chance of a rare spawn in there? Throwing one random tool or lowkeyrandom spawn into them would do the trick

NanakoAC:
Whats the purpose of the various scrap related machines?
burner, pile ripper, recycler, stacker, etc

How are these obtainable ingame? Can you add some comments about what they do?

code/modules/scrap/scrap_refine.dm
```diff
@@ -0,0 +1,52 @@
...
+	for(var/i in 1 to reag_num)
+		if(reagents.total_volume == reagents.maximum_volume)
+			break
+		reagents.add_reagent(pick(chemical_reagents_list), rand(1, reagents.maximum_volume))
+	pixel_x = rand(0, 16) - 8
+	pixel_y = rand(0, 8) - 8
+
+/obj/item/stack/sheet/refined_scrap
+	name = "refined scrap"
+	desc = "This is ghetto gold! It could be used as fuel or building material. Even central Command would give cargo points for this."
```
NanakoAC: We don't have central command

LudwigVonChesterfield:
Most of the comments could be addressed by having a look through the code...
I didn’t add scrap to lowkeyrandom as I use that in scrap piles to generate loot, and then scrap piles would generate more scrap piles inside which is slightly meh.
After each dig scrap pile shuffles it’s contents and displays part of it to user, if you put something in and dig there’s a chance it will disappear, but after more digging you’ll see it.
Scrap pile “exhausts” not when it’s out of “not broken” items, but when it’s out of unrefined scrap.
The machines can turn unrefined scrap into refined scrap, and that’s pretty much it.

code/_helpers/game.dm
```diff
@@ -1,5 +1,11 @@
 //This file was auto-corrected by findeclaration.exe on 25.5.2012 20:42:31
 
+#define RANGE_TURFS(RADIUS, CENTER) \
```
NanakoAC: there's already a trange define which does this

NanakoAC:
I said add it to junk list, not lowkeyrandom
use lowkeyrandom inside the scrap

even If you're using junk in its contents you can do some magical stuff to the item to spawn proc to prevent nesting

It looks wierd being a pile when it has nothing in that pile, are you sure the exhausting when out of scrap is working? I did about ten digs on one pile and it seemed to keep going.
 Perhaps there should be less scrap in a pile, or more items, or more scrap yielded from each dig, so they run out about the same time

Thats definitely not pretty much it for the machines, you don't need four machines for that

LudwigVonChesterfield:
Yes, it taes more than 10 digs, you can see in code how many exactly. Yes I am sure that it shouldn't exhaust. Items in a scrap pile are those that weren't damaged by time/nature/etc, most of the scrap pile is however... Scrap. Scrap piles shouldn't be chests full of goodies, rather - trash.

Four machines are doing different things.
Scrap pilripper rips scrap-piles/scrap-cubes/items apart and turns it into unrefined scrap, the other scrap-machine turns unrefined scrap into refined scrap, scrap-stacker stacks scrap(duh), and there's a scrap generator that turns scrap into energy.

NanakoAC:
the issue with the depleting is that you've got two systems of giving things to users here. The contents of the pile appear in an inventory window, but the scrap is just dropped at your feet.

Can you make the scrap appear in the inventory instead? can still maintain the one scrap per dig thing.
The important thing is, the user should never dig and end up with an empty inventory screen there. if its empty then the pile should be depleted and gone, thats whats most intuitive. so keep sticking scrap in there until its exhausted

NanakoAC:
This project has dragged on for far too long, and dan/ludwig has now been MIA for six days since the last comment.

Given these circumstances, and the importance of this as new content, i'm taking over this branch as of now. A new PR will be made

PR #2749, opened 2019-01-28
Telebears: A challenging new Major Event

NanakoAC:
This PR ports and reworks one of my favourite projects from aurora, the Bluespace Bears

The artwork and name should both be considered placeholders. I'm aware that bluespace bears sounds like a silly meme, and i'm open to ideas for more monstrous names. Shazbot is working on better art for them

But that aside, they are fully functional and feature complete. Details:

The bears are a powerful mid-high end monster, designed to challenge skilled players.  They have an agile and unpredictable combat style, focused on rapid shortranged teleporting and attacking from various different angles, as well as generally higher damage and health pool than most monsters. 

Because they were originally designed for aurora - a heavy RP environment - they were also designed not to murderbone. While they will warp around the ship in search of targets, they have quite a long warning period where they'll growl and intimidate before launching an assault.  Civilians can easily run away before they get angry, go and hide somewhere out of sight (in a locker works well) or even lie down and play dead until the creature moves on.
In short, they are an easily avoidable hazard for noncombatants, and a challenging threat for those who engage them in combat.

In battle, reactive teleporting makes the choice of weaponry important. Their ability to blink out of the way mid-burst means that rapid fire weapons are generally ineffective, but slow, heavy hitters like fireaxes, sniper rifles and explosives will work wonders. Skill, reactions and planning are required to fight them effectively, up close they are a disorienting and confusing foe.

The bears come with a broad range of high quality audio that i sourced years ago, really atmospheric roars and growls that signal the distant presence of a monster somewhere on ship. With a bit of luck, you can track them via their vocalisations

Bear attack is a Major Event, adding more content to that much needed pool. Given many recent rounds that have gone for many hours, and seen players run out of content, i feel like this kind of new foe is what we really need right now.
They're also rewarding! Once slain, the corpses can be harvested for their hide, an item that doubles as a trophy and well rounded head armor.

Also includes some tweaks and interactions with beartraps. Luring one into a trap is a very effective strategy, they will escape eventually but at the cost of hefty damage

Includes some minor bugfixes to superior animal and teleport code, mostly just return values not working
Fixes an exploit where people could spend time setting a beartrap on the floor, then pick it up in the deployed state and drop a deployed trap instantly anywhere. Now it undeploys when picked up


NanakoAC:
the art is still not ready, and this PR won't be finished, closing

PR #2756, opened 2019-01-29
Shield Fix

NanakoAC:
Fixes an issue where the shield generator could break by a lockout var being set and not reset


code/modules/shield_generators/shield_generator.dm
```diff
@@ -157,6 +157,7 @@
 
 	// The generator is not turned on, so don't generate any new tiles.
 	if(!running)
+		generatingShield = FALSE
```
drexample: Looks legit.

PR #2758, opened 2019-01-29
Restructuring ship areas.

SpadesSlick:
This PR is needed for my further work on coloured light in department areas. In order to do this right, said areas should be grouped by those departments. So there is a problem that PR solves.

The area structure was an absolute mess. I sure that 66% of areas existed in a code **don't used right now, and will never will.** This is just a dead weight. Some areas were put into department subtype, some were just like this one ``/area/acting``, and some were put in ``/eris/``. Deck names were wrong, maint had a wrong type of areas, and generally everything was on fire.

This PR does this:
* Almost all of the unused areas were removed. It was at least HALF of all of the areas.
* Almost all of the ship areas are now in a separate file under ``/eris/`` type.
* Department structure were provided was areas to group them together.
* Maint areas now have new sprites with Deck and Section numbers on them.
* Mass replacement was used on map to rework those areas. This process cannot be map merged by the tools we have now.


SpadesSlick:
image
🤔

SpadesSlick:
Description updated.

PR #2761, opened 2019-01-30
Nerf space lube

Uvash:
Now lube can dry...


NanakoAC:
its not a bug, lube is water that doesn't dry

Look at the code that was edited, theres a var thats passed in just to check if the floor is wet with lube (wet value of 2) and if so to not dry that

All that Uvash did is bizarrely lobotomise that check, making the passed var not do anything.

This PR deliberately introduces a bug, it is being closed

PR #2765, opened 2019-01-31
Tape recorder rework

EvenInDeathIStillServe:
Ports /tg/ standalone cassette tapes for tape recorders. Now you can record, take out the tape, label it, store it, and use the recorder again.

I also added wire interaction for the recorders. Pulse the wires to stop, record, play, or wipe the tape. Works for signalers too.

I decreased the tape capacity from 1 hour to 20 minutes since recorders can record to many tapes. Might be better to further decrease to 10 minutes but I'm not sure.

The tapes are available on the basic autolathe designs disk.

Please review & test.


NanakoAC:
casette tapes in 2500 AD seems really silly.

Couldn't you make them record onto little usb sticks instead, which could be read from any computer?

NanakoAC:
Going even farther, i'm not sure that having a seperate object for this even makes sense.

having a Microphone component for modular computers (including PDAs), a conversation recording program for computers, and having these things preinstalled on the detective's PDA, would make the most sense

EvenInDeathIStillServe:
I think tape recorders have a really cool aesthetic. There's something special when you're being interrogated and the detective sets a recorder on the table. I definitely agree that the storage could be modernized to USB sticks and made compatible with computers, but I really, really wouldn't like PDAs replacing the recorders.

NanakoAC:
well having the recorder device isn't bad, can you do this then, but make it use usb sticks and be properly compatible with computers?

EvenInDeathIStillServe:
Yes, I can do that. I'll need a sprite for the USB sticks though. I think I can handle the recorder's sprite as long as I know how the stick looks like.

NanakoAC:
Ah a little misunderstanding there, it already exists, fully sprited and implemented afaik
just make it accept and write to these

/obj/item/weapon/computer_hardware/hard_drive/portable

NanakoAC:
ironhammer inspector lockers don't actually contain a tape recorder, please add one to them

Ejecting the data crystal needs a rightclick verb
We could do with an overlay to indicate that a crystal is plugged in

NanakoAC:
slightly out of scope, but eject portable storage just dumps the datacrystal into the computer's tile, and its hard to see

can you make it go into the user's hand if possible, and also change its layer so it draws above the computer>?

NanakoAC:
the audio program should come preinstalled on the computers in the inspector's office

NanakoAC:
I'm thinking, remove the print transcript button from the recorder device, so that people need to use a computer to print off a transcript

we need more reasons to use them

NanakoAC:
almost perfect, one last problem

I ejected the datacrystal while playing audio on a console

[00:54:12] Runtime in audio_player.dm,31: Cannot read null.storedinfo
  proc name: play audio (/datum/computer_file/program/audio/proc/play_audio)
  usr: Bluespace Technician (nanakoac) (/mob/living/carbon/human/bst)
  usr.loc: The floor (88,149,3) (/turf/simulated/floor/tiled/dark/gray_perforated)
  src: /datum/computer_file/program/a... (/datum/computer_file/program/audio)
  call stack:
  /datum/computer_file/program/a... (/datum/computer_file/program/audio): play audio()
  /datum/computer_file/program/a... (/datum/computer_file/program/audio): Topic("src=\[0x21022295];PRG_toggleau...", /list (/list))
  /datum/topic_manager/program (/datum/topic_manager/program): Topic("src=\[0x21022295];PRG_toggleau...", /list (/list))
  Audio Player (/datum/nano_module/program/audio): Topic("src=\[0x21022295];PRG_toggleau...", /list (/list), /datum/topic_state/default (/datum/topic_state/default))
  /datum/nanoui (/datum/nanoui): Topic("src=\[0x21022295];PRG_toggleau...", /list (/list))
  NanakoAC (/client): Topic("src=\[0x21022295];PRG_toggleau...", /list (/list), /datum/nanoui (/datum/nanoui))
  NanakoAC (/client): Topic("src=\[0x21022295];PRG_toggleau...", /list (/list), /datum/nanoui (/datum/nanoui))

PR #2766, opened 2019-01-31
Animated Speechbubbles

mattroks101:
Adds a small animated speech bubble for when someone says something. Ported from mid 2015 Hippiestation. It's a very nice effect I don't know why it got lost.

### DIFF
--- a/code/modules/mob/living/say.dm
+++ b/code/modules/mob/living/say.dm
@@ -235,7 +235,7 @@ proc/get_radio_key_from_channel(var/channel)
 			speech_bubble_recipients += M.client
 		M.hear_say(message, verb, speaking, alt_name, italics, src, speech_sound, sound_vol)
 
-	flick_overlay(speech_bubble, speech_bubble_recipients, 30)
+	animate_speechbubble(speech_bubble, speech_bubble_recipients, 30)
 
 	for(var/obj/O in listening_obj)
 		spawn(0)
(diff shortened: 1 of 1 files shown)


GrayRachnid:
Can you fix speech bubbles being under catwalks and health huds while you're at it?

mattroks101:
No. Speech bubbles themselves aren't the problem, only the typing indicator. I looked at the typing indicator code and I have no idea why it's doing that it's on the right layer.

GrayRachnid:
oh yeah that's what I meant. Huh, idk then.

LudwigVonChesterfield:
clouds will be seen through walls if you use mesons or something of that kind.

mattroks101:
You sure about that? This doesn't change the recipients list at all. If the person can't see you talking in the first place they shouldn't be seeing this effect. Did you test to see if that was the case?

code/modules/mob/living/say.dm
```diff
@@ -245,6 +245,22 @@ proc/get_radio_key_from_channel(var/channel)
...
 	return TRUE
 
+
+/proc/animate_speechbubble(image/I, list/show_to, duration)
+	var/matrix/M = matrix()
+	M.Scale(0,0)
+	I.transform = M
+	I.alpha = 0
+	for(var/client/C in show_to)
+		C.images += I
```
mattroks101: @LudwigVonChesterfield This is only showing it to the people in the show to list. Not just any random person.

PR #2772, opened 2019-02-01
Nerf space lube 2

Uvash:
Space cleaner can clean lube.


LethalGhost:
That's also make floor dry.

NanakoAC:
Ghost has a good point here, spacecleaner shouldn't make the floor wet, it should just change spacelube's super wetness (wet = 2) to normal wetness (wet = 1) and schedule an undrying proc

PR #2773, opened 2019-02-01
Antigrav generator

Camper12345:
Machinery designed for disabling gravity at the area in which it placed.


code/game/machinery/antigrav.dm
```diff
@@ -0,0 +1,124 @@
...
+		area = null
+
+	if(on)
+		stop_anim()
+
+	on = FALSE
+	use_power = 1
+	update_icon()
+
+/obj/machinery/antigrav/Process()
```
NanakoAC: There's no parent call here. IIRC the machinery/Process is where power usage is handled

code/game/machinery/antigrav.dm
```diff
@@ -0,0 +1,124 @@
...
+			user << SPAN_WARNING("Fasten \the [src] to the floor first.")
+	else
+		stop()
+
+/obj/machinery/antigrav/attackby(var/obj/item/I, var/mob/user)
+	src.add_fingerprint(usr)
+	if (I.has_quality(QUALITY_BOLT_TURNING))
+		if (anchored)
+			if(!on)
+				user << SPAN_NOTICE("You begin to unfasten \the [src] from the floor...")
```
NanakoAC: Isn't there generic/base behaviour for this? We shouldn't need to keep putting wrench checks into every new machine

code/game/objects/items/weapons/circuitboards/machinery/power.dm
```diff
@@ -28,4 +28,16 @@
 	req_components = list(
 		/obj/item/weapon/cell/large = 3
 	)
-	matter = list(MATERIAL_PLASTIC = 2, MATERIAL_GLASS = 3)
\ No newline at end of file
+	matter = list(MATERIAL_PLASTIC = 2, MATERIAL_GLASS = 3)
+
+/obj/item/weapon/circuitboard/antigrav
+	name = T_BOARD("antgrav generator")
```
NanakoAC: typo here, should be anti

code/game/objects/items/weapons/circuitboards/machinery/power.dm
```diff
@@ -28,4 +28,16 @@
...
+	build_path = /obj/machinery/antigrav
+	board_type = "machine"
+	origin_tech = list(TECH_ENGINEERING = 4, TECH_MAGNET = 4)
+	req_components = list(
```
NanakoAC: these components don't really make much sense, what would a grav blocker need with a matter bin?
 did you copy them from something else?

It'd be better to use subspace components like ansible crystal

code/game/machinery/antigrav.dm
```diff
@@ -0,0 +1,124 @@
+/obj/machinery/antigrav
+	name = "antigrav generetor"
+	desc = "It temporarily disables gravity around."
+	icon_state = "GraviMobile"
+	density = 1
+	anchored = 0
+	use_power = 1
+	idle_power_usage = 10
+	active_power_usage = 10000
+	circuit = /obj/item/weapon/circuitboard/autolathe
```
NanakoAC: Autolathe circuit here, i guess thats where it was copied from :D

code/game/machinery/antigrav.dm
```diff
@@ -0,0 +1,124 @@
+/obj/machinery/antigrav
+	name = "antigrav generetor"
+	desc = "It temporarily disables gravity around."
+	icon_state = "GraviMobile"
+	density = 1
+	anchored = 0
+	use_power = 1
+	idle_power_usage = 10
```
NanakoAC: Whats the explanation for the idle power usage? Any reason not to make it 0 ?

code/game/area/Space Station 13 areas.dm
```diff
@@ -41,6 +41,8 @@ NOTE: there are two lists of areas in the end of this file: centcom and station
 	var/used_environ = 0
 
 	var/has_gravity = 1
+	var/cached_gravity = 1		//stores updated has_gravity even if it's blocked
```
NanakoAC: Is there really a need for this cached gravity var? 

Couldn't the gravity generator just tell the area to update its gravity to what it should be (by checking for the presence of a gravity generator on the ship) when it moves?

code/game/machinery/antigrav.dm
NanakoAC: then why are there even checks for power there?

if power isnt used from process, theres no purpose in those checks. that would just make some things process forever if they use power despite not needing to

NanakoAC:
is this ready for review?

NanakoAC:
several comments unaddressed

code/game/machinery/antigrav.dm
Camper12345: This is most basic behaviour for this purpose, as far as i know.

code/game/area/Space Station 13 areas.dm
Camper12345: It needed for the case when gravity generator changes area's gravity while it's blocked.
This is much better than looking for the gravity generator on map.

NanakoAC: Its an example of a special case (antigrav generators) using a generic and common object (area datums) to store data

Searching for a gravity generator on the map may be a little less efficient, but it'd be a better design.
In any case, for all kinds of practical purposes, we definitely need to have a reset_gravity proc. Where an area can abandon whatever gravity states have been set upon it, re-examine all the factors in the game that would affect it, and set its gravity to what it should be

code/game/machinery/antigrav.dm
NanakoAC: Sadly you are correct, i made an issue for this a while ago #1770 
We will just have to accept this for now

code/game/area/Space Station 13 areas.dm
Camper12345: Do you think we need more advanced structure for storing gravity state?

Camper12345: I think searching for the factors of anything like that isn't a good thing. It's complex, not very effective and a huge possibility for different bugs.

NanakoAC: this is not true at all, its both the simplest and least error prone solution

NanakoAC:
Alright I've fixed this

PR #2778, opened 2019-02-02
Cloning activation fix

NanakoAC:
Fixes a nasty issue where the cruciform wouldn't properly activate when attached to a newly cloned person, resulting in them being unable to be targeted by rituals


TorinoFermic:
The testing went well and the cruciform should work fine by forcefully dragging the actual ghost into the newly cloned.

PR #2780, opened 2019-02-02
Add stats transfering on cloning

### LINKED ISSUE #2746: Stats are lost on cloning
Newly cloned bodies start with nothing in stats

these should be stored on the mind datum probably

hitsedesen:
closes #2746


code/modules/core_implant/cruciform/modules.dm
```diff
@@ -69,6 +70,7 @@
 	if(ishuman(implant.wearer))
 		var/mob/living/carbon/human/H = implant.wearer
 		dna = H.dna
+		stats = H.stats
```
LethalGhost: That's possible what smth will (or already) affect dead body stats.
It's better to make copy of it (values only, not mods).

NanakoAC:
Cruciform is the wrong place for this. Store it in the mind datum instead

the cruciform stores a reference to that

NanakoAC:
and a direct reference is better than making a clone of the stats, because it will be kept updated. We'll be adding ways for players to change their stats during a round

LethalGhost:
> a direct reference is better than making a clone of the stats

@nanako
But that will bound dead body stats to cloned person one.

NanakoAC:
oh no, there has been a terrible misunderstanding here.

I didn't mean move all stat interactions to the mind, just have mind hold a pointer to it
Stats still need to be on the mob for the sake of mindless NPCs which still need stats support

And Ghost, i misunderstood what you meant. A clone should be made, but only at the time the new body is created, so that was a misunderstanding too. i think this is the way things were before

NanakoAC:
I remain convinced that the mind datum is the best place for this, and that this PR is ready to go

PR #2790, opened 2019-02-03
Full Auto Firing Mode

NanakoAC:
This PR adds a shiny new feature that i've always noted was conspicuously absent from SS13

Real, true, fully functional full auto firing mode.
You hold LMB to start firing, release it to stop firing, aiming follows your cursor while holding the fire button.  Smooth, reactive, and flawless - ping times permitting, anyways. You have precise enough control to fire a single round even in full auto mode by tapping the button
It Just Works™

Some videos demonstrating this functionality. All depicting an L6 SAW firing at 600 RPM:
https://drive.google.com/file/d/1ciHRElkzDOJWKmWl0k2SSQFCarb4qMXN/view?usp=sharing
https://drive.google.com/file/d/1KmFggq-qcntIvYAgORbok31UgU9KBuCp/view?usp=sharing
https://imgur.com/jZNtTcb

Currently, this new firing mode has been implemented on a few ballistic weapons which are specifically described as automatic, as well as all LMGs and SMGs. I'm not sure if all of these are actually obtainable ingame though. The full list of weapons with this firing mode is:
	FS Atriedes SMG
	FS Idaho SMG
	C-20r SMG 
	Dallas AR (I don't think this one exists ingame)
	L6 SAW (This is the LMG that the serbian mercenaries get. It will later be replaced by the PK, and that will have 
	full auto fire too.)
	Excelsior Maxim LMG
	STS 35 AR (not sure if obtainable)
	Z8 Bulldog Carbine

All of these weapons get a firing rate of 400 rounds per minute in full auto mode, except the LMGs which gets 600 instead. Big guns need an advantage to make them worthwhile	

By specific request from Clock, the AK used by serbians and excelsior does NOT have full auto fire. Its description has been updated to indicate this.

In practical testing against mobs, full auto fire proves somewhat inaccurate and wasteful, as it should be. But in the right circumstances, with a steady hand, it can make for a very powerful tactic

To accomplish this, Sahqo's click handlers are repurposed, given new functionality, and used to intercept a few click events. The gun updates its click handler whenever anything changes that could affect it (dropped, safety toggled, swapped hands, etc). 

Some new functions are added in this process. Items now have the swapped_to and swapped_from procs, which are called whenever the user swaps hands to make their active hand the one that has/doesn't have that item. This code is designed for future plans of allowing people with >2 arms through the use of bionics or mutation

In future, I plan to expand farther on this code, to not only add support for rapid fire pulsing lasers, but also for single continuous beam lasers which deal damage several times per second to anything they're pointed at. It will also need some adaptation to work for mechas once i work on those.


Voldirs:
>Z8 Bulldog Carbine
This one has 10 rounds in magazine. No point in auto-fire. Either mags should be able to hold more rounds either auto-fire should be removed for it.
>FS Atriedes SMG
And this one has 44 damage per bullet. It's the most powerful weapon in game for now and now it's even better. I'll do ammo rebalance later myself, I think
>FS CAR Sol
It's already better than AK, because it deals slightly more damage, but now it's even better because of capability to fire in auto-mode. Basically, AK is trash. It's a bad decision to make default weapon of operatives to be able to mow down whole hallway.

NanakoAC:
good feedback, looks like a few weapons need nerfing

Camper12345:
If you drop the gun during full auto shooting, you're still able to shoot it, even if lmb was released.

NanakoAC:
Removed full auto from the sol

Also added in a new damage multiplier var that can be used to alter damage on a per gun basis

All of voldirs other comments are balance issues that we will address in the upcoming ammo rebalance

NanakoAC:
> 
> 
> If you drop the gun during full auto shooting, you're still able to shoot it, even if lmb was released.

oh shit, this is not good
will fix asap

NanakoAC:
alrighty, bugs fixed, should be ready to go

PR #2792, opened 2019-02-03
Adds custom vending machines

EvenInDeathIStillServe:
What it says up there ^^

You can now print a custom vendomat circuit board using the circuits autolathe disk, or buy one from the asters disk vendor.

Once built, the custom vendomat can be registered to someone by swiping their ID in it. The owner can unlock and lock the vendomat by swiping their ID. The vendomat can be assigned to a specific department, allowing members of the department to manage it, and transfering profits to the department account. An account number can be specified if you want to transfer the profits to a specific account other than the department or the one linked to your ID.

When the vendomat is unlocked, it can be stocked with any item. Once an item is added, the owner can change the price for it.

If the vendomat is unlocked, you can vend items for free to destock it. When it runs out of an item, you can remove the entry for it in the item list.

Profits from sales go to the owner of the vendomat.

You can unregister the vendomat by clicking "Unregister" when it is unlocked and the maintenance panel is open.

The vendomat can be remodeled and renamed using the appropriate verbs as long as it's unlocked.

The vendomat can be disassembled when the panel is open using a welder. This is a long process, so bystanders can react. The disassembled vendomat will drop all items in it, and the vendomat circuit board. The vendomat can be emagged in order to override current ownership, effectively unregistering the owner and allowing the next ID swiped to take his place.


NanakoAC:
Lots of questions and recommendations:
1. Is there anything to stop someone hacking it and registering themselves as the new owner? Or alternatively, is there any way to accomplish this ?
2. Can it be deconstructed if the owner is gone? 
3. We need an ability to enter an arbitrary account number to send the money to
4. We also need a nice clean interface to pick a department account to send money to. Because anyone should be able to make profit for their department account

TorinoFermic:
If you can add the ability to add additional users to allow like a dept to restock the vendor. This way, Service dept only have manage one vendor.

EvenInDeathIStillServe:
> 1. Is there anything to stop someone hacking it and registering themselves as the new owner? Or alternatively, is there any way to accomplish this ?

Not yet. I can make it so that using an emag unregisters the owner, and allows you to register it in your name. If you don't want this to happen, place the vendor in a public place where saboteurs will be easily spotted.
> 2. Can it be deconstructed if the owner is gone?

Currently, it can be unwrenched and moved. I suppose I could make it so that using a welder lets you deconstruct it, dropping all its contents, although it will be a long process so if you're doing this in a public place without authorization, you will be spotted.
> 3. We need an ability to enter an arbitrary account number to send the money to

I can do that. You'll still need to use your ID card to lock and unlock it, but when it's unlocked you can set it to transfer funds to a specific account.
> 4. We also need a nice clean interface to pick a department account to send money to. Because anyone should be able to make profit for their department account

Same thing with the above, except you'll get a different button which gives a list of department accounts which you can choose to receive the profits.

> If you can add the ability to add additional users to allow like a dept to restock the vendor. This way, Service dept only have manage one vendor.

I can add another button to the management screen that lets you set it so that people from a department can access it.

code/game/machinery/vending.dm
```diff
@@ -122,6 +127,11 @@
 	var/seconds_electrified = 0 //Shock customers like an airlock.
 	var/shoot_inventory = 0 //Fire items at customers! We're broken!
 
+	var/custom_vendor = FALSE //If it's custom, it can be loaded with stuff as long as it's unlocked.
+	var/locked = TRUE
+	var/datum/money_account/machine_vendor_account //Owner of this vendomat. It's the account of the person who registered it.
```
NanakoAC: I don't see a reason for machine_vendor_account and earnings_account to be two different vars

code/game/machinery/vending.dm
```diff
@@ -423,7 +503,7 @@
  */
 /obj/machinery/vending/proc/credit_purchase(var/target as text)
 	var/datum/transaction/T = new(currently_vending.price, target, "Purchase of [currently_vending.product_name]", name)
-	T.apply_to(vendor_account)
+	T.apply_to(vendor_department ? vendor_department : machine_vendor_account)
```
NanakoAC: I don't think you can apply a transaction directly to a department name like this, you'd need to fetch the relevant department account

code/game/machinery/vending.dm
```diff
@@ -521,20 +611,81 @@
...
+			if(!newaccount)
+				src.status_message = "No account specified. Vendomat will transfer its profits to its owner. Organization will be prioritized if set."
+				earnings_account = null
+			else
+				src.status_message = "This Vendomat will now transfer its profits to the specified account, owned by [newaccount.owner_name]."
+				src.status_error = 0
+				earnings_account = newaccount
+
+		else if (href_list["setdepartment"])
+			var/newdepartment = input("Which organization should be considered the owner of this Vendomat? This will also allow members to manage it.", "Vendomat Department", null) in list("Privately Owned", "Ironhammer Security", "Moebius Laboratories (Medical Division)","Moebius Laboratories (Research Division)","Church of NeoTheology","Asters Guild","Technomancer League")
```
NanakoAC: nooooo this is awful

Look at the economy subsystem and departments.dm, You'll want to loop through subtypes of datum/department and retrieve names/accounts from there

code/game/machinery/vending.dm
```diff
@@ -521,20 +611,81 @@
...
+			var/datum/money_account/newaccount = get_account(input("Please enter the number of the account that will receive profits from this Vendomat.", "Vendomat Account", null) as num)
+			if(!newaccount)
+				src.status_message = "No account specified. Vendomat will transfer its profits to its owner. Organization will be prioritized if set."
+				earnings_account = null
+			else
+				src.status_message = "This Vendomat will now transfer its profits to the specified account, owned by [newaccount.owner_name]."
+				src.status_error = 0
+				earnings_account = newaccount
+
+		else if (href_list["setdepartment"])
```
NanakoAC: This should call a proc to set the department instead of being done inline in the topic call.

In addition to this, we need the ability to set a compiletime var for a department, and have the vendor automatically set itself to that department when initialized

code/game/machinery/vending.dm
```diff
@@ -1048,10 +1199,75 @@
...
+
+/obj/machinery/vending/custom/verb/remodel()
+	set name = "Remodel Vendomat"
+	set category = "Object"
+	set src in oview(1)
+
+	if(locked)
+		usr << SPAN_WARNING("[src] needs to be unlocked to remodel it.")
+		return
+	var/choice = input(usr, "How do you want your Vendomat to look? You can remodel it again later.", "Vendomat Remodeling", null) in list("Generic", "Security", "Electronics", "Research", "Medical", "Engineering", "Engineering 2", "Tools", "Shady", "Fridge", "Alcohol", "Frozen Star", "NeoTheo", "Asters Power Cells", "Asters Disks")
```
NanakoAC: big switches like this are bad design, use an associative list instead. like 

list("Security" = "sec", "Electronics" = "cart")
etc
And don't declare the list inline with the input proc like that, but instead make it a #define at the top of this file, and #undef it at the bottom

code/game/objects/items/weapons/circuitboards/machinery/vending.dm
```diff
@@ -0,0 +1,8 @@
+obj/item/weapon/circuitboard/vending
+	name = "circuit board (custom vendomat)"
+	build_path = /obj/machinery/vending/custom
+	board_type = "machine"
+	origin_tech = list(TECH_DATA = 2)
+	req_components = list(
+		/obj/item/stack/cable_coil = 10
```
NanakoAC: This should require some more interesting components. At the very least, a matter bin (for storing goods) and a manipulator (for dispensing them)

EvenInDeathIStillServe:
Made it so that custom vendomats advertise the items they sell, like slogans for normal vending machines. Also, activated the advertisements which were defined but not implemented. They now appear on the listings screen for vending machines.

NanakoAC:
https://www.youtube.com/watch?v=2k0SmqbBIpQ
https://en.wikipedia.org/wiki/Feature_creep

NanakoAC:
I built a custom vendomat and did nothing with it, its spamming runtime errors 

[04:30:18] Runtime in vending.dm,793: pick() from empty list
  proc name: Process (/obj/machinery/vending/Process)
  src: Custom Vendomat (/obj/machinery/vending/custom)
  src.loc: the floor (113,111,2) (/turf/simulated/floor/tiled/dark/gray_perforated)
  call stack:
  Custom Vendomat (/obj/machinery/vending/custom): Process(20)
  Machines (/datum/controller/subsystem/machines): process machinery(1)
  Machines (/datum/controller/subsystem/machines): fire(1)
  Machines (/datum/controller/subsystem/machines): ignite(1)
  Master (/datum/controller/master): RunQueue()
  Master (/datum/controller/master): Loop()
  Master (/datum/controller/master): StartProcessing(0)

seems like once a minute, why is it processing like this at all?

NanakoAC:
I can spam-swipe an ID endlessly and make a lot of noise, there's no indication as to whether each swipe is locking or unlocking or doing god knows what https://i.imgur.com/uFpjwBT.png

NanakoAC:
Setting vendomat account number can't be cancelled, and entering nothing won't allow it to be dismissed 
https://i.imgur.com/cuBN5jK.png

NanakoAC:
Being able to manually set a price for items is good, but it should also suggest a default price, obtained by calling thing.get_item_cost()

NanakoAC:
https://i.imgur.com/xxDP9HP.png I don't understand whats gone wrong here
selling items to it doesnt seem to be working

PR #2804, opened 2019-02-06
Fix for brig timers not working with flasher & doors

EvenInDeathIStillServe:
Okay so basically, the machinery subsystem gets initialized AFTER atoms, that means the list of machines is empty when the brig timer spawns. I think searching the whole world for the brig doors and flasher is a bit too much, but that's what it already did to look for the locker and shower.

### DIFF
--- a/code/game/machinery/doors/brigdoors.dm
+++ b/code/game/machinery/doors/brigdoors.dm
@@ -36,11 +36,11 @@
 	..()
 
 	spawn(20)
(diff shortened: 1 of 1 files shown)


NanakoAC:
This is still really awful spaghetti code. Having a spawn in New and searching the whole world is not good

LethalGhost:
Agree with @NanakoAC 
`New` should be replaced with `Initialize` or even `lateInitialize` due to subsystems init order.

PR #2809, opened 2019-02-07
Bay Porting: Phase 1, Part 4: Login and Click Catchers

NanakoAC:
This ports the contents of bay's Login.dm, and an interesting feature along the way;

Click catchers are a set of invisible screen objects that only catch clicks on blackspace tiles. That is, the byond system blackness that represents line of sight

Clicks onto these blackspaces are passed to the turf underneath. This means that players can now throw things and fire projectile weapons at black areas. This is quite good for gunplay.

There will be some issues, i'm not certain what they are yet, but probably someone will find a nice new exploit related to blackspace clicking. This is par for the course, and we'll fix those things as we go

Also includes one other small feature from bay, middleclick now swaps hands.
This feature is disabled if you're using a hardsuit with an active module, since middleclick works on the suit instead. but it's available most of the time

I didnt think this feature was worth the time, but it seems a lot of people disagreed, so its ported by popular demand


NanakoAC:
woops merge screwup somehow, will fix

PR #2812, opened 2019-02-07
Cruciform organ damage fix

NanakoAC:
Fixes issues where installing the cruciform would puncture the target's lungs
now it deals five small bursts of damage, none of which are high enough to cause internal damage - assuming that the target was healthy beforehand.

I checked future bay code, this will work there too, its future proof


NanakoAC:
text change done
oops needs rereviewed now :<

PR #2814, opened 2019-02-08
Fixes the change made to aimbot detection that broke detection in some cases

### LINKED ISSUE #2813: The middle click drag aimbot exploit is still present in this repo because of a change made when porting the fix
In https://github.com/discordia-space/CEV-Eris/pull/2770#pullrequestreview-198937153 @NanakoAC suggested @TorinoFermic change a bit of code because they didn't understand it. (they applying to both people)

it was changed from

```dm
	var/list/L = params2list(params)
	var/dragged = L["drag"]
	if(dragged && !L[dragged])
		return
```
to
```dm
	// Fixes the middle mouse exploit aimbot
	var/list/L = params2list(params)
	var/draggedMiddle = L["drag"]	// Returns anything pressed down during the left click event as we are in Click() method by left or middle click
	var/left = L["left"]		// Obivously checking if the left got clicked and not held down.

 	// Deny the exploit when middleclick is pressed during the left click event
	if(draggedMiddle == "middle" && left)
		return
```

The changes broke detection, and any user can bypass detection and use the aimbot exploit still by holding with a different button, say dragging with left click and middle clicking (valid on some if not most guns), or running a special command via one of many methods to bypass control freak to disable the right click menu, allowing them to hold with right click and click with left click, bypassing detection.

Lesson here @NanakoAC and @TorinoFermic: when porting security exploit fix prs, do not fuck with code you don't understand until after you understand it.

MrStonedOne:
fixes #2813 by reverting the code to the original version in bay and /tg/ and @TorinoFermic original pr

It's was not suppose to be checking for the middle mouse at all, because that's not how the exploit works, it's just what people decided to call the exploit.

The exploit can be triggered using any combination of mouse buttons where one is dragging, and one is clicking.

In byond, the clicked mouse button is not given as a key that holds a value indicating what mouse button was clicked, but rather the present of one of 3 keys (`left`, `middle`, `right`) that have a positive value. For drags during a click event how ever, the mouse button that started the drag is given as the value to the key `drag`.

So what the original code **did** was grab the value in the `drag` key, and check for a key of the same name in the main `params` list, to figure out if the same mouse button was clicked that started a drag (caused by the click event being under a different queue then drag update events, so you can release a drag and click.). This way it works for any combination of mouse buttons.


MrStonedOne:
Allowed not allowed table:

|                               | No mouse button clicked | Left mouse button clicked | Middle mouse button clicked | Right mouse button clicked |
|------------------------------:|:-----------------------:|:-------------------------:|:---------------------------:|:--------------------------:|
|     No mouse button held down | -                       | Allowed                   | Allowed                     | Allowed                    |
|   Left mouse button held down | n/a                     | Allowed                   | Not Allowed                 | Not  Allowed               |
| Middle mouse button held down | n/a                     | Not  Allowed              | Allowed                     | Not  Allowed               |
|  Right mouse button held down | n/a                     | Not  Allowed              | Not  Allowed                | Allowed                    |

optimumtact:
you sound offended

MrStonedOne:
I am offended! They touched GLORIOUS /tg/ chad code and ruined it!

but no we talked in discord, and I made some slightly incorrect assumptions about why the change was made.

PR #2838, opened 2019-02-08
Bull shotgun fix

Garen7:
changes: 
  - bugfix: "Bull shotguns ammo overlay no longer breaks when you have shots in the chamber and fully load the gun."
  - rscdel: "IH Commander no longer has 2 berets in his locker since there should only ever be one commander.

I also renamed the getAmmo() proc to get_ammo() since snake case is the standard.


code/game/objects/structures/crates_lockers/closets/secure/security.dm
```diff
@@ -80,7 +80,6 @@
 	new /obj/item/weapon/gun/energy/gun/martin(src)
 	new /obj/item/clothing/accessory/holster/waist(src)
 	new /obj/item/weapon/melee/telebaton(src)
-	new /obj/item/clothing/head/beret/sec/navy/hos(src)
```
NanakoAC: i only see one in this list and you've removed it, where is this second beret?

Garen7: cb6f54c5-fdc8-4114-8901-5fbf1fe9debf

code/modules/projectiles/guns/projectile/shotgun/bull.dm
```diff
@@ -48,22 +61,22 @@
...
-				chambered = AC
-	reload = 1
+		if(other_chambered != null)
+			chambered = other_chambered //load the main barrel we use with the side barrel.
+			other_chambered = null
 
-/obj/item/weapon/gun/projectile/shotgun/bull/unload_ammo(user, allow_dump)
+/obj/item/weapon/gun/projectile/shotgun/bull/unload_ammo(mob/user, allow_dump)
 	if(chambered)
+		to_chat(user, SPAN_NOTICE("You unchamber [(other_chambered ? 2 : 1)] round\s from [src]."))
```
Wesreidau:
```suggestion
		to_chat(user, SPAN_NOTICE("You unchamber [(other_chambered ? 2 : 1)] shell\s from [src]."))
```

NanakoAC: this code for a second barrel seems more than a little snowflake-y, i'll have a thorough look through, but i'm not convinced this is the best way to do it.

Perhaps a better approach would be to implement a barrels var for all guns

Garen7:
@NanakoAC that's true and its totally fine, I'll take out the second chamber code probably tomorrow and just leave the overlay fix and the removal of the second beret.

Garen7:
nevermind I just reverted that stuff today

Issue #2840, opened 2019-02-09
Various Things check for Wrench instead of Bolt Turning


Boredone:
So, list of the stuff I found doing a quick check of wrenchs. Might not be 100% of them all, but should be most.
Also noticed some checks for crowbars instead of prying in them, so might end up updating this for them too.

supply_beacon
portable flasher
apc frame
pipes
medical stand
electricchair
synthesizer
bee hive
AI Core with no AI
bed
heat exchanger
meters
cash_register
stools
floorlayer (Do we even have this in game?)
holographic stuff
seed storage
deployable barriers
recharger
wheel chair
jukebox
pipe dispenser
power generator
bookcases
more pipes
pianos
mulebots
portable turrets

PR #2841, opened 2019-02-09
Bed dismantle runtime fix and tool quality standardization

Carbonhell:
Beds and subtypes were runtiming when dismantled due to the attempted creation of a /material (datum) type, this has been resolved and they now get dismantled properly, and drop some steel. Also, to dismantle/remove the padding, the checks now use tool qualities instead of strict type-checks for basic tools.
Addresses a tiny part of #2840 , might want to remove beds from the list

### DIFF
--- a/code/game/objects/structures/stool_bed_chair_nest/bed.dm
+++ b/code/game/objects/structures/stool_bed_chair_nest/bed.dm
@@ -102,7 +102,7 @@
 		return TRUE
 
 /obj/structure/bed/attackby(obj/item/weapon/W as obj, mob/user as mob)
-	if(istype(W, /obj/item/weapon/tool/wrench))
+	if(QUALITY_BOLT_TURNING in W.tool_qualities)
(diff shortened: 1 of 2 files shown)


code/game/objects/structures/stool_bed_chair_nest/bed.dm
```diff
@@ -102,7 +102,7 @@
 		return TRUE
 
 /obj/structure/bed/attackby(obj/item/weapon/W as obj, mob/user as mob)
-	if(istype(W, /obj/item/weapon/tool/wrench))
+	if(QUALITY_BOLT_TURNING in W.tool_qualities)
```
NanakoAC: We have procs for this, things like has_quality
Make use of them rather than checking vars directly

Carbonhell: Done.

Carbonhell:
Also removed an unnecessary qdel call since I added it in dismantle().

PR #2798, opened 2019-02-04
Ammo rebalance, new antag ammo, sprites

Voldirs:
1)Firearm damage rebalance
Basic idea was to separate low damage smg/pistol rounds, medium damage rifle rounds, high damage revolver/rare pistol rounds and to exclude some strange things like .357 being more powerful than .44 magnum.
This also includes rubber bullets changes, they got less brute damage in general and some pain damage rebalance. 12g beanbag is a beast now

2)New ammo type
I replace old ammo with "high-velocity" variant in uplink. So now you have 2 choices of getting ammo as antag with uplink. Either you can find/buy common ammo or diskette, either you can spend some telecrystals in exchange for much more punch and armor piercing capabilities in a shot.

3)Colored stripes on rubber or high velocity mags
Blue stripe is for rubber ammo, Red one is for high-velocity ammo.
scrnshot1

PR #2843, opened 2019-02-09
Ammo rebalance, new antag ammo, sprites. Attempt 2

Voldirs:
Read description and stuff in #2798


code/modules/projectiles/projectile/bullettypes.dm
```diff
@@ -1,5 +1,12 @@
 /obj/item/projectile/bullet/c9mm
-	damage = 18
+	damage = 22
+	sharp = 0
+	can_ricochet = TRUE
+
+/obj/item/projectile/bullet/c9mmh
+	damage = 1.2*22
```
NanakoAC: having math in here like this is wierd, best to just precalculate it and write it in.  26.4

Voldirs: That's for future, so people know how these numbers got calculated. Does it look so weird?

NanakoAC: yea this is a bad method

If you want to make it based on the base class, then make the HV bullets a subclass, and have them modify the values in their New proc
and also make 1.2 a define

NanakoAC:
Copy of the PM i sent you this morning 

I have an idea on how to setup the ammo code
do it like this:
```
#define HV_MULT 1.2


#define DAMAGE_762 20
#define AP_762 15
/obj/item/projectile/bullet/a762
    damage = DAMAGE_762
    armor_penetration = AP_762
    
/obj/item/projectile/bullet/a762/hv
    damage = DAMAGE_762 * HV_MULT
    armor_penetration = AP_762 * HV_MULT
```
Make defines like that for each ammotype, that way the actual numbers aren't being copypasted around, and its easily maintainable
and right at the bottom of the file make sure to do...
```
#undef HV_MULT
#undef DAMAGE_762
#undef AP_762
#undef everything_else_you_defined```

Voldirs:
>Make the HV bullets a subclass of the main bullet

They have no var shared. Why doing so?

PR #2844, opened 2019-02-09
Fixes input lag for hear_talk

### LINKED ISSUE #2796: Issue with recorder
https://i.imgur.com/wgIud36.png

I started the recorder, spoke some stuff, signed some stuff in a non verbal language ( probably unrelated)

stopped the recorder, then quickly started and stopped it again once more

the whole conversation shows twice in the transcript though

EvenInDeathIStillServe:
Speechbubble code used sleep() instead of spawn(), meaning that say() code had to wait for the speech bubble to end before calling hear_talk().

Fixes #2796

### DIFF
--- a/code/modules/mob/living/say.dm
+++ b/code/modules/mob/living/say.dm
@@ -254,11 +254,11 @@ proc/get_radio_key_from_channel(var/channel)
 	for(var/client/C in show_to)
 		C.images += I
 	animate(I, transform = 0, alpha = 255, time = 5, easing = ELASTIC_EASING)
-	sleep(duration-5)
-	animate(I, alpha = 0, time = 5, easing = EASE_IN)
-	sleep(5)
-	for(var/client/C in show_to)
-		C.images -= I
+	spawn(duration-5)
+		animate(I, alpha = 0, time = 5, easing = EASE_IN)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/say.dm
```diff
@@ -254,11 +254,11 @@ proc/get_radio_key_from_channel(var/channel)
 	for(var/client/C in show_to)
 		C.images += I
 	animate(I, transform = 0, alpha = 255, time = 5, easing = ELASTIC_EASING)
-	sleep(duration-5)
-	animate(I, alpha = 0, time = 5, easing = EASE_IN)
-	sleep(5)
-	for(var/client/C in show_to)
-		C.images -= I
+	spawn(duration-5)
```
LudwigVonChesterfield: for God's sake, if you think this is optimization look into callback.dm, and think twice

EvenInDeathIStillServe:
Done. Replaced the spawn with addtimer, and a qdel after a spawn with QDEL_IN.

LudwigVonChesterfield:
actually there was an alternative solution of making the entirety of creating speech bubble a seperate proc,
and making that proc a set waitfor = FALSE.

but oh well

PR #2846, opened 2019-02-09
Fix for supermatter monitor gasses

EvenInDeathIStillServe:
Game tracked "phoron" instead of "plasma", meaning the plasma percent was always 0%. Fixed and changed the "PH" for phoron to "PZ" for plasma.
Removed the H2 tracking because we don't have that gas.


GrayRachnid:
All other consoles use TX for plasma. Might want to use that.

EvenInDeathIStillServe:
Done

NanakoAC:
noo don't use TX, its dumb
PZ is fine

GrayRachnid:
:thonk:

PZ sounds very stupid if I were to be honest.

EvenInDeathIStillServe:
TX, standing for Toxin, makes it sound like Plasma is a BAD thing that you DON'T want interacting with your engine. PZ, however, sounds like an eco-friendly energy source that definitely belongs in a absurdly high pressure & temperature environment.

PR #2847, opened 2019-02-09
Allows crowbarring armory showcases

EvenInDeathIStillServe:
You can now crowbar armory showcases if the power is out.


NanakoAC:
but they already can be crowbarred. You apply crowbar to glass until it breaks.

NanakoAC:
These display cases are only supposed to open with the proper authority and under good conditions.

In all other cases, you smash them. Coming to the armoury and finding it smashed indicates something is up, this sets a mood. Allowing them to be unsmashingly forced open under any circumstances would contravene this design

If you need in and theres no power, break it!

PR #2850, opened 2019-02-10
Fixes fuhrer roaches summoning spiders

### LINKED ISSUE #2825: BUG: Fuhrer roach calls spiders as reinforcements
Title. This ends up in roaches fighting with spiders and assistants finishes whoever remains. I think that's unintended and reinforcement should only call roaches.

MadmanMartian:
closes #2825

:cl:
fix: Fixes fuhrer roaches summoning spiders
/:cl:

### DIFF
--- a/code/game/objects/structures/burrows.dm
+++ b/code/game/objects/structures/burrows.dm
@@ -29,6 +29,7 @@
 
 	//Vars for migration
 	var/processing = FALSE
+	var/wanted_type = null //What mob we would like to come to this location
 	var/obj/structure/burrow/target //Burrow we're currently sending mobs to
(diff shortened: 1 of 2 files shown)


NanakoAC:
I'm somewhat dubious on this, attracting many kinds of creatures opens up the not indistinct possibility that they'll have a spontaneous gang war on arrival, creating an interesting story.

I'm curious, do you think this is worth removing? What are the benefits of preventing it?

MadmanMartian:
Personally I've never experienced this (As I've never played on Eris).

I just looked through your issues and found one that looked fun to fix and made a remedy, as I was making a content PR.

NanakoAC:
general consensus is that this is desireable behaviour, its fun and doesn't need fixed, so we won't be merging this PR

PR #2852, opened 2019-02-10
Light color rework

SpadesSlick:
Now areas have a specific var for color that lights should use. It allow us to more carefully and precisely create an atmosphere where we need it.

image
image
image
image


NanakoAC:
This is a step in the right direction, but I thought we were also going to be toning down the range/power of those lights?

They're still too bright and well lit

Issue #2817, opened 2019-02-08
Black spriteless flooring


RandomColonist:
You can encounter this when you remove and then simply reattach flooring:
![blackflooring](https://i.imgur.com/tXYI6lY.png)

Garen7:
should be fixed after #2857

PR #2857, opened 2019-02-12
Black Floor Fix

### LINKED ISSUE #2574: Bug - Floor Tile Sprites Inconsistent
Experienced with nearly every floor tile is some form of this; two or three tiles can be placed together, but if you create a square of four, the last one placed will not properly display the innermost quadrant. This is not entirely consistent nor were all tiles checked, but in a broad sampling I was unable to find a tile that could work for a large construction project.

A number of "dark" tiles did not have sprites while still in object form and could not be placed as floor tiles. Ongoing from #476

Golden floor tiles also cannot be placed.

bugfloortile

NanakoAC:
fixes #2574
Fixes floors being black after repair, and also removes some old obsolete vars


Garen7:
Also fixes #2817

PR #2858, opened 2019-02-12
Fixes lottery taking into account departments, and wild html tags

### LINKED ISSUE #2832: BUG: Lottery
[Common] CEV Eris System Announcer states, "Lottery: Nyx Daily wishes to congratulate `<b></b>` for winning the Nyx Stellar Slam Lottery, and receiving the out of this world sum of 15000 credits!"

Carbonhell:
Fixes #2832 and removes the chance for departments to be chosen as winners.

### DIFF
--- a/code/game/gamemodes/events/money_lotto.dm
+++ b/code/game/gamemodes/events/money_lotto.dm
@@ -24,8 +24,11 @@
 /datum/event/money_lotto/start()
 	winner_sum = pick(5000, 10000, 15000)
 	if(all_money_accounts.len)
-		var/datum/money_account/D = pick(all_money_accounts)
-		winner_name = D.owner_name
+		var/list/private_accounts = all_money_accounts.Copy()
+		for(var/i in department_accounts)
+			private_accounts.Remove(department_accounts[i])
+		var/datum/money_account/D = pick(private_accounts)
+		winner_name = D.get_name()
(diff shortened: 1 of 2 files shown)


code/game/gamemodes/events/money_lotto.dm
```diff
@@ -24,8 +24,11 @@
 /datum/event/money_lotto/start()
 	winner_sum = pick(5000, 10000, 15000)
 	if(all_money_accounts.len)
-		var/datum/money_account/D = pick(all_money_accounts)
-		winner_name = D.owner_name
+		var/list/private_accounts = all_money_accounts.Copy()
+		for(var/i in department_accounts)
```
NanakoAC: this is unnecessary, you can just directly subtract a whole list, like this

private_accounts -= department_accounts

html/changelogs/lottery.yml
```diff
@@ -0,0 +1,4 @@
+author: Carbonhell
+delete-after: True
+changes: 
+  - bugfix: "The lottery cannot be win by departments anymore, and the output message has been fixed."
```
mattroks101: Be won.

code/game/gamemodes/events/money_lotto.dm
Carbonhell: Department accounts is a list with numbers as keys and the actual accounts as value, i thought simply doing -= would attempt to subtract the keys, huh

NanakoAC: oh i see, disregard this comment then

PR #2860, opened 2019-02-13
Bullet-type ammunition refactoring & Rubber Rebalance.

Kurgis:
Refactors all bullet ammunition to use a damage define, and further categorized the list by type, and alphabetized order of appearance to correspond with the define list. ~~This is meant purely as a refactor, if there has been noticeable balance changes caused please point them out.~~ Of course, due to the calculations, damage will be changed to a negligible extent (some bullets do 6.6 damage instead of 7 for example).

Adds rubber damage multiplier and rubber agony multiplier to handle rubber bullet damage and agony, respectively.
Capitalized and moved the defines so it doubles as a table of contents, file is starting to need it.
Moved all multipliers to multiply the damage value, instead of being multiplied by the damage value.
Adds practice multiplier to handle both practice bullet damage and agony.


Kurgis:
Ready for review.

Kurgis:
Changes requested through Discord.

Kurgis:
Ready for review.

code/modules/projectiles/ammunition/bullets.dm
```diff
@@ -8,20 +8,20 @@
 	caliber = "9mm"
 	projectile_type = /obj/item/projectile/energy/flash
 
+/obj/item/ammo_casing/c9mmh
```
NanakoAC: this isn't in proper hierarchy.
it should be 
/obj/item/ammo_casing/c9mm/hv
/obj/item/ammo_casing/c9mm/rubber

etc

code/modules/projectiles/ammunition/bullets.dm
```diff
@@ -31,7 +31,7 @@
 /obj/item/ammo_casing/a10mmh
 	desc = "A 10mm high-velocity bullet casing."
 	caliber = "10mm"
-	projectile_type = /obj/item/projectile/bullet/hv/a10mm
+	projectile_type = /obj/item/projectile/bullet/a10mm/high_velocity
```
NanakoAC: no need tor fully writing out "/high_velocity", /hv will do fine

code/modules/projectiles/projectile/bullettypes.dm
```diff
@@ -1,197 +1,246 @@
...
+#define DAMAGE_32 20
+#define DAMAGE_45 26
 //Carbines and rifles
-#define damage_556 16
-#define AP_556 25
-#define damage_65 18
-#define AP_65 20
-#define damage_762 20
-#define AP_762 15
+#define DAMAGE_10X24 30
```
NanakoAC: add in a few linebreaks to seperate the general categories. Just hit enter twice before each comment

code/modules/projectiles/projectile/bullettypes.dm
```diff
@@ -1,197 +1,246 @@
...
-	embed = 0
-	sharp = 0
+/obj/item/projectile/bullet/c38/rubber
+	name = "rubber bullet"
+	damage = DAMAGE_38 * RUBBER_DAMAGE_MULTIPLIER
+	agony = DAMAGE_38 * RUBBER_AGONY_MULTIPLIER
+	armor_penetration = ARMOR_PENETRATION_REVOLVER * RUBBER_PENETRATION_MULTIPLIER
+	embed = FALSE
+	sharp = FALSE
+	can_ricochet = FALSE
```
NanakoAC: I'm not seeing the logic in making rubber rounds not ricochet. Rubber is bouncy!

code/modules/projectiles/projectile/bullettypes.dm
```diff
@@ -203,38 +252,53 @@
 	spread_step = 10
 
 /obj/item/projectile/bullet/shotgun/practice
-	name = "practice"
-	damage = 5
-	embed = 0
-
-/* Miscellaneous */
+	name = "practice slug"
+	damage = DAMAGE_SLUG * PRACTICE_MULTIPLIER
```
NanakoAC: I'm thinking all the practise rounds should have their damage set to zero.

Make them do small amounts of agony isntead, but there's no practical application for making them do actual damage

code/modules/projectiles/ammunition/bullets.dm
Kurgis: Failed to notice this, will resolve.

code/modules/projectiles/ammunition/bullets.dm
Kurgis: Would this not go against the Naming conventions concerning abbreviations listed in the Contribution Guide? The object is not a variable within a proc.

code/modules/projectiles/projectile/bullettypes.dm
Kurgis: Will remove ricochet vars so they are inherited from the parent, the new-found ability for rubber bullets to ricochet won't be listed in the change-log, however. If you'd be able to include it in Rosalimo's big changelog he commits, that'd be appreciated.

code/modules/projectiles/projectile/bullettypes.dm
Kurgis: Practice rounds don't serve a real purpose. The only purpose they server separately from rubber, is they don't do significant agony damage. This would allow the person being engaged to not get pain stunned, which would allow for a live fire training exercise. There's just one problem with this, the practice rounds are actually lethal. They don't serve much of any significant role that can't be handled by blank shots, or rubber shots. I personally believe they should be removed entirely, and replaced completely with the likes of the blank round.

Kurgis: Will remove the blank shotgun bullet, and make all practice rounds `nodamage = TRUE`.

Kurgis: Misread your initial comment, will make the requested change.

code/modules/projectiles/ammunition/bullets.dm
NanakoAC: In this case, its a bit too unwieldy
Those conventions are largely for variable names, not typepaths.
Besides which, there are some real technical limitations with overlong paths

Kurgis: Considering that, I'll abbreviate the path down to `/hv`.

Kurgis:
Resolved conversations, ready for review.

code/modules/projectiles/ammunition/magazines.dm
```diff
@@ -1,88 +1,21 @@
...
-/obj/item/ammo_magazine/c45smg/rubber
-	name = "smg magazine (.45 rubber)"
-	icon_state = "smg45r"
-	ammo_type = /obj/item/ammo_casing/c45r
+/obj/item/ammo_magazine/a10mm/highvelocity
+	name = "magazine (10mm high-velocity)"
+	icon_state = "12mmh"
+	ammo_type = /obj/item/ammo_casing/a10mm/hv
 
 /obj/item/ammo_magazine/mc9mm
```
NanakoAC: Why does this magazine not specify a calibre?

code/modules/projectiles/ammunition/magazines.dm
```diff
@@ -149,26 +82,77 @@
...
+	name = "magazine (10mm x 24 caseless)"
+	icon_state = "10x24"
+	mag_type = MAGAZINE
+	caliber = "10x24"
+	matter = list(MATERIAL_STEEL = 8, MATERIAL_PLASTIC = 1)
+	ammo_type = /obj/item/ammo_casing/c10x24
+	max_ammo = 99
+	multiple_sprites = 1
 
 /obj/item/ammo_magazine/a556
```
NanakoAC: no caliber here, probably others

Please check calibers on all these magazines

Garen7:
hey @Kurgis when you made this PR did you mean to do the following changes to rubber bullet damage?
- .45 rubbers damage from 45 to 26
- .44 rubbers damage from 73 to 45
- .50 rubbers damage from 85 to 50
- and etc
(damage including agony)
It seems like this would be something to include in your description or changelog. Aren't rubber bullets supposed to have higher overall damage than normal bullets since agony fades?

Kurgis:
It was a requested change, a change-log of it should be added by Rosalimo in the future mentioning it. The reason it wasn't initially put into the changelog, was because the PR "evolved" past what I originally intended.

PR #2862, opened 2019-02-13
Fixes several GC failures

Carbonhell:
Fixes all the gc failures pointed by @NanakoAC that happen a few seconds after roundstart. I have kept a test server up for ten minutes, and none appeared.


code/game/mecha/equipment/tools/medical_tools.dm
```diff
@@ -401,6 +402,13 @@
 		create_reagents(max_volume)
 		synth = new (list(src),0)
 
+	Destroy()
+		syringes = null
```
NanakoAC: setting these vars null may not be adequate. For single object references, this is why we have QDEL_NULL

no such macro exists for lists yet (maybe you should add one) so for a list you have to loop through and qdel each thing in it, then set the list equal to list()

code/game/objects/items.dm
```diff
@@ -67,8 +67,16 @@
 	//Does not affect damage dealt to mobs
 
 /obj/item/Destroy()
-	qdel(hidden_uplink)
-	hidden_uplink = null
+	QDEL_NULL(hidden_uplink)
+	QDEL_NULL(blood_overlay)
+	origin_tech = null
```
NanakoAC: afaik many of these are just strings or lists of strings, there's no need to do this.

examine each one carefully, strings and numbers can be left alone, only references to datums need to be nulled

Carbonhell:
Gonna finish the PR in around 12 hours, after sleeping.

code/game/mecha/equipment/tools/medical_tools.dm
Carbonhell: a macro does exist, QDEL_NULL_LIST, i used that instead of adding a new one, although it's different from /tg/ and I'm not sure which one is better.

Carbonhell:
Ready

Issue #2818, opened 2019-02-08
BUG: Excelsior machinery issues


RandomColonist:
1. Ghost can use verb that toggles excelsior shield generator mode if they are near it.
2. When player passes near generator, it creates minor lag as new tab appears and dissapears. Maybe it would be wise to move that verb to Objects tab?
3. Excelsior turret can't be unwrenched even if it's disabled, instead telling me to turn it off when i try to unwrench disabled turret.
4. Excelsior shield generator don't want to feed from ship powernet when directly connected to it via cable and there is power in cable underneath it (and machinery is wrenched ofc), but feeds just fine from local powernet hooked to the disel generator.
5. Excelsior implant reconstructor can accept multiple implants while reconstructing, resulting in producing only one implant.

RandomColonist:
Added new entry.

PR #2867, opened 2019-02-14
Bugfixes for Excelsior machinery

EvenInDeathIStillServe:
Fix for #2818 

Fixes all the bugs mentioned in that issue. Could not confirm the bug regarding the shieldgen with powernets, as placing the wallgen on an APC terminal allowed it to function properly.


code/game/machinery/excelsior/field.dm
```diff
@@ -26,8 +26,12 @@
 	return
 
 /obj/machinery/shieldwallgen/excelsior/verb/toggle_stun()
+	set category = "Object"
 	set name = "Toggle stun mode"
-	set src in range(1)
+	set src in view(1)
+
+	if ( usr.stat || usr.restrained() )
```
NanakoAC: there's an incapacitated() proc which handles all of this

EvenInDeathIStillServe:
Done.

code/game/machinery/portable_turret.dm
```diff
@@ -299,7 +299,7 @@ var/list/turret_icons
 					qdel(src) // qdel
 
 		else if((istype(I, /obj/item/weapon/tool/wrench)))
-			if(enabled || raised)
+			if(enabled)
```
NanakoAC: Hmmm, this would introduce a severe balance problem

PR #2920, opened 2019-02-17
Recharger Construction/Deconstruction

imw-Oculus:
- Allows Recharger circuits to be built at the autolathe (Circuit Design Disk)
- Allows Rechargers to be constructed
- Allows Rechargers to be deconstructed


code/game/machinery/recharger.dm
```diff
@@ -19,9 +20,15 @@ obj/machinery/recharger
 	var/icon_state_charging = "recharger1"
 	var/icon_state_idle = "recharger0" //also when unpowered
 	var/portable = 1
+	var/screwed = FALSE
```
NanakoAC: this var is never used and not necessary as far as i can tell

PR #2931, opened 2019-02-19
Beartrap Fix and Tweaks

NanakoAC:
Fixes resisting out of beartraps by adding a new permission proc.

Beartraps intercept it and punish you for trying
Beartraps are also now harder, and do some screenshake
beartraps will now catch the victim even if their leg armor blocks the initial damage.


code/game/objects/items/weapons/traps.dm
```diff
@@ -157,8 +157,17 @@ Freeing yourself is much harder than freeing someone else. Calling for help is a
...
 
-
-
+//Attempting to resist out of a beartrap will not work, and you'll get nothing but pain for trying
+/obj/item/weapon/beartrap/resist_buckle(var/mob/user)
+	if (user == buckled_mob && !user.stunned)
+		//We check stunned here, and a failure stuns the victim. This prevents someone from just spam-resisting and instantly killing themselves
+		if (user.client)
+			fail_attempt(user)
+			to_chat(user, SPAN_WARNING("Struggling out of this isn't going to work, you'll need to try to release \the [src] with your hands or a tool"))
```
mattroks101: Would it be possible for resisting out to activate the proc for escaping with your hands, or with a tool if you're holding one?

NanakoAC: It would be entirely possible, i chose not to for a reason.

traps are supposed to be a hugely inconvenient, annoying, and harrowing experience. Their raison d'etre is to provide a chilling effect and force you to rethink your course of action. This starts with the process of escaping it. Attempting to resist is immediately punished, you have to find the little sprite and click on it

NanakoAC:
Updated to fix robot interactions, also made one tangential change: Robots are no longer unable to interact with stuff while buckled.

This was a silly and arbitrary check in the middle of click code, it makes no sense. 
The lockdown function from consoles doesn't use it, so i don't predict any side effects

PR #2933, opened 2019-02-19
Links crew record criminal status with HUDs

### LINKED ISSUE #2876: Crew records and the SecHUD are not connected.
Setting someone to arrest on Crew records doesn't update their HUD. You have to shift click them and change it from there, which should not be the only way to change the status of crewmembers.

Carbonhell:
Fixes #2876

### DIFF
--- a/code/modules/modular_computers/file_system/reports/crew_record.dm
+++ b/code/modules/modular_computers/file_system/reports/crew_record.dm
@@ -216,15 +216,11 @@ FIELD_LONG("Exploitable Information", antagRecord, access_syndicate, access_synd
 
 /datum/report_field/options/crew_record/criminalStatus/set_value(given_value)
 	..()
-	for(var/a in data_core.security)
-		var/datum/data/record/E = a
-		var/datum/computer_file/report/crew_record/C = owner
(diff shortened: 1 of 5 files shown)


code/modules/modular_computers/file_system/reports/crew_record.dm
```diff
@@ -214,6 +214,18 @@ FIELD_LONG("Qualifications", skillset, access_heads, access_heads)
 // ANTAG RECORDS
 FIELD_LONG("Exploitable Information", antagRecord, access_syndicate, access_syndicate)
 
+/datum/report_field/options/crew_record/criminalStatus/set_value(given_value)
+	..()
+	for(var/a in data_core.security)
```
NanakoAC: Datacore is deprecated, we're trying to phase it out. Everything should be done through the crew record system now.

Instead of setting the var in the datacore, you should change the huds so that they read from crew records

Carbonhell:
@NanakoAC Done. This probably fully breaks the security record console,since it's fully based on data_core, but I don't get the point of updating it since the crew records console program does what it does, better.
Might aswell just delete the outdated console.

Issue #2926, opened 2019-02-18
Excelsior autolathes broke AGAIN


GrayRachnid:
![image](https://user-images.githubusercontent.com/6613514/52976683-64298900-3398-11e9-9ab9-a215d459c0e2.png)

@NanakoAC

NanakoAC:
how did this happen?

TorinoFermic:
Fixing it.

PR #2935, opened 2019-02-20
Excelisor autolathe build requirements fix after first building.

TorinoFermic:
### General
Fixes : #2926 

## Changelog
🆑 TorinoFermic
fix: Added back missing required components for the excelsior autolathe circuit board after having built.
/🆑


NanakoAC:
I added that, don't comment it out.
The point of it is to allow excelsior to build their autolathe without needing to scavenge for fiddly parts, so they can get a base up and running anywhere. the autolathe is used to manufacture those parts for all other machines

if its not working, find out what broke my code, don't comment it out

TorinoFermic:
PR Changed to NanakoAC's suggestion with proper fixes.

PR #2936, opened 2019-02-20
Ports radials from /tg/, and applies it to borg icon choice

nichlas0010:
Ports radial menus from /tg/, except for the optional tooltips since tooltips don't exist on eris ~~yet~~

Lets add_overlay use images instead of just icon_states, to let add_overlay work with icon files that aren't the src object's icon. Also moves it to the /atom helperfile, rather than having it in the closetfile 🤔 

Adds the popleft proc which functions like the pop proc, except taking from the start of the array rather than the back


NanakoAC:
Trying it out in practise, i'm not sure if this is an improvement.

You click on the item to get the menu (doing which is itself pretty fiddly) and the menu appears not where you click, but around your character, you have to move halfway across the screen to click anything

In the case of sniper rifles at least, this is a definite downgrade. I tried using the penetrator, the process goes:

Switch to hand holding rifle
Click on rifle
Move mouse to middle of screen and click bolt to open the bolt
Switch to hand containing ammo
move mouse back down to the bottom to click the gun with ammo in hand
switch to rifle hand
Click on rifle
Move mouse to middle of screen and click bolt to close the bolt
move mouse back down to the bottom to click the gun and bring up the menu a third time
Move mouse to middle of screen to click Zoom
Finally after completing this long and arcane ritual, you can fire

Now do it all again to reload and fire another shot.
This is a major useability downgrade, its really not good.

NanakoAC:
This is going to need some playtesting and discussion

nichlas0010:
You know, instead of clicking the item in your hand, you could always just use the attack_self hotkey, which while on hotkey mode is Z, and while off hotkey mode is ctrl+Z. That way, you don't need to be moving your mouse from the middle of the screen at all!

nichlas0010:
Either way, I don't feel particularly strongly about the radial for the sniper, since I just needed something to show that it worked. I'm fine removing that change and just having the framework merged first.

SpadesSlick:
Yeah, the point of this PR should be in framework, not in actual use. We can use in whatever way we want later.

nichlas0010:
Removed the sniper change, only the framework remains now.

nichlas0010:
Here it is in action :)

nichlas0010:
Except there's a popup going "hey do you like the icon?" at the end now

PR #2950, opened 2019-02-22
makes the telecomms consoles scan for any telecomms machines on the Z-level

nichlas0010:
requested by @GrayRachnid


NanakoAC:
I seem to remember saying something about level data on another PR, it would apply here too

nichlas0010:
Yeah, forgot to close this PR too 😄

NanakoAC:
why close it, can't you improve it?

nichlas0010:
I've no clue how level data works, and I couldn't personally I see a reason why the usage of map_data in is_on_same_plane_or_station() is wrong, so I just moved on to other projects 🤷‍♂️

PR #2953, opened 2019-02-22
makes the prisoner console check all connected Zs

nichlas0010:


code/game/machinery/computer/prisoner.dm
```diff
@@ -32,7 +32,8 @@
 			var/turf/Tr = null
 			for(var/obj/item/weapon/implant/chem/C in world)
 				Tr = get_turf(C)
-				if((Tr) && (Tr.z != src.z))	continue//Out of range
```
NanakoAC: same plane or station?

Planes are used for visual sorting, they shouldn't be involved in this
And we're on a ship not a station

what exactly does this function do? Also please rename it

nichlas0010: the function compares the two Z-levels given, and first checks whether they're identical, then checks whether they're both ship-levels. If either of those are true, it returns true, otherwise it returns false.

Why it's called **plane** or station, I've no clue.

NanakoAC: ok well this is no good.

Camper recently added a new level_data system which better stores data about exactly what zlevels are connected and how. Look into that and use it

nichlas0010: Well as far as I can tell, it already uses the map_data system, so I'm not sure what you want me to do.

PR #2954, opened 2019-02-22
fixes magazines not dropping to the ground properly

nichlas0010:
put_in_hands returns false if it gets put on the floor, meaning the magazine got dropped but not unlinked if there wasn't space in the user's hands.


NanakoAC:
this seems to be a change, not a fix? I'm not seeing how it was broken before

GrayRachnid:
Right now if you eject the magazine while your other hand is full, it drops a fake magazine on the floor while the real one remains inside the gun.

code/modules/projectiles/guns/projectile.dm
```diff
@@ -212,10 +212,10 @@
 //attempts to unload src. If allow_dump is set to 0, the speedloader unloading method will be disabled
 /obj/item/weapon/gun/projectile/proc/unload_ammo(mob/user, var/allow_dump=1)
 	if(ammo_magazine)
-		if(!user.put_in_hands(ammo_magazine))
```
nichlas0010: This line meant that if put_in_hands returned false, the rest of the proc wouldn't run. If put_in_hands put it on the tile beneath the user (because their hands are full) put_in_hands would return false, meaning the magazine would drop to the ground but not get unliked through the rest of the proc below.

PR #2966, opened 2019-02-22
adds select equipment to the right click menu on humans for admins.

nichlas0010:
because going into the fun tab, then selecting the mob in an inputbox of all the mobs is bad 👎

### DIFF
--- a/code/modules/admin/verbs/debug.dm
+++ b/code/modules/admin/verbs/debug.dm
@@ -377,11 +377,12 @@ ADMIN_VERB_ADD(/client/proc/cmd_debug_tog_aliens, R_DEBUG, FALSE)
 
 
 ADMIN_VERB_ADD(/client/proc/cmd_admin_dress, R_FUN, FALSE)
-/client/proc/cmd_admin_dress()
+/client/proc/cmd_admin_dress(mob/living/cabon/human/M in GLOB.human_mob_list)
 	set category = "Fun"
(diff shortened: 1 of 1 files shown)


NanakoAC:
I don't think this one will be merged, couple reasons:

1. Byond sucks, and verbs add lag to the rightclick menu. The more, the worse. This may not apply since it was already a verb. uncertain
2. The rightclick menu is far too cluttered with useless trash already that are rarely or never used, like

- Man up
- Jump to mob
- Direct Narrate
- Subtle Message

This list needs to become smaller, not larger

3. Outfit selection is very much a niche feature you'll use once in a blue moon. It belongs in the VV dropdown menu, or in the player panel, or where it is now.
if the list of all mobs is a problem then fix that list to only show human players

GrayRachnid:
Outfit selection is not a niche feature that is used once in a bluemoon. Infact, it was the first feature I ever asked for and you straight up denied it. The fact that nich is asking for it too right after he started coding shows that it's not a niche feature. Anyone who's come from a /tg/ codebase knows how useful this is, and making it easily accessible is very important.

GrayRachnid:
@NanakoAC please consider reopening this.

NanakoAC:
It would be a lot better to improve the menu for picking who to set an outfit on. I still believe this is too niche to be on the rightclick menu

getting a list of the actual players without monkeys would make it pretty simple to choose

PR #2977, opened 2019-02-23
Datumizes world/topic & admin/topic

nichlas0010:
Only two mechanic changes have been made to world/Topic, that being the removal of ?players and the changing of ?ping returning 1 more client than there really is.

?players was removed because it wasn't used and was just a clone of ?ping, except without an increment, and ?ping was changed to return the number of clients rather than the number of clients +1, cause that makes sense.

all the admincaster datums have been merged into one, using admincaster as the href

Port of: https://github.com/tgstation/tgstation/pull/31543/files


nichlas0010:
did about 2/3rds of the admin topic so far 😄

ACCount12:
Nice work. We need to port a lot of refactors to modernize the codebase. Guess I'll start by porting mine too.

nichlas0010:
Ready for review :)

NanakoAC:
This is risky, change the target branch to Development please, we'll try it out on the test server first

PR #2990, opened 2019-02-25
Sets up manual links.

Kurgis:
Has #3049 as a dependency.

Moves the majority of manuals over to wiki manuals, these connect right to the wiki by checking what's set in the config. The example config shows the needed links to make them work. Only a few manuals which didn't fit, or are better as is, were left with the old manual system. This means that most used manuals are no longer hardcoded, and a PR isn't required every time for a spelling mistake.

Default window size can now be edited on books.

Adds a book icon for ironhammer paragraphs, removes various unused icons which aren't planned on being used again.


Kurgis:
Ready for review.

code/game/objects/items/weapons/manuals.dm
```diff
@@ -20,13 +20,14 @@
 		</head>
 
 		<body>
-		<iframe width='100%' height='97%' src="[config.wikiurl]Guide_to_construction&printable=yes&remove_links=1" frameborder="0" id="main_frame"></iframe>
+		<iframe width='100%' height='97%' src="http://wiki.syndi-cat.com/index.php?title=Guide_to_construction&printable=yes&remove_links=1" frameborder="0" id="main_frame"></iframe>
```
NanakoAC: Not a good design

Not needing to edit configs is good, but that doesn't mean full urls should be hardcoded everywhere
At the very least, wiki link should be a define
Ideally, make a proc for a wiki link that also adds stuff like the printable and remove links, so that you only need to supply an article name

Kurgis: I'd need to know what the config is going to be set to, otherwise there are issues which will make the link not work. For example if the config has http://wiki.syndi-cat.com/index.php/Main_Page or http://wiki.syndi-cat.com/index.php/ set as its wikiurl, it won't pull up a page. So I'd need to know what the wikiurl is going to be set as, since it will need to be http://wiki.syndi-cat.com/index.php?title= I'm unaware if any other url works.

Kurgis:
I'd like to also acknowledge a myriad of issues that arise with using this oldcode, as certain page layouts could allow you to browse the entire wiki using an in-game book. This could be done with something as simple as a table of contents for related mechanics, which is far less than ideal. I'm going to look into alternatives, or better ways of doing this.

Kurgis:
This code kills the map, and will require map changes to remove non-existent manuals.

Kurgis:
Ready for review.

NanakoAC:
It does indeed kill the map. I'm afraid we can't merge this right now.
Partly due to a policy of code integrity, but also because i -literally- can't merge it if it doesn't pass Travis

Map is currently locked for significant remodelling by clock, so this PR will have to be on hold until after that is done

Kurgis:
Good, gives time for the articles to be made for the manuals to be linked to.

Kurgis:
Due to the way the wiki is now being set-up alongside Gammastation, I'll likely have to rethink how this is done. PR will likely be closed to create a new one, but I'm going to keep this up until then.

Kurgis:
Can't figure out how to squash my commits, if anyone has some pointers please DM me over Discord.

SpadesSlick:
Any progress will be made here?

SpadesSlick:
Oh right, map. It's not locked now, but you have only 5 days until it will be locked again.

Kurgis:
Couldn't make the changes within that time, when the map is no longer locked, DM me again if you may.

PR #3000, opened 2019-02-25
Bandanas & Balaclava Additions

EvenInDeathIStillServe:
Adds bandanas that can be worn on the head or the face. Added them to loadout, and clothing closets.

Gave the tactical balaclava a new state, where it covers only the mouth. Also fixed a bug where you showed up as Unknown even if your face was uncovered with it. Adjusting the bandana now has an action button on the HUD.

Fixes some misspellings.


EvenInDeathIStillServe:
Yoink, PR #3000 get!

code/modules/clothing/masks/balaclava.dm
```diff
@@ -12,23 +12,33 @@
...
-			gas_transfer_coefficient = initial(gas_transfer_coefficient)
-			body_parts_covered = initial(body_parts_covered)
-			icon_state = initial(icon_state)
-			user << "You pull balaclava up to cover your face."
-		update_wear_icon()
+/obj/item/clothing/mask/balaclava/tactical/proc/adjust_mask(mob/living/carbon/human/user)
+	if(!istype(user))
+		return
+	if(!user.incapacitated())
+		open = (open + 1) % 3
```
drexample: ...why?

EvenInDeathIStillServe: It's one line of code, instead of 3 lines that you would get if you manually set the state after the If's.

NanakoAC: Sorry, Drexample is right here.

While reducing unnecessary code is usually a plus, turning several readable lines of code into a confusing cipher is not good. Generally this sort of thing should only be done if necessary, for performance critical code (like lighting). Any human initiated action generally doesn't count as such

make this simpler and more readable please

EvenInDeathIStillServe:
Done.

PR #3005, opened 2019-02-26
Hivemind nodes and mobs fixes

Macleod962:
Whitelist updated with pipe meter
Nodes isn't spawns too close to each other now
New nodes takes wireweeds around them properly
Mobs dies once again
Infested machinery now saves pixel position

Hivemind now saves assimilated machinery curcuit if possible and drop it on destruction. Otherwise, machinery will be spawned back
Whitelist updated with cryopod to prevent player spawn issues


code/modules/hivemind/wires.dm
```diff
@@ -188,43 +188,52 @@
 //assimilation process
 /obj/effect/plant/hivemind/proc/assimilate(var/atom/subject)
 	if(istype(subject, /obj/machinery) || istype(subject, /obj/item/modular_computer/console))
+		//new node creation
+		if(hive_mind_ai.hives.len < 10)
```
NanakoAC: Hardcoding these numbers in here is a bad idea, make them defines

Macleod962:
Fixed.
Added a few more changes, changelog updated

PR #3008, opened 2019-02-27
Fixes #2996, #2998, and ion trails

### LINKED ISSUE #2996: Hardsuit Mounted Drills think they need fuel
Hardsuit mounted drills work fine, draining power and all until you use them to the point where a regular drill would be out of fuel. At this point, they become useless as it is impossible to refuel them.

I believe the fact that they need fuel at all is a bug, though it may not be. The fact that you cannot refuel them at all would be the bug in that were not the case.

### LINKED ISSUE #2998: Look up doesn't work in space
Instead of showing the zlevel above you, it just prints "You see space" to chat

this feature is pretty vital for climbing around the hull

CakeQ:
Fixes ion trails lasting forever
Fixes #2996
Fixes #2998


code/game/objects/items/weapons/tools/pickaxe.dm
```diff
@@ -55,6 +55,10 @@
 	use_fuel_cost = 0.15
 	max_fuel = 120
 
+/obj/item/weapon/tool/pickaxe/diamonddrill/nofuel
```
NanakoAC: this feels kinda hacky, but it'll do for now. The drill still uses power right?

Also change the name and path to something application specific like /diamonddrill/rig

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1438,7 +1438,7 @@ var/list/rank_prefix = list(\
 			if(client.eye == shadow)
 				reset_view(0)
 				return
-			if(istype(above, /turf/simulated/open))
+			if(istype(above, /turf/simulated/open) || istype(above, /turf/space))
```
NanakoAC: Instead of either of these, check above.is_hole

That var will be true on space tiles, openspace tiles, and in future on any other flooring which is wide open

PR #3011, opened 2019-02-27
Tile stack fixes.

Kurgis:
Tweaks the debug tile stack icon, fixes white grayplatform stacks having the debug sprite, fixes steel platform tile stacks having their colors inverted. Removes matter from the tiles parent, so cyborgs can't produce steel.


NanakoAC:
I'm not quite understanding the change for robots here.

While i do agree that them being able to convert energy to matter is a bit silly, its not a bug. Its a feature that's working as intended (according to the intentions of the bay coders who designed it)

I would support removing this feature and requiring robots to eat and store stacks of material, but that's something that should be done properly.  The change you've made here seems to be a hacky way to do it, which introduces logical inconsistencies

Kurgis:
Quite, though I'm ill-prepared to do so within this PR. I'm going to remove the matter changes from the floor tile parent so that this may be merged.

Kurgis:
Ready for review.

PR #3031, opened 2019-02-28
Hivemind: Wireweeds and machines fixes

### LINKED ISSUE #3021: HIVEMIND | Flood in chat, with use of little items
image
I can click, and click, and click with that on Hivemind structures, and it will flood chat with message, of useless usage of it.

### LINKED ISSUE #3022: HIVEMIND | Hivemind wires can't grow under some structures.
image

### LINKED ISSUE #3023: HIVEMIND | Wires sprite connection problem
1e6ec9e91739a4460cda17a8e368f95d

### LINKED ISSUE #3024: HIVEMIND | Hivemind monsters are dumb.
Hivemind creatures can be placed into disposal unit, and they will never leave it. Also, they can't unbuckle by themselves. Hivemind humans, when trying to look like they are dead, attack windows, so that uncover them as dead bodies. Also, when they trying to look like they are dead, still use their ability to scream.

### LINKED ISSUE #3027: HIVEMIND | Hivemind structures doesn't vanish, when wires is removed.
image

### LINKED ISSUE #3028: HIVEMIND | Some problems, with recovering of machinery.
1e6ec9e91739a4460cda17a8e368f95d
When i destroyed transformed APC, it moved from wall, to floor.

Macleod962:
==Machines:
Now they can slowly regenerate their health.
Some of them requires wireweeds on their turf and will slowly dying if not found them. Fixes #3027
Assimilated machinery now spawns properly. Fixes #3028
Ineffective attack message now less spammy. Fixes #3021
==Mobs:
Now they can go out disposal units and unbuckle themselves.
Himan now doesn't attack surroundings during fake death.
Fixes #3024
==Wireweeds:
Sprites now updates properly and spreading should look much better and smoother. Fixes #3023
They also can pass some not anchored density structures, like closets or crates. Fixes #3022
Wires now use only not maintenance burrows and break them open.
If they dying they doesn't break doors anymore.


code/modules/hivemind/machines.dm
```diff
@@ -1,6 +1,7 @@
 //Hivemind various machines
 
-#define HIVE_FACTION "hive"
+#define HIVE_FACTION 			"hive"
+#define REGENERATION_SPEED 		4
```
NanakoAC: After extensive testing with the blobs, i'm going to say that the core needs to have no regen.

In this case, the nodes, the source object, whatever. The big thing where all the wires come from. Make sure it has no regen, if players make a suicidal dive to damage it, any hits they land should have a permanant effect

Every other machine having regen is fine, and give the cores a larger health pool to compensate if you like

code/modules/hivemind/machines.dm
```diff
@@ -175,7 +219,8 @@
 		. = ..()
 		take_damage(I.force)
 	else
-		visible_message(SPAN_WARNING("[user] is trying to hit the [src] with [I], but it seems useless."))
+		if(prob(25))
```
NanakoAC: this probably shouldn't be luck based.

I recommend implementing Resistance, used on creatures, windows and airlocks. Reduce incoming damage by a flat amount

code/modules/hivemind/mobs.dm
```diff
@@ -107,6 +107,17 @@
 
 	closet_interaction()
 
+	//if somebody put us into disposal unit, let's go out
+	if(istype(loc, /obj/machinery/disposal))
```
NanakoAC: why not just check if the loc is any kind of /obj?

code/modules/hivemind/wires.dm
```diff
@@ -244,24 +238,22 @@
...
-				new saved_circuit(created_machine)
-			else
-				created_machine.assimilated_machinery_path 	= subject.type
-				created_machine.assimilated_machinery_dir 	= subject.dir
-			qdel(subject)
+			created_machine.consume(subject)
 
 	//Corpse reanimation
-	if(istype(subject, /mob/living) && !istype(subject, /mob/living/simple_animal/hostile/hivemind))
+	if(isliving(subject) && !istype(subject, /mob/living/simple_animal/hostile/hivemind))
```
NanakoAC: make a proc for hivemind creatures

code/modules/hivemind/mobs.dm
Macleod962: But some of them has a special code for such things. If we just change mob's loc isn't it cause issues with this object?

Macleod962:
Updated

PR #3039, opened 2019-03-02
Fail chance on tool actions is now rounded down

ACCount12:
This makes fail messages look more sane (no more 21.155372% fail chances) and also removes sub-1% failures. Having a sub-1% operation fail is worse than my XCOM luck.


ACCount12:
I stand by this PR being as it is.

Having decimals displayed is dumb. It's information that's overly technical and largely useless. You can't even use this info to track your tool status because it only displays after you fail the operation. As for decimals of a percent being used in calculations - the difference is small enough to be negligible (less than 1% worst case), and after enough fails like that, memes stop being memes and become player frustration.

NanakoAC:
It shows a gradual shift over repeated attempts, and i think players prefer having that information than not having it.  Like i said, i'm not opposed to rounding, we don't need six digits of precision. I think two is a good compromise

But this is rounding for display. 
Rounding before doing the prob check is a definite no

ACCount12:
I don't see any good reason to not round before check.

Jc1469:
To take XCOM as an example, XCOM rounds up the displayed chance. If you have 99.6, it's rounded visually to 100. But you can still get the 0.4% fail chance.  
And to this, I simply call upon the late and great Reggie.
https://www.youtube.com/watch?v=FdxEUL18ySY

Funny memes on reddit are cool and all, some people go "haha unlucky". Other people go "that's fucking dumb."

XCOM is dumb.
Don't be like XCOM.
Vaccinate today.

NanakoAC:
Its a loss of precision, and precision will be important when people are making small buffs to their stats

I've offered a couple of compromises. Think on them and re-evaluate this PR. It will not be merged in its current state

GrayRachnid:
You have 100 numbers to deal with, that's already enough for the average player to make calculations. Using decimals to include another million options isn't going to make this game any more interesting. Please think about what you're saying by rejecting this PR.

seekin4u:
Do you really want to be as fucking XCOM, is it worth of that or what? Truncated 0.6555532% will help you only once per 100 tries, it is good enough.
Every time 23,666578543% appears on my monitor i cry like a bitch, it is a good improvment with a very little chances loss.

NanakoAC:
This PR has not been rejected, it simply has changes requested. Rounding the numbers for external display is fine, using rounded numbers for internal calculations is bad

the cases of <1% fail chance can be handled as an exception

nichlas0010:
Or, how about you just take the easy way out of this by rounding it up instead 🤔

NanakoAC:
Here's what needs to happen for this PR to be merged:

1. Move the rounding to after the prob call is done. It should not be rounded before internal calculation
2. Fix the rounding so that it doesn't erroneously display 0 or 100 in cases where thats not true. Rounding up is probably simplest, or some special case checks

Do these things and it will be approved

NanakoAC:
Due to obstinate noncooperation from the author, this PR is rejected.

Any more reasonable coder is welcome to make a PR with similar changes

nichlas0010:
~~_or maybe it's obstinate noncooperation from the headdev_~~

ACCount12:
Well, seems like there is no one left to disagree with this now.

PR #3043, opened 2019-03-02
Removes trademark, and memetindo direct reference.

Kurgis:


NanakoAC:
the metroid sprites, are they actually ripped from a game, or just inspired by one?

Kurgis:
Hard to tell since most Metroid sprites were already removed from the code at the point it was pulled for Eris, I'll check Bay's commit history.

Kurgis:
The Phazon sprite (removed in this pr from the mob icon file) was created by Doohl, added to Baystation12 in 2011, seems to of been ported from /tg/ station.
The gib Metroid sprites removed from the mob.dmi file I can't find the source for, I will make the assumption they were sprited by another /tg/-goer in 2010-11, or taken from a sprite sheet.
After looking at a asset sheet from Metroids for Game Boy, it appears they took a Metroid sprite, stuck it into DM, resized it to be 32x32, and then made a new sprite based of the image. This image, however, was already removed before this PR.

While the Metroids are no longer in-game, being replaced by slimes, you can see the past mechanics. When a Metroid would attach to you for example to deal damage, so now does a slime attach to you to do the same. Metroid cores replaced with slime cores, and so on. Only thing left is Doohl's Phazon Metroid sprite, and the death sprites for another Metroid mob which was already removed. I've come to find that the Phazon mecha itself is inspired by Metroid after looking through all this, and you can see the shape a familiar character in its finished frame.

PR #3048, opened 2019-03-03
Robot and AI clickfix

NanakoAC:
My clicking fix yesterday wasnt comprehensive enough, robots and AI were still broken

Now fixed
also removed a silly config check, we're not ever going to turn off click cooldowns and having the check there just adds extra work to every click


Anonmare:
Can I get a speedmerge on this? Kind of an important fix

PR #3049, opened 2019-03-03
Adds language config.

Kurgis:
Allows you to connect to both parts of the wiki using the config, and not hardcode.


config/example/config.txt
```diff
@@ -175,6 +175,9 @@ GUEST_BAN
 ## set a server location for world reboot. Don't include the byond://, just give the address and port.
 #SERVER server.net:port
 
+## Language being used for the server, deals with external links. Use either "En" for ENGLISH, or "Ru" for RUSSIAN. Do not capitalize the second letter.
```
NanakoAC: why is the capitalisation important?

Kurgis: Because of how the wiki has been set-up.
http://wiki.syndi-cat.com/index.php

PR #3050, opened 2019-03-03
Makes openspace clickable

Camper12345:
Openspace turf is now clickable, so players should be capable of building lattice over it.


Camper12345:
It uses a plain black icon with alpha 1 because modifying object's alpha variable also affects its vis_contents transparency, making whole tile black.

NanakoAC:
isn't this reverting what you did with vis contents? I thought part of the point of that was being able to click stuff on lower levels, for example to fire guns downwards?

Camper12345:
To build floor over openspace you should click an openspace turf. Also, clicking on object displayed on vis_contents isn't doing anything

RyanSmake:
This is not what you should have done. There is a BYOND var that enforces clicks. 
To quote the documentation:

> This may be used to control how mouse operations on an object are interpreted. A click or mouse movement over an object's icon normally applies to that object only if it is the top-most object that is not transparent at the position of the mouse. Setting `mouse_opacity` to 0 would cause the object to be ignored completely, and setting it to 2 causes it to always be chosen over any lower-level objects, regardless of the transparency of its icon.

PR #3069, opened 2019-03-05
Adds Computer Parts Disk

EvenInDeathIStillServe:
Adds a computer parts autolathe disk, courtesy of Moebius Laboratories.

The disk can be acquired the same way you can get other disks.
The disk contains designs for standard computer parts, and frames for portable modular computers. Higher grade parts will still need to be made at R&D.
Portable modular computers can now be disassembled for materials in the autolathe.


NanakoAC:
I have several questions and concerns related to balance:
1. Will recycling a computer also recycle all its components too? It should
2. What are the material costs for computer components like? They should all be using small amounts of less common materials like silver
3. Tesla link is a concern, i don't want it to be easily obtainable. Make sure there's some gold in its matter costs

EvenInDeathIStillServe:
Recycling a computer gives you all the component materials, including silicon.
Prices are 1 metal + 1 plastic for most things, 0.5 silver for advanced parts.
Tesla link is 1 metal + 1 plastic + 1 gold.

PR #3083, opened 2019-03-07
Implant fixes

EvenInDeathIStillServe:
You can now put implants back into empty cases. You can now scoop up an implant by using an implanter on it, if the implant isn't in a case.
Fixes a runtime error when using an implanter on an empty case.


code/game/objects/items/weapons/implant/implant.dm
```diff
@@ -19,6 +19,18 @@
...
 
+/obj/item/weapon/implant/attackby(obj/item/weapon/I as obj, mob/user as mob)
+	..()
+	if (istype(I, /obj/item/weapon/implanter))
+		var/obj/item/weapon/implanter/M = I
+		if(is_external())
+			return
+		if(!M.implant)
+			user.drop_from_inventory(src)
+			forceMove(M)
```
LethalGhost: It's better to replace previous lines with `if(user.unEquip(src, M))`

code/game/objects/items/weapons/implant/implantcase.dm
```diff
@@ -62,9 +60,14 @@
...
-			if(implant && !M.implant)
+			if(implant && !implant.is_external() && !M.implant)
 				implant.forceMove(M)
 				M.implant = implant
 				implant = null
 				update_icon()
 				M.update_icon()
+	else if (istype(I, /obj/item/weapon/implant) && !implant)
+		user.drop_from_inventory(I)
+		I.forceMove(src)
```
LethalGhost: Same here.

EvenInDeathIStillServe:
Done.

LethalGhost:
@EvenInDeathIStillServe 
Ugh. Sry, i missed important note - `Mob.unEquip(Item)` will return `null`  if `Item` is not equipped on `Mob`.
But `unEquip` is propper way to remove items from mob.

EvenInDeathIStillServe:
I tested it and it works properly. When would a player be able to use attackby with an item if they weren't holding it?

LethalGhost:
@EvenInDeathIStillServe 
Check situation:
`/obj/item/weapon/implant` is placed on table. And user attack it with `/obj/item/weapon/implanter`

PR #3085, opened 2019-03-08
Roaches can be taken off surgery tables

EvenInDeathIStillServe:
Roaches are no longer permanently stuck on operating tables/altars when laid upon them.


mattroks101:
Hang on, roaches are carbon?

EvenInDeathIStillServe:
Yes. They're superior animals. They, along with spiders, are carbon mobs.

PR #3142, opened 2019-03-17
Attack Log viewer

FlattestGuitar:
pic

Can reach it from the player panel, as well as the bigger list of players (which is also incidentally called Player Panel?). No more clicking through VV to read those ugly logs.

I considered adding a search function, but disappearing logs would be annoying as hell. You can ctrl+f in there anyway.

### DIFF
--- a/code/modules/admin/admin.dm
+++ b/code/modules/admin/admin.dm
@@ -31,6 +31,32 @@ proc/admin_notice(var/message, var/rights)
 
 ///////////////////////////////////////////////////////////////////////////////////////////////Panels
 
+/datum/admins/proc/view_log_panel(mob/M)
+	if(!M)
+		to_chat(usr, "That mob doesn't seem to exist! Something went wrong.")
+		return 
+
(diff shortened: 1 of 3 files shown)


code/modules/admin/admin.dm
```diff
@@ -31,6 +31,32 @@ proc/admin_notice(var/message, var/rights)
...
+	body += "<body><center>Logs of <b>[M]</b><br>"
+	body += "<a href='?src=\ref[src];viewlogs=\ref[M]'>REFRESH</a></center><br>"
+
+
+	var/i = length(M.attack_log)
+	while(i > 0)
+		body += M.attack_log[i] + "<br>"
+		i--
+
+	usr << browse(body, "window=\ref[M]logs;size=500x500")
```
FlattestGuitar: The ref is there so you can have multiple windows open at once.

seekin4u: don't we have user_browse or something like that?
Pure output without wrapping is not so good

Issue #3041, opened 2019-03-02
Remains can be found in rubbish piles, but cant be removed - so the pile cant be cleared


driecg36:
This item isn't interactable with, so it just keeps shifting around in the trash and preventing you from clearing it.

driecg36:
https://i.imgur.com/8BGn7Xr.png this is the item

Kurgis:
This appears to be the relevant line: https://github.com/discordia-space/CEV-Eris/blob/master/code/game/objects/effects/decals/remains.dm

These items are meant as decals, and I don't believe they should ever be spawning in containers.

PR #3161, opened 2019-03-19
Removing unclickable item from scrap loot list

Kernstall:
Closes: #3041


ACCount12:
I don't think that actually fixes anything. There is still a bunch of `/obj/effect/decal/cleanable` objects in spawn list.

Kernstall:
image
There was no visible problem with decals cause they somehow never get to var/obj/item/weapon/storage/internal/updating/loot
must be related to this part of code and mechanic of hidden loot.
image

exclude list looks working fine and this will fix most of the cases this item found in scrap...Except the circle /obj/random/junk/nondense -> /obj/random/tool -> /obj/random/lowkeyrandom -> /obj/random/junk, but that mess has low chance and needs much more attention

PR #3167, opened 2019-03-21
Surgery is now radialized

Kmc2000:
Removes ugly input boxes when taking organs out / separating organs.


ACCount12:
We may want to simplify the way radial menus are built in the future, but this is out of scope of this PR.

PR #3210, opened 2019-03-23
Vending Machine Economy Changes

XanderDox:
Changes prices across vendors to be more reasonable and realistic within on our current economy. The changes were based on a document by Regins, and consideration for the cost of actually making the item.

https://docs.google.com/document/d/13b1mjcAhTd6mCkLdbzbs6Xnr00cQCCN1WVOyvYWgO68/edit?usp=sharing < Changes can be found here, there are 1-3 differences based on when I disagreed with the sheet, but these are only be 10-20 credits.


XanderDox:
Closing due to being told it won't be accepted  due to disagreements on how to resolve the issue of the economy.

PR #3214, opened 2019-03-23
[BALANCE] AMR ammo and TC cost changes

TorinoFermic:
## General
As of discussed in Discord with ACCount and Gray.

If we make the sniper without hitscan on, therefore giving any alert targets a chance to escape and to survive.

## Changelog
🆑 TorinoFermic
tweak: AMR and ammo has their costs and contents changed.
/🆑


Stealthkibbler:
goodbye sweet prince

PR #3223, opened 2019-03-27
Job Skills

Regis-Emiel:
I talked to Xander about this and his response after discussion was that we disagree on what the job skills should be at. I like for most of the departments with exceptions for heads to be similar in their competence, meaning that the service department don't have weirdness like the Gardener have 20 Robust but the Bartender having 0.

So it's basically just a gradient and the idea is, you don't feel punished for picking certain jobs as most of them give you some functionality with the skill system. This is my first PR so be gentle

### DIFF
--- a/code/game/jobs/civilian.dm
+++ b/code/game/jobs/civilian.dm
@@ -1,0 +1,204 @@
+#define LOYALTY_CIVILIAN "As a civilian, your only loyalty is to yourself and your livelihood.<br>\
+		You just want to survive, make a living, and get through the day. You shouldn't try to be a hero, or throw your life away for a cause. Nor should you hold any loyalties. Civilians should be easily corruptible, willing to take bribes to do anything someone wants and stay quiet about it."
+
+/datum/job/bartender
+	title = "Bartender"
+	flag = BARTENDER
+	department = DEPARTMENT_CIVILIAN
+	department_flag = SERVICE
+	faction = "CEV Eris"
+	total_positions = 2
+	spawn_positions = 2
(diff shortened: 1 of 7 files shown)


XanderDox:
It appears that you're adding a lot of files instead of editing existing ones in commit aa757a4. Are you meaning to do this?

RandomColonist:
Why IH Commander has 10 mechanic while not really needing it while Gunny lacks mechanic which he could use for brig maintenance (also could use bio because he sits in brig all round while medspecs can be away). Also operative, possible war vet and paid professional mersenary still equally robust to a fucking cook or janitor who cook food and sweep floor.

Regis-Emiel:
@XanderDox 
Yeah I asked for help with this before, I really don't know how the PR stuff works. I just knew the skills were in disarray and that it's relatively easy to look inside the DM files to change those numbers around. Though nobody will upload a PR for me.

@RandomColonist 
You haven't seen the original stats have you? There were a lot of civilian / non security roles that had equal robust stats to security, now none of them are equal to them - they are lower. 

An example was the Gardener had 20 Robust and 15 Tough. They now have 10 Robust and 10 Tough. 
I tried to make the existing stuff more reasonable and then round that out, which even then I was told was a big stretch. Also the Head Roles just by default are always stronger than other departmental roles, that's something I didn't want to change and I just wanted to fix what was more or less broken. I was not given permission to do much else. This may even be beyond what I'm allowed to do as it is.

Regis-Emiel:
i have been informed that i dun goofd with the pr

i will now attempt to pr correctly good sirs

PR #3232, opened 2019-03-27
Boozeomat drinks are free again & knives screwdriving

EvenInDeathIStillServe:
Makes it so that the boozeomat doesn't have auto-pricing anymore, and gives low screwdriving quality to knives.


RandomColonist:
You can open basically any door now with knife and crowbar, just think about it.

PR #3235, opened 2019-03-28
Changelog update

ACCount12:


html/changelog.html
```diff
@@ -405,83 +428,10 @@ <h3 class="author">Nanako updated:</h3>
 				<li class="tweak">Majorly reworked how material costs are calculated for protolathe, autolathe and circuit imprinter designs. Some costs were adjusted in the process</li>
 				<li class="tweak">Made the prices of ordering materials in the guild far cheaper. Also added a bulk metal crate for large quantities of steel at once.</li>
 			</ul>
-
-			<h2 class="date">20 January 2019</h2>
```
mattroks101: Was this removed on purpose?

mattroks101: Looking at the other changelog PRs it doesn't look like any other mass removed parts of the changelog, did the tool mess up here? The server does use this HTML file for the ingame changelog right?

ACCount12: Yes. 

I didn't touch the file, I just ran the generator tool, so I have no idea why was it removed. It could be that at some point, someone added it manually, and so, there was no matching entries for it when changelog regenerated.

mattroks101: Yeah it's really weird that it removed such a large section of the changelog, but I don't see any duplicates of this section so hopefully we can find some way of keeping this section of the changelog without messing the tool up.

ACCount12: I reviewed the changelog tool code. Apparently, it auto-removes old changelog entries from HTML file, intentionally so. And this one is more than a month old.

The related entries are still present in `.all_changelog.yml`.

Case closed.

PR #3237, opened 2019-03-30
Enables Cargo Exports

EvenInDeathIStillServe:
Changes:
Re-enables cargo exporting. If an item has an export datum defined, it will be used. If it doesn't but has a price_tag, its export value will be set to that.
Adjusts prices for most cargo packs: generally everything was made cheaper, except surplus weapon crate which was bumped up to 3000 credits.
Space cleaner now removes color modifiers from stuff, for example if paint was used on it or if it's an old rusty item.

Fixes some bug and oversights:
R&D console apparently required you to use an EMP if you wanted to disable the safety protocols. It now uses emags.
Closets and crates no longer spawn with an invisible cover that only appears after you toggle it open/closed. 
Changes some instances where the supply console refered to Eris as a station.


Voldirs:
But there is white paint for washing colors away. You know painting people is fun, why taking it away with so easy way to wash paint away?

EvenInDeathIStillServe:
I would be inclined to agree if the coloring was not obnoxious as hell.

I used to prank people on the Fallout server by spamming them with spray tan until they became yellow. Sure, it was funny, but it ruins your character's appearance.

I know you're just joking, man, but I had to say that. I even wanted to make it so that SHOWERS got rid of coloration, but went against it since it would be too easy.

ACCount12:
DNM as requested by author.

PR #3248, opened 2019-04-02
[QoL] Remove contraband hacking on a platitude of vendors

driecg36:
It's silly that the guild needs to hack their own vendors every round in order to sell their products. Even more so when the "contraband" products are usually the best sellers. They always end up getting hacked, so it's nothing more than an inconvenience at best and a significant frustration at worst. 

Also feels just like a hold-over from vanilla ss13, as it makes about 0 sense in our universe.

### DIFF
--- a/code/game/machinery/vending.dm
+++ b/code/game/machinery/vending.dm
@@ -1009,15 +1009,16 @@
 	/obj/item/weapon/storage/box/shotgunammo/flashshells = 10,
 	/obj/item/weapon/storage/box/shotgunammo/blanks = 10,
 	/obj/item/clothing/accessory/holster = 5,
-	/obj/item/weapon/storage/pouch/pistol_holster = 5)
-
-	contraband = list(/obj/item/ammo_magazine/sl38 = 20,
+	/obj/item/weapon/storage/pouch/pistol_holster = 5,
+	/obj/item/ammo_magazine/sl38 = 20,
 	/obj/item/ammo_magazine/cl32 = 20,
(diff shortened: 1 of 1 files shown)


RandomColonist:
Vendors are not for guildsmen in general, they are for everyone else. Guild can order/print those, if they need too, but it's much easier to take some dosh from cap and hack-buy it then order it/print it. Problem is not the contraband restriction, actually. They will still buy things from their own vendors. Instead i purpose a few solutions that will fit better:
- Add an internal switch to vendors which enables/disables contraband (which can be toggled only if you have guild/merchant access) with still a way to hack it. Will ease pain in the ass for guildsmen.
- Rename it from contraband to restricted access items or something and lock it behind code blue/red.

driecg36:
I'm gonna put what I said into discord here as well for the purpose of record keeping.

1. It is a QoL change, as the vendors invariably get hacked every round. This will thus have little effect on balance if any.
2. It makes 0 sense in the lore that asters and Frozen star would lock away their most profitable products behind a contraband wire. This is a "sweep under the rug" mechanic that players try to ignore best they can, as there is no way to justify this from the setting.

However, I am willing to add other contraband items to these vendors to make up for the lack of contraband in these machines (perhaps the disk cloner for the print-o-mat, and maybe some more exotic types of ammo like 14.5mm and 10mm under the FS vendor). I have no idea what I would put for the powermat, however.

ACCount12:
Unnecessary, and not even a QoL change on the top of that.

PR #3257, opened 2019-04-03
Cargo Crate Revamp/Rebalance

driecg36:
Most of our cargo crates were holdovers from oldbay that didn't really make logical sense in the setting, as well as having wildly varying prices for seemingly no reason. So, I did the following -

1. Removed access requirements from the majority of crates - Asters wouldn't lock themselves out of their own property. Gameplay wise, it just adds the effort of asking someone to come open it, or more often, just shooting it open yourself. I did leave it for "high risk" crates like the supermatter and a few different explosives, which asters probably wouldn't want opened by themselves or on accident.

2. Reorganized the security crates, which were just a mess. The old "weapons crate" was literally just a code holdover from bay that shipped a random assortment of guns. I replaced this with a few gun "packs" from FS and NT, the two main gun suppliers in our settings. I adjusted the prices to what I thought they would reasonably be. Also added a surplus ammo crate. Also created dedicated ballistic and energy armor crates as well as a surplus IH crates, adding helmets to all of them.

3. Adjusted a lot of prices, especially in engineering. Quadrupled the cost of a new supermatter from 5k to 20k (though lore wise it should probably be higher). I also added more of a few items in crates I thought were too expensive for the price (1500 for three boxes of slugs you can buy for 300 a pop at the vendor? Really?)

4. Separated the surplus RIG and surplus space suit into two different categories and added a rig module crate. Also added an expensive plasteel crate to engineering.

Open to any suggestions for new crates or balanced prices and quantities.


Voldirs:
"IH Energy Armor"
This doesn't sound good, it sounds like energized armor, not energy-resistant armor

Maybe add some pouch crates?

Boredone:
Voidsuits tend to be notably more common then Hardsuits, so I'd recommend atleast pumping up the price for the hardsuit crate.

Also while you are dealing with the contents anyway, glance suggests the new IH guns didn't get added to the Surplus Weapons crate, so could fix that as well while you are messing with the crates.

RandomColonist:
Adding more things for any price to guild = straight buff because Captain funds them for 100000~ each time. Also plasteel should cost higher, and FS assault pack should contain new SMGs and BR,  not PDWs and should cost higher.

driecg36:
Implemented all these comments, more or less.

code/modules/cargo/packs.dm
```diff
@@ -128,20 +128,48 @@ var/list/all_supply_groups = list("Operations","Security","Hospitality","Enginee
...
-	crate_name = "Weapons crate"
-	access = access_brig
+	crate_name = "FS Handgun Pack"
+	group = "Security"
+	
+/datum/supply_pack/fsassault
+	name = "FS Assault Pack"
+	contains = list(/obj/item/weapon/gun/projectile/automatic/ak47/fs,
+			/obj/item/weapon/gun/projectile/automatic/wt550,
+			/obj/item/weapon/gun/projectile/automatic/wt550)
```
Voldirs: I'm not sure if these smgs are using proper updated sprite for gun and ammo, maybe it's better to use some other option here

driecg36: My bad, I confused the Idaho (the top mounted 9mm one, I believe) and the Wt550.

PR #3264, opened 2019-04-04
[BALANCE] Excelsior antag won't be given to Heads or Ironhammer.

TorinoFermic:
## General
Changed so the antag Excelsior won't apply to heads of staff by the storyteller.
Excelsiors can forcefully implant heads or IH just fine.

## Changelog
🆑 TorinoFermic
tweak: Excelsior will not apply to head of staff or Ironhammer for the storyteller picking.
/🆑

### DIFF
--- a/code/game/gamemodes/roleset/faction/excelsior.dm
+++ b/code/game/gamemodes/roleset/faction/excelsior.dm
@@ -12,11 +12,3 @@
 
 	req_crew = 6
 	leaders = -1 //Every excelsior spawned directly is a leader. Non leaders are those recruited during gameplay
-
-	story_ineligible = list(JOBS_SECURITY, JOBS_COMMAND)
-
-// Code to prevent a role from being picked by the storyteller.
-/datum/storyevent/roleset/faction/excelsior/antagonist_suitable(var/datum/mind/player, var/datum/antagonist/antag)
-	if(player.assigned_role in story_ineligible)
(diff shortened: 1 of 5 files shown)


driecg36:
+1 - it's not fun when heads become excelsior and ruins the spirit of the mode IMO (revolution of working class against heads). Seems like excelsior was always meant to be this way.

TorinoFermic:
I have properly changed the code to make use the underutilized function for it.

code/game/gamemodes/roleset/faction/excelsior.dm
```diff
@@ -13,4 +13,10 @@
 	req_crew = 6
 	leaders = -1 //Every excelsior spawned directly is a leader. Non leaders are those recruited during gameplay
 
+	story_ineligible = list(JOBS_SECURITY, JOBS_COMMAND)
 
+// Code to prevent a role from being picked by the storyteller.
+/datum/storyevent/roleset/faction/excelsior/antagonist_suitable(var/datum/mind/player, var/datum/antagonist/antag)
```
ACCount12: Move this code up to `/datum/storyevent/roleset/faction`, so `story_ineligible` actually works for other antags.

TorinoFermic: Done, not sure about the story_ineligible being unique for each factions.

PR #3268, opened 2019-04-05
Recoil and aim increased when dual wielding

TorinoFermic:
## General
To stop the abuse of full auto smgs

## Changelog
🆑 TorinoFermic
tweak: Recoil and aim are doubled when dual wielding on 2 guns.
/🆑

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -76,6 +76,7 @@
 	var/muzzle_flash = 3
 	var/list/dispersion = list(0)
 	var/requires_two_hands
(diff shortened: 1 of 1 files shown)


Boredone:
Not 100% sure, but glancing at it kinda suggests that either there will be no recoil penalty if your gun is silenced, or no advantage to dual wielding with a silenced one recoil wise.
Which feels a bit iffy since silencers apparently help recoil wise currently.

TorinoFermic:
The code while it runs should trigger each gun and their attributes. I'll fix for the combo dual wield + silencers

PR #3277, opened 2019-04-05
Enables Cargo Exports (For real this time)

EvenInDeathIStillServe:
Lots of changed files: mostly just adding a price_tag to stuff, and adding matter defines to guns that don't have it. Will need more PRs to add price tags.

Enables cargo exports.
Removes pre-defined export datums: datums get generated when you try to sell an object for the first time and get deleted when the selling process ends.
Surplus crate cost bumped up to 3000.
Plasteel crate is now 50->60 sheets for 2500->3000.
Most supply crates had their costs lowered.

Misc things: Space cleaners removes color modifier from stuff.

Fixed some bugs: R&D console gets hacked when emagged, not EMPed. Closets and crates no longer spawn with invisible covers. Changes some instances where the supply console refered to Eris as a station.


code/modules/cargo/exports.dm
```diff
@@ -48,7 +50,21 @@ Credit dupes that require a lot of manual work shouldn't be removed, unless they
...
+				cost += item_value
+				sold_str += " [thing.name]"
+				if(!dry_run)
+					var/datum/export/new_export = new()
+					new_export.unit_name = thing.name
+					new_export.cost = item_value
+					new_export.include_subtypes = FALSE
+					new_export.export_types = list(thing.type)
+					new_export.sell_object(thing, contraband, hacked)
+					SSsupply.exports.Add(new_export)
```
ACCount12: Why even bother generating export datums for that? Move "sell_object" elsewhere and use it without invoking export datums.

EvenInDeathIStillServe: It's to keep track of the number of exported items of the same type when making the export log in the supply console. Is there a better way to do this?

PR #3292, opened 2019-04-08
You can no longer have 1 word names

EvenInDeathIStillServe:
Yeah, because if your Exultant goes only by the name "Buttercup", there's something terribly wrong.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station.dm
@@ -18,3 +18,12 @@
 
 /datum/species/human/get_bodytype()
 	return "Human"
+
+/datum/species/human/sanitize_name(var/name)
+	var/sanitized_name = sanitizeName(name)
+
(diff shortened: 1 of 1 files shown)


ACCount12:
Does this still allow actors to have silly names?

EvenInDeathIStillServe:
They will need to have a 2 word name. Sure, no Periwinkle, but they can do Mr. Periwinkle.  I don't think snowflake code is necessary for one job?

GrayRachnid:
Make it so you get a prompt to name yourself when you spawn in as an actor similar to how Cyborgs and AI get it.

driecg36:
I would also like an exemption for assistants. I think one-word names are appropriate for them as well, considering they can be stowaways or tunnel rats.

drexample:
Is this PR getting worked on?
If not, it will be closed in a week's time.

drexample:
Well. The week has passed. Closing.

PR #3293, opened 2019-04-08
Players' high-priority role shows up in lobby tab

EvenInDeathIStillServe:
image


EvenInDeathIStillServe:
image

After discussion, we figured it would be better if it showed character names instead of Ckeys.

mattroks101:
Alright I guess that's fair.

EvenInDeathIStillServe:
image
This is for admins using the Who verb

EvenInDeathIStillServe:
I removed the Lobby tab altogether. The Status tab now takes care of the duty of the Lobby tab when you're in the lobby.
I made it so that you can see the storyteller in the status tab when you're ingame.

PR #3302, opened 2019-04-10
[TWEAK] Remains of stash will be hidden under the floor

TorinoFermic:
## General
Remains are hidden under the current floor. Not anymore of self-defeating stashes.

## Changelog
🆑 TorinoFermic
tweak: Remains of stash are hidden under the floor.
/🆑

### DIFF
--- a/code/modules/stashes/stash_datum.dm
+++ b/code/modules/stashes/stash_datum.dm
@@ -280,6 +280,12 @@ This file contains the underlying code for stash datums
 			//Spawn it in the thing
 			results += new a(T)
 
+	// Better hide remains under the tile.
+	if (istype(T, /obj/item/remains/human))
+		var/obj/item/remains/human/R = T
+		R.level = 1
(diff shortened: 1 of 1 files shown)


code/modules/stashes/stash_datum.dm
```diff
@@ -280,6 +280,16 @@ This file contains the underlying code for stash datums
 			//Spawn it in the thing
 			results += new a(T)
 
+	// Going thru the list to hide under the floor
+	for(var/h in results)
+		// Better hide remains under the tile.
+		if (istype(h, /obj/item/remains))
+			var/obj/item/remains/R = h
```
ACCount12: In this case, `for(var/obj/item/remains/R in results)` would be a better option. It basically folds istype into the loop.

TorinoFermic: Ah, it would clear up how the dm code works.

PR #3305, opened 2019-04-10
Small rig additions

RandomColonist:
- Wizard and old Nuke Operatives voidsuits are no longer normal size, but bulky in line with any other voidsuit. You can find them in maint and they both have pretty good armor, how's they more compact? It was made by bay for nuke ops, but it's not used as antag-only items anymore, so we might as well put it in line with other voidsuits.
- IH Combat RIG now have 1.35 slowdown (same as serb voidsuit). IH voidsuit now have standart 1 slowdown. You will have to choose now between more protection or more speed.
- IH Combat RIG lost their thermals and combat chem injector but gained storage.
- Eased slowdown on Technomancer Exultant Advanced Voidsuit Magboots because he was still slow with 2, made it 1. TE deserves some love.


code/modules/clothing/spacesuits/rig/suits/combat.dm
```diff
@@ -42,7 +42,7 @@
 	helm_type = /obj/item/clothing/head/helmet/space/rig/ihs_combat
 	suit_type = "ironhammer hardsuit"
 	armor = list(melee = 65, bullet = 60, laser = 60, energy = 55, bomb = 45, bio = 100, rad = 70)
-	slowdown = 1.2
+	slowdown = 1 //was 1.2 but why the fuck use magical numbers?
```
seekin4u: зочем такие же магические комменты оставлять?

RandomColonist: самое то

ACCount12:
Seems like a lot of balance tweaks, and the only one I agree with is slowdown reduction on Exultant magboots.

RandomColonist:
> Seems like a lot of balance tweaks, and the only one I agree with is slowdown reduction on Exultant magboots

So you're okay with IH and captain grabbing rig for thermals to valid-hunt? I can remove 1.2 change, but thermals should go. They already have chance to spawn random rigsuit modules near them, including said thrrmals, but it's not guaranteed.

Regarding SCAFs - 2 armor addition/removal is unneccesary, i can remove those if you really dislike it.

Regarding voidsuits - merc rig spawns in maints frequently, and i can't see why it should have normal size AND considerably better armor. But can remove those changes as well upon request.

AMM101902:
If you are going to reduce the IH rig slowdown, you might as well reduce their voidsuit's slowdown as well, probably just to 1. It is currently at 1.5 slowdown, and with the IH rig having better protection and armor there is little reason to use it.

RandomColonist:
> If you are going to reduce the IH rig slowdown, you might as well reduce their voidsuit's slowdown as well, probably just to 1. It is currently at 1.5 slowdown, and with the IH rig having better protection and being faster there is little reason to use it.

Thanks, great idea, i will raise slowdown of RIG to 1.5 and lower void slowdown to 1 to at least try to stop valid-hunters who grab rig and big gun from roundstart to ruin fun for antags.

ACCount12:
I'm okay with thermals being removed too. IMO, it's a shitty mechanic for crew to have. Other codebases restrict them to antags for this exact reason.

code/modules/clothing/spacesuits/rig/suits/combat.dm
```diff
@@ -52,8 +52,9 @@
 
 
 	initial_modules = list(
-		/obj/item/rig_module/vision/thermal,
+		//obj/item/rig_module/vision/thermal, //no
```
TorinoFermic: It would best to not comment out lines. They are instead removed out.

ACCount12: Agreed. No reason to keep in dead lines like that.

TorinoFermic:
I think internal storage module could at least be added for IH RIG.

RandomColonist:
> I think internal storage module could at least be added for IH RIG.

Already added.

Steelpoint:
The big issue I find with the IH Void/Hardsuit debate is that I feel the IH Voidsuits are utterly inferior to the Hardsuit entirely outside of aesthetics.

They both have (seemingly) the same pathetic movement speed, but the Hardsuit has access to a range of useful modules, and has superior armour (barely). Whereas the Voidsuit offers nothing, it barely has better armour than your standard armour vest but it confers a massive slowdown and nothing else.

At least the result of this PR will mean both Hard and Voidsuits are so close to similarly uselessness that you can justify picking one or the other.

----

Otherwise, I do agree with removing the thermals. I would suggest adding in a night vision module to compensate and to give the player some reason to take a Hardsuit

RandomColonist:
Fixed, implemented suggested changes, alrered description, ready to merge.

PR #3315, opened 2019-04-11
Some augment typo fixes, description additions and a new multitool

greggbot:
So I started this off as a simple typo fix since "muscules" and "surgial" has been annoying me for ages.
Then it turned into this.
So I've fixed the typos, added descriptions to the modules, and then added a new multitool designed for miners, because at that point I just figured "fuck it, why not?"


code/modules/research/mechfab_designs.dm
```diff
@@ -111,57 +111,63 @@
...
+
 //Prosthesis ====================================
 
 /datum/design/item/mechfab/prosthesis
 	category = "Prosthesis"
 	materials = list(MATERIAL_STEEL = 18)
 
 /datum/design/item/mechfab/prosthesis/r_arm
-	name = "right arm"
+	name = "Right arm"
```
mattroks101: Aren't these improper nouns? How does this show up in game?

ACCount12: Those are design datums, not items. They only show in HTML/Nano UIs.

mattroks101: Oh I see.

greggbot: Yeah, most of the datum names are capitalized so I went ahead and made it consistent for prosthetics/augments.

PR #3317, opened 2019-04-11
Tweak firemodes of AK-47

BASIC-BIT:
- Remove 3 round burst from AK-47
- Add Full Auto (400 RPM) to AK-47

I feel as these are the more "iconic" AK-47 mechanics - Safe, Semi-Auto, and Full Auto.  Obviously this isn't important but why not try ¯\\\_(ツ)\_/¯

### DIFF
--- a/code/modules/projectiles/guns/projectile/automatic/ak47.dm
+++ b/code/modules/projectiles/guns/projectile/automatic/ak47.dm
@@ -22,7 +22,7 @@
 
 	firemodes = list(
 		SEMI_AUTO_NODELAY,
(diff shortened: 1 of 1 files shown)


Keekenox:
Fuck yeah.

RandomColonist:
AK-47 supposed to be in line with Sol being carbine rather than AR. That was discussed with project leader and can't be merged without his allowance. My personal suggestion: either just rise up price of AK or add auto-fire to Excelsior AK only without removal of three - round burst, rising price of only Exc AK.

BASIC-BIT:
Yeah, I really should have labelled this PR as a suggestion and not a real "tweak" - It would also maybe be interesting to use "short burst" rather than "3 round burst" - "3 round burst" implies a gun mode, while "short/long burst" imply manual trigger control.

SpadesSlick:
Outdated due #3453.

PR #3320, opened 2019-04-12
Small fix around runtime in rig_attackby

soslow4u:


code/modules/clothing/spacesuits/rig/rig_attackby.dm
```diff
@@ -58,10 +58,8 @@
 								user << "You detatch \the [cell] from \the [src]'s battery mount."
 								for(var/obj/item/rig_module/module in installed_modules)
 									module.deactivate()
-								if(user.r_hand && user.l_hand)
+								if(!user.put_in_hands(cell))
 									cell.forceMove(get_turf(user))
```
LethalGhost: `put_in_hands(..)` already call that if both hands are not empty.

soslow4u: Done in next commit

PR #3441, opened 2019-04-15
Bank account numbers are no longer public

EvenInDeathIStillServe:
I had some fun yesterday teaching the crew about how secure their bank accounts really are.

Crew records no longer show bank account numbers unless you have FO access.
Security records found in security filing cabinets (there's one in the FO's office) now contain crew account numbers.

Want to be immune to bank account theft? Raise your account security level to at least level 1 so you need the PIN to access it.

### DIFF
--- a/code/modules/modular_computers/file_system/reports/crew_record.dm
+++ b/code/modules/modular_computers/file_system/reports/crew_record.dm
@@ -192,7 +192,7 @@ FIELD_LIST_EDIT("Status", status, GLOB.physical_statuses, null, access_moebius)
 
 FIELD_SHORT("Species",species, null, access_change_ids)
 FIELD_SHORT("Email",email, null, access_change_ids)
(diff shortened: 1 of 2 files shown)


ACCount12:
Doesn't fix the issue really. The issue is that level 0 is supposed to be your ID + PIN or your account number + PIN. For some baycode reason, it doesn't work that way.

EvenInDeathIStillServe:
In my opinion, this "issue" is beneficial when it comes to gameplay. It opens up new possibilities for illegal activities, and teaches people to actually take proper security measures.

ITCups:
crew records are not available to everyone.
so it considered secure info.
Instead of this you should open account numbers only to guild members.
Currently security records only available to IH and medical records to moebus, make the same with acc numbers

ITCups:
And you should be able to access account only using number.
It should be okay to access account using ID, but in order to access without it you have to know number and pin

PR #3449, opened 2019-04-16
Grinder radial

Kmc2000:
https://media.discordapp.net/attachments/265411250341543936/567668365943111690/unknown.png?width=1091&height=444


ACCount12:
Do we seriously need everything to have a radial menu slapped onto it?

For some things, it's a good fit. For some, as the gun example has shown, it's terrible. This? Closer to guns I think.

ITCups:
I think yes, having browser window as interface is horrible. We should aim for better interactions between world and player.
It not nessesary should be radial menu, but its a right direction.

Kmc2000:
@ACCount12
I got asked to do it, take if you want.
In principle, when you have 2-3 actions as most on an item, a radial is perfect.

PR #3452, opened 2019-04-17
Targeting using numpad

ITCups:


code/_onclick/hud/screen_objects/base_screen_objects.dm
```diff
@@ -181,54 +181,56 @@
 	name = "damage zone"
 	icon_state = "zone_sel"
 	screen_loc = ui_zonesel
+	var/selecting = BP_CHEST
```
LethalGhost: That's outdated var.
Currently targeted organ is hold in mob var not in screen object one.

PR #3456, opened 2019-04-17
(Ready for review)Recoil rework pt 1 - Crosshair

Kmc2000:
https://cdn.discordapp.com/attachments/265411250341543936/568135305735962624/recoilcursor.mp4

Adds a crosshair that changes with your gun's recoil
Recoil now affects accuracy

Security, captain and barman have a skill which reduces recoil.


code/modules/projectiles/dynamic_cursor.dm
```diff
@@ -0,0 +1,29 @@
...
+			H.client.mouse_pointer_icon = scaled
+
+/obj/item/weapon/gun/dropped(var/mob/living/user)
+	if(user.client)
+		user.client.mouse_pointer_icon = initial(user.client.mouse_pointer_icon)
+	. = ..()
+
+/obj/item/weapon/gun/afterattack(atom/A, mob/living/user, adjacent, params)
+	update_cursor(user)
+	. = ..()
```
ACCount12: Just `equipped` and `dropped` are not enough. What if hands are switched? What if the object is qdeleted while in hands?

PR #3457, opened 2019-04-18
Minor fix for tac reload mesasge on empty gun

TorinoFermic:


code/modules/projectiles/guns/projectile.dm
```diff
@@ -148,10 +148,10 @@
 				if(AM.ammo_mag != ammo_mag && ammo_mag != "default")
 					to_chat(user, SPAN_WARNING("[src] requires another magazine.")) //wrong magazine
 					return
-				if(tac_reloads)
+				if(tac_reloads && !ammo_magazine)
 					unload_ammo(user)	// ejects the magazine before inserting the new one.
 					to_chat(user, SPAN_NOTICE("You tactically reload your [src] with [AM]!"))
```
ACCount12: Does that actually work? This code would only trigger when there is no mag inside the gun. So, tac reloads would be broken.

TorinoFermic: I am wondering why I did add "!" to this ammo_magazine. I think the DM code is messing with my mind. Fixed.

PR #3463, opened 2019-04-19
Making changelings playable again

DerFlammenwerfer:
Somebody already beat me to the LMG fix, so I made this instead.


SpadesSlick:
Making a button to literally win any fight cannot make antag more "playable" by any means.

PR #3464, opened 2019-04-19
Adds roach taming

Kmc2000:
Walk up to a roach with ambrosia vulgaris in your hand for a chance to make it friendly and rideable!
Due to bay shitcode, I was unable to make them controlled by riding....
Seriously. Fuck these movement handler things.


soslow4u:
What about wheelchair code?
https://github.com/discordia-space/CEV-Eris/blob/33b9b69753ca13acbe40c36b5e5000aaf781eb88/code/datums/movement/mob.dm#L130

Kmc2000:
Yeah I tried a bunch of methods, including that.
The only way it would work with is if I directly controlled the roach, which defeats the purpose.

PR #3466, opened 2019-04-19
A less bullshit rework of Death Sting

DerFlammenwerfer:
Removes the stun entirely from the death sting and makes it inject frost oil. Kills you by freezing you to death, which cools you down fairly slowly but ultimately is just as debilitating as the cold makes it difficult to move.

Also increased the chemical cost of the Paralyse sting from 30 to 40, duration is unchanged. I'll rework it into a brain-damaging or unstable mutagen sting when grabcuff is nerfed but for now the instant stun sting has to remain for the sake of balance.


SpadesSlick:
It's Travis failed.

DerFlammenwerfer:
Probably because I wrote this after a recent merge

PR #3472, opened 2019-04-20
Miner QoL Changes

granodd:
Makes a miner's life a little easier.

**tweak:** stacking machine now defaults refined material stack size to 120
**added:** option of stacking to 120 sheets in addition to the already existing 1, 5, 10, 20, 50
**added:** miner lockers now come with one (1) pair of mesons and an accompanying small cell
**tweak:** replaced the jackhammer with an actual mining drill so that the fuel tanks inside the locker actually have use
**removed:** second fuel tank attachment and random tool attachment spawn from miner locker
**tweak:** replaced mining lantern with heavy duty flashlight


SpadesSlick:
I think you better remove tool upgrades at all, they are not suppose to be here.

Also, do we have a new sprite for heavy flashlight? Or it's an ugly and old one?

maps/CEVEris/_CEV_Eris.dmm
```diff
@@ -8142,7 +8142,7 @@
 "daD" = (/obj/structure/shuttle_part/mining{icon_state = "1,4"},/turf/space,/area/shuttle/mining/station)
 "daE" = (/turf/simulated/shuttle/wall/mining{icon_state = "2,4"},/area/shuttle/mining/station)
 "daF" = (/obj/structure/cable/yellow{d1 = 4; d2 = 8; icon_state = "4-8"},/obj/structure/cable/yellow{d1 = 1; d2 = 4; icon_state = "1-4"},/turf/simulated/shuttle/floor/mining{icon_state = "4,4"},/area/shuttle/mining/station)
-"daG" = (/obj/machinery/power/terminal{icon_state = "term"; dir = 1},/obj/structure/cable/yellow{d2 = 8; icon_state = "0-8"},/obj/structure/ore_box,/turf/simulated/floor,/area/shuttle/mining/station)
+"daG" = (/obj/machinery/power/terminal{icon_state = "term"; dir = 1},/obj/structure/cable/yellow{d2 = 8; icon_state = "0-8"},/obj/structure/reagent_dispensers/fueltank,/turf/simulated/floor,/area/shuttle/mining/station)
```
ACCount12: The map is on lockdown right now, with no map changes being accepted. Revert this.

Otherwise, I'm ready to accept this PR.

code/modules/mining/machine_stacking.dm
```diff
@@ -68,7 +68,7 @@
 	var/input_dir = null
 	var/output_dir = null
 	var/list/stack_storage
-	var/stack_amt = 50; // Amount to stack before releassing
+	var/stack_amt = 120; // Amount to stack before releassing
```
ACCount12: No real need for that `;`. I get that it was there before, but it's still useless. Remove it.

granodd: I'm too used to looking at C# code that I probably didn't even notice it. I'll get rid of that.

granodd:
Undid the fuel tank on the Hulk and removed the motor and guard attachments from the mining locker while keeping the fuel tank attachment. This gives the miners some incentive to seek upgrades either through scavenging or through bartering.

ACCount12:
There is still some map change hanging in the diff. Did you put whatever was there back?

granodd:
I removed the fuel tank and made sure that the lines of code concerning that tile were the same as they were before it was added. I'll be sure to keep map edits in their own commits in the future should I ever do them to make reverting them easier.

ACCount12:
изображение
no you didn't

granodd:
I'll see if I can revert the commits then and manually re-add the changes to the locker then.

TorinoFermic:
You can do "git reset --soft HEAD~4" then unstage these map changes, create new commit with your mesasge then git push --force

granodd:
I can't get tortoisegit to properly cooperate with me on reverting these map changes, so to save myself time and effort, I'm going to close this current PR/Branch and redo it without any of the map edits.

PR #3480, opened 2019-04-20
Patch to make stealthy stings stealthy, rework of death sting into cr…

DerFlammenwerfer:
…yo sting

- Removes global 'you feel a tiny prick' warning from stings. Custom warnings per sting unchanged.

~ Death Sting is now a Cryo Sting, injecting frost oil instead of Lexorin. Unlike lexorin, freezing to death can actually kill you within two minutes. Clever chemists can take advantage of the low body temperature.

~Increases the chemical cost of para-sting, duration unchanged.

### DIFF
--- a/code/game/gamemodes/changeling/changeling_powers.dm
+++ b/code/game/gamemodes/changeling/changeling_powers.dm
@@ -702,7 +702,7 @@ var/list/datum/dna/hivemind_bank = list()
 //	spawn(10)	src.verbs += verb_path
 
 	src << SPAN_NOTICE("We stealthily sting [T].")
-	T << SPAN_WARNING("You feel a tiny prick.")
+//	T << SPAN_WARNING("You feel a tiny prick.")
 	log_attack("[src] use [src.client.CH.handler_name] on [T]")
 	if(!T.mind || !T.mind.changeling)
 		return T //T will be affected by the sting
@@ -784,7 +784,7 @@ var/list/datum/dna/hivemind_bank = list()
 
 /mob/proc/changeling_prepare_paralysis_sting()
 	set category = "Changeling"
-	set name = "Paralysis Sting (30)"
+	set name = "Paralysis Sting (40)"
 	set desc="Sting target"
 
 	check_CH("Paralysis Sting",/datum/click_handler/changeling/changeling_paralysis_sting)
@@ -792,7 +792,7 @@ var/list/datum/dna/hivemind_bank = list()
 
 
 /mob/proc/changeling_paralysis_sting(atom/A)
-	var/mob/living/carbon/T = changeling_sting(30,A)
+	var/mob/living/carbon/T = changeling_sting(40,A)
 	if(!T)	return 0
 	T << SPAN_DANGER("Your muscles begin to painfully tighten.")
 	T.Weaken(20)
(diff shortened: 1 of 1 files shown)


DerFlammenwerfer:
There we go, Travis passed this time.

code/game/gamemodes/changeling/changeling_powers.dm
```diff
@@ -916,20 +916,18 @@ var/list/datum/dna/hivemind_bank = list()
...
 /mob/proc/changeling_DEATHsting(atom/A)
-	var/mob/living/carbon/T = changeling_sting(40,A)
+	var/mob/living/carbon/T = changeling_sting(50,A)
 	if(!T)	return 0
-	T << SPAN_DANGER("You feel a small prick and your chest becomes tight.")
-	T.silent = 10
-	T.Paralyse(10)
-	T.make_jittery(1000)
-	if(T.reagents)	T.reagents.add_reagent("lexorin", 40)
+	T << SPAN_DANGER("You feel a small prick and your blood begins to freeze in your veins.")
```
mattroks101: These should be `to_chat()`

code/game/gamemodes/changeling/changeling_powers.dm
```diff
@@ -702,7 +702,7 @@ var/list/datum/dna/hivemind_bank = list()
 //	spawn(10)	src.verbs += verb_path
 
 	src << SPAN_NOTICE("We stealthily sting [T].")
-	T << SPAN_WARNING("You feel a tiny prick.")
+//	T << SPAN_WARNING("You feel a tiny prick.")
```
seekin4u: Please remove such things, if you don't need chat output than just remove it

Stealthkibbler:
Could you make the ranged sting more obvious, the ling is literally shooting the stinger out of its mouth according to the flavour text yet it turns your stings into ranged and still silent.

code/game/gamemodes/changeling/changeling_powers.dm
DerFlammenwerfer: > These should be `to_chat()`

The reason it uses SPAN_DANGER is to make it a warning text that only appears on the target's client, complete with italic font and red coloring.

DerFlammenwerfer:
> Could you make the ranged sting more obvious, the ling is literally shooting the stinger out of its mouth according to the flavour text yet it turns your stings into ranged and still silent.

All the stings that have warnings will still deliver those warnings to the target, ranged or no. Additionally, wearing any armor/suit thick enough block a fired syringe (such as from a syringe revolver) also appears to block the ranged sting, and tells the user that they cannot find a path to sting the target. Riot armor, for example, as well as most voidsuits seem to do this.

code/game/gamemodes/changeling/changeling_powers.dm
DerFlammenwerfer: I'll keep that in mind in the future.

code/game/gamemodes/changeling/changeling_powers.dm
TorinoFermic: The reason why we changing from << to to_chat is to avoid hardcoded lines like this shit. Basically T << SPAN_DANGER( is same as to_chat( T , SPAN_DANGER(.....))

granodd:
Frost oil doesn't actually do any damage. It just makes you cold. The way you worded "clever chemists can take advantage of low body temperature" assumes that the changeling is a chemist, and even then, forcibly getting chems into someones body as a chemist without a syringe gun isn't stealthy or very easy.

Make the cryosting inject water as well so that it freezes inside the person. At least I'm fairly certain that's what other servers have their cryosting do.

drexample:
This just nerfs death sting into oblivion, replacing it with a useless one, and, while changeling stings are stealthy, meaning of the stealth is that surrounding people don't see that. 
No matter what shape or coating of your stinger is, if you pump someone full of chems, they are going to notice that. Not to mention that this PR is more than 3 months old.
Please, either rework the PR to provide improvement to changeling rounds experience, or it will be eventually closed.

drexample:
Seven days have passed, no replies, PR closed.

PR #3483, opened 2019-04-21
Cyberpunk intensifies

soslow4u:
hmmm5

Ruh-oh, you would need only ID to break your local holoplant for hot holodancer


ACCount12:
I know it's not your mistake, but can you fix this `emaged` thing? `emagged` is the proper word.

PR #3486, opened 2019-04-22
[Ready for Review] Adds Player Notes

GeneriedJenelle:
Ports from baystation, will update as this goes along.

### DIFF
--- a/code/modules/admin/admin.dm
+++ b/code/modules/admin/admin.dm
@@ -135,7 +135,7 @@ ADMIN_VERB_ADD(/datum/admins/proc/show_player_panel, null, TRUE)
 		<A href='?_src_=holder;warn=[M.ckey]'>Warn</A> |
 		<A href='?src=\ref[src];newban=\ref[M]'>Ban</A> |
 		<A href='?src=\ref[src];jobban2=\ref[M]'>Jobban</A> |
-		<A href='?src=\ref[src];notes=show;mob=\ref[M]'>Notes</A> "
+		<A href='?src=\ref[src];notes=show;mob=\ref[M]'>Notes</A>
 
 	"}
 
--- a/code/modules/admin/admin.dm
+++ b/code/modules/admin/admin.dm
@@ -29,31 +29,6 @@ proc/admin_notice(var/message, var/rights)
 /datum/admins/SDQL_update(var/const/var_name, var/new_value)
 	return 0
 
-/datum/admins/Topic(href, href_list)
-	..()
-	if (usr.client != src.owner)
-		world << "\blue [usr.key] has attempted to override the admin panel!"
-		log_admin("[key_name(usr)] tried to use the admin panel without authorization.")
-		return
-
-	//Player Notes
-	if(href_list["notes"])
-		var/ckey = href_list["ckey"]
-		if(!ckey)
-			var/mob/M = locate(href_list["mob"])
-			if(ismob(M))
-				ckey = M.ckey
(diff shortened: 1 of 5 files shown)


code/modules/admin/player_notes.dm
```diff
@@ -0,0 +1,83 @@
+//This stuff was originally intended to be integrated into the ban-system I was working on
+//but it's safe to say that'll never be finished. So I've merged it into the current player panel.
+//enjoy				~Carn
+
+#define NOTESFILE "data/player_notes.sav"	//where the player notes are saved
+
+obj/admins/proc/notes_show(var/ckey)
```
TorinoFermic: obj is missing the first / at line 7.

code/modules/admin/player_notes.dm
```diff
@@ -0,0 +1,83 @@
...
+	usr << browse("<head><title>Player Notes</title></head><body>[notes_gethtml(ckey)]</body>","window=player_notes;size=700x400")
+
+
+obj/admins/proc/notes_gethtml(var/ckey)
```
TorinoFermic: Also this too.

code/modules/admin/player_notes.dm
```diff
@@ -0,0 +1,83 @@
+//This stuff was originally intended to be integrated into the ban-system I was working on
+//but it's safe to say that'll never be finished. So I've merged it into the current player panel.
+//enjoy				~Carn
+
+#define NOTESFILE "data/player_notes.sav"	//where the player notes are saved
+
+obj/admins/proc/notes_show(var/ckey)
+	usr << browse("<head><title>Player Notes</title></head><body>[notes_gethtml(ckey)]</body>","window=player_notes;size=700x400")
```
TorinoFermic: We are using to_chat(), they have to be replaced.

GeneriedJenelle:
https://github.com/Baystation12/Baystation12/commit/c94b35dc00a0cb5d9e51ffc9378c6c79f6d753c1#diff-43cbbc3a371c8391b464134749f84924

Porting from here, I'll do it in a moment.

GeneriedJenelle:
image

**edit:** So I've made the objs into datums to modernize the code a bit, and i've added a notes button to your PP menu. Only issue is the player panel for eris is custom and I didn't want to mess with it since you had custom code. If you guys get the modern version of the player panel you should be able to get the modern notes code,

That said, this version should be bug free and works similarly to modern code anyhow, just different layout. Serves it's function. Let me know if it's okay.

code/modules/admin/admin.dm
```diff
@@ -29,12 +29,37 @@ proc/admin_notice(var/message, var/rights)
 /datum/admins/SDQL_update(var/const/var_name, var/new_value)
 	return 0
 
+/datum/admins/Topic(href, href_list)
+	..()
```
ACCount12: Move this to `topic.dm` so it's easy to clean it up later.

GeneriedJenelle: Now done!

ACCount12:
Would have to be ported to datums later, but it's good enough for now.

code/modules/admin/topic.dm
```diff
@@ -22,6 +22,26 @@
 		message_admins("[usr.key] has attempted to override the admin panel!")
 		return
 
+	//Player Notes
+	if(href_list["notes"])
```
nichlas0010: m a n e

PR #3487, opened 2019-04-22
Excelsior implants disguise

comma:
Excelsior implant now shows up as either Death Alarm or Chem Implant on Adv Scanner
Gotta pry it out to be sure it's commiestuff.

dreamseeker_2019-04-22_21-03-11


Pariah919:
Neither of these are really used (Death alarms are used by literally nobody but IH.) and to anyone with half a brain will just make people think its a commie. Esp for the chemical implants.

comma:
Sure, but now it requires half of brain, 'unknown body' is a no-brainer. See, progress.

PR #3491, opened 2019-04-23
New syringe sprites by Keekenox

soslow4u:
He doesn't know how to git and github still, so I had to do it cuz it would be left untouched after.

viewport

inhands
empty on table

Discuss


granodd:
Looks kind of large but I like the sprite. Very clear what it is versus the tiny thing that most servers have them be. Fits the server's aesthetic good as well.

PR #3494, opened 2019-04-24
Fungo Di Artigleria tweak

RandomColonist:
Removes polyacid and plasma from list of possible chems for maintenance shrooms. Keep in mind that it still contains mutagen, space drugs, mercury, radium, thermite, synaptizine, toxin, slime jelly, cyanide, mindbreaker, sleep toxin, blattedin, chloral hydrate, so it can be pretty fucking harmful still.

Summed up reason for this change:
image


ACCount12:
Do not comment the fucking lines out. How many times does everyone have to say that? If you remove a line, REMOVE a line.

PR #3495, opened 2019-04-24
Fixes buggy recoil cursor

Kmc2000:
I didn't know recoil could be decimals :flatcat:


Kmc2000:
Need to add something else, one sec

PR #3499, opened 2019-04-25
Aim stat fixes and renaiming.

SpadesSlick:
* Aim is now Vigilance. Description was eddited, and later it will be a stat will hepl protects from insanity.
* That stat was fixed across all job.
* Serbs now have this stat.


Pariah919:
Is it not going to affect gun skills in the future? If not wouldn't removing the part about aiming better be fine? Or is that just there until you fully rework the stat?

SpadesSlick:
It's going to. A bit later we will look at actual balance of this thing, to make it work better. Because of how guns are important to our codebase, you should be able to use guns at 0 level of stats and not be terrible at it.

PR #3500, opened 2019-04-25
Unnerfing stungloves.

SpadesSlick:
Looks like this last nerf was too much.

### DIFF
--- a/code/game/objects/items/weapons/stunbaton.dm
+++ b/code/game/objects/items/weapons/stunbaton.dm
@@ -14,7 +14,7 @@
 	attack_verb = list("beaten")
 	price_tag = 500
 	var/stunforce = 10
-	var/agonyforce = 50
+	var/agonyforce = 30
 	var/status = FALSE		//whether the thing is on or not
 	var/hitcost = 100
 	var/obj/item/weapon/cell/cell = null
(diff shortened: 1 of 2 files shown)


Pariah919:
There's a good reason I'd argue why they were nerfed to being 30. The fact they dealt 50 agony is quite hilarious considering they STACK with pulled punches and intents like grab. Having batons deal more agony makes sense and gives them a purpose while considering that I've had times playing where IH literally ignored my laser/gunshots and fucking stungloved me in the face with 90 agony (or 60 if they don't know to aim for head) has been quite a few times. Though negativity bias.
I'd say 35 or 40 is a better number than 50.

SpadesSlick:
Then stacking should be removed.

I have no intentions of giving batons any purpose, they just there for the sake of filling up the armoury. What IH need is a melee weapon that can give them any reason to go to melee combat, because of how much ranged options they have.

Stealthkibbler:
Batons are also able to crack airlocks open pretty quickly, so their use extends past actual combat, not to mention because of the agony they do, when used as a harmbaton against roaches they kill them extremely fast.

TorinoFermic:
It turns out the stunbaton's stun var is too high and might be removed while keeping the agony intact.

PR #3502, opened 2019-04-25
Vigilance checks on psy related stuns.

SpadesSlick:
* You can now resist effect of two techhive structures with Vigilance stat.
* You can now resist both drawback and actual effect of ``Searing Revelation`` litany (crusader AoE stun).


SpadesSlick:
All of those were tested properly and works.

PR #3511, opened 2019-04-26
Fixes door.dm,420

soslow4u:
No more `door.dm,420: Cannot execute null.set opacity().`
Glass multi-airlocks weren't properly initialized


soslow4u:
>Muh panic when switched branches and compiled in middle of conversation

PR #3520, opened 2019-04-28
Chemistry Sprites

granodd:
Updates the chem heater (for future use) and changes beakers to better fit Eris' style.

eris beakers

image


SpadesSlick:
How do beakers look with chems inside them?

granodd:
@SpadesSlick 

image

Though do keep in mind that the lids have been edited to get rid of the bright sheen on its left side. I might re-add that, as well as making the beaker markings in the beaker's reagent overlay a little darker just so they're visible.

image

SpadesSlick:
Lids still need more work. Maybe a different color.

granodd:
What color might you suggest? There's only so much that can be done with their shape due to how small they are, which curved lines don't typically play well with at a pixel level

SpadesSlick:
Try more detailed one with different shape.

fernandos33:
i thini they are really cool the one thing would be nice is id the little lines qould matxh with how much liquis therw is too. like 60u would match the middle line and the 120u would match the last. i THINK its something you can change easily and would be really usefull for people rushing chems

ACCount12:
Any further changes should go into a separate PR.

PR #3521, opened 2019-04-28
Window defines

Kurgis:
Extends some of the defines in weapons to cover windows.


Kurgis:
MFW I somehow merged a work branch into my master.
MFW I have no idea how to remove a select commit from my codebase.

PR #3522, opened 2019-04-28
Expands uses of the resistance value.

Kurgis:
Windows now use the defines set out in defines weapons.

Slightly tweaks said define values.

Fixes most glass airlocks having an incorrect resistance value, making them stronger than regular airlocks.


Kurgis:
Ready for review.

PR #3523, opened 2019-04-29
Manual Redux

Kurgis:
Requires eventual replacement of existing books with map changes, but not in this PR.

Properly connects a new set of wiki manuals, which will connect to the wiki instead of the broken system the manuals currently use. Includes language support, and the wiki should connect to the Russian version of the page if it's on the Russian server and vice-versa.

It just works.


Kurgis:
It just works, ready for review.

Kurgis:
Also as a note:

EN server config: `WIKIURL` needs to be "http://wiki.syndi-cat.com/index.php" it is currently "http://wiki.syndi-cat.com/index.php/Main_Page_ErisEn"
`LANGUAGE` config needs to be properly set if it has not already.

RU server config: `WIKIURL` needs to be "http://wiki.syndi-cat.com/index.php" it is currently "http://wiki.syndi-cat.com/index.php/Main_Page_ErisRu"
`LANGUAGE` config needs to be properly set if it has not already.

PR #3525, opened 2019-04-29
Fixes lighting_turf.dm,63

soslow4u:
Runtime was caused by improper initialization - turfs without any light would get no `lightning_corner` datums


soslow4u:
Forgot to add: has heavy impact on maint shrooms as it would calculate proper luminosity-alike value  for checks instead of placing shroom on tile after runtime

PR #3530, opened 2019-05-01
Antag RIG buff

soslow4u:
- New item: **RIG Autodoc module**
Light version of beloved machine that would fix your traitorous ass when whole ship is against you.
Costs **16 TC**, you think it worth it? Let me know if this is too much.
- Added basic chemical dispenser for uplink. Just **8 TC** for 80 units of cool chems for all your needs.
   - Tricordrazine
   - Tramadol
   - Dexalin+
   - Spaceacillin
   - Dylovene and many more!

### DIFF
--- a/code/modules/clothing/spacesuits/rig/modules/utility.dm
+++ b/code/modules/clothing/spacesuits/rig/modules/utility.dm
@@ -82,6 +82,7 @@
 
 	device_type = /obj/item/weapon/rcd/mounted
 
+
 /obj/item/rig_module/device/New()
 	..()
 	if(device_type) device = new device_type(src)
@@ -380,4 +381,18 @@
 	jets.holder = null
 	jets.trail.set_up(jets)
 
-/obj/item/rig_module/foam_sprayer
+/obj/item/rig_module/autodoc
+	name = "mounted autodoc Jr."
+	desc = "A complex syrgery system for almost all your needs. Can't treat organ damage, but capable to fix inernal bleeding."
+	icon_state = "autodoc"
+	usable = 1
+	selectable = 0
+	toggleable = 0
+	disruptive = 1
+
+	engage_string = "Configure"
+
+	interface_name = "Autodoc"
+	interface_desc = "Module with set of instruments that is capable to preform basic syrgery on user"
+
+
--- a/code/modules/clothing/spacesuits/rig/modules/utility.dm
+++ b/code/modules/clothing/spacesuits/rig/modules/utility.dm
@@ -163,16 +163,12 @@
 		for(var/chargetype in charges)
 			var/datum/rig_charge/charge = charges[chargetype]
 			if(charge.display_name == R.id)
-
 				var/chems_to_transfer = R.volume
-
 				if((charge.charges + chems_to_transfer) > max_reagent_volume)
 					chems_to_transfer = max_reagent_volume - charge.charges
-
 				charge.charges += chems_to_transfer
 				input_item.reagents.remove_reagent(R.id, chems_to_transfer)
 				total_transferred += chems_to_transfer
-
 				break
 
 	if(total_transferred)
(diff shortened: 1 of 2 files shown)


code/modules/clothing/spacesuits/rig/modules/utility.dm
```diff
@@ -380,4 +377,102 @@
...
+	if(!.)
+		return
+	to_chat(holder.wearer, SPAN_NOTICE("With quiet hum internal machinery starts checking your body..."))
+	var/mob/living/carbon/human/H = holder.wearer
+
+	for(var/obj/item/organ/external/E in H.bad_external_organs)
+		to_chat(holder.wearer, "Found wounded organ: "+SPAN_WARNING("[E]"))
+		var/datum/rig_surgery/patchnote = new()
+		for(var/datum/wound/internal_bleeding/IB in E.wounds)
+			patchnote.surgery_type += DR_RIG_IB
```
LethalGhost: That's bit value. 
```suggestion
			patchnote.surgery_type |= DR_RIG_IB
```

code/modules/clothing/spacesuits/rig/modules/utility.dm
```diff
@@ -380,4 +377,102 @@
...
+	if(!.)
+		return
+	surgery_list.Cut()
+
+/obj/item/rig_module/autodoc/Process()
+	. = ..()
+	for(var/datum/rig_surgery/patchnote in surgery_list)
+		var/obj/item/organ/external/ex_organ = patchnote.ex_organ
+		if(!ex_organ)
+			CRASH()
```
LethalGhost: How about `surgery_list -= patchnote` to prevent continuous crashing?

code/modules/clothing/spacesuits/rig/modules/utility.dm
```diff
@@ -380,4 +377,102 @@
...
+		if(patchnote.surgery_type & DR_RIG_IB)
+			for(var/datum/wound/IB in ex_organ.wounds)
+				if(prob(50) && IB.internal)
+					ex_organ.wounds -= IB
```
LethalGhost: Why not use `IB.heal_damage(SOME_VALUE)`?

LethalGhost:
В целом.
По моему, стоит заменить 50% шанс на мгновенное лечение каждой раны.
Сделать, например так - в худе рига есть кнопка открывающая интерфейс модуля.
В нем кнопка "сканировать" (или сканирование происходит автоматически).
После сканирования мы получаем список ран носителя рига.
Возле каждого элемента списка кнопка "лечить". Лечение происходит не мгновенно (как и операции) и, возможно, требует носителя стоять неподвижно некоторое время.

Дискас.

ENG:
How about replace 50% chance of immidiatly healing (for each wound) with 
- Special module UI which display all your wounds
- You can heal each wound by clicking "heal" button near it in module UI
- Healing take time (like surgery) and may be require RIG wearer to stay still

Let the discussion begin!

code/modules/clothing/spacesuits/rig/modules/utility.dm
soslow4u: Well, it was late night brainfart about gibbed organs in action, it honestly should just remove patchnote at all.

code/modules/clothing/spacesuits/rig/modules/utility.dm
soslow4u: Because of `prob()` proc

soslow4u:
1. > Immediately healing.

What if organ has just too much brute/burn value? Is it wise to spend just 500 units of power to get rejuvenate analog? Shrapnel is still a threat, inner organ damage should be too.

2. > RIG UI with each organ.

Another minigame to get your wounds healed under action? I find it not wise to make player click additional buttons when he is getting hit.
But maybe yes, just need to make sure it would be even used by antags

3. > Need to stand still.

Yeah, I had same idea, but then decided to discuss it there. By the way, low healing values/rng rates were set in sake for keeping module active in fight for short amount of time.

tl;dr: need more ideas/screams/discussion about sanity of such suggestion

LethalGhost:
>Another minigame to get your wounds healed under action? I find it not wise to make player click additional buttons when he is getting hit.

Right now it can heal you extremely fast even mend fractures. There is no reason to not use it within a battle.

UI will make it more controllable and balanced - it'll force you to stop for healing. At least for complex operations. We will be able to set different powercost/time for different operations.

That change will force user to not "hope for a luck" but use that tool wisely.

soslow4u:
Well then, to be done.

PR #3531, opened 2019-05-01
IB rebalance

soslow4u:
Tweaked a bit IB numbers, it would take twice more time to build up nasty numbers and would let you live for like one or two more minutes.


fernandos33:
oh yes yes pleaseeeee yesss amazing

PR #3536, opened 2019-05-02
Cryopod antirobbery system

soslow4u:
Discuss

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -455,7 +455,7 @@
 	//Eject any items that aren't meant to be in the pod.
 	var/list/items = src.contents
 	if(occupant)
-		if(usr != occupant)
+		if(usr != occupant && !occupant.client)
 			to_chat(usr, SPAN_WARNING("It's locked inside!"))
 			return
 		items -= occupant
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -455,7 +455,7 @@
 	//Eject any items that aren't meant to be in the pod.
 	var/list/items = src.contents
 	if(occupant)
-		if(usr != occupant && !occupant.client)
+		if(usr != occupant && !occupant.client && occupant.stat<2)
 			to_chat(usr, SPAN_WARNING("It's locked inside!"))
 			return
 		items -= occupant
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -455,6 +455,9 @@
 	//Eject any items that aren't meant to be in the pod.
 	var/list/items = src.contents
 	if(occupant)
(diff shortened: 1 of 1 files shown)


GrayRachnid:
Speedmerge

Kurgis:
Cryopod only removes the occupant from the round into cryogenic storage once the occupant either ghosts or logs off. Far as I'm aware, cryopods can't be opened or interacted with, so someone can jump into a cryopod to be effectively invulnerable. Correct me if I'm wrong, but otherwise, there's no possible way to remove someone from a cryopod except by ejecting the occupant.

Kurgis:
The cryopod lock should check whether the occupant is logged on, and still in their body. It should only lock if the person is ghosted, offline, or otherwise outside of their character's body.

soslow4u:
Resolved

code/game/machinery/cryopod.dm
```diff
@@ -455,6 +455,9 @@
 	//Eject any items that aren't meant to be in the pod.
 	var/list/items = src.contents
 	if(occupant)
+		if(usr != occupant && !occupant.client && occupant.stat<2)
```
mattroks101: What is the stat greater than 2?

mattroks101: Like, stat 2 is dead, do we have a special one for stuff greater than 2?

soslow4u: Well, look at this one
https://github.com/discordia-space/CEV-Eris/blob/9ea20b5c4dc30d9cc0f4caaa6abbee2067b5d754/code/game/machinery/cryopod.dm#L254-L269

mattroks101: Oh I'm really fucking dumb. I misread it. God I hate this stat shit sometimes.

LethalGhost: Related to previous conversation
>Occupant is living and has no client.

actually means `occupant.stat != DEAD` ~~(or may be `! (occupant.stat & DEAD)`)~~
Anyway 2 is magick number.
```suggestion
		if(usr != occupant && !occupant.client && occupant.stat != DEAD)
```

soslow4u: Well, I tried to reuse code, as calling `incapacitated()` would be abused with just using `sleep` verb
https://github.com/discordia-space/CEV-Eris/blob/907d44d858d5fe0644724a7b8da7936f73e9eac1/code/__defines/mobs.dm#L154-L166
As there is no value for DEAD case. Honestly, have no idea how would it check if mob is dead and took the easiest option was available

LethalGhost: Check "Suggested change".

TorinoFermic: This is what I used on the excelsior chat to share the contents to ghosts with AntagHUD enabled.

PR #3537, opened 2019-05-02
Improve default NanoUI UI re-open behavior.

PJB3005:
If you were to click on a machine with its UI open, the UI flashes briefly and updates.

This commit makes the default behavior to update the UI (without that annoying flicker) and move the window in question to the top.

So if you have 5 different computers open you can click them in the map window to focus the correct window.

~~Don't merge yet I still gotta test something.~~


zamolxius:
based and redpilled

PJB3005:
Alright tested it on Windows, works like a charm.

PR #3552, opened 2019-05-07
Fixes welder tanks exploding when they shouldn't

### LINKED ISSUE #3551: Re-fueling a drill which is turned on will cause the drill to react to the weldertank like a welder, causing it to explode.
Obviously an important issue caused by #3534. Will try to find fix.

Kurgis:
Drills will no longer kill you on re-fuel.

This code is tested in-game, only works if the tool uses fuel, and has an active welding tool quality.

fixes #3551


Kurgis:
Ready for review, and requesting speedmerge.

PR #3557, opened 2019-05-09
[TWEAK] New tool and ability to force open powered doors.

TorinoFermic:
## General
Pneumatic crowbar subject to change in discussion.
The new tool cannot placed via the code because the Paramedic locker is fairly generic.

## Gallery
Tool : halligan

Screenshot : dreamseeker_2019-05-09_15-46-10

Video in use : 
[Video of it in action](https://imgur.com/emKgQZR.mp4)

## Changelog
🆑 TorinoFermic
tweak: Halligan bar and pneumatic crowbar will open powered doors.
/🆑


Kurgis:
After looking at real-life examples of a Halligan Bar, how is it more effective at digging than a shovel or jackhammer?

TorinoFermic:
Well, more like collapse the burrow than digging for the halligan. Sadly the code isn't easy to fix.

drexample:
This feature does seem nice, but the implementation makes it essentially an infinite use door emag, that doesn't even leave trace that the door was opened.

Make the prying rely more on robustness (as in - make the failure chance higher) and break airlock electronics after prying the door open. That would make sense, since you're essentially lodging it into mechanism and breaking it, which would require a lot of force.

Kurgis:
I disagree, such a feature change will make it nigh no different than an RNG emag. Not only will it open doors into areas you want to be, it will also break the door, forcing an area open until someone can bring a RCD to remove it.

drexample:
@Kurgis can't argue with that. Maybe make it leave easily distinguishable marks on an airlock instead? 
And I still think it should rely more on robustness - higher failure chance, coupled with already long use time seems like a good enough drawback in itself.

Kurgis:
I think a noticeable mark would be good, perhaps an overly of chipped door paint exposing the metal. Examining would give further details of the airlock's frame being exposed.

Also, perhaps a critical failure on forcing an airlock open would cause the bolts to close? Since every time a tool failure occurs on an object the fail rate increases, eventually if they are unable to pass the check they will be unable to re-roll their door open pass due to the bolts dropping. Unlike the airlock electronics breaking, this will only require a multi-tool to fix, instead of going through deconstruction. One possible function I could imagine from this, is someone intentionally failing checks to bolt a door, but I think such an interaction would lead to more divergent play than anything.

drexample:
Agreed. This solution seems good and balanced. I don't see forcing the bolts down on purpose a problem either.

drexample:
@TorinoFermic please, fix the conflicts and add the requested changes.

drexample:
@TorinoFermic one last chance to respond, before the PR is closed.

TorinoFermic:
Added suggested changes, please advise me if overlays is required to add visual damage to this airlock. This takes time for me to do sprites.

Kurgis:
Personally while I do think adding overlays would add some in-game QoL, I don't think it's necessary. I hope to see this in the game soon, and that it's merged without further requests.

SpadesSlick:
Opening doors in one click will be too powerful thing, there is a way to hack them, it present a danger for a first time for people without gloves, and it worked fine so far.

And the random chance to damage airlock or even lock it makes it worse then a multitool, so there is no point.

Kurgis:
> Maintainer: Make changes A, because otherwise this will be too powerful.
>
> Project Lead: Since this is making changes A, there's no point for this feature, PR closed.
>

_Nice._

PR #3561, opened 2019-05-11
Tables

RandomColonist:
gif
- Your sprite appears slightly higher when you are on the table or low wall now. Shamelessly stolen from (jesus forgive my sins) Infinity codebase.


mattroks101:
Wouldn't it just be easier mess with the table's crossed like this:
```
/obj/structure/table/Crossed(var/mob/living/M as mob)
	if(!flipped && istype(M))
		M.pixel_y = 12

/obj/structure/table/Uncrossed(var/mob/living/M as mob)
	if(!flipped && istype(M))
		M.pixel_y = 0```
Rather than checking if the mob is on a table every time they move? Is there any benefit to this method?

ACCount12:
It would cause issues if the table was deleted while the mob was on it.

Honestly, both approaches are bad and I don't think I'll be merging this PR.

PR #3564, opened 2019-05-12
NT access and roles update

ACCount12:
* NeoTheology now has special access levels that are granted by cruciform. None of those are mapped in yet, wait for the big NT update for that.
* Once it's mapped in, you would be able to open church doors with nothing but your cruciform and your faith. Opening doors with a cruciform you brutally ripped out of some religious fool's chest works too, but you wouldn't do that, would you?
* NeoTheology is now restricted from command positions and Ironhammer jobs. You can still convert to NT ingame, but spawning as an NT security officer is no longer allowed.
* NeoTheology is now capitalized as NeoTheology. Always.


driecg36:
I very much like all of these changes, but I'm not so hot about ALL of command being banned from Neotheology. IH, Moebius, and the technomancers I get, but I feel like the rest would be fine as NT (yes even the captain). There's some interesting RP in between balancing the interests of your department and that of the church as a religious head, but that's just my two cents.

Kurgis:
There has been interesting RP with the above banned roles while they were filled with NT characters, but a player shouldn't have to balance this serious dis-balance of interest every round. In nearly all cases, NeoTheologians will accept the word of an Inquisitor as being equivalent to the word of ~~god himself~~ an admin, they will turn this power to absolutely ruin Command, IronHammer, and The Crew in general. I've only encountered one instance where a player went a different direction to roleplay out a scenario, after an Inquisitor ordered a NeoTheologian Command Staffer to kill someone. These encounters I would assume are the exception, not the rule. When you decide to play with the Religion field set to "None" as Captain, you shouldn't be immediately set to fail due to your First Officer and Commander being NeoTheologians.

There's also the additional problem of Inquisitors not having role restrictions set, so that even Command Staff or IronHammer can become Inquisitors. Hilariously, I don't believe there's a check for your preference field, only cruciform, so a force-converted MEO could turn out to of been an Inquisitor the entire time!

driecg36:
@Kurgis I personally think that some command roles should still be allowed to be NT, as they do not have the power alone to completely give the inquisitor a free pass to do anything. This is especially tolerable as if Ironhammer is NT free, it means that a captain letting the inquisitor do whatever they want is a one way ticket to a heavily armed mutiny. I do agree with your point of the first officer, so let me propose which command roles should be NT or not.

Acceptable for NT - Captain and Merchant 
Not Acceptable for NT - Commander, MEO, MBO, FO, TE.

This would serve to solve the problem of overly pious command staff (which I personally don't see that often anyway) while still allowing the unique and interesting RP of a Neotheo head in some cases. The captain isn't able to do much if he doesn't have the support of IH, and the merchant is usually more interested with profit than with religion anyway.

SpadesSlick:
@driecg36 at this point this is balance problem we are solving there. This all could be reworked later, when we will start working on the setup rework, but for now we need simple and concrete solution.

driecg36:
Again I don't think there's anything unbalanced about an NT captain or an NT merchant, considering neither can get away with much if they don't have IH on their side. NT already feels pretty weak as of recently and I don't think it needs to be nerfed that hard (perhaps when the NT update drops that will change).

code/modules/client/preference_setup/occupation/occupation.dm
```diff
@@ -462,8 +466,9 @@ datum/category_item/player_setup_item/proc/prune_occupation_prefs()
 	if((GLOB.using_map.flags & MAP_HAS_RANK)\
 	   && (!pref.char_rank || !mil_branches.is_spawn_rank(pref.char_branch, pref.char_rank, S)))
 		pref.char_rank = "None"
+	*/
```
mattroks101: We should get around to purging this branch stuff one of these days.

ACCount12: Yeah, it's not really useful to us. Doesn't seem to be adaptable to our corporatocracy.

PR #3566, opened 2019-05-13
Explode this PR its wrong

fernandos33:
…is PR thing not sure if i did it right
 https://i.imgur.com/M02oEFL.gifv


fernandos33:
so all i did on VS was to open and read the code in module/mining/ ore.dm and ore-datum.dm and in weapons/bag.dm but now here it says i deleted other code... does anyone understand what happened?

PR #3569, opened 2019-05-13
[tweak] Punks Smoke

fernandos33:
well this adds 4 cigars as contraband for the cigarette machine (costing 25) , and  more important adjusted the price from 300 to 12 so people are able to fuck their lungs at will (we are in cyberpunk yeah smoking is expensive in real life but... even them not that much)


ACCount12:
Too much of a price decrease. Set it to around 100, 50 at the minimum.

fernandos33:
will do in matter of minutes

ACCount12:
Now the prices of matchbox and cheap lighter don't make any sense.

Set cigar price to 200, cigarettes to 100, cheap lighter to 10 and matchbox to 5.

fernandos33:
made a mistake in the description it sets the price for the ones you said so

zamolxius:
why would a box of matches be 5? a damn bread box is 50.  cheap lighter should be at least 50-60

ACCount12:
Word "cheap" in the name doesn't clue you in?

An entire welding tool costs 40, and it's both more complex and more useful than a mass produced plastic lighter.

fernandos33:
ok set and done this pr to moving to the export little mistake now

PR #3570, opened 2019-05-14
Export fix

### LINKED ISSUE #3554: [BUG] Sell() Proc Only Calls One (1) Time
Selling things on the cargo shuttle only works one time before the Sell() proc stops properly working. In a controlled test of one crate with five sheets of plasma, the first was sold and credits were earned, but attempting to sell the second crate yielded no change in credits to the Guild account, and (likely) no change in the cargo history of items sold.

This can be recreated easily on the live server by selling anything, and then attempting to sell anything afterwards. The cargo history screen being stuck on the first ever shipment you sold.

fernandos33:
Fixes #3554


fernandos33:
damn mistake made here

PR #3572, opened 2019-05-14
export fix

fernandos33:
it is that

### DIFF
--- a/code/controllers/subsystems/supply.dm
+++ b/code/controllers/subsystems/supply.dm
@@ -71,7 +71,7 @@ SUBSYSTEM_DEF(supply)
 		points += E.total_cost
 
 	msg += "<br>Total exports value: [points] credits.<br>"
-	E.export_end()
+	exports.Cut()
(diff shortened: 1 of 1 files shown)


code/controllers/subsystems/supply.dm
```diff
@@ -69,7 +69,7 @@ SUBSYSTEM_DEF(supply)
 
 		msg += "[export_text]<br>"
 		points += E.total_cost
-		E.export_end()
+		exports.Cut()
```
ACCount12: You did that in a loop, and you needed to do it after the loop is done. It would break the loop otherwise.

granodd: Just remove the tab before `exports.Cut()` to bring it in line with the `for` loop before it so that it's not part of the loop and this will work. Tested on a local server and it works wonderfully.

fernandos33:
uhuulll (thanks account you basically did it all)
awesome

PR #3588, opened 2019-05-19
Assembly fix

RandomColonist:
No one besides me even tried to use them, but they was broken because someone copy-pasted mindlessly. 

- You can open integrated electronics assemblies now again.


ACCount12:
oh wow that's A LOT of merge commits

PR #3592, opened 2019-05-21
Fixes for a recent update

ACCount12:
Fixes the issues that surfaced in the recent update: NT folks spawning with no crux, vendomats not showing disk names, etc.


ACCount12:
A new BYOND feature, see the F1 docs.

Basically, it's an alternate way to write chains like if(H && H.mind && H.mind.assigned_job). If H is null, `H?.mind` returns null instead of throwing a runtime in your face, and so it goes.

PR #3595, opened 2019-05-24
Retro Laser

Kurgis:
Removed old retro laser sprites, including the laser tag versions. Adds a new retro laser sprite done by Watler, adds new laser tag sprites using said retro laser that I palette changed.

retrolaser
lasertagblue
lasertagred
retroinhand


Kurgis:
Ready for review.

RandomColonist:
It was.large handgun before, but you made it look like carbine. Thus, you need to remove possibility to insert it into holster pouch and increase it w_class on 1 as well.

Kurgis:
> It was large handgun before, but you made it look like carbine. Thus, you need to remove possibility to insert it into holster pouch and increase it w_class on 1 as well.

Thank you, this passed my mind.

Kurgis:
Inserting the retro gun into the pistol holster has been addressed, this should be good to go.

Kurgis:
In commit 1f4f485, the title I placed may of cause misconception. It only affects the laser tag guns so people can't put them in their pocket.

RandomColonist:
As i said, you might want to slightly buff w_class of Cog because now it's literally pistol-sized, but if that's intended - good.

Kurgis:
The w_class of the Cog is already set to `ITEM_SIZE_NORMAL` despite the icon being pistol-shaped, so it's already the same size as other laser rifles.

RandomColonist:
Very weird, this way you can carry around multiple laser rifles in your satchel. I will correct that myself in some time then, thanks for the feedback.

Kurgis:
This is ready for review.

Kurgis:
I've actually deleted the repository hosting this PR, and GitHub Editor isn't going to be able to fix .dmi conflicts.

PR #3598, opened 2019-05-25
Gun Icon Refactor

Kurgis:
Gun icons are now located in a dedicated folder under objects, they'll be separated by their respective parent item. A laser gun has energy as its parent, and would be located in `energy.dmi` in the `gun` folder for example.


Kurgis:
This is ready for review.

PR #3601, opened 2019-05-25
New Ian sprites.

Kurgis:
Watler's 2017 corgi sprites, in addition of a puppy sprite I did based off it.

corgi
corgiside


Kurgis:
Checked with the corgi clothes so that'd they still fit.

PR #3608, opened 2019-05-28
Blob grenade improve

ITCups:
Blob nade now spawn blob seed that will search for food and when it reach certain mass it will turn into mature blob.


ACCount12:
Baystation grab port here? Wow, that's some PR feature creep.

ITCups:
i know right ? ill separate this PR on completion, but cant guarantee much.

SpadesSlick:
This PR will not be accepted at this state anyway, please slit it different PRs.

PR #3611, opened 2019-06-02
[FIX] Serb double spawning fix

### LINKED ISSUE #2835: BUG: Serbian objectives
Serbian objectives shown to all players two times instead of one - first time when shuttle start to fly back, second time when shuttle arrives back at base. In other words - every time it starts exept for when it leaves base.

### LINKED ISSUE #3423: [BUG] Gibbing despawns items. Again.
BUG: Everything embedded in roaches disappear after gibbing

TorinoFermic:
## General
Not sure, but I don't have the easy way to test this fix.

Fixes #2835
## Changelog
🆑 TorinoFermic
bugfix: Serbs shouldn't spawn more than once for now.
/🆑

### DIFF
--- a/code/game/gamemodes/roleset/faction/roleset_faction.dm
+++ b/code/game/gamemodes/roleset/faction/roleset_faction.dm
@@ -71,6 +71,7 @@
 	if (success_quantity >= target_quantity)
 		//Yay, all antags successfully spawned
 		return TRUE
+
 	else
 		//Welp, we didn't manage to spawn as many as desired
 		log_and_message_admins("Storyteller Warning: Antagonist Spawning unsuccessful for antagonist [role_id], [antag] \n \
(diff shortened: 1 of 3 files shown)


Kurgis:
This would retroactively fix #2835

TorinoFermic:
Thanks, I corrected my desc.

code/game/gamemodes/storyevent.dm
```diff
@@ -71,7 +71,7 @@
 		if (report) report << SPAN_NOTICE("Failure: The event is disabled")
 		return FALSE
 
-	if ((ocurrences_max) > 0 && (ocurrences >= ocurrences_max))
+	if (ocurrences_max > 0 && ocurrences >= ocurrences_max)
```
ACCount12: This couldn't have fixed anything. See BYOND F1 docs on operators: `>` and `>=` always have priority over `&&`, with or without `()`. Them having priority over `&&` also seems to be desired behavior here.

If you want to fix this issue, it would require further investigation.

code/game/gamemodes/roleset/faction/roleset_faction.dm
```diff
@@ -82,6 +82,8 @@
 		if (success_quantity > 1)
 			success_percent = success_quantity / target_quantity
 		cancel(severity, success_percent)
+	
+	return TRUE
```
ACCount12: I don't think it should return TRUE if 0 antags were spawned.

TorinoFermic: This is fixed, but I assumed the occurrences strictly meant the event did occur and will never again but I was wrong.

TorinoFermic:
I hope the change is the right one.

PR #3614, opened 2019-06-04
Pulse rifle changes

Kurgis:
This rebalances the Cassad, pulse rifle, and pulse rifle destroyer. Cassad is now a child of the pulse rifle, old Cassad type needs to be deleted with mapmerger. Changes references to `/pulse_rifle` to just `/pulse`. Gun force now uses define.

Pulse rifle is now a large item, this affects the Cassad which also gets a minuscule power cost increase. Adjusted the pulse rifle's energy usage drastically, so it is no longer unusable.

Pulse rifle destroyer now has two fire modes, DESTROY and RAPID. One destroys as per usual, the other destroys twice as fast, for ~2.3x the energy cost.


RandomColonist:
Can you also remove Zoom from pulse rifle? I don't really see why it would have those, as this is CQC gun for breaking walls doors and burning through skulls. We already have dedicated energy sniper rifle for it (Valkirye).

Kurgis:
Sure.

code/modules/projectiles/guns/energy/cassad.dm
```diff
@@ -1,4 +1,4 @@
-/obj/item/weapon/gun/energy/cassad
+/obj/item/weapon/gun/energy/cassad //TODO: Replace all in-game instances with "/obj/item/weapon/gun/energy/pulse/cassad" once mapmerger is available, and then delete.
```
ACCount12: See `maps/map_shims.dm` on how to deal with path changes while the map is locked.

Kurgis: Ah, thank you, I've forgotten you've added this tool.

Kurgis:
Request changes made, this is ready for review.

PR #3617, opened 2019-06-05
Excelsior machine gun fix

Kurgis:
Works so you can mess around with it on the test server.


ACCount12:
Aaaand it just conflicted with your other PR. Thanks gun icons.

PR #3620, opened 2019-06-06
Makes HUD overlays more translucent

RandomColonist:
Direct port of https://github.com/Baystation12/Baystation12/pull/14428.
meson
thermal
Night vision overlay picture should be here, but it appears that we do not use it at all.


ACCount12:
Holy fuck, can you replace master to remove all that merging from your repo?

PR #3622, opened 2019-06-06
Removes PDA meta

Kurgis:
Restores the tesla link to PDAs, so they effectively don't die. Removes the special PDA cell, which would be "mined" by Science every round. The tesla link only works if the area controller is operative and powered, which should keep this balanced in a way. Nanako had a different design decision for the PDAs, but with ACCount's recent modular computer cell change I can see this is no longer the case, and I believe that this may be a better solution.

Reduces tesla link's gold matter amount from 1 to 0.25, as it was too high. Every modular computer has a tesla link, so that's a material sheet per deconstruction. It's not a high-tier component.


ITCups:
I suggest add new tesla link type that can only be inserted inside of PDA/tablet. Cuz free tesla link might affect balance in future hacking rework.

Kurgis:
https://docs.google.com/document/d/1kKHEgMnbpKRLBPihMz25GYlL5-m529m7LZHd2FalacQ

After checking the design doc, I was unable to find a mention of the Tesla Link, besides an entry labeled "Tech Link". Tesla links are a required piece of every non-handheld modular computer, it's a basic component found in everything, the problem will extend far beyond PDA/tablets into modular computers itself.

ITCups:
I mean having free tesla link that can be transferred from PDA to device capable of hacking is strong. You can basically have portable hacking station that don't need recharging.
If we to separate tesla link into basic and advanced version we can avoid that problem.

ACCount12:
Can you please not? Tesla Links are supposed to be rare gear, and the fact that stationary consoles need those to function is oversight, not intention.

PR #3623, opened 2019-06-07
Grey Railings, railings improvement and crafting fixes

ITCups:


code/game/atoms.dm
```diff
@@ -622,7 +622,7 @@ its easier to just keep the beam vertical.
 
 //This proc is called when objects are created during the round by players.
 //This allows them to behave differently from objects that are mapped in, adminspawned, or purchased
-/atom/proc/Created()
+/atom/proc/Created(var/mob/user)
```
ACCount12: `var/` is unnecessary in proc declarations, see contribution guidelines. I wouldn't delay the PR over it, just keep that in mind for the future.

PR #3630, opened 2019-06-14
Biomatter solidifier

Macleod962:
Added solid biomatter sheets, so it's possible to create things with that.
Added solidifier. That machine converts liquid biomass into useful sheets.
Sprites are done by Granodd.


code/modules/biomatter_manipulation/solidifier.dm
```diff
@@ -0,0 +1,110 @@
...
+	var/active = FALSE
+	var/port_dir = NORTH
+	var/obj/structure/reagent_dispensers/biomatter/container
+
+
+/obj/machinery/biomatter_solidifier/update_icon()
+	if(active)
+		icon_state = "solidifier_on"
+	else
+		icon_state = initial(icon_state)
```
LethalGhost: It's better to use `icon_state = initial(icon_state) + "_on"` to keep ability to redefine icon state for child types.

code/modules/biomatter_manipulation/solidifier.dm
```diff
@@ -0,0 +1,110 @@
...
+				if(locate(/obj/item/stack/material/biomatter) in loc)
+					for(var/obj/item/stack/material/biomatter/stack_on_my_loc in loc)
+						//if this isn't full, we use that stack(current)
+						if(stack_on_my_loc.amount < stack_on_my_loc.max_amount)
+							current_stack = stack_on_my_loc
+							break
+
+				if(current_stack)
+					current_stack.amount += 1
+					current_stack.update_strings()
```
LethalGhost: That should be handled with at least `stack.add(AMOUNT)`

code/modules/biomatter_manipulation/solidifier.dm
```diff
@@ -0,0 +1,110 @@
...
+					if(current_stack.amount == current_stack.max_amount)
+						state("Stack is ready.")
+						ping()
+				else
+					current_stack = new(loc)
+
+			else
+				abort("Insufficient amount of biomatter.")
+		else
+			abort("Container of liquid biomatter required.")
```
LethalGhost: I suggest to rewrite structure of that proc to keep abort/message calls neart check like
```
if !check_1()
	error_1()
else if !check_2()
	error_2()
else
	go()
```

instead
```
if check_1()
	if check_2()
		go()
	else
		error_2()
else
	error_1()
```

code/modules/biomatter_manipulation/solidifier.dm
```diff
@@ -0,0 +1,110 @@
...
+			playsound(src, 'sound/machines/airlock_ext_open.ogg', 60, 1)
+			to_chat(user, SPAN_NOTICE("You dettached [tank] from [src]."))
+			container = null
+			toxin_attack(user)
+		else
+			to_chat(user, SPAN_WARNING("There are already connected container."))
+
+
+/obj/machinery/biomatter_solidifier/attack_hand(mob/user)
+	user.setClickCooldown(DEFAULT_ATTACK_COOLDOWN*2)
```
LethalGhost: Seems like *Костыль*.

Macleod962:
Updated

code/modules/biomatter_manipulation/solidifier.dm
LethalGhost: With `stack.add(AMOUNT)` line
`current_stack.update_strings()`
is no longer required.

Macleod962:
Updated

code/modules/biomatter_manipulation/solidifier.dm
```diff
@@ -0,0 +1,109 @@
+//Biomatter compressor
+//This machine converts liquid biomatter to solid one(sheets)
+//Working with this also required bio protection cloths
+
+#define BIOMATTER_PER_SHEET 		10
+#define CONTAINER_PIXEL_OFFSET 		6
+
+/obj/machinery/biomatter_solidifier
+	name = "Biomatter compressor"
```
ACCount12: Better keep object names lower case.

Macleod962:
Updated again

PR #3632, opened 2019-06-25
Base for adding craftingRecipes, stats, etc. using items

ITCups:
Universal interface to apply some effects for mobs.
can be stored in vars and utilized as knowledge, mutations etc.


cev_eris.dme
```diff
@@ -2078,7 +2080,7 @@
 #include "code\modules\projectiles\ammunition\magazines.dm"
 #include "code\modules\projectiles\guns\alien.dm"
 #include "code\modules\projectiles\guns\energy.dm"
-#include "code\modules\projectiles\guns\hatton.dm"
+#include "code\modules\projectiles\guns\Hatton.dm"
```
ACCount12: ?

ITCups: hell if i know, my VS code does this. Let it be i wont rename it back in every PR

PR #3633, opened 2019-06-30
[TWEAK] Climbing into delay is now affected by your stat Vigilance

TorinoFermic:
## Changelog
🆑 TorinoFermic
tweak: Vigilance will affect how fast you climb onto/into disposals, ladders, tables and railings.
/🆑

### DIFF
--- a/code/_onclick/other_mobs.dm
+++ b/code/_onclick/other_mobs.dm
@@ -52,7 +52,8 @@
 
 			visible_message(SPAN_WARNING("[src] starts climbing onto \the [A]!"))
 			shadow.visible_message(SPAN_WARNING("[shadow] starts climbing onto \the [A]!"))
-			if(do_after(src, 50 * src.stats.getDelayMult(STAT_VIG, STAT_LEVEL_BASIC), helper))
+			var delay = 50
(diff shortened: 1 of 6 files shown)


code/_onclick/other_mobs.dm
```diff
@@ -52,7 +52,8 @@
 
 			visible_message(SPAN_WARNING("[src] starts climbing onto \the [A]!"))
 			shadow.visible_message(SPAN_WARNING("[shadow] starts climbing onto \the [A]!"))
-			if(do_after(src, 50, helper))
+			var delay = 50
```
LethalGhost:
```suggestion
			var/delay = 50
```

LethalGhost: Please keep codestyle solid with all other files.

TorinoFermic: How I could have missed this detail, although that's my experience of Java leaking in my code.

TorinoFermic:
Fixed, anything else to fix like changing the min of 66% for something better ?
I had put that otherwise IH will have instant climbing on everything like tables, etc.
All stats checks uses the max of STAT_LEVEL_EXPERT which is 40 stats points. And the reduction is capped to 66% of original timer.

PR #3641, opened 2019-07-14
Update merge

ACCount12:
Merges update branch into master. This is likely to break biomass material icons, I'll have to fix that later.


ACCount12:
Should be ready now.

SpadesSlick:
To be clear, this is NOT NT update, but a collection of small shit and mechanics that were put on hold until they were ready.

PR #3643, opened 2019-07-15
New Parallaxes

granodd:
Adds 6 new parallaxes, bringing the total up to 7. Each round, a random one will be selected.


code/_onclick/hud/parallax.dm
```diff
@@ -13,7 +17,7 @@
 	var/obj/parallax_screen/parallax_screen
 	var/list/layers = list()
 
-	New(mob/M)
+	/obj/parallax/New(mob/M)
```
ACCount12: I'm fairly sure that the tabs here are fucked up.

granodd: I thought the same thing, but moving their indentation back any seems to break everything in that code block there. It was like that when I came to the code and I didn't do too much to it aside from changing the `New()` to be the type path version of itself. I might just change it back as that `New()` looks like it's inside the body of the parallax object itself.

For all intents and purposes, this does the job of choosing the parallax and assigning it for everyone to see as opposed to giving every mob a random parallax. This code can be adjusted in the future if need be.

granodd:
The six new parallaxes, for visual reference.

image

PR #3650, opened 2019-07-25
Cleaning up list helpers, reorganizing defines, helpers and macros.

Ghommie:
What the heck? We don't have a pull request template format? This is very disappointing, but I'm digressing.

So, what I'm doing here is exactly what's said on the tin.
Reorganizing the top half of the topmost part of the code folder to be less confusing.

- ` __defines` -> `__DEFINES`,
- `_helpers` -> `__HELPERS`,
- deleted `_macros.dm` and `_marcos` folder,
- redistributing its contents to `__DEFINES` and `__HELPERS`,
- `lists.dm` -> `_lists.dm` (also cleaning it up),
- `math_physics.dm` -> `maths.dm`, killing off its helper counterparts and moving the atmosia related defines to `atmos.dm`.

Why? The repo contains several duplicate and copycat helper functions, its list sorting procs are still using the old methods vs Carnie's TimSort and math helpers procs vs macros.

Forgive me for making my first PR in this repository to be so unatomized, I got carried on with the maths defines a bit too much and only noticed so as I looked at the ++/-- line changes just now, I'll revert that bit if requested.

Draft pull request as local testing will be required because of the likeliness of mistakes being due to its size. Feel free to review nevertheless.


SpadesSlick:
Do you planning in making it not draft?

Ghommie:
Oh snaps, I'll be back on it in a day or so, sorry about it.

Ghommie:
Reminder to ping me if this causes issues.

PR #3653, opened 2019-07-28
Resprites Advanced Trauma/Burn Kits and gives them charge indicators

Michiyamenotehifunana:
TraumaBurnKit

This PR will replace the sprites of the Advanced Trauma Kit and Advanced Burn kit. These kits are found in Advanced First Aid Kits. In addition, they now have a "ammo" counter, which will indicate the amount of uses left. It should be noted that the empty state won't ever be visible, since the kits will simply be deleted when it's used up.

c8yXVMZO58

In-game preview:
dreamseeker_SYIvyEYcIB

dreamseeker_2JPcIJApuc

All assets included in this PR are licensed under CC-BY-SA unless otherwise noted.
---
### ADDENDUM
This PR will now add two new vars to item stacks: Consumable, and Splittable. They are true by default. If the consumable var is true, the item stack will be deleted upon exhausting all of its charges. If the Splittable var is false, the item stack cannot be split.

The Advanced Trauma and Burn kit have both of the aforementioned vars set to false.


drexample:
Please merge and resolve the conflicts before we merge.

Michiyamenotehifunana:
Conflicts resolved, retested and working fine.

PR #3636, opened 2019-07-01
burrow rework

ITCups:
- new sprites
- burrows now separated into closed ones (cracks) and open, closed burrows are often ignored when choosing target for migrations
- burrows can be closed with explosions
- burrow can be closed with steel sheets
- spawns scrap pile with random loot if closed with digging tool

removed unnecessary loging

PR #3663, opened 2019-08-04
[WIP] Burrows Rework

mistyLuminescence:
- [x] Better loot- destroying a burrow now spawns a junk pile to search through!
- [ ] Less annoying ways of destroying (click in-hand to destroy slowly, or use tool to destroy quickly?) Consider reducing base health to make destruction speed faster.
- [ ] Fancier sprites?


mistyLuminescence:
... apparently this is already a thing #3636 we should really remove it from the task list

PR #3671, opened 2019-08-11
pixel aiming system

Derven:
shooting on a limb, one can get into any closest limb from the one that was selected on the doll, provided that the mouse is pointed there when pressed. Also depends on the direction of the target.


LudwigVonChesterfield:
just saying... this could be accomplished with visual_contents to enable clicking on the limbs... that way you wouldn't have to make a ladder of ifs for each possible pixel layer...

Derven:
изображение

PR #3672, opened 2019-08-11
pixel aiming system

Derven:
shooting on a limb, one can get into any closest limb from the one that was selected on the doll, provided that the mouse is pointed there when pressed. Also depends on the direction of the target.


Kurgis:
Due to the way players stand, I'd imagine this system is bound to be somewhat awkward.

PR #3679, opened 2019-08-15
overmap comet

Derven:
Сomet event that will actually fly, while shitting with it tail all around

https://cdn.discordapp.com/attachments/220251965609869312/610901395792592926/unknown.png

https://recordit.co/UOLmhhKGxf

### DIFF
--- a/code/modules/overmap/events/movable.dm
+++ b/code/modules/overmap/events/movable.dm
@@ -151,9 +151,6 @@
 
 /obj/effect/overmap_event/movable/comet/cometmedium/Initialize()
 	. = ..()
-	//start_x = pick(2, maps_data.overmap_size - 1)
-	//start_y = pick(2, maps_data.overmap_size - 1)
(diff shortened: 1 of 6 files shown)


code/modules/overmap/events/movable.dm
```diff
@@ -0,0 +1,195 @@
...
+/obj/effect/overmap_event/movable/comet/cometmedium
+	movable = 0
+	eventtype = /datum/overmap_event/meteor/comet_tail_medium
+	icon_state = "meteor1"
+
+/obj/effect/overmap_event/movable/comet/cometmedium/Initialize()
+	. = ..()
+	//start_x = pick(2, maps_data.overmap_size - 1)
+	//start_y = pick(2, maps_data.overmap_size - 1)
+	//moving_vector = pick(NORTH, SOUTH, WEST, EAST)
```
mattroks101: Why are these commented out?

PR #3693, opened 2019-08-17
Moves power cell icons into dedicated file.

Kurgis:
## About The Pull Request

Power.dmi: Deletes portgen legacy icons.

Moves power cell icons into their own file from power.dmi.

## Why It's Good For The Game

Power cells were in Power.dmi back when there was only four of them, now there's over 40+ states. This makes it more manageable.


Kurgis:
Placing this into a different PR so I don't have to fix the inevitable conflict.

PR #3699, opened 2019-08-21
Megafauna, broken version

ACCount12:
This is an outline of a megafauna PR, with lots of pieces missing. It wouldn't work as is.


SpadesSlick:
The branch will remain there for further reference, but the draft must go away.

PR #3711, opened 2019-08-28
Components basics

LethalGhost:
## About The Pull Request

Add basic mechanics for Components.
Already can be used but have no signal/call implemented
Also. Add proc:
`/proc/stack_trace(msg)` `/datum/proc/stack_trace(msg)` - `CRASH(msg)` wrapper
`/proc/type2parent(child)`, `/proc/type2top(the_type)` 

## Changelog
:cl:
code: Add components basic mechanics
/:cl:


code/__defines/components.dm
```diff
@@ -0,0 +1,291 @@
...
+//////////////////////////////////////////////////////////////////
+
+// /datum signals
+#define COMSIG_COMPONENT_ADDED "component_added"				//when a component is added to a datum: (/datum/component)
+#define COMSIG_COMPONENT_REMOVING "component_removing"			//before a component is removed from a datum because of RemoveComponent: (/datum/component)
+#define COMSIG_PARENT_PREQDELETED "parent_preqdeleted"			//before a datum's Destroy() is called: (force), returning a nonzero value will cancel the qdel operation
+#define COMSIG_PARENT_QDELETED "parent_qdeleted"				//after a datum's Destroy() is called: (force, qdel_hint), at this point none of the other components chose to interrupt qdel and Destroy has been called
+
+// /atom signals
+#define COMSIG_PARENT_ATTACKBY "atom_attackby"			        //from base of atom/attackby(): (/obj/item, /mob/living, params)
```
Carbonhell: Imo it's more confusing to add those defines but not implement them properly than just leaving them out so that future coders immediately notice the absence of the signal.

code/_helpers/unsorted.dm
```diff
@@ -1242,6 +1242,13 @@ var/list/FLOORITEMS = list(
 /proc/crash_with(msg)
 	CRASH(msg)
 
+//gives us the stack trace from CRASH() without ending the current proc.
+/proc/stack_trace(msg)
```
Carbonhell: Why is this needed both as a global proc and as a datum's?

LethalGhost: Keep that from original source.
Actually agree with you. We need only `CRASH(msg)` or `crash_with(msg)` for real.

PR #3726, opened 2019-09-05
Batch crafting

PlinHost:
## About The Pull Request
It's back.
img
The plus button prompts user for amount.
img

## Changelog
:cl:
add: Added batch crafting to craft menu for certain recipes
/:cl:

### DIFF
--- a/code/datums/craft/recipe.dm
+++ b/code/datums/craft/recipe.dm
@@ -68,51 +68,24 @@
 	return CS.apply(I, user, target, src)
 
 /datum/craft_recipe/proc/build_batch(mob/living/user, amount)
+	if(!amount)
+		return
 	if(steps.len > 1)
 		warning("Attempted to batch craft a multi-step recipe: [name]")
 		try_build(user)
 		return
 
-	if(!can_build(user, get_turf(user)))
-		return
-
-	var/datum/craft_step/CS = steps[1]
-
-	var/resamt = 0
-	while(resamt < amount)
-		var/obj/item/I = CS.find_item(user)
-
-		if(!I)
-			to_chat(user, SPAN_WARNING("You can't find required item!"))
+	var/obj/item/CR = try_build(user)
+	while(--amount)
+		var/obj/item/stack/S = try_build(user)
+		if(!S)
 			break
-
-		//Robots can craft things on the floor
-		if(ishuman(user) && !I.is_held())
-			to_chat(user, SPAN_WARNING("You should hold [I] in hands for doing that!"))
-			break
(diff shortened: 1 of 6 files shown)


code/datums/craft/recipe.dm
```diff
@@ -67,6 +67,52 @@
 	var/datum/craft_step/CS = steps[step]
 	return CS.apply(I, user, target, src)
 
+/datum/craft_recipe/proc/build_batch(mob/living/user, amount)
+	if(steps.len > 1)
+		warning("Attempted to batch craft a multi-step recipe: [name]")
```
ITCups: do we really need this unnecessary logging ?

code/datums/craft/recipe.dm
```diff
@@ -67,6 +67,52 @@
 	var/datum/craft_step/CS = steps[step]
 	return CS.apply(I, user, target, src)
 
+/datum/craft_recipe/proc/build_batch(mob/living/user, amount)
```
ITCups: make this proc utilize try_build() instead of copypasting code

PlinHost: That would result in lots of stacks containing only one item spawning when the result is a stack. This proc puts all items in one stack in that case.

code/datums/craft/recipe.dm
PlinHost: User asks for fifty but gets one because some coder messed up a recipe? Yeah why should we log that

ITCups: mention that then

code/datums/craft/recipe.dm
ITCups: make try_build() return result if succeeded and stack all stackables in build_batch()

PlinHost: And where should the spawn code from try_build go?

ITCups: it stays there, you just merge all stackables that try_build() returned

code/datums/craft/recipe.dm
PlinHost:
```suggestion
		warning("A multi-step recipe has BATCH_CRAFT flag: [name]. It should not!")
```
Like that?

code/datums/craft/recipe.dm
PlinHost: resolved in 54bced1e

code/datums/craft/recipe.dm
ITCups: yeah good enough

PR #3748, opened 2019-09-10
Cyborg cuffs are removable, dead simplemobs don't die repeatedly

Hatterhat:
## About The Pull Request
Fixes two bugs, thanks to @rpilcrow for fixing these so I could steal his code for it. Yoink.
## Why It's Good For The Game
Bugs are bad. Might as well not have them.
## Changelog
:cl:
fix: Cyborg-applied handcuffs should be able to be resisted out of, now.
fix: Dead simplemobs now die once instead of every tick.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/storage/storage.dm
+++ b/code/game/objects/items/weapons/storage/storage.dm
@@ -484,18 +484,14 @@
 
 	return ..()
 
-/obj/item/weapon/storage/verb/quick_empty(var/turf/target)
+/obj/item/weapon/storage/verb/quick_empty()
 	set name = "Empty Contents"
 	set category = "Object"
 
 	if((!ishuman(usr) && (src.loc != usr)) || usr.stat || usr.restrained())
 		return
 
-	var/turf/T
-	if(isturf(target))
-		T = target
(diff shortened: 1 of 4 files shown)


code/game/objects/items/weapons/storage/storage.dm
```diff
@@ -484,18 +484,14 @@
 
 	return ..()
 
-/obj/item/weapon/storage/verb/quick_empty(var/turf/target)
+/obj/item/weapon/storage/verb/quick_empty()
```
ITCups: this was done intentionally, so boxes could be emptied when clicking on turfs
remove this

Hatterhat: alright so turns out the other guy had no idea that you _could_ empty quick_empty capable items by clicking on turfs, he said he'll get back to me on that eventually

Hatterhat: trying to empty contents onto the floor with the verb opens a menu that lets you drop stuff onto turfs in viewrange which is something that should be addressed though

PR #3758, opened 2019-09-14
Quick-emptying isn't long-range now

Hatterhat:
## About The Pull Request
Yet another fix stolen from @rpilcrow; thanks, buddy! Makes dumping things out be one tile range only, and the "empty contents" bag verb only dump out at your feet.
## Why It's Good For The Game
Dumping out a bag of stuff through a wall probably isn't the best plan.
## Changelog
:cl:
fix: It is now impossible to dump things out at long range. Probably.
/:cl:


code/game/objects/items/weapons/storage/storage.dm
```diff
@@ -477,30 +477,31 @@
...
-	else
-		T = get_turf(src)
+	var/turf/T = get_turf(src)
 	if(!istype(T))
 		return
+	dump_it(T)
+
+/obj/item/weapon/storage/proc/dump_it(var/turf/target) //he bought?
+	if(!isturf(target))
+		return
```
LethalGhost: Shouldn't here be `Adjacent` check?

Hatterhat: src being set to in view(1) about 12 lines or so above handles it being adjacent, far as I know. tested locally, it's fine, unless some extra janky stuff is happening

LethalGhost: It's only checks range, but adjacent checks ability to touch to.
For example if there's a windor/window on the way `view(1)` will allow interactions but `adjacent` will not.

Hatterhat: alright fair enough i'll get something up for that

Hatterhat: i'm dumb as shit and have no idea how i'd try to slot in an `Adjacent` check

Hatterhat: The `Adjacent` check seems unnecessary; I've tested it and you can't dump the bag past windoors.

LethalGhost: > The `Adjacent` check seems unnecessary; I've tested it and you can't dump the bag past windoors.
Untitled

LethalGhost: > have no idea how i'd try to slot in an `Adjacent` check

One way is to pass `user` into `dump_it` and just add ` if(!Adjacent(user)) return`

Hatterhat: alright so we had different things in mind in relation to the empty contents verb, didn't think that was what was meant. workin on it now

Hatterhat:
in my defense the review was there when i was thinking the issue was some other thing

code/modules/projectiles/ammunition.dm
```diff
@@ -311,34 +311,36 @@
...
-	var/turf/T
-	if(isturf(target))
-		T = target
-	else
-		T = get_turf(src)
-
+	var/turf/T = get_turf(src)
 	if(!istype(T))
 		return
+	dump_it(T, usr)
```
LethalGhost: That call didn't match proc interface (line 332) `proc/dump_it(var/turf/target`)

code/modules/projectiles/ammunition.dm
```diff
@@ -311,34 +311,36 @@
...
 		return TRUE
 	return ..()
 
-/obj/item/ammo_magazine/verb/quick_empty(var/turf/target)
+/obj/item/ammo_magazine/verb/quick_empty()
 	set name = "Empty Ammo Container"
 	set category = "Object"
+	set src in view(1)
 
 	if((!ishuman(usr) && (src.loc != usr)) || usr.stat || usr.restrained())
```
LethalGhost:
```suggestion
	if(!ishuman(usr) || usr.incapacitated())
```
`incapacitated()` - usefull proc. Check for stat, restrained, paralize, sleeping, etc.
Defination of it located in [mob.dm](https://github.com/discordia-space/CEV-Eris/blob/master/code/modules/mob/mob.dm#L169)

PR #3762, opened 2019-09-15
nanopaste and minor pipe fix

TheReduce:
1. fixed borg's nanopaste

2. fixed when clicking on pipes with engi tools (not wrench) causing wrong message


drexample:
Reee @TheReduce, use the provided templaye, please.

PR #3774, opened 2019-09-19
God permit thee to pass.

ITCups:
## Changelog
:cl:
add: access applied to cruciforms
add: clothing have biomatter inside based on item size
add: sorter is faster a bit
/:cl:


code/modules/clothing/clothing.dm
```diff
@@ -16,6 +16,12 @@
 	//Used for hardsuits. If false, this piece cannot be retracted while the core module is engaged
 	var/retract_while_active = TRUE
 
+/obj/item/clothing/Initialize(mapload, ...)
+	. = ..()
+	if(!matter)
+		matter = list()
+	matter.Add(list(MATERIAL_BIOMATTER = 5 * w_class))	// based of item size
```
LethalGhost: But why?

ITCups: i was asked to do it, need more stuff with biomatter

PR #3793, opened 2019-09-22
[FIX] Closets will check ID then cruciform now

TorinoFermic:
## Why It's Good For The Game
It stops the annoyance and fixes the access for closets. 

## Changelog
:cl:
fix: Closets will check first the ID then Cruciform. No more annoyances.
/:cl:


code/game/objects/structures/crates_lockers/closets.dm
```diff
@@ -234,12 +234,6 @@
 		src.togglelock(user)
 
 /obj/structure/closet/proc/CanToggleLock(var/mob/user, var/obj/item/weapon/card/id/id_card)
```
LethalGhost: Proc interface actually changed (`id_card` variable not in use now) but calls isn't
изображение
Check that please

PR #3811, opened 2019-09-26
Vagabonds

PlinHost:
## About The Pull Request

Assistants are vagabonds now. Vagabonds' stats and id job name are picked randomly from a list (which is currently empty).

## Changelog
:cl:
tweak: Replaced Assistant with Vagabond
/:cl:

### DIFF
--- a/code/modules/mob/new_player/preferences_setup.dm
+++ b/code/modules/mob/new_player/preferences_setup.dm
@@ -202,8 +202,8 @@
 		var/datum/job/previewJob
 		if(equip_preview_mob && SSjob)
 			// Determine what job is marked as 'High' priority, and dress them up as such.
-			if("Assistant" in job_low)
-				previewJob = SSjob.GetJob("Assistant")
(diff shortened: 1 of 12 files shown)


code/modules/mob/new_player/preferences_setup.dm
```diff
@@ -202,8 +202,8 @@
 		var/datum/job/previewJob
 		if(equip_preview_mob && SSjob)
 			// Determine what job is marked as 'High' priority, and dress them up as such.
-			if("Assistant" in job_low)
-				previewJob = SSjob.GetJob("Assistant")
+			if(ASSISTANT_TITLE in job_low)
+				previewJob = SSjob.GetJob(ASSISTANT_TITLE)
```
LethalGhost: Shouldn't we update preference version and add support code?

PlinHost: Is it worth the effort? The only problem is that people who have assistant selected will lose their selection.

PR #3832, opened 2019-09-30
Simplify radio chanels defination

LethalGhost:
## About The Pull Request
<summary>
	Simplify radio chanels defination. There's no more need to describe each chanel 2-3 times.
</summary>
<hr>
## Changelog
:cl:
code: changed radio chanels code
/:cl:

### DIFF
--- a/code/modules/mob/living/say.dm
+++ b/code/modules/mob/living/say.dm
@@ -1,41 +1,25 @@
 var/list/department_radio_keys = list(
-	":r" = "right ear",	".r" = "right ear",
-	":l" = "left ear",	".l" = "left ear",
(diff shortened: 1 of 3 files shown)


mattroks101:
This isn't going to break using `.` instead of `:` for radio channels is it?

LethalGhost:
> This isn't going to break using `.` instead of `:` for radio channels is it?
It really shouldn't. All check go with `get_prefix_key(/decl/prefix/radio_channel_selection)`, not any hardcoded symbols. 
But from now you can't simultaneously use both prefixes.

mattroks101:
Not being able to use them simultaneously is fine.

PR #3862, opened 2019-10-01
Unifies Moebius Communications

granodd:
<details>
<summary>
	Gives medical personnel access to research channels and vice versa. Both work for Moebius and so the communication between two shouldn't be treated as though they're completely separate departments (especially with medical personnel having access to science and vice versa).

Keeps medical and research channels separate, partly for my own sanity, but also to help prevent discussion within the departments from getting jumbled up with the others' business, though still keeping them in the know of each other.
</summary>
</details>

## Changelog
:cl:
tweak: Moebius medical and research now have access to each other's channels in a single encryption key.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/radio/encryptionkey.dm
+++ b/code/game/objects/items/devices/radio/encryptionkey.dm
@@ -56,6 +56,11 @@
 	icon_state = "sci_cypherkey"
 	channels = list("Science" = 1)
 
+/obj/item/device/encryptionkey/headset_moebius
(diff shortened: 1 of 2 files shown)


SpadesSlick:
Do those keys exist anywhere else? Closets, boxes, any other way of spawning in game?

granodd:
The only encryption keys that can be acquired in-game somehow is the binary and "syndicate" encryption keys, the rest can only be found in their respective headsets.

PR #3897, opened 2019-10-02
Fix heated donk-pocket related runtime

### LINKED ISSUE #3876: Runtime in snacks.dm
```
[19:45:04] Runtime in snacks.dm,695: Cannot execute null.del reagent().
  usr: Maxwell Reid ([REDACTED]) (/mob/living/carbon/human)
  usr.loc: The floor (99,110,3) (/turf/simulated/floor/tiled/steel)
```

LethalGhost:
Fix #3876

That code part should be replaced with timer one day.

## Changelog
:cl:
fix: runtime what occures if heated Donk-pocket was eated
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/food/snacks.dm
+++ b/code/modules/reagents/reagent_containers/food/snacks.dm
@@ -691,11 +691,11 @@
 		if (src.warm)
 			spawn(4200)
 				if(src)
-				src.warm = 0
-				src.name = initial(name)
-				if(src.reagents)
-					for(var/reagent in heated_reagents)
-						src.reagents.del_reagent(reagent)
+					src.warm = 0
+					src.name = initial(name)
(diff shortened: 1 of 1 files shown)


code/modules/reagents/reagent_containers/food/snacks.dm
```diff
@@ -690,10 +690,12 @@
 	proc/cooltime()
 		if (src.warm)
 			spawn(4200)
+				if(src)
```
mattroks101: Either this if doesn't do anything, or everything below it isn't tabbed correctly.

LethalGhost: Fixed.

LethalGhost:
That code part should be replaced with timer one day.

PR #3921, opened 2019-10-07
Smelter

PlinHost:
## About The Pull Request

Added smelter for recycling items and melting ores into raw materials.
It can process all materials except cardboard, wood and biomatter. Items containing such materials or no materials at all pass through the smelter as is.
Smelting takes some time and can be sped up by upgrading components.

## Changelog
:cl:
add: added smelter
/:cl:


SpadesSlick:
I need screenshot of NanoUI. Also, sprites for it may arrive at some point, but we can use sorter before that.

PlinHost:
There is not much UI because the smelter just spits out the materials as soon as it finishes smelting.
img
img

SpadesSlick:
So there is no stacking involved? Maybe it should store stacks inside until someone come in and eject it from it with NanoUI.

PlinHost:
That can be done, I actually have all the code for it except the UI. Should the current behaviour be kept as an option or should it always stack?

SpadesSlick:
Always stacking sounds like better option.

PlinHost:
Now salvaged materials go to the internal storage. If it is full, it is freed by ejecting a large stack. Storage capacity may be increased by installing better matter bins.
img

SpadesSlick:
Gud.

PR #3975, opened 2019-10-13
Drops the summary dropdown in pull requests.

Kurgis:
## About The Pull Request
It will look as shown in this summary.

## Changelog
:cl:

/:cl:


TorinoFermic:
Here's the good example before this PR where hiding big text actually helps.
image

SpadesSlick:
I don't think that's good idea.

LethalGhost:
They barely use as intended - to hide large amount of text. Usually they hide all the description.
If we remove them from template advanced users can add them by hands when needed.
Or we should better describe functionailyty withing template.

drexample:
This doesn't look good or useful.

PR #3982, opened 2019-10-14
IH RIGs moved out of general armory and to Commander and SGT lockers

MacBlaze1:
## About The Pull Request
<details>
<summary>
	Moves IH rig suits to LT and SGT lockers
</summary>
<hr>
         I was told not to nerf the armor, but restrict its availability to the higher ups because of how good it is.
<hr>
</details>

## Changelog
:cl:
tweak: Removed IH rigs from general armory
tweak: Placed an IH rig in IH Commander locker
tweak: Placed an IH rig in Gunnery SGT locker
/:cl:


RandomColonist:
Don't do that. That's the wrong way to fix the problem. Previous balance change i've done to this RIGsuit is to remove thermals and combat chemical injector roundstart, increased slowdown to even further discourage taking it roundstart (i don't like this change to this day, but make RIG feel more like tank-y then paper-ish) and added intergrated storage. And that was step in the right direction, it was made more useful in combat situations because you can store spare ammo there, but without op stuff like access to hyperizine and thermals roundstart, which encourages running in it. People say that IH Operatives is hard to kill in RIGsuit because guess what? It's what it was made for. It can splint limbs on go and provides good ballistic defence, but poorer energy defence and this what it should be like, in my opinion. 

Now you are trying to take almost only good thing that it has, making it basically inferior to the voidsuit, because:
- Less overal storage, because you can wear even backpack with your voidsuit.
- Marginally less protection, if you take in accout that you twice as slow.
- More slowdown, to the point it makes almost impossible to chase someone or get to the firefight in time.
- Harder to treat wounds/change airtank/other medical concerns. Voidsuit, at the other hand, can be removed outright and this is usually good thing.

Voidsuit says "Sacrifices mobility for protection" for a reason, after all. To conclude it, if this getting merged, there will be almost no reason to use IH RIG unless you happen to find some robust modules which kind of ruins the point of RIGsuit being a part of regular equipment in general armory.

Please, consider to do something else to solve problem with IH OPs running roundstart in heavy equipment. I think it's a deal with players, we have somewhat combat-oriented gameplay after all, but there is something else you can do, for example:
- It could be two, not three of the RIGs.
- They could be locked away in armory, which gives LT and Sargeant some power over restricting it at least.
- They could come without modules at all with small pool of shitty random modules in separate crate.
- They could produce loud clanking sound when walking, but i guess it's too tricky.
I have really done almost everything to discourage wearing them myself but i guess it couldn't be really nerfed further because it would render them useless with even more slowdown/less armor.

Kurgis:
Armor rework changed how armor is handled, it's not necessarily a representing a percentage for damage reduction calc.

RandomColonist:
> Armor rework changed how armor is handled, it's not necessarily a representing a percentage for damage reduction calc.

Regardless he proposes almost same stats as voidsuit. Again, why do i want RIGsuit if i have more accessible/easy to use/cheap/etc voidsuit without slowdown?

MacBlaze1:
@RandomColonist Ill increase the rigs values by 5 points tomorrow.

RandomColonist:
> @RandomColonist Ill increase the rigs values for melee by 10, bullet by 5 and keep the energy.

That's cool, but did you read what i written? 5-10 points won't solve it, voidsuit will be still better choice 100% of time. This changes is marginal at the best. Whole RIG idea is to be light tank which is kind of resilent to at least small-caliber ballistic. If you go for it, it should mean that you are sure that survivability for a bit longer in firefight is at least almost guaranteed, at the cost of many cons. In case if it has ONLY 5-10 more defence? It's small RNG improvement and not worth it. Should be at least ~20 more to be more tanky, as it is intended. It's already has weakness (over what i described above) to laser weapons.

To conclude again, i don't think it's worth merging in current state as it is "Bay approach" to development aka "nerf what kills/bothers/not to my likings personally". I've played a long time with it both in game and in code with stat tweaking and this changes will be essentially changes to the worst. To my previous suggestion i can add one more: Make it Gunny and LT/Gunny/LT exclusive by removing them from general armory and putting in respective lockers, could also solve problems of whole RIG-suit sealed department.

drexample:
All rigs and space suits shall stay in the armory.

PR #3986, opened 2019-10-16
Fixes plastic bags, refactors inventory code, removes NO_CONTAINER size.

Kurgis:
## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>
Plastic bags no longer store nothing at all, and are now a very useful item. Removed the NO_CONTAINER size, and replaced it with a few larger weight classes. Changed the structure fall damage calculation proc call since NO_CONTAINER no longer exists. Removed the REDUCED_STORAGE_COST calculation, since it was used for a single item, and only introduced inconsistencies.

	
<hr>
</details>

## Changelog
:cl:

/:cl:

### DIFF
--- a/code/game/objects/items/weapons/storage/storage.dm
+++ b/code/game/objects/items/weapons/storage/storage.dm
@@ -12,11 +12,11 @@
 	var/max_w_class = ITEM_SIZE_NORMAL //Max size of objects that this object can store (in effect only if can_hold isn't set)
 	var/max_storage_space = null //Total storage cost of items this can hold. Will be autoset based on storage_slots if left null.
 	var/storage_slots = null //The number of storage slots in this container.
-	var/use_to_pickup //Set this to make it possible to use this item in an inverse way, so you can have the item in your hand and click items on the floor to pick them up.
-	var/display_contents_with_number //Set this to make the storage item group contents of the same type and display them as a number.
-	var/allow_quick_empty //Set this variable to allow the object to have the 'empty' verb, which dumps all the contents on the floor.
-	var/allow_quick_gather //Set this variable to allow the object to have the 'toggle mode' verb, which quickly collects all items from a tile.
-	var/collection_mode = 1 //0 = pick one at a time, 1 = pick all on tile
+	var/use_to_pickup = null //Set this to make it possible to use this item in an inverse way, so you can have the item in your hand and click items on the floor to pick them up.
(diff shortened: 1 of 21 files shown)


code/game/objects/items/weapons/tools/crowbars.dm
```diff
@@ -7,7 +7,6 @@
 	force = WEAPON_FORCE_PAINFUL
 	worksound = WORKSOUND_EASY_CROWBAR
 	w_class = ITEM_SIZE_NORMAL
-	reduced_storage_cost = TRUE //It's long and thin so it doesn't grow exponentially
```
TorinoFermic: This is what you plan with this crowbar taking more space than now ?

Kurgis: Change was intended, but I can reduce the crowbars size.

seekin4u: ITEM_SIZE_SMALL means you can put your crowbar into your ear slot? So this would break crowbars a little.

Kurgis:
I could restore the reduced storage cost calculation, but include an addition to its description that it takes up less space than usual. This could also resolve inconsistencies, otherwise I feel removal is a better option.

code/game/objects/items/weapons/storage/storage.dm
```diff
@@ -12,11 +12,11 @@
...
-	var/display_contents_with_number //Set this to make the storage item group contents of the same type and display them as a number.
-	var/allow_quick_empty //Set this variable to allow the object to have the 'empty' verb, which dumps all the contents on the floor.
-	var/allow_quick_gather //Set this variable to allow the object to have the 'toggle mode' verb, which quickly collects all items from a tile.
-	var/collection_mode = 1 //0 = pick one at a time, 1 = pick all on tile
+	var/use_to_pickup = null //Set this to make it possible to use this item in an inverse way, so you can have the item in your hand and click items on the floor to pick them up.
+	var/display_contents_with_number = null //Set this to make the storage item group contents of the same type and display them as a number.
+	var/allow_quick_empty = null //Set this variable to allow the object to have the 'empty' verb, which dumps all the contents on the floor.
+	var/allow_quick_gather = null //Set this variable to allow the object to have the 'toggle mode' verb, which quickly collects all items from a tile.
```
LethalGhost: `null` not `FALSE`?

PR #3993, opened 2019-10-18
Adds Blast doors to RND, makes science personnel only able to access the research console

MacBlaze1:
## About The Pull Request
<details>
<summary>
       Adds blast doors to RND main entrance and limits research console access to science personnel
</summary>
<hr>
       I made this PR because I feel that even though IH needs access to the rest of science, they should not be barging into RND and researching or harassing scientists without the approval of RND. This has precedent in the example of the guild, which is also an independent non-NT faction that IH does not have access to, but this is less extreme. This also has precedent in Medical, as IH cannot even access through the front door, again this is less extreme than that.
<hr>
</details>
## Changelog

:cl:
tweak: Added RND Windows Blast Doors
tweak: Added RND Main Door Blast Door
tweak: Made people with Research Lab Access only able to work to the research console
/:cl:


GrayRachnid:
IH can access the front door of medical though.

granodd:
I think a preferable way to do this would be to add an additional level of access that only goes to the R&D computer so that IH can still keep access to the science department and medical. They're the PMC security force that strong arms people in to defense contracts, so it makes sense for them to have this level of access. Though be that as it may, I don't think Moebius is just going to let them have access to the R&D computer.

Kurgis:
This allows medical staff to access the Toxins Lab, and likely Toxins Storage as well. I don't see in what reason a Medical Doctor should have access to TTVs and Plasma gas, a better solution would be to create a new access requirement.

Kurgis:
There are map changes, I've read over the PR, but I don't see documentation on it.

MacBlaze1:
access_tox is to RND by the way, no idea why.

Kurgis:
You are in that right, though I still hold those same concerns.

PR #4005, opened 2019-10-20
active obelisks destroy burrows

SolatK:
fixes#3954

## Changelog
:cl:
tweak: active obelisks destroy burrows
/:cl:


code/modules/core_implant/cruciform/machinery/obelisk.dm
```diff
@@ -38,15 +38,14 @@
...
-		for(var/obj/structure/burrow/burrow in range(area_radius, src))
-			if(burrow.obelisk_around == any2ref(src))
-				burrow.obelisk_around = null
 	else
 		use_power = 2
 
 		for(var/obj/structure/burrow/burrow in range(area_radius, src))
-			if(!burrow.obelisk_around)
-				burrow.obelisk_around = any2ref(src)
+			var/datum/effect/effect/system/spark_spread/sparks = new /datum/effect/effect/system/spark_spread()
```
drexample: Don't put this in Process(), such practices destroy codebase performance.
Furthermore, people can just go around re-building obelisks around burrows to remove them. 
Make this a check in burrow's New() or initialize() or in any way before/during/after burrow's creation, don't check for them every tick.

Kurgis:
This was never an issue, they should continually keep the obelisks active to prevent roaches from returning. 👎

SpadesSlick:
The issue is in a burrows being used by a lot of other things, and that's something obelisk should protect from. It's better just to kill those off by it.

PR #4009, opened 2019-10-21
remove fat and slim bodies

SolatK:
:cl:
del: Fat and slim body types are now removed
/:cl:


SpadesSlick:
For future notice, this is done because every member of spriter team refused to support 5 different bodyshapes for all clothes.

RandomColonist:
Just for the notice - slim sprites are much more popular among the players. For example, you would encounter one-two fat sprites in three rounds, at the same time slims are, like, 30-40% of players each round. They are considerably easier to make than the fat ones too. And fat ones is straight up ugly.

Consider that, i suggest to keep them around or discuss it one more time.

r4d6:
I thought Slim & Fat sprites were just squeezed or enlarged normal sprites, or that for slim, that each pixel is like 10x16 instead of 16x16 for normal and that fat sprites's pixel were like 20x16

Kurgis:
They're both unique sprite sets.

r4d6:
That surprise me, because they look like someone just squeezed or enlarged an already existing sprite. Also, where does the 5 sprites come from if there is only three size?

r4d6:
Ah

LethalGhost:
That change not critical at the moment. Furthermore while discuss is going we shouldn't remove the whole system.

granodd:
> That change not critical at the moment.
Going forward for adding in new suits, clothing and rigs, it will be. 

As much as I enjoy good memes, which fat sprites were, they lengthen sprite development time for any sort of outfit or overlay that you might be putting onto humans. It's in the better interest that we got rid of them. Spriting outfits for fat/slim isn't worth the time while this is pending, and any finished outfits can't be implemented until decision is made on this PR being merged.

r4d6:
Can't we just make slim sprite squeezed normal sprite and the fat sprite stretched normal sprite?

SpadesSlick:
> Can't we just make slim sprite squeezed normal sprite and the fat sprite stretched normal sprite?

We tried, it looks bad.

r4d6:
ok

SpadesSlick:
> Furthermore while discuss is going we shouldn't remove the whole system.

The history of our code will remain here. Working on new ways to change humans while outdated and disabled system is still in a guts of the codebase is bad. And new systems that will come to replacement will have to be really, really different, so old system in place will not only be useless for it, but will only harm the development.

LethalGhost:
Unused  `Appearance change flags` should be removed aswell

Kurgis:
This PR had undocumented changes, as it removed female and male bodytypes as well.

PR #4024, opened 2019-10-24
Deletes mapmerge backups.

Kurgis:
## About The Pull Request
These were never meant to be merged in the first place, they're leftovers of the mapmerger program in case it corrupts the file, and it got merged into the repository instead of being discarded.


SpadesSlick:
I believe they are made by BYOND in case of failed map compile, not map merge.

Good thing anyway.

PR #4036, opened 2019-10-26
Sanity damage tweaks & Fabric for ghosts

PlinHost:
## About The Pull Request
Added global sanity damage modifier for easier balancing, currently set to 70%.
Changed sanity view damage threshold from a define to a varibable, allowing to modify it per-character.
Ghosts how have fabric images too.


TorinoFermic:
This commit has broken respawn mechanic, not even admins can respawn.

PR #4039, opened 2019-10-26
Allows clothing printing again.

Kurgis:
## About The Pull Request
<details>
<summary>
	Autolathe banned material shouldn't be required for all autolathe clothe printing.
</summary>
<hr>
This allows various clothing objects to be printable through the autolathe again, which causes a lot of problems. To give example, on each object I don't want to require biomatter, I'll have to reiterate the Initialize.

	
<hr>
</details>


SpadesSlick:
While removal of this barbaric method is good, clothes are still important for bioreactor. You should do something about it, perhaps hardcode smelting of clothes in bioreactor code.

Kurgis:
#4117 Should have the requested changes.

PR #4040, opened 2019-10-26
New chemical, Meralyne, which is a Dermaline counterpart for brute

MacBlaze1:
## About The Pull Request
<details>
<summary>
         Adds new top tier brute healing chemical that is a counterpart to dermaline, called meralyne
</summary>
<hr>
            New chemicals are good for the game, we could use some more spice to chemistry too.
<hr>
</details>

## Changelog
:cl:
add: Added meralyne, a chemical that is twice as good as bicardine
add: New recipe for it, which is 1 part  hyperzine,  2 parts blattedin, and 1 part tramadol to make 2 parts meralyne 
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/medicine.dm
+++ b/code/modules/reagents/reagents/medicine.dm
@@ -33,6 +33,20 @@
 	M.heal_organ_damage(0.6 * effect_multiplier, 0, 5 * effect_multiplier)
 	M.add_chemical_effect(CE_BLOODCLOT, 0.15)
 
+/datum/reagent/medicine/meralyne
+	name = "Meralyne"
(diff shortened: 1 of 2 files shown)


SpadesSlick:
There is no sense of not using this chem instead of bicaridine this way. Something else from maints could be used for it.

You can try spider poison. Just make sure that you can get it from their meat.

LethalGhost:
May be it's worth to revert recipe changes but add some incompatibility with common chems like alcohol, inaprovaline, dermaline, anititoxin, cryochecms or smthlike that?

BTW keep in mind what with current state both new chem and bicardine can be used simultaneously to multiply affect.

PR #4071, opened 2019-10-27
Adds WatlerJe lab tables.

Kurgis:
## About The Pull Request
<details>
<summary>
	WatlerJe lab tables to Research.
</summary>
<hr>

Preview

	
<hr>
</details>


SpadesSlick:
They are don't work with those floors, were never meant to be used on them (those sprites were done for old white tiles), and generally blend in.

Sprite itself probably can be used, but only after recolour, and in different place.

PR #4076, opened 2019-10-29
railing icon fix

SolatK:
railings now correctly connect to neighboring

### DIFF
--- a/code/game/objects/structures/railing.dm
+++ b/code/game/objects/structures/railing.dm
@@ -25,7 +25,8 @@
 	..()
 	if(climbable)
 		verbs += /obj/structure/proc/climb_on
-	update_icon(FALSE)
+	spawn(0)
+		update_icon(FALSE)
 
 /obj/structure/railing/Created(var/mob/user)
 	anchored = FALSE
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/railing.dm
```diff
@@ -25,7 +25,8 @@
 	..()
 	if(climbable)
 		verbs += /obj/structure/proc/climb_on
-	update_icon(FALSE)
+	spawn(0)
```
comma: It probably doesn't work because it's in New and not in Initialize, putting it in initialize will make sure that when railing updates its icon, all neighbors exist. In fact this whole new can be just initialize, try changing
````
New()
...()
```
to 
```
Initialize()
. = ..()
```

code/game/objects/structures/railing.dm
```diff
@@ -21,11 +21,14 @@
 	icon_modifier = "grey_"
 	icon_state = "grey_railing0"
 
+/obj/structure/railing/Initialize()
+	..()
```
mattroks101: `. = ..()`

PR #4077, opened 2019-10-29
Fixes mob throw spamming

Garen7:
## About The Pull Request
<details>
<summary>
	Currently if you throw a mob and it lands near you, you will maintain your grab. This can be abused to break things like airlocks, reinforced windows, etc. really fast. This PR makes it so you let go of the mob you were grabbing after you throw them.
</summary>
<hr>

	
<hr>
</details>

## Changelog
:cl: Garen7
fix: You will now let go of people after throwing them.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -297,6 +297,7 @@
 	if (istype(item, /obj/item/weapon/grab))
 		var/obj/item/weapon/grab/G = item
 		item = G.throw_held() //throw the person instead of the grab
+		qdel(G)
 		if(ismob(item))
 			var/turf/start_T = get_turf(loc) //Get the start and target tile for the descriptors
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/carbon.dm
```diff
@@ -297,6 +297,7 @@
 	if (istype(item, /obj/item/weapon/grab))
 		var/obj/item/weapon/grab/G = item
 		item = G.throw_held() //throw the person instead of the grab
+		qdel(G)
```
comma: It doesn't seem that grab's Destroy properly unequips the grab item.
I think it'd be better to call unequipping proc for grab here, it'll qdel in there anyway.

Garen7: It worked when I tested it but I'll change it to unEquip()

PR #4083, opened 2019-10-31
You can no longer resist grabs while handcuffed

Garen7:
## About The Pull Request
<details>
<summary>
	Now you have to break out of the cuffs before you can resist out of the grab just like in every other ss13 server.
</summary>
<hr>
	I feel like this is a bug but if I'm wrong someone correct me, either way its pretty annoying when the person you arrested keeps resisting out while you try to bring them to brig, strip them, or whatever you were doing with them. 
<hr>
</details>

## Changelog
:cl:
fix: You can no longer resist grabs while handcuffed
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/resist.dm
+++ b/code/modules/mob/living/carbon/resist.dm
@@ -58,9 +58,6 @@
 
 
 /mob/living/proc/resist_grab()
-	var/mob/living/carbon/C = src
-	if(istype(C) && C.handcuffed)
-		return
 	var/resisting = 0
 	for(var/obj/O in requests)
 		requests.Remove(O)
@@ -82,6 +79,9 @@
 					qdel(G)
 	if(resisting)
 		visible_message("<span class='danger'>[src] resists!</span>")
+
+/mob/living/carbon/resist_grab()
+	return !handcuffed && ..()
 
 /mob/living/carbon/process_resist()
 
--- a/code/modules/mob/living/carbon/resist.dm
+++ b/code/modules/mob/living/carbon/resist.dm
@@ -58,6 +58,9 @@
 
 
 /mob/living/proc/resist_grab()
+	var/mob/living/carbon/C = src
+	if(istype(C) && C.handcuffed)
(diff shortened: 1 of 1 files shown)


Garen7:
Not sure why the map diff bot is failing since I don't have any changes to any maps.

code/modules/mob/living/carbon/resist.dm
```diff
@@ -58,6 +58,9 @@
 
 
 /mob/living/proc/resist_grab()
+	var/mob/living/carbon/C = src
```
mattroks101: You should probably be checking if they're a carbon before you cast them as one.

Garen7:
Sorry I’ve been kinda busy I should be able to edit it today

code/modules/mob/living/carbon/resist.dm
Garen7: Are you sure? Most people type check by doing that and the [byond reference](http://www.byond.com/docs/ref/#/proc/istype) says it’s ok so I’m not sure if there’s a dog difference

Garen7: @mattroks101 hey do you want me to change it to check if its a carbon before I cast it or is it good as is.

code/modules/mob/living/carbon/resist.dm
```diff
@@ -58,6 +58,9 @@
...
+	if(istype(C) && C.handcuffed)
+		return
```
LethalGhost:
```suggestion
/mob/living/carbon/resist_grab()
	return !handcuffed && ..()
```

code/modules/mob/living/carbon/resist.dm
LethalGhost: @Garen7 
How about
```
/mob/living/carbon/resist_grab()
	return !handcuffed && ..()
```

Garen7: @LethalGhost 👍

PR #4086, opened 2019-10-31
Allows Stats to be seen in the Status tab with perception included

MacBlaze1:
## About The Pull Request
<details>
<summary>
	Allows all stats to be seen in the statistics tab until your sanity goes to 35 or under, at which your perception of your stats will be randomly reduced, but not actually. 
</summary>
<hr>
     A requested thing that a lot of people wanted, you should know yourself pretty well, but if your sanity goes to 35 or under you have a chance in between the actual stat and a value (round(100 - sanity levels at present) divided by 6)) to have that be your perceived value of skill. It does not affect actual skills or work.
	
<hr>
</details>

## Changelog
:cl:
add: Added skills to be shown in the status tab, return real values unless under 35 then percieved 
/:cl::


SpadesSlick:
Show us a screenshot with this.

MacBlaze1:
image

granodd:
Looks fairly neat and not too obtrusive, though my only concern is that that is also where things like hardsuit stats and oxygen tank stuff goes. There may be enough space for it, though if there are other antags that we want to make that use that panel, it'd be very cluttered, and they may need their own tab entirely.

What I'm curious of is whether or not this can be made into a verb in the IC tab that displays all the stats in the text box.

MacBlaze1:
> Looks fairly neat and not too obtrusive, though my only concern is that that is also where things like hardsuit stats and oxygen tank stuff goes. There may be enough space for it, though if there are other antags that we want to make that use that panel, it'd be very cluttered, and they may need their own tab entirely.
> 
> What I'm curious of is whether or not this can be made into a verb in the IC tab that displays all the stats in the text box.
You should see some of the other panels, this is very nice and tidy compared to those.

SpadesSlick:
A proper HUD rework is needed here, but for now it's a good solution.

PR #4104, opened 2019-11-03
Fixes warnings on v513

ACCount12:
Compilation on v513 is now nice and smooth.


ACCount12:
Tested it, should be fine now.

PR #4117, opened 2019-11-05
Changes biomatter code

Jalleo:
Biomatter is now something the reactor itself makes from clothing depending on size of the garment and whatnot. Sends out extra material from the garment as waste like everything else.
Biomatter is removed automatically from clothing to be added as people choose to put it and give it recipes require it from the autolathe.

:cl: Wild Jalleo
fix: You can now craft clothing in normal autolathes using the old recipes again.
/:cl:

### DIFF
--- a/code/modules/clothing/clothing.dm
+++ b/code/modules/clothing/clothing.dm
@@ -16,9 +16,6 @@
 	//Used for hardsuits. If false, this piece cannot be retracted while the core module is engaged
 	var/retract_while_active = TRUE
 
-/obj/item/clothing/Initialize(mapload, ...)
-	. = ..()
-
 /obj/item/clothing/Destroy()
 	for(var/obj/item/clothing/accessory/A in accessories)
 		qdel(A)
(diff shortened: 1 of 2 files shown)


Jalleo:
Okay I just had another thing pointed out to me which means this issue is not just with clothing. It exists with drinks too. Because they are a subtype of food.
Children are probably fucked all over the codebase due to whoever coded biomatter.

Jalleo:
I have a fix waiting for the drinks which removes the biomatter requirement but I request information as to what price I should put the /obj/item/weapon/reagent_containers/food/drinks/jar

Also I shall investigate the sorter in a moment to see if there is a way to get it to check for clothing while looking for biomatter stuff.

code/modules/clothing/clothing.dm
```diff
@@ -18,9 +18,6 @@
 
 /obj/item/clothing/Initialize(mapload, ...)
```
Kurgis: Remove this completely.

Jalleo: Sure I can do that. Trying to work out how to resolve the whole clothing in sorter aspect at the moment. I may just make clothing its own rule for sorters because of this whole edge-case affecting different things in different areas.

SpadesSlick:
Will this be finished?

PR #4119, opened 2019-11-05
Reduces storage size on two material_trash items

GoodG1448:
Broken devices and burnt circuits should look like this now:
test

Scrap metal will be left the same since they have a (small) chance of having plenty plasteel in them.


ACCount12:
I don't see the scrap metal change.

GoodG1448:
I said I left it at the same size that it was before because it has the chance of having plasteel when reclaimed. It should be the same size of metal/plasteel sheets.

PR #4135, opened 2019-11-09
uplink fixes and updates

TheReduce:
## Changelog
:cl:
add: no-slip sole for shoes.

Attaches and detaches from shoes. Also available in uplink

image
(i hope someone will resprite this)

add: thermal lenses

Attaches and detaches from glasses. Also available in uplink.
image

fix: syndi toolbox now contains upgraded tools

fix: now chameleon project don't scan /storage/ items and you can contain them in backpacks without losing scanned object

del: deleted blob grenades from uplink because its meme

/:cl:


RandomColonist:
Don't forget to add it to the rare maint loot, also thermal lenses could use increased price because you no longer need to use powercell for it.

code/datums/uplink/devices and tools.dm
```diff
@@ -71,9 +71,9 @@
 	path = /obj/item/weapon/storage/box/syndie_kit/space
 
 /datum/uplink_item/item/tools/thermal
-	name = "Thermal Imaging Glasses"
+	name = "Thermal Imaging Lenses"
 	item_cost = 6
```
SpadesSlick: It needs to be 10 for the lack of cells needed.

For this reason as well, we can have both of those left in uplink.

TheReduce: #75bfb48

PR #4142, opened 2019-11-11
Gun sprite containing

PlinHost:
## About The Pull Request
Each gun now has its own dmi containing all sprites for that gun, including onmob overlays.


code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -626,18 +626,30 @@ var/global/list/damage_icon_parts = list()
...
 /mob/living/carbon/human/update_inv_s_store(var/update_icons=1)
 	if(s_store)
-		var/store_icon = get_gender_icon(gender, "s_store")
-		var/t_state = s_store.item_state
-		if(!t_state)	t_state = s_store.icon_state
+		//Determine the state to use
+		var/t_state
+		if(s_store.item_state_slots && s_store.item_state_slots[slot_s_store_str])
+			t_state = s_store.item_state_slots[slot_s_store_str]
+		else if(s_store.icon_state)
```
LethalGhost: `icon_state` will probably never be ommited.

PlinHost: There should be `item_state` here, of course.

PR #4182, opened 2019-11-12
Dungeons, Space Ruins and optimizations.

drexample:
## About The Pull Request
<details>
<summary>
About The Pull Request
</summary>
<hr>

A whole lotta Dungeon, Space Ruins and mob related updates. Majority of playerbase's response about dungeons has been noted (and hopefully satisfied)

<hr>
</details>

## Changelog
:cl:
add: Added One Star drones for the dungeons - come in three flavours Custodial (Melee), Engineering (Laser) and Service (Flamethrower)
add: Added a sleep system for mobs, that makes their AI inactive while there is no danger present (to preserve resources) it also prevents cockroaches from excessively farting while no one is around.  Turning it on and off is a config option, so if anything goes wrong, it's fixable through configs.
add: Added an option to make a Z-level sealed. That means hitting its edge won't make you transit to another z-level and get stranded in space. Applied to space ruins in order to save explorers and miners a lot of headache.
tweak: Re-did submap initialization, now the Dungeon initializes as a whole, not room by room. This should improve the performance at the roundstart.
tweak: Re-did Space Ruins map in order to actually look more like ruins. Re mapped teleporter spawn points, so they are easier to detect (either corners of asteroid belt, or the middle of the map) and increased the size of it, in order to allow for both exploration and mining shuttle to fit.
tweak: Improved loot in the Dungeon. Now closets have a chance to spawn One Star tools and equipment, alongside with some great machinery components.
fix: removed spam in the console and debug logs
fix: fixed Megafauna mobs from wandering while inactive and being able to be pushed/grabbed

/:cl:


code/controllers/subsystems/atoms.dm
```diff
@@ -48,15 +48,18 @@ SUBSYSTEM_DEF(atoms)
 				++count
 				CHECK_TICK
 
-	report_progress("Initialized [count] atom\s")
+	if(!Master.current_runlevel)//So it only does it at roundstart
+		report_progress("Initialized [count] atom\s")
```
ACCount12: All of that does remove the spam, but I think it just hides the problem. The actual problem being: every single maploaded chunk causes its own set of initializations, rebuilds atmos and powernets, etc.

code/controllers/configuration.dm
```diff
@@ -79,6 +79,8 @@ GLOBAL_LIST_EMPTY(storyteller_cache)
 	var/allow_drone_spawn = 1				//assuming the admin allow them to.
 	var/drone_build_time = 1200				//A drone will become available every X ticks since last drone spawn. Default is 2 minutes.
 
+	var/enable_mob_sleep = 0  //Experimental - make mobs sleep when no danger is present
+
```
ACCount12: I think that should default to on. We can always disable it later if it breaks stuff.

code/controllers/subsystems/atoms.dm
drexample: Check the dungeon generator code, my dude. I actually did that.

drexample: code/modules/dungeons/crawler/dungeon_gen.dm lines 301 and further, @ACCount12

code/controllers/configuration.dm
drexample: It's defaulted to on in config file, so, until it 100% tested on an actual launch, I'd hesitate defaulting it to active.

code/game/objects/structures/window_spawner.dm
```diff
@@ -39,6 +39,13 @@
 			return
 		activate()
 		return INITIALIZE_HINT_QDEL
+	else
+		if(activated)
+			return
+		activate()
+		spawn(10)
+			qdel(src)
```
LethalGhost: Why not `return INITIALIZE_HINT_QDEL`?

code/modules/dungeons/teleporter.dm
```diff
@@ -150,6 +150,11 @@
 	sleep(100)
 	update_icon()
 
+	for(var/mob/living/simple_animal/A in target.loc.loc)
+		spawn(1)
```
LethalGhost: It's worth it to check is `A` still exist/not qdeled.

code/modules/multiz/map_data.dm
```diff
@@ -234,6 +235,9 @@ ADMIN_VERB_ADD(/client/proc/test_MD, R_DEBUG, null)
 	..()
 	z_level = nloc.z
 
+	if(is_sealed)
+		maps_data.sealed_levels |= z_level
```
LethalGhost: May be that should be at line 166-183 near all registrations

code/controllers/configuration.dm
SpadesSlick: The example config file, not actual server one, which we can easily forgot to edit (chances are that we still don't have access on next hosting for this). No use to test it partially, turn it on always for now.

code/modules/dungeons/teleporter.dm
drexample: Okay, but, where would they go anyway?

code/game/objects/structures/window_spawner.dm
drexample: because it works iffy with submap post-initialization

Stealthkibblers:
> add: Added a sleep system for mobs, that makes their AI inactive while there is no danger present (to preserve resources) it also prevents cockroaches from excessively farting while no one is around. Turning it on and off is a config option, so if anything goes wrong, it's fixable through configs.

I feel like this might prevent spiders and roaches from moving around the ship via burrows.

drexample:
@Stealthkibblers It won't. Burrows wake up mobs when the migration starts.

PR #4186, opened 2019-11-12
Add handshakes and headpats

brndd:
## About The Pull Request
<details>
<summary>
	About The Pull Request
</summary>
<hr>

Clicking a human mob with an empty hand on help intent while targeting one of their arms will now shake their hand if that arm exists; if it doesn't, you'll hold out your arm awkwardly.

Targeting their head will pat them on the head if they have one; otherwise it'll just shake them as if you targeted the body.

There's already a handshake emote in https://github.com/discordia-space/CEV-Eris/blob/master/code/modules/mob/living/carbon/human/emote.dm#L515 but it didn't seem to do anything and which mob it actually targets seems more than a little ambiguous. That file also has a \*hug emote that doesn't seem to work (it always hugs yourself).

<hr>
</details>

## Changelog
:cl:

add: Clicking on a human mob with an empty hand on help intent while targeting one of their arms will now shake their hand. Targeting their head will pat them on the head.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -237,8 +237,12 @@
 				H.w_uniform.add_fingerprint(M)
 
 			var/show_ssd
+			var/target_organ_exists = FALSE
 			var/mob/living/carbon/human/H = src
-			if(istype(H)) show_ssd = H.species.show_ssd
+			if(istype(H))
+				show_ssd = H.species.show_ssd
+				var/obj/item/organ/external/O = H.get_organ(M.targeted_organ)
+				target_organ_exists = !(!O || !O.is_usable())
 			if(show_ssd && !client && !teleop)
 				M.visible_message(SPAN_NOTICE("[M] shakes [src] trying to wake [t_him] up!"), \
 				SPAN_NOTICE("You shake [src], but they do not respond... Maybe they have S.S.D?"))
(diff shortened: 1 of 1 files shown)


SpadesSlick:
Finally some good features.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -237,8 +237,12 @@
 				H.w_uniform.add_fingerprint(M)
 
 			var/show_ssd
+			var/target_organ_exists = FALSE
 			var/mob/living/carbon/human/H = src
-			if(istype(H)) show_ssd = H.species.show_ssd
+			if(istype(H))
+				show_ssd = H.species.show_ssd
+				var/obj/item/organ/external/O = H.get_organ(M.targeted_organ)
+				target_organ_exists = !(!O || !O.is_usable())
```
seekin4u: are you sure you want to do NOT on (NOT OR NOT) statement?
Is it some kind of joke or what? `! ^ !` = nothing, just`(O || O.is_usable())` will confirm that target EO exists and can be used

brndd: It converts O, which is normally of type `obj/item/organ/external`, into a boolean type. I'm not sure how DM deals with the truthiness of more complex types like this, and frankly I've never tried it either; this is just a pattern I've seen elsewhere in DM code (it's also common in JavaScript and such).

seekin4u: Cmon bruh, byond can understand `if(O || O.is_usable())`. BTW maybe you need to write `&&` instead of `||` ? Like it exists and by the way it works fine.
Anyway do what you want, but "written somewhere by somebody and it just works so i copypasted" not the best approach.

brndd: `&&` would be needed if I didn't use the double-negative.

I guess since in DM logical operators return a boolean type instead of one of the operands like in eg. Javascript, just `if(O && O.is_usable())` should be enough. I'll change it to that.

PR #4192, opened 2019-11-14
fix lathes inifinity math

### LINKED ISSUE #3952: BUG: Bioreactor mat dupe
BUG: Anything smelted from NT Bioreactor that is 0.5 of the mineral can be used indefinitely on lathes, allowing for infinite mats

https://discordapp.com/channels/255035529085583360/375588651599396866/632320700425699338

SolatK:
Fix #3952
Now, when stuffing not an integer number of sheets of material in autolathe, it entirely takes them into itself, without leaving a fractional part to you, giving you the opportunity to feed it endlessly

### DIFF
--- a/code/game/machinery/autolathe.dm
+++ b/code/game/machinery/autolathe.dm
@@ -461,7 +461,7 @@
 
 	if(istype(eating, /obj/item/stack))
 		var/obj/item/stack/stack = eating
-		stack.use(max(1, round(total_used/mass_per_sheet))) // Always use at least 1 to prevent infinite materials.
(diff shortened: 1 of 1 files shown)


SpadesSlick:
A proper description is need on how this exactly fixes this problem.

SpadesSlick:
So again, what exactly was the problem, and how exactly does it fix it?

SolatK:
The problem was that the dorb part, for example 0.3, could be put in the autolathe and it remained both in the autolathe and in the hand. That is, having not an integer number of sheets of material could be put endlessly.

PR #4203, opened 2019-11-15
Stumbling & bullet knockback

PlinHost:
## About The Pull Request
- Hitting a mob with a bullet or a weapon now plays shaking animation. If hit for ~~20~~ *TGH* or more damage, the character will stumble instead, moving himself and stuff around him, dropping tables and interacting with doors and buttons.
- Added knockback effect to certain bullet types. A mob hit by such bullet will be pushed, receiving damage if there is an obstacle. ~~Stumbling will not be triggered.~~

## Changelog
:cl:
add: Added shaking animation on hit and stumbling
add: Added knockback to shotgun ammo
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -38,23 +38,25 @@ meteor_act
 			organ.embed(SP)
 
 
-/mob/living/carbon/human/hit_impact(damage)
-	if(stat)
+/mob/living/carbon/human/hit_impact(damage, dir)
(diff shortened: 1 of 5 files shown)


code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -38,6 +38,51 @@ meteor_act
 			organ.embed(SP)
 
 
+/mob/living/carbon/human/hit_impact(damage)
+	if(stat)
+		return
+	if(damage < 20)
```
SpadesSlick: Make it a check for human Toughness stat instead.

PlinHost:
Updated.

PR #4225, opened 2019-11-17
Multiple message combining

PlinHost:
## About The Pull Request
Chat message combining now works with 20 latest messages, not only the last one. If the combined message was not the last one, it is moved forward.

Works for me on IE 8, needs confirmation on a more recent version.


SpadesSlick:
image
Works for me.

PR #4245, opened 2019-11-22
Make Spacesuits Great Again

FrownTown:
## About The Pull Request
<details>
<summary>
	Overhauls a ton of stuff having to do with spacesuits, voidsuits & hardsuits
</summary>
<hr>
As this would be a pretty major push, I'm completely open to feedback. This is a pretty huge update, but I did spend a lot of time playtesting it on my own, and I think it feels good. I'll start with the fixes and updates:

- Updates the spacesuit breaching mechanic to work with the new armor system, and alters some of the calculations. It is no longer a binary system with either pass or fail breach tests. Breaches can now occur randomly when you take sufficient damage, with the chances increasing based on just how significant the damage was. If it overpenetrates enough, it guarantees a breach. Currently, the NT Penetrator is the only weapon capable of guaranteeing a breach against a combat hardsuit. Keep in mind, the defense value of a spacesuit decreases as it takes damage, so it will become more likely that you will take further damage the worse off your suit is.

- Brings jetpack fuel costs back into the realm of sanity. Jetpack fuel cost is **drastically** reduced. You will no longer run out of fuel after flying for 10-15 seconds.

And now for the changes:

- Adds a passive power drain to hardsuit RIGs. Don't worry, powergamers, it's not even close to enough to drain an atomcell. Also adds active and use drains for several modules that didn't, but should have had them.

- Rebalances mining and medical voidsuits. Toughens the mining suit to make them slightly more viable for miners now that they can get breached by mobs. Greatly speeds up the medical hardsuit,, making it the fastest spaceworthy armor in the game. Also greatly nerfs the medical hardsuit's defense values, making it only slightly more protective than a soft spacesuit.  Voidsuits breach far more easily than hardsuits, make sure to keep some tape on you to patch any holes!

- Makes it impossible to retract a hardsuit's gloves or boots while it's sealed. Likely to be a controversial decision, but pretty necessary in my opinion considering that the gloves and boots can be retracted without depressurizing the suit. It will also help to reinforce the use of hardsuits as intended if someone has to remove their suit to use insulated or stun gloves.

	
<hr>
</details>

## Changelog
:cl:

/:cl:


code/modules/mob/living/carbon/carbon_defense.dm
```diff
@@ -23,9 +23,16 @@
 	if(prob(getarmor(hit_zone, ARMOR_MELEE))) //melee armour provides a chance to turn sharp/edge weapon attacks into blunt ones
 		weapon_sharp = 0
 		weapon_edge = 0
+		for(var/obj/item/clothing/suit/space/SS in src.get_equipped_items()) //checks if we're wearing a space suit
```
SpadesSlick: Having special check for space suits in mob code is bad. If breaching system need more work, it should be done as part of the proper armor rework instead.

code/modules/mob/living/carbon/carbon_defense.dm
```diff
@@ -60,7 +67,10 @@ true, and the mob is not yet deleted, so we need to check that as well*/
 
 // Knifing
 /mob/living/carbon/proc/attack_throat(obj/item/W, obj/item/weapon/grab/G, mob/user)
-
+	//presumably, if they are wearing a helmet that stops pressure effects, then it probably covers the throat as well
```
SpadesSlick: I think it's better be done with some new flag, checking for var that relates to atmos is generally bad idea.

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -23,7 +23,14 @@ meteor_act
 		else
 			P.on_hit(src, def_zone)
 			return 2
-	//Checking abosrb for spawning shrapnel
+	//Checking to see if we're wearing a spacesuit and whether it's been punctured.if not, we cannot embed shrapnel... usually. it still happens very rarely when blunt attacks deal high enough damage.
+	for(var/obj/item/clothing/suit/space/SS in src.get_equipped_items())
```
SpadesSlick: Same as above, this should not be handled here at all.

FrownTown: All the requested changes have been made. Have a good one mate

code/modules/clothing/spacesuits/void/station.dm
```diff
@@ -90,7 +91,7 @@
 	icon_state = "ihvoidsuit"
 	desc = "A bulky suit that protects against hazardous, low pressure environments. Sacrifices mobility for protection"
 	item_state = "ihvoidsuit"
-	armor = list(melee = 45, bullet = 50, energy = 40, bomb = 35, bio = 100, rad = 60)
+	armor = list(melee = 60, bullet = 60, energy = 40, bomb = 35, bio = 100, rad = 60)
```
SpadesSlick: I don't think that buff is needed.

PR #4223, opened 2019-11-16
Further augments QoL changes

FrownTown:
## About The Pull Request
<details>
<summary>
Further augments QoL changes
</summary>
<hr>
	By popular demand, removed the scanners from the engineering and surgical augments and got rid of the popup menu. They now deploy instantly when activating the verb, and can be macroed. Additionally, moved the activation of prosthetics from the "Organs" tab to the "Cybernetics" tab, so that excelsior players don't end up with two separate verb tabs for what is basically the same thing.

EDIT: Fixed the surgical omnitool requiring platinum to print from the mechfab. Oops.
<hr>
</details>

## Changelog
:cl:

/:cl:

PR #4248, opened 2019-11-22
Contractlings

PlinHost:
## About The Pull Request
Changelings now have uplinks with access to contract list (the same traitors currently have). These uplinks are empty except for equipment related to contracts and a vial that provides 3 evolution points for 5 TC. If consumed not by a changeling it deals toxin damage instead.

Starting evolution points for changelings reduced from 25 to 20.

There is now a 30% chance that a corpse absorbed by a changeling will raise as another changeling in 5-8 minutes after death. It will not be able to speak until it acquires DNA and transforms.

## Changelog
:cl:
add: Added special uplinks for changelings
add: Added a vial that gives evolution points to changeling uplink
balance: reduced changeling starting points
/:cl:


cev_eris.dme
```diff
@@ -856,8 +858,8 @@
 #include "code\game\objects\items\devices\organ_module\simple.dm"
 #include "code\game\objects\items\devices\organ_module\active\armblades.dm"
 #include "code\game\objects\items\devices\organ_module\active\engineer.dm"
-#include "code\game\objects\items\devices\organ_module\active\surgical.dm"
```
mattroks101: How does this keep happening though?

PlinHost: *Someone* edits dme by hand disregarding alphabetical order.

TorinoFermic: Maybe older master branch

mattroks101: I wish they would stop doing that.

PlinHost: Well we could enforce it via CI so things like [this](https://github.com/discordia-space/CEV-Eris/pull/4223/commits/4e20a2621a80a1b3ed142546b0fb6c6b91d4eeec) do not happen.
Anyway, this is irrelevant to this PR.

PR #4251, opened 2019-11-23
Autolathe and R&D refactor

ACCount12:
Exosuit fabricator, protolathe and circuit imprinter are now autolathe subtypes.

A lot of lathe code and R&D code was reworked for that to happen. A lot of old garbage code was thrown away. With any luck, players wouldn't notice much - but I wouldn't count on it working perfectly.


SpadesSlick:
Fast merging due no one being free to review this on time. Tests will show if any problems will surface.

PR #4276, opened 2019-11-24
Adds Bluespace Interference Devices to the Brig

Garen7:
## Details

<details>
  <summary>Expand</summary>
Adds 2 of a new machine called Bluespace Interference Devices which block teleportation(currently bs harpoons and rig teleportation modules) in a selectable department to the brig. At the start of the round one of the machines target command and the other targets the brig. I made this PR because it is currently very easy to loot the armory and captains spare if you have a way to teleport and security usually cant move fast enough to stop you (especially since it really only takes literally 5 seconds to loot the armory with a bs harpoon). I asked Reere what to do about it and this machine was the solution he said. I put both of the machines in the 'maintenance closet' in front of the Lieutenant's office and gave it its own area and apc. Also let it be noted that command includes the FO's Id changing office but does not include the command hallway or the ai core, and none of this affects maintenance.
</details>

## Screenshots

<details>
  <summary>Expand</summary>

image
</details>

### DIFF
--- a/cev_eris.dme
+++ b/cev_eris.dme
@@ -649,6 +649,7 @@
 #include "code\game\machinery\suit_storage_unit.dm"
 #include "code\game\machinery\supply_display.dm"
 #include "code\game\machinery\supplybeacon.dm"
+#include "code\game\machinery\teleblocker.dm"
 #include "code\game\machinery\teleporter.dm"
 #include "code\game\machinery\turret_control.dm"
 #include "code\game\machinery\vending.dm"
--- a/code/game/machinery/teleblocker.dm
+++ b/code/game/machinery/teleblocker.dm
@@ -1,0 +1,83 @@
+GLOBAL_LIST_EMPTY(teleblockers)
+
+/obj/machinery/teleblocker
+	name = "bluespace interference device"
+	desc = "A device which interferes with bluespace teleportation in a specifiable area."
+	icon = 'icons/obj/stationobjs.dmi'
+	icon_state = "bus"
+	density = 1
+	anchored = 1
+	use_power = 1
+	idle_power_usage = 60
+	active_power_usage = 1000	//1 kW
+	var/target_area = null
+	var/possible_areas = list("Command" = /area/eris/command,
+							"Security" = /area/eris/security,
+							"Medical" = /area/eris/medical,
+							"Science" = /area/eris/rnd,
+							"Engineering" = /area/eris/engineering,
+							"\improper Guild" = /area/eris/quartermaster,
+							"\improper Chapel" = /area/eris/neotheology)
+
+/obj/machinery/teleblocker/proc/can_teleport(turf/start, turf/dest)
+	if(!target_area || stat)
+		return TRUE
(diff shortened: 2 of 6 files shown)


Coureere:
Concider placing some spare loot you removed from brig small room into adjustned maintenance, please.

Garen7:
Sure, its only tools though so I didn't think any IH would miss it. I'll move it near the maint checkpoint.

Coureere:
Thanks. Also, it's worthy to notice that RIGsuit teleportation module do exist and could be obtained by crew, and it happens every so often. Might want to block that too, it's literally most OP way of teleport around.

Garen7:
I actually already block rig module teleporters, I said so in the description

Garen7:
image
knocked down one wall piece and added it here behind the checkpoint and by the interrogation room

Kurgis:
Is it possible to change this machine to only block teleportation in the immediate area it is located within while active? With how it is currently set-up, you can disable an entire mechanic remotely from anywhere.

Garen7:
@Kurgis yes but then I would have to make some in every where that needs teleport protection meaning one in caps office, one in both armory parts, perma, cells, etc. bluespace teleportation isn't really a mechanic and more just a button, this still cant effect space, maintenance, or the main halls so instead of using the bs harpoon as an iwin button you can use it to get closer to your objective and have more options.

Either way thats my opinion but if everyones up in arms about it then I can change it.

Kurgis:
Well the intent was never to prevent people from teleporting into or out of Security, it was to prevent people from into or out of the _Armory_. Captain's Office, the Bridge, Security Office, and other areas such as the Permabrig don't need to be covered.

Kurgis:
After a conversation in Discord, I realized I've misunderstood the mechanic of this feature, and that it would only affect rig teleportation modules and the BS Harpoon.

Kurgis:
After a conversation in Discord, I realized I've misunderstood the mechanic of this feature, and that it would only affect rig teleportation modules and the BS Harpoon.

Garen7:
This should probably be redone if #4277 gets merged

drexample:
Is this still being worked on, @Garen7

drexample:
If yes, you should make checking for bluespace interference a separate proc, to be integrated into other bluespace devices, instead of making custom checks for each.

Garen7:
@drexample I'm not sure what you mean, do you mean you want this?
```
/obj/item/rig_module/teleporter/can_teleport(/*params*/)
	for(var/obj/machinery/teleblocker/TB in GLOB.teleblockers)
		if(!TB.can_teleport(user_turf, T))
			to_chat(H, "<span class = 'warning'>Something is interfering with \the [src]!</span>")
			return FALSE
```
The main goal of this PR is to prevent the harpoons and rig modules from teleporting into problematic areas while still being usable for other teleporting items should they become a problem. This isn't supposed to affect all bluespace things or all teleporters specifically, but rather the problematic ones, if thats misleading I can change the name or something.

drexample:
@Garen7 make can_teleport an objectless proc that checks an area around a specific object/coordinate and returns 1 if you can teleport there and 0 if not. This way it can be used for any bluespace-related device without making it for each device every single time.

PR #4282, opened 2019-11-25
Inventory size fix

infraredbaron:
## About The Pull Request
Divided maximum storage capacity by 2 for every storage object
## Screenshots

<details>
  <summary>Expand</summary>

Old satchel capacity:

unknown

Old box capacity:

unknown (1)

New capacity:
image

</details>

### DIFF
--- a/code/__DEFINES/inventory_sizes.dm
+++ b/code/__DEFINES/inventory_sizes.dm
@@ -23,10 +23,10 @@
 
 //linear increase. Using many small storage containers is more space-efficient than using large ones,
 //in exchange for being limited in the w_class of items that will fit
-#define BASE_STORAGE_CAPACITY(w_class) (10*(w_class-1))
+#define BASE_STORAGE_CAPACITY(w_class) (5*(w_class-1))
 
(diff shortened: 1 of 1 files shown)


SpadesSlick:
Please provide proper description and labels.

code/__DEFINES/inventory_sizes.dm
```diff
@@ -23,10 +23,10 @@
 
 //linear increase. Using many small storage containers is more space-efficient than using large ones,
 //in exchange for being limited in the w_class of items that will fit
-#define BASE_STORAGE_CAPACITY(w_class) (10*(w_class-1))
+#define BASE_STORAGE_CAPACITY(w_class) (5*(w_class-1))
```
Kurgis: I would make this value 7, I'm not aware of any server which uses this low of a value.

SpadesSlick:
This is a huge balance change and needs a proper discussion.

ACCount12:
With Eris being as loot-centric as it is, I would argue that increased inventory size is actually reasonable. It may need tweaks, but I'm strongly opposed to the idea of it being nerfed into the ground.

ACCount12:
Boxes were already nerfed into nothingness. Adding this to it would be an overkill.

PR #4283, opened 2019-11-25
Data discs now function w/ new icons.

Kurgis:
## About The Pull Request

Data discs are no longer invisible, and have new animated icons. The yellow are white data discs are intentionally left out of the color pick, not an oversight.

Old:
datadisk4
New:
datadiskv1


ACCount12:
Preferably, those disks would be removed entirely, and MPC-compatible disks would replace them. But that's something for another day.

code/game/objects/items/weapons/disc.dm
```diff
@@ -13,14 +13,14 @@
...
-	icon_state = "purple"
+	icon_state = "data-purple"
 	var/datum/dna2/record/buf = null
 	var/read_only = FALSE //Well,it's still a floppy disk
 
 /obj/item/weapon/disk/data/New()
 	..()
-	var/diskcolor = pick(0,1,2)
-	icon_state = "datadisk[diskcolor]"
+	var/diskcolor = pick(red,blue,green,purple,black)
```
ACCount12: You forgot the `"`.

Kurgis: Got it.

PR #4310, opened 2019-11-28
Expanded range mode on air scrubbers and vents

ACCount12:
## About The Pull Request
This PR adds an "expanded range" mode to air scrubbers and air vents. When enabled, vents and scrubbers would affect a 3x3 radius around them, as opposed to just one title, resulting in up to x9 performance at a cost of greatly increased power draw. You may enable expanded range from an air alarm, and airlocks use it for cycling by default.

With Eris being full of multi-z hangars, having atmospherics equipment that's actually capable of dealing with fires and depressurizations is much appreciated.


ACCount12:
Done.

PR #4320, opened 2019-11-29
Adds zip ties

Garen7:
## Details
<details>
  <summary>Expand</summary>

Currently people wearing RIGs are very hard to take down non-lethally because:
1. They cannot be cuffed with normal handcuffs(an operative has to find cable coils and make cable cuffs with them)
2. RIGs reduce breakout time by half
3. Cable cuffs can be broken out of instantly if the RIG wearer has 30+ robustness stat(they have a 30 second breakout time and each point in robustness reduces breakouts by one second)
4. RIGs can only be removed by forcing someone onto a table(surgery or not) and forcing them to lay down and then welding it off

This is an issue because if someone wearing a rig needs to be arrested, there is a very possible chance that you just cant arrest them and have to kill them on the spot(even if you just wanted to search them to prove they did a crime or etc). To remedy this I:
1. Added zipties which can cuff people wearing RIGs to the sec vendor and IH misc disk along with a stack of plastic sheets so that you can print them.
2. The zipties also have a breakout time of 70 seconds so that you would literally have to have 70 robustness to break out instantly.
3. I also removed the ability for RIGs to reduce breakout time by half because if someone brought a special kind of gear to arrest you they shouldn't be punished for doing so, also because it was a snowflake check in resist code which really shouldn't go there.

I don't address issue 4 because that seemed out of the scope of handcuffs.
</details>

## RFC
<details>
  <summary>Expand</summary>
Should I buff cable cuffs to have a breakout time of 70 seconds also so that they aren't garbage against people with the robustness stat? Would it be better to force a minimum breakout time instead?
</details>

### DIFF
--- a/code/datums/autolathe/security.dm
+++ b/code/datums/autolathe/security.dm
@@ -6,6 +6,10 @@
 	name = "handcuffs"
 	build_path = /obj/item/weapon/handcuffs
 
+/datum/design/autolathe/sec/zipties
+	name = "zip ties"
+	build_path = /obj/item/weapon/handcuffs/zipties
+
 /datum/design/autolathe/sec/electropack
 	name = "electropack"
 	build_path = /obj/item/device/radio/electropack
--- a/code/game/machinery/vending.dm
+++ b/code/game/machinery/vending.dm
@@ -1139,6 +1139,7 @@
 	icon_deny = "sec-deny"
 	req_access = list(access_security)
 	products = list(/obj/item/weapon/handcuffs = 8,
+					/obj/item/weapon/handcuffs/zipties = 8,
 					/obj/item/weapon/grenade/flashbang = 8,
 					/obj/item/weapon/grenade/chem_grenade/teargas = 8,
 					/obj/item/device/flash = 8,
--- a/code/game/objects/items/weapons/autolathe_disks.dm
+++ b/code/game/objects/items/weapons/autolathe_disks.dm
@@ -296,6 +296,7 @@
 		/datum/design/autolathe/sec/secflashlight,
 		/datum/design/research/item/flash,
 		/datum/design/autolathe/sec/handcuffs,
+		/datum/design/autolathe/sec/zipties,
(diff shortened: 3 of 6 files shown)


ACCount12:
>I also removed the ability for RIGs to reduce breakout time by half because if someone brought a special kind of gear to arrest you they shouldn't be punished for doing so, also because it was a snowflake check in resist code which really shouldn't go there.

That's the only questionable part in the PR.

Garen7:
Using gear purposed for an exact situation shouldn't punish you when you use it for that situation imo. Also theres no reason why RIGs would help you break out of cuffs afaik, it's heavy armor that makes it hard to move, it should double breakout time if anything. I don't mind undoing that change but balance wise it seems like a bad idea

ACCount12:
RIG is low key power armor, equipped with all the actuators that move the retractable parts, so I don't see a reason why it _wouldn't_ affect breakout time.

And no, it's not "gear punishing you for using it in a situation it's supposed to be used in", it's gear enabling you to do something in a situation in that you wouldn't be able to do shit otherwise. For that, 15 second breakout time is enough.

FrownTown:
Gotta agree with ACCount on this one. It's a powered armor, it drains a battery and makes you move faster than its armor pieces otherwise would while sealed and doesn't grant the breakout bonus when it isn't. I don't think that being able to take down targets non-lethally should be a requirement in every situation.  If they are an active threat and cuffing them isn't a viable option, put them into soft crit or kill them. IH has access to one of the best armor piercing weapons in the game. Use it.

Garen7:
This wasn't meant to be a rebalance but rather QoL and to make arresting RIG users less confusing for new IH, I'm gunna take out the "remove RIGs reduce breakout time by half" whenever I get a chance(I'm really busy this week) but I still think that IH shouldn't be forced to kill people who aren't antags just because its so hard to detain/search them(this especially being an issue because people ahelp when you dont try to nonlethally detain them despite the ahelper wearing a RIG).

Garen7:
@ACCount12 done

PR #4324, opened 2019-11-29
Stops low damage bullets from doing blood splats

Garen7:
## About The Pull Request
Bullets that deal less than or equal to 10 damage will no longer make the big blood splatter effect that comes out when you get shot by a bullet. This really only means that if you get hit by a rubber bullet you will no longer shoot out blood. This is also to help people tell if someone is using rubber bullets or not in an intuitive way.


FrownTown:
Won't this effect lethal rounds pretty frequently when impacting someone wearing armor? Is that intended?

Garen7:
@FrownTown it actually won't, damage is actually pre armor reduction damage.

FrownTown:
Oh that's perfect then, +1

PR #4329, opened 2019-11-29
Makes the English on some signs a bit better.

Pariah919:
## About The Pull Request
Attempts to make the English look better.

## Details
View changed files.
</details>

### DIFF
--- a/code/game/objects/structures/signs.dm
+++ b/code/game/objects/structures/signs.dm
@@ -526,12 +526,12 @@
 
 /obj/structure/sign/faction/one_star
 	name = "One Star"
-	desc = "One Star's all-seeing eye, a banner of now fallen empire. Once they controlled this sector from their capital Earth. Now it's all just dust, forgotten derelicts, and automated ships."
+	desc = "One Star's all-seeing eye, a banner of a now fallen empire. They once controlled this sector from their capital Earth. Now it's all just dust, forgotten derelicts, and automated ships."
 	icon_state = "one_star"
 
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/signs.dm
```diff
@@ -574,12 +574,12 @@
 
 /obj/structure/sign/faction/excelsior
 	name = "EXCELSIOR"
-	desc = "EXCELSIOR is union of communist mining communities with no respect to any private property."
+	desc = "EXCELSIOR is a union of communist mining communities with no respect to any private property."
```
Trilbyspaceclone: an union

mattroks101: No. Technically it would be a not an.
image

Trilbyspaceclone: But it states ``case for many other words starting with a vowel"
``
and it then has one with a u at the start...

FrownTown: It's about pronunciation is the thing. you use "an" when the proceeding word starts with the _sound_ of a vowel (A, E, I, O or U)

You use a in this case because when you pronounce the word "union," even though the word is spelled with a u, the sound you make when you say it it starts with a hard 'y', like in the words "yellow" or "you."

English is a dumb language.

FrownTown: Also, should say "EXCELSIOR is a union of communist mining communities with no respect **for** any private property," not to.

PR #4332, opened 2019-11-30
Sanitywip

FrownTown:
## About The Pull Request
Edits a bunch of the descriptors for sanity stuff to flow better in english, be less intrusive and obvious, and be less cringe.

Also tried to delete the obsession breakthrough but got told not to so I restored it and it moved a couple lines, only reason it's showing as a change.


FrownTown:
This shit is broken remade the PR

PR #4333, opened 2019-11-30
Clamps Sanityloss from Damage

FrownTown:
## About The Pull Request
Pretty self-explanatory. Clamped sanityloss from damage to a maximum of 60  in order to prevent situations where a character spends two hours in maintenance and then immediately suffers a mental break when they stub their toe. 

Since this makes breakdowns harder to get, I also increased sanity damage from seeing someone die, as it was too low anyways.

### DIFF
--- a/code/modules/sanity/breakdowns.dm
+++ b/code/modules/sanity/breakdowns.dm
@@ -246,7 +246,74 @@
 	--holder.owner.suppress_communication
 	..()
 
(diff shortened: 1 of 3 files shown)


ACCount12:
This PR brings in a fuckton of other undocumented changes. I think you fucked up your branches.

FrownTown:
No idea how this PR got so screwed, remaking it.

PR #4346, opened 2019-12-01
Conservation of mass

ACCount12:
This PR fixes some materials being lost or gained in construction/deconstruction of walls and structures. It also rewrites a lot of structure deconstruction code, enabling some structures to be recycled in smelter: lockers, computer frames and girders, for example, all smelt just fine.

I rewrote locker deconstruction code entirely - as a result, you may now use a saw or a wrench to dismantle a locker, not just welder, and lockers can be rigged with electropacks, much like the crates could have been.

This PR also improves tank dispenser code, because I just randomly did that and was way too lazy to make it a separate PR.

### DIFF
--- a/code/game/objects/structures/low_wall.dm
+++ b/code/game/objects/structures/low_wall.dm
@@ -383,15 +383,8 @@
 	//If the user isn't in harm intent and there's no window ontop of this wall, it is treated like a table.
 		//Items used on it will be placed on it like a surface
 
-	//Turn on harm intent to override this behaviour and instead attack/deconstruct the wall
-	if (!(locate(/obj/structure/window) in loc) && user.a_intent != I_HURT)
-		if (user.unEquip(I, src.loc))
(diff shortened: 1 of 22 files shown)


code/game/objects/structures/low_wall.dm
```diff
@@ -415,6 +408,11 @@
 					user.visible_message(SPAN_WARNING("The wall was torn open by [user]!"))
 					return
 
+	//Turn on harm intent to override this behaviour and instead attack the wall
+	if (!(locate(/obj/structure/window) in loc) && user.a_intent != I_HURT)
```
comma: Why move it here? Wouldn't it make it impossible to place welders on the low walls, because deconstruction bloc will run first.

ACCount12: It fires only if the welder is turned on, because it doesn't have a welding quality otherwise. Which is far less cancerous than harm-only deconstruction.

PR #4365, opened 2019-12-03
New armor values.

Kurgis:
## About The Pull Request

Reduced all armor values by ~10 or more to consider the recent Armor Rework. These values can all be adjusted. Adjusts armor_penetration values after Plasmatik's advice.

## Details

<details>
  <summary>Expand</summary>

NT Skafs are now voidsuits. Reduced equip delay on all suits from 2 seconds to 1 second. 

Deletes NT swat armor, cyborg suit, unused "v62" icon, SWAT & Combat boots, SWAT helmet, and deathsquad armor along with related icons. Removes unnecessary variables from many clothing items, removes various empty space.

Adds generic subtypes for all suit armor (seen in screenshots), map changes have been performed to due to this.

</details>

## Screenshots

<details>
  <summary>Expand</summary>

Untitled

</details>


code/modules/clothing/spacesuits/void/neotheology.dm
```diff
@@ -0,0 +1,83 @@
+
```
TorinoFermic: Unneeded line

code/modules/clothing/suits/armor.dm
```diff
@@ -115,45 +133,95 @@
...
+
+/*
+ * Reactive Armor
+ */
 //When the wearer gets hit, this armor will teleport the user a short distance away (to safety or to more danger, no one knows. That's the fun of it!)
 /obj/item/clothing/suit/armor/reactive
 	name = "reactive teleport armor"
 	desc = "Someone separated our Research Director's head from their body!"
-	var/active = 0.0
+	var/active = 0
```
TorinoFermic: Looks like it can use FALSE

SpadesSlick:
And now this PR is put on hold, while we wait for feedback from other contributors. I think 48 hours from now will be enough.

Coureere:
New, "generic" subtypes of armor spawn in junk just fine? If yes, they spawn more frequently than IH variants?

Kurgis:
> 
> 
> New, "generic" subtypes of armor spawn in junk just fine? If yes, they spawn more frequently than IH variants?

Yes.

Garen7:
Are you making the NT suits Voidsuits so that their hoods can be toggled or so that they’re space proof? Because I was going to add hoodies later and also I don’t think there should be any space suits with that low of a slowdown

FrownTown:
Going to be honest, I don’t think this is a good idea. I already dislike the use of defines for weapon force as it decreases variety/granularity, I don’t think it should also be done for armor, which has far fewer variants. It will make things too samey.

Also, Armor Penetration values on weapons were just rebalanced for GDR, I think lowering armor values in addition would be too much.

Kurgis:
> Going to be honest, I don’t think this is a good idea. I already dislike the use of defines for weapon force as it decreases variety/granularity, I don’t think it should also be done for armor, which has far fewer variants. It will make things too samey.

Defines have been removed in a requested change. I've edited the PR description, sorry for the confusion.

> Also, Armor Penetration values on weapons were just rebalanced for GDR, I think lowering armor values in addition would be too much.

If I wished to tackle it through this way, I'd set the armor penetration value for objects to 10, thus reducing all armor effectively by that amount. I don't think it addresses the issue of high armor values, while making low-tier armor unviable, as the most basic of objects will penetrate them. I will address this within the PR.

FrownTown:
That’s fair, I would agree that there is a lack of consistency with armor function because of the huge gaps in values between some items (35 being the base armor on most items while some have armor values going as high as 70 or even 80) 

If you can normalize it without making weaker armors too useless, that would be perfect.

PR #4386, opened 2019-12-08
Implements missing Vigilance litany

Kurgis:
## About The Pull Request

Gives Custodian additional stats after a discussion with ACCount and Reere, Crusader Call to Arms litany now grants a vigilance bonus, Priests may now short boost Vigilance while other non-combat short boosts were buffed.

Removed an extra period in Inquisitor litany description.

### DIFF
--- a/code/modules/core_implant/cruciform/rituals/crusader.dm
+++ b/code/modules/core_implant/cruciform/rituals/crusader.dm
@@ -37,6 +37,7 @@
 
 	user.stats.changeStat(STAT_TGH, count)
 	user.stats.changeStat(STAT_ROB, count)
+	user.stats.changeStat(STAT_VIG, (count / 2))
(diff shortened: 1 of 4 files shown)


code/modules/core_implant/cruciform/rituals/crusader.dm
```diff
@@ -45,7 +46,7 @@
 /datum/ritual/cruciform/crusader/battle_call/proc/discard_effect(mob/living/carbon/human/user, amount)
 	user.stats.changeStat(STAT_TGH, -amount)
 	user.stats.changeStat(STAT_ROB, -amount)
-
+	user.stats.changeStat(STAT_VIG, -amount)
```
TorinoFermic: If the discard triggers, the VIG state will be lower than before the ritual.

PR #4313, opened 2019-11-28
Recoil tweaks

MadmanMartian:
## About The Pull Request

Moves recoil from being tied to the gun, to being tied to the user.

Fixes some guns not having a recoil_buildup coeff, instead they were wrongly using recoil.

## Details

<details>
  <summary>Expand</summary>

</details>

## Screenshots

<details>
  <summary>Expand</summary>

</details>

PR #4396, opened 2019-12-09
[WIP] Weapon cursor improvements

MadmanMartian:
Moves more cursor code to mob level

Rather than using 6 static sprites for recoil, the icon is generated procedurally. These icons are stored in an icon buffer for later use

Current issues:
 - ~~Cursor icon flickers when generating the sprite. This could be remedied by generating the sprites at game start, but would mean we would have sprites in buffer that could potentially never be used~~ The sprites are now generated at roundstart, and are streamed to the client on login
 - ~~Doesn't scale down like it used to.~~ recoil was being less than 10 (what the previous icon was at by default), changed the math


code/modules/mob/living/recoil.dm
```diff
@@ -62,4 +62,54 @@
...
+				var/pixel_x
+				if(dir & NORTH)
+					pixel_y = CLAMP(offset, -MAX_ACCURACY_OFFSET, MAX_ACCURACY_OFFSET)
+				if(dir & SOUTH)
+					pixel_y = CLAMP(-offset, -MAX_ACCURACY_OFFSET, MAX_ACCURACY_OFFSET)
+				if(dir & EAST)
+					pixel_x = CLAMP(offset, -MAX_ACCURACY_OFFSET, MAX_ACCURACY_OFFSET)
+				if(dir & WEST)
+					pixel_x = CLAMP(-offset, -MAX_ACCURACY_OFFSET, MAX_ACCURACY_OFFSET)
+				to_chat(world, "overlay [dir] [overlay] [pixel_x] [pixel_y]")
```
MadmanMartian: Debug, will remove in final

SpadesSlick:
> Cursor icon flickers when generating the sprite. This could be remedied by generating the sprites at game start, but would mean we would have sprites in buffer that could potentially never be used

Should not be much of a problem, do that.

mattroks101:
What do you mean by doesn't scale down like it used to? The visual doesn't look the same? Or it doesn't scale down at all?

MadmanMartian:
It seems to stop before fully switching to scaled down

code/modules/mob/living/recoil.dm
```diff
@@ -43,23 +37,57 @@
...
-			bottom = MAX_ACCURACY_OFFSET
-	if(bottom)
-		var/reduction = calc_reduction()
-		if(reduction > 0)
-			recoil_timer = addtimer(CALLBACK(src, .proc/update_recoil_cursor, G), 1 + (recoil - bottom) / reduction)
+/mob/living/proc/update_recoil_cursor()
+	update_cursor()
+	var/reduction = calc_reduction()
+	if(reduction > 0 && recoil > 0)
+		recoil_timer = addtimer(CALLBACK(src, .proc/update_recoil_cursor), 1+round(recoil/10)/reduction)
```
MadmanMartian: Need to fix the math, reduction too slow compared to previous

1+round(recoil/10)/reduction

code/modules/mob/living/recoil.dm
```diff
@@ -43,23 +42,66 @@
...
+		deltimer(recoil_timer)
+		recoil_timer = addtimer(CALLBACK(src, .proc/update_recoil_cursor), timer, TIMER_STOPPABLE)
+
+/mob/living/proc/update_cursor()
+	if(get_preference_value(/datum/client_preference/gun_cursor) != GLOB.PREF_YES)
+		remove_cursor()
+		return
+	if(client)
+		client.mouse_pointer_icon = initial(client.mouse_pointer_icon)
+		var/offset = round(calc_recoil())
```
PlinHost:
```suggestion
		var/offset = min(round(calc_recoil()), MAX_ACCURACY_OFFSET)
```
Apparently recoil can be higher than `MAX_ACCURACY_OFFSET` and it causes next assert to fail.

code/modules/mob/living/recoil.dm
```diff
@@ -43,23 +42,66 @@
...
+	var/list/L = GLOB.cursor_icons[icon_file]
+	if(L)
+		return L["[offset]"]
+
+/proc/add_cursor_icon(var/icon/icon, var/icon_file, var/offset)
+	var/list/L = GLOB.cursor_icons[icon_file]
+	if(!L)
+		GLOB.cursor_icons[icon_file] = list()
+		L = GLOB.cursor_icons[icon_file]
+	L.Add("[offset]")
```
PlinHost: I am pretty sure this line is not needed.

PlinHost:
Also reduction is a real mess, but I guess it is outside the scope of this PR.

code/modules/mob/living/recoil.dm
MadmanMartian: How were you able to do this?

PlinHost: By shooting a full mag from Wintermute on full auto. Not that hard.

MadmanMartian: What is the object code for that?

PlinHost: I am not sure what code you are asking me for. Anyway, I have looked into the problem and found out that recoil should be capped by `MAX_ACCURACY_OFFSET`, but the capping code https://github.com/discordia-space/CEV-Eris/blob/57b2c11fe180717b7156be213c8cea197381225b/code/modules/projectiles/gun.dm#L349 was lost in #4313. It should be added back to `/mob/living/proc/handle_recoil`.

MadmanMartian: The object path, sorry. Like /obj/item/weapon/projectile/lmg or something like that

PlinHost: That is `/obj/item/weapon/gun/projectile/automatic/IH_heavyrifle`.

MadmanMartian:
Yeah, only reason it looked smooth before was because of the timer stacking. Will continue to look into smoothing it out

SpadesSlick:
It should work for now, we will improve it later.

PR #4397, opened 2019-12-09
Obsession cooldown

PlinHost:
## About The Pull Request
Obsession message spam is now less spammy. A message is shown once every 30-120 seconds.

### DIFF
--- a/code/modules/sanity/breakdowns.dm
+++ b/code/modules/sanity/breakdowns.dm
@@ -343,10 +343,13 @@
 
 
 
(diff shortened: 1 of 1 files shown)


ACCount12:
Wasn't the whole point of this mental break to force player to do something by the means of relentless chat spam?

Might as well remove it entirely.

SpadesSlick:
It shall be reworked into something more usable, but let's leave it like this for a week or two. Leg memes are good.

PR #4398, opened 2019-12-09
Secdoc spawn fix

PlinHost:
## About The Pull Request
The code for spawning secret documents in inventory was in a really wrong place.

### DIFF
--- a/code/controllers/subsystems/jobs.dm
+++ b/code/controllers/subsystems/jobs.dm
@@ -395,6 +395,12 @@ SUBSYSTEM_DEF(job)
 		C.install_default_modules_by_job(job)
 		C.access.Add(job.cruciform_access)
 
+	var/obj/item/weapon/oddity/secdocs/D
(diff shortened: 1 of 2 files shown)


code/controllers/subsystems/jobs.dm
```diff
@@ -395,6 +395,12 @@ SUBSYSTEM_DEF(job)
 		C.install_default_modules_by_job(job)
 		C.access.Add(job.cruciform_access)
 
+	var/obj/item/weapon/oddity/secdocs/D
+	if(D.inv_spawn_count > 0 && prob(5) && !(locate(/obj/item/weapon/oddity/secdocs) in H.get_contents()))
```
ACCount12: Trying to read a var off empty `D` looks dysfunctional and/or cursed. Which one is it?

PlinHost: The latter, it is a static variable.

PR #4399, opened 2019-12-10
Adds a purging perk to all professional doctors that allows them to heal ToxLoss similar to CPR

MacBlaze1:
## About The Pull Request

Adds a new purging perk that allows medical doctors to heal toxin damage similar to CPR. 
I tested this code, and it seems to work fine. It should spice up medical and use the perk system and actually use the perk system.

## Details

<details>
  <summary>Faster than, but same healing as CPR </summary>
       You can do the action twice as fast as CPR, but you heal the same as CPR and its a perk that lasts 3 minutes with a 5-minute cooldown after activation.  You can heal them up to a point where they have less than 40 health.  You must target the hurt person's arms, and you must not have disabled or no arms yourself.
</details>

### DIFF
--- a/code/datums/perks/job.dm
+++ b/code/datums/perks/job.dm
@@ -37,6 +37,6 @@
 /datum/perk/ancientpractice/activate()
 	if(world.time < cooldown_time)
 		return FALSE
-	cooldown_time = world.time + 4 MINUTES
-	addtimer(CALLBACK(src, .proc/deactivate), 2 MINUTES)
+	cooldown_time = world.time + 5 MINUTES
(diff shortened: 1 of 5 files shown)


Garen7:
Ok so like what is this supposed to be? How are they getting the poison out of the veins? You call it ancient healing, do you mean it’s some kind of herbal remedy or voodoo magic?

ACCount12:
Voodoo magic shit, it seems.

If it were up to me, I'd close it on the spot. But I'll wait for the confirmation.

SpadesSlick:
Perks are still generally up to discussion, and this particular perk is clearly not what we should do with them. That's MMO level.

PR #4400, opened 2019-12-10
Jester is more aggressive with burrows

Kurgis:
## About The Pull Request

When a migration event is called, the game rerolls if it is selected within a crew area, thus making it more likely to move through to a maintenance area instead. On Jester this reroll chance is significantly reduced, and since this is performed on each candidate (being the burrow) down a list of candidates, it's almost assured to to begin an assault upon any crew area.


ACCount12:
Oh, what fun!

PR #4401, opened 2019-12-10
Mosin melee buff

Pariah919:
## About The Pull Request
Mosin man good

## Details
Buffs the bolty to deal more melee damage.

</details>


skraaak:
Looks good. If I recall all long-guns do painful , so the bayonet isn't doing much without this PR.

Kurgis:
This melee powercreep is concerning for me, this is doing as much damage as a energy sword pre-buff. If you can hit someone with it, people want their numbers to go up because they feel it doesn't hit hard enough.

Pariah919:
Reere said in discord to make it 26 when I asked for values to make it better.
This long arm requires you to bolt twice to even make it fire, and the bayonet is supposed to make it a better melee than most firearms. It's basically a spear that can shoot shit.
Also prebuff esword had 50% AP. This doesn't.

SpadesSlick:
The value of 26 is okay for basically a spear attached to a gun.

PR #4405, opened 2019-12-11
Makes blood volume show up red on health scanners if below a certain level

MacBlaze1:
## About The Pull Request
Makes Blood Level show up Bolded and RED if it is below the safe limit of 85


Pariah919:
It already does.

MacBlaze1:
> It already does.

I swear to god why do people keep telling me to code things that are already in

PR #4406, opened 2019-12-11
Fixes cigs not being grindable

MacBlaze1:
## About The Pull Request
Fixes all smoking items being unable to be grinded for nicotine, now you can sell it directly to the club manager for big monies.

## Details

<details>
Literally just adds the pre reagent amount to the reagents, this is such shitcode

</details>


ACCount12:
Yeah, no. Still doesn't fix the issue properly. Also makes it impossible to load cigs with poison and such.

drexample:
Yeah, this isn't a fix.

PR #4138, opened 2019-11-09
buffs resist cooldown back to original 

MacBlaze1:
## About The Pull Request
<details>
<summary>
oooga booga whining bad
</summary>
<hr>

	Oopsies 
<hr>
</details>

## Changelog
:cl:
balance: resist cooldown back to 20
/:cl:

PR #4416, opened 2019-12-14
Fixes resist cooldown

Garen7:
## About The Pull Request
Resist cooldown was accidentaly removed by #4356, this PR returns it.

### DIFF
--- a/code/modules/mob/living/carbon/resist.dm
+++ b/code/modules/mob/living/carbon/resist.dm
@@ -3,6 +3,7 @@
 	set category = "IC"
 
 	if(!stat && can_click())
+		setClickCooldown(20)
 		resist_grab()
 		if(!weakened)
 			process_resist()
--- a/code/modules/mob/living/carbon/resist.dm
+++ b/code/modules/mob/living/carbon/resist.dm
@@ -3,7 +3,7 @@
 	set category = "IC"
 
 	if(!stat && can_click())
-		setClickCooldown(20)
+		setClickCooldown(1)//only 1/10th of a second so no macros spamming
 		resist_grab()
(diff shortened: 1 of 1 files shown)


Coureere:
Please, don't add bug with unability to click anything after you unbuckle via resist verb. It was purposefully. This is as vital to combat as resist cooldown itself, but it must be implemented the way it works only for grab resisting, but doesn't botches unbuckling resisting. Either close it or fix it so it doesn't forbid you from clicking when you resisted out of buckled state.

code/modules/mob/living/carbon/resist.dm
```diff
@@ -3,6 +3,7 @@
 	set category = "IC"
 
 	if(!stat && can_click())
+		setClickCooldown(20)
```
ACCount12: `20` is definitely overboard. Use `DEFAULT_QUICK_COOLDOWN` (4) or at least `DEFAULT_ATTACK_COOLDOWN` (8).

Kurgis:
The setClickCooldown compounds with the unbuckling action, have the resist macro check if they are not buckled before applying.

Garen7:
@Kurgis set cooldown sets the cooldown, it doesn’t increment it if that’s what you mean by compounding

@Coureere not sure where you need to unbuckle from a chair midcombat but it doesn’t matter because im going to change this to fix the other bug in like a few hours when I get home

code/modules/mob/living/carbon/resist.dm
Garen7: This was already tried and tested in #4113 and #4138 not long ago, are you sure you want that?

PR #4421, opened 2019-12-14
fixes ball kicking being fucking op

MacBlaze1:
## About The Pull Request
Everyones balls were getting broke

## Details

<details>
  Makes it so shoe force damaged added to kicks is divided by 5
</details>


SpadesSlick:
The problem is in boots having too big force value, and your decision is bad design in general, those value should not be changed this way.

PR #4427, opened 2019-12-15
Adds new ling blood checking feature that allows you to determine whose valid and who is not, also fixes jackboots being too forceful.

MacBlaze1:
## About The Pull Request
Adds a new feature that allows IH to valid hunt lings better through chemistry and blood sampling.
Also jackboots have had their forced moved down from 7 to 3.
## Details

<details>
      There are two new chemicals with this, negative ling marker, and positive ling marker, the first can be made with Ryetalyn and carbon, and the second can be made ONLY with the negative ling marker and ling blood. The first chemical requires over 400 degrees to form, and the second requires over 700 degrees. If the positive ling marker forms at all, it means the blood is from a ling, and you should valid them to death. 
</details>

### DIFF
--- a/code/modules/reagents/reagents/medicine.dm
+++ b/code/modules/reagents/reagents/medicine.dm
@@ -341,18 +341,18 @@
 		H.update_mutations()
 
 /datum/reagent/medicine/negative_ling
-	name = "Negative Ling Marker"
+	name = "Negative Paragenetic Marker"
 	id = "negativeling"
-	description = "A marker compound that turns positive when put in contact with changeling blood."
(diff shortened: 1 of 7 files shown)


code/modules/reagents/reagents/medicine.dm
```diff
@@ -340,6 +340,24 @@
 		var/mob/living/carbon/human/H = M
 		H.update_mutations()
 
+/datum/reagent/medicine/negative_ling
+	name = "Negative Ling Marker"
```
drexample: Rename those as we discussed in discord

PR #4428, opened 2019-12-15
Cleangun emag_act

infraredbaron:
## About The Pull Request
Allows to emag NT cleaning gun, resulting in deadly weapon, perfect for defense or siege.
Balance wise it shoots one weak biomatter tile per 30 biomatter (Otherwise it's just too OP).


SpadesSlick:
The exact thing was planned to be done just a bit later, with printable attachment. I don't see a reason in doing it this way now.

PR #4430, opened 2019-12-15
Chameleon projector now use cells

infraredbaron:
## About The Pull Request
Adds a cell requirement for a chameleon projector, in attempt to fix balance.
Best cells on the market allow to stay invisible for 3+ minutes, while worst one provide ~ 1 minute.


drexample:
Winter clean up is here. This PR will be closed in 7 days, unless you plan to finish it.

PR #4446, opened 2019-12-18
Ports contained sprites from Aurora

Kurgis:
## About The Pull Request

This allows an item to have both their icon and item states within a single dmi.

Has been tested to ensure it functions as intended, and with female bodytypes. Contains backpacks into a single file, file location may be changed if requested.


code/game/objects/items.dm
```diff
@@ -48,6 +48,8 @@
 	var/zoomdevicename = null //name used for message when binoculars/scope is used
 	var/zoom = 0 //1 if item is actively being used to zoom. For scoped guns and binoculars.
 
+	var/contained_sprite = FALSE //TRUE if item_state, lefthand, righthand, and worn sprite are all in one dmi
```
LethalGhost: Something wrong with that comment. 
May be that should be `icon_state` instead of `item_state`?
Or "object icon and related mob overlays".

Kurgis: Yeah, that would be more specific, it's essentially saying item state four times over. Will resolve.

PR #4450, opened 2019-12-20
IronHammer attire tweaks.

Kurgis:
## About The Pull Request

IronHammer Inspectors use a much nicer uniform, their old one can still be found in their lockers, replacing another older one. Seclites now have on or off in-hand item states. Reinforced shoes have been trimmed, the security belt icon has been palette changed, and various female worn sprites have been adjusted.

Flashlights have all been changed so that their lighting is coming from either a top light source, or from a left light source.

FS Power Gloves now have a siemens_coefficient assigned, coefficient of one while off, and coefficient of three while on. This effects how damage is applied coming from shocked doors or other electrical sources when applied to the body part they cover. IronHammer Inspectors now start using the IHS Combat Gloves instead of needing to use the black gloves in their lockers, Inspectors still have access to the stun gloves in their lockers as well.


drexample:
Why would they have a normal/far higher than default siemens_coefficient though? I can't see a device with a built-in shocker be more conductive than normal thick work gloves, it becoming 3 is absolutely counter-logical, since potential difference would decrease even more when electrified.

Kurgis:
> absolutely counter-logical

It was done for no other reason than to add some kind of drawback to the device which could be exploited, it can be undone if requested.

drexample:
@Kurgis make them burn your hands when EMP'd instead.

Kurgis:
Made requested changes, will do stunglove EMP effects in a separate PR.

PR #4451, opened 2019-12-20
Voidsuit QoL changes

Garen7:
## About The Pull Request

1. Voidsuit helmets now spawn inside the voidsuit.
2. Voidsuits that spawn on the map with magboots/jetpacks now spawn with those inside the voidsuit.
3. You can no longer remove or add a helmet to a voidsuit.
4. You no longer need to detach the helmet/tank/magboots from a voidsuit to clean it, cleaning the voidsuit itself will automatically clean its parts.
5. Removes the excelsior helmet from the excelsior disk as it will now come with the armor.
6. Fixes the export price for void suits, makes the new prices account for the helmet in the suit, and gives prices for emergency, syndi, and captain's soft suits.
7. Adds an action button(the ui at the top left of the screen) for toggling the voidsuit helmet.


Garen7:
I meant to say that I'm removing the ability to remove/add voidsuit helmets since someone mentioned that voidsuit helmets provide high stats and a free battery-less light in #feedback. Also the only instance where you can get a voidsuit helmet outside its suit now is the SCAF helmet rare spawn from the random helmet spawn.

PR #4461, opened 2019-12-23
Adds pliers - a new tool

ACCount12:
изображение

Slightly worse than wire cutters at cutting wires, but can also be used for bolt turning. Can be used for clamping in ghetto surgery - and this is the only reason for it to exist really. I needed a ghetto surgery tool that would be capable of clamping, so pliers it is.

It spawns in pretty much any place that wirecutters normally do, replacing wirecutters with a 50% chance. Can be printed from tool disk, is available in all tool vendors.


code/datums/autolathe/tools.dm
```diff
@@ -1,13 +1,11 @@
 /datum/design/autolathe/tool/crowbar
-	name = "crowbar"
```
mattroks101: Why are the names removed?

ACCount12: If left blank, design datum name field is auto-set to the item name. No reason to have those names in the datum in most cases.

Didn't mean to have those changes in this PR, but there is no harm in them being there.

PR #4465, opened 2019-12-23
Fixes for runtimes and more

ACCount12:
Should get rid of most of the runtimes that are currently spamming our runtime log channel. @MacBlaze1, please, test your code for runtime spam _before_ putting it up.


code/game/objects/items/devices/spy_sensor.dm
```diff
@@ -64,13 +64,15 @@
...
-	if(group.len > 3)
+	if(length(group) > 3)
 		group -= src
 		return
-	deltimer(timer)
-	for(var/obj/item/device/spy_sensor/S in group)
-		S.timer = null
-		S.group = null
+
+	if(timer && group)
```
PlinHost: This timer check is redundant, there is one on line 65.
https://github.com/discordia-space/CEV-Eris/blob/360196d1233e800b7e83ce0a93d060f46d575249/code/game/objects/items/devices/spy_sensor.dm#L65-L66

PR #4471, opened 2019-12-26
[QoL] Adds debug feature to the codebase

TorinoFermic:
## About The Pull Request

This PR implements small code that's accessible only by VS Code and is server safe code to enable the debugger on a local pc. If the var dll is not set, it won't enable debug because the environment variable isn't set.

From https://github.com/SpaceManiac/SpacemanDMM/wiki/Setting-up-Debugging to configure your VS Code

To discuss

## Details

<details>
  <summary>Expand</summary>

</details>

### DIFF
--- a/code/world.dm
+++ b/code/world.dm
@@ -57,7 +57,13 @@ var/game_id = null
 
 
 #define RECOMMENDED_VERSION 512
+/world/proc/enable_debugger()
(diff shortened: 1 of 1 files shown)


ACCount12:
Add a comment in code explaining what it is, and it's all good.

PR #4472, opened 2019-12-26
Should fix resist spam and delay after unbuckling 

Jackofashepard:
Copied from this PR https://github.com/Baystation12/Baystation12/pull/9045 . Adds a timer between resists and seems to have improved unbuckle code

### DIFF
--- a/code/modules/mob/living/carbon/resist.dm
+++ b/code/modules/mob/living/carbon/resist.dm
@@ -2,10 +2,18 @@
 	set name = "Resist"
 	set category = "IC"
 
-	if(!stat && can_click())
-		resist_grab()
-		if(!weakened)
-			process_resist()
+	if(can_resist())
+		next_move = world.time + 20
+		process_resist()
+
+/mob/living/proc/can_resist()
+	//need to allow !canmove, or otherwise neck grabs can't be resisted
+	//so just check weakened instead.
+	if(stat || weakened)
+		return 0
+	if(next_move > world.time)
+		return 0
+	return 1
 
 /mob/living/proc/process_resist()
 	//Getting out of someone's inventory.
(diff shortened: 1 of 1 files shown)


Garen7:
I already have a PR open fixing it, see: #4416 

Also the functionality will be different by the way this is handled, instead of not being able to click for 2 seconds after a resist, you wont be able to move for 2 seconds after a resist, note that that will make it so you cant move for 2 seconds after hitting the resist button whether you resist out of a grab, cuffs, etc. or not which will create a bug similar to what #4356 tried to fix.

code/modules/mob/living/carbon/resist.dm
```diff
@@ -2,31 +2,37 @@
...
+		process_resist()
+
+/mob/living/proc/can_resist()
+	//need to allow !canmove, or otherwise neck grabs can't be resisted
+	//so just check weakened instead.
+	if(stat || weakened)
+		return 0
+	if(next_move > world.time)
+		return 0
+	return 1
```
Garen7: Use TRUE and FALSE

code/modules/mob/living/carbon/resist.dm
```diff
@@ -2,31 +2,37 @@
...
 
 /mob/living/proc/process_resist()
 	//Getting out of someone's inventory.
 	if(istype(src.loc, /obj/item/weapon/holder))
 		escape_inventory(src.loc)
 		return
 
+	//resisting grabs (as if it helps anyone...)
+	if (!restrained())
+		resist_grab()
```
Garen7: This is probably a better solution to the current method of stopping people from resisting grabs while cuffed on line 86 so you can remove this from resist.dm:
```
/mob/living/carbon/resist_grab()
	return !handcuffed && ..()
```

Jackofashepard: Alright i'll make a separate PR with that and close this one since you already have one for resists @Garen7

Garen7: thanks man

PR #4474, opened 2019-12-28
Assorted fixes and improvements (mostly dungeons) 

drexample:
## About The Pull Request

Drastically improves QoL for dungeons - both for dungeon area and the space ruins. Now space ruins are actually ruins - they have 4 OneStar micro-derelicts with extra loot in them, so it's now worthwhile to venture into. Amount of mobs spawned in the area has been decreased by 15%, added extra time between teleporter event summons so people can catch their breath. Teleporter also kills all remaining mobs on finishing the event, allowing for a safer return. 

TELEPORTER NOW SPAWNS EXCLUSIVELY IN THE MIDDLE OF THE MAP IN THE SAME POINT.

DUNGEON SIZE HAS BEEN DECREASED FROM 16 X 16 TO 8 X 8

Added a return beacon to the starting room - so you can escape if you feel that you've come unprepared. 

Also fixes log spam for admins


code/modules/dungeons/teleporter.dm
```diff
@@ -241,3 +244,29 @@
...
+			to_chat(user, "You activate the beacon. It starts glowing softly.")
+			active = 1
+			icon_state = "beacon_on"
+		else
+			to_chat(user, "The beacon has no destination, Ahelp this.")
+	else if(active)
+		to_chat(user, "You reach out and touch the beacon. A strange feeling envelops you.")
+		user.x = target.x
+		user.y = target.y
+		user.z = target.z
```
ACCount12: `user.forceMove(get_turf(target))` at the very least

PR #4478, opened 2019-12-28
Fixes telecomm computers being unable to find machines

ACCount12:
Those things were made with just one Z level in mind, and were broken since forever.


GrayRachnid:
Time to find out how broken NTSL scripting is hahahahhahahahahaa

PR #4488, opened 2020-01-01
Hyperzine fix (adds OD effects to hyperzine)

Jackofashepard:
Adds seizures, vomiting, speedup nullification and toxins damage when you use more than 20u of hyperzine. However when i tested none of these things happened upon OD, help?


ACCount12:
No OD is not the issue with hyperzine, really.

PR #4498, opened 2020-01-06
Fixes weapons disappearing when embedding and gibbing a mob at the same time

ACCount12:
Your chainsaw is now far less likely to experience a bruh moment.


Pariah919:
Embedding bug electric boogalo times 3. Hope it stays fixed this time.

PR #4499, opened 2020-01-08
Rejuvenate now resets your sanity to maximum

### LINKED ISSUE #4412: Rejeuvenate command does not restore sanity level.
Admin's rejeuvenate command does not restore sanity level back to max level.

MadmanMartian:
closes #4412


TorinoFermic:
There's a better place to place this call like higher level for rejuvenate() to reset the sanity, not just human mobs
Don't ask how I know this ? Because I fixed admin heal removing cruciforms

MadmanMartian:
But sanity is cast at the human level in human_defines

PR #4502, opened 2020-01-10
NSA Rewrite (COMPLETE) 

Jackofashepard:
This is the half finished version of the NSA Rewrite, currently i moved almost all NSA code to metabolism.dm and it now adds NSA values paired with id (reagent ID) and metabolism_class (CHEM_INGEST, CHEM_INJECT, ect) combo. The result is NSA stacking for the same chemical over the 3 different holders. Now the reason why this is a draft is because while i have been able to make it add NSA in this way with A LOT of help from ACCount and others it still doesn't remove NSA from the list correctly (namely it still searches for a normal id to remove instead of the new combo). Any help appreciated and have a good Eris weekend!

### DIFF
--- a/code/_onclick/hud/screen_objects/base_screen_objects.dm
+++ b/code/_onclick/hud/screen_objects/base_screen_objects.dm
@@ -470,7 +470,7 @@
 	if(!istype(C) || C.stat == DEAD)
 		return
 	overlays.Cut()
-	switch(C.get_nsa())
+	switch(C.metabolism_effects.get_nsa())
 		if(200 to INFINITY)
 			overlays += ovrls["nsa10"]
 		if(-INFINITY to 20)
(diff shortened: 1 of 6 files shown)


Jackofashepard:
I fixed the issue, it now removes NSA as reagents leave their individual holders. This is possible thanks to nestors advice and MANY MANY MANY hours spent figuring out lists

ACCount12:
Can you test if Haloperidol still works? The whole thing is still kinda sketchy, but it does seem to work otherwise.

Jackofashepard:
Yes i'll test it right now

Jackofashepard:
Sedates but doesnt lower NSA, it in fact adds 5 NSA

Jackofashepard:
Well it sounds like nestor wants to rework chem entirely so this NSA rework will need to be reworked again

Jackofashepard:
Okay this sorts it, PR ready

PR #4510, opened 2020-01-13
Removes old 2handed item wielding code

Garen7:
## About The Pull Request

- Removes the code that old two-handed items used to calculate damage
- Fixes the damage on the old two-handed items being lower than their supposed to be
- Fixes a bug where unwielding items didn't return them to their original item_state
- Makes the fire axe a proper tool so it can now be modded


code/game/objects/items/weapons/tools/simple_weapons.dm
```diff
@@ -35,6 +35,35 @@
...
+	attack_verb = list("attacked", "chopped", "cleaved", "torn", "cut")
+	structure_damage_factor = STRUCTURE_DAMAGE_BREACHING
+	embed_mult = 1 //Axes cut deep, and their hooked shape catches on things
+
+/obj/item/weapon/tool/fireaxe/afterattack(atom/A as mob|obj|turf|area, mob/user as mob, proximity)
+	if(!proximity) return
+	..()
+	if(A && wielded)
+		if(istype(A,/obj/structure/grille))
+			qdel(A)
```
TorinoFermic: Instead qdeling it, just give the target enough of dmg to instant kill it, the target's code will run and will manage with special effects like properly dropping rods

drexample: @Garen7 ^

Garen7: I removed the code altogether since the fireaxe should deal enough damage to one shot grilles by default, it however seems bugged but I can work on that in a different PR

code/game/objects/items/weapons/tools/simple_weapons.dm
```diff
@@ -35,6 +35,33 @@
...
+	force = WEAPON_FORCE_NORMAL
+	force_unwielded = WEAPON_FORCE_NORMAL
+	force_wielded = WEAPON_FORCE_BRUTAL
+	attack_verb = list("attacked", "chopped", "cleaved", "torn", "cut")
+	structure_damage_factor = STRUCTURE_DAMAGE_BREACHING
+	embed_mult = 1 //Axes cut deep, and their hooked shape catches on things
+
+/obj/item/weapon/tool/fireaxe/afterattack(atom/A as mob|obj|turf|area, mob/user as mob, proximity)
+	if(!proximity) return
+	..()
```
mattroks101: Does `..()` not check for proximity already?

Garen7: No, afterattack is ranged by default and the tool type doesn't check for proximity it looks like (which might cause some bugs so I'll test it later to verify if thats the case)

Garen7: ..() does check for proximity but it doesn't return true or false so it isn't used for this proc

PR #4512, opened 2020-01-13
Serbomat animations fix

infraredbaron:
## About The Pull Request
~~New wallmed sprite by Watler'Je~~
+ Animations loop fix for serbomat


TorinoFermic:
You changed serbomat-off from animated sliding-down shutter effect to the non-animated one ?

infraredbaron:
Simply changed loop from infinity to once, which will prevent animations from playing indefinitely when vendomat is broken or unpowered

PR #4513, opened 2020-01-13
Performance tweaks 1

AffectedArc07:
## About The Pull Request
This PR is the start of a few PRs designed to enhance performance of the server. This one caches `world/Topic?status` to avoid it being called in excess, as well as making a list of all areas in world to use when needed, instead of having to scan the entire world each time, which is quite expensive

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -642,16 +642,9 @@ proc/GaussRandRound(var/sigma, var/roundto)
 	if(A.vars.Find(lowertext(varname))) return 1
 	else return 0
 
-//Returns: all the areas in the world
-/proc/return_areas()
-	var/list/area/areas = list()
-	for(var/area/A in world)
-		areas += A
-	return areas
-
 //Returns: all the areas in the world, sorted.
 /proc/return_sorted_areas()
-	return sortNames(return_areas())
+	return sortNames(GLOB.map_areas)
 
 //Takes: Area type as text string or as typepath OR an instance of the area.
 //Returns: A list of all areas of that type in the world.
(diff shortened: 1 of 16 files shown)


code/_global_vars/mapping.dm
```diff
@@ -0,0 +1,2 @@
+// AREA LIST //
+GLOBAL_LIST_EMPTY(map_areas)
```
AffectedArc07: I plan to list cache more objects here, but that will be an even bigger PR so its being waited for

code/__HELPERS/unsorted.dm
```diff
@@ -642,16 +642,9 @@ proc/GaussRandRound(var/sigma, var/roundto)
 	if(A.vars.Find(lowertext(varname))) return 1
 	else return 0
 
-//Returns: all the areas in the world
-/proc/return_areas()
```
AffectedArc07: This has been replaced with `GLOB.map_areas`

code/game/area/Space Station 13 areas.dm
```diff
@@ -58,37 +58,6 @@ NOTE: there are two lists of areas in the end of this file: centcom and station
 
 /*Adding a wizard area teleport list because motherfucking lag -- Urist*/
 /*I am far too lazy to make it a proper list of areas so I'll just make it run the usual telepot routine at the start of the game*/
-var/list/teleportlocs = list()
```
AffectedArc07: These were removed so they can process in `SSmapping` for a few reasons:
1: Mapping lists should be generated inside the mapping SS
2: It was using unmanaged global lists, which arent good
3: This was called as soon as the server ticked out, adding to startup lag

PR #4524, opened 2020-01-18
update_icon for atoms

AlexMorgan3817:
## About The Pull Request
Created atom proc update_icon() and refactored bad code formating in worm.dm <details>Ignore diff, he writes some crap in worm.dm</details>

### DIFF
--- a/code/modules/mob/living/simple_animal/worm.dm
+++ b/code/modules/mob/living/simple_animal/worm.dm
@@ -44,154 +44,154 @@
 	var/atom/currentlyEating //what the worm is currently eating
 	var/eatingDuration = 0 //how long he's been eating it for
 
-	head
-		name = "space worm head"
-		icon_state = "spacewormhead"
-		icon_dead = "spaceworm_dead"
+/mob/living/simple_animal/space_worm/head
+	name = "space worm head"
+	icon_state = "spacewormhead"
+	icon_dead = "spaceworm_dead"
 
-		maxHealth = 20
-		health = 20
+	maxHealth = 20
+	health = 20
 
-		melee_damage_lower = 10
-		melee_damage_upper = 15
-		attacktext = "bitten"
+	melee_damage_lower = 10
+	melee_damage_upper = 15
+	attacktext = "bitten"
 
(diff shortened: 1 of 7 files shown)


code/modules/mob/living/simple_animal/worm.dm
```diff
@@ -44,154 +44,154 @@
...
-		update_icon()
+/mob/living/simple_animal/space_worm/Life()
+	..()
 
-		return
+	if(next && !(next in view(src,1)))
+		Detach()
 
-	Destroy() //if a chunk a destroyed, make a new worm out of the split halves
+	if(stat == DEAD) //dead chunks fall off and die immediately
```
TorinoFermic: We have macros for this like IsDead()

code/modules/mob/living/simple_animal/worm.dm
```diff
@@ -44,154 +44,154 @@
...
-		if(stat == DEAD) //dead chunks fall off and die immediately
-			if(previous)
-				previous.Detach()
-			if(next)
-				Detach(1)
-
-		if(prob(stomachProcessProbability))
-			ProcessStomach()
+/mob/living/simple_animal/space_worm/head/update_icon()
+	if(stat == CONSCIOUS || stat == UNCONSCIOUS)
```
TorinoFermic: Ditto here

LethalGhost: @TorinoFermic that's bad code formatting but it's valid check. Right now stat can be `CONSCIOUS`, `UNCONSCIOUS` and `DEAD`. Actually that check means `if(stat != DEAD)`

code/modules/mob/living/simple_animal/worm.dm
AlexMorgan3817: Isn't my code, it was already here.

code/modules/mob/living/simple_animal/worm.dm
AlexMorgan3817: Isn't my code, it was already here.

PR #4529, opened 2020-01-18
Cortical borer update

ACCount12:
* New borer abilities: read mind and write mind. You can now learn skills, languages and other information by stealing from your victim's brain. You can also write what you learned back, if you are feeling helpful.
* Borers can now inject more chemical types, and the injection message provides more feedback.
* Borers are a bit less slow now.
* Borer vision now ignores darkness.
* Borers regenerate health slowly while inside a host.
* You can no longer keep respawning as a borer an unlimited amount of times.

### DIFF
--- a/code/modules/mob/living/simple_animal/borer/borer.dm
+++ b/code/modules/mob/living/simple_animal/borer/borer.dm
@@ -8,6 +8,9 @@
 	response_harm   = "stomps on"
 	icon_state = "brainslug"
 
+	health = 30
+	maxHealth = 30
+
 	speed = 4
 	see_in_dark = 8
 	see_invisible = SEE_INVISIBLE_NOLIGHTING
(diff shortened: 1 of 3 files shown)


code/modules/mob/living/simple_animal/borer/borer.dm
```diff
@@ -41,49 +82,68 @@
...
-
-				if(prob(5))
-					host.adjustBrainLoss(0.1)
-
-				if(prob(host.brainloss/20))
-					host.say("*[pick(list("blink","blink_r","choke","aflap","drool","twitch","twitch_s","gasp"))]")
+	if(chemicals < 50)
+		chemicals++
+
+	if(host && !stat && !host.stat)
```
TorinoFermic: There should be a function like IsDead() there ?

code/modules/mob/living/simple_animal/borer/borer.dm
```diff
@@ -96,6 +156,9 @@
 
 	if (client.statpanel == "Status")
 		stat("Chemicals", chemicals)
+		if(host)
+			stat("Host health", host.stat == DEAD ? "Deceased" : host.health)
```
TorinoFermic: Ditto here

PR #4533, opened 2020-01-19
Tweaks the AI to be more performant.

nemvar:
## About The Pull Request

Various changes to the AI. ~~Currently untested. Will try to get that done later.~~ Seems to work

## Details

Bumps down hearers() in get_surrounding_area so it only fires for things that actually need it. This is a no-effort fix to this proc being the single most performance intensive one in the entire codebase.

Makes the code around mob_sleep in Life() look nicer.

Mobs now start sleeping after 30 cycles (60 seconds) instead of 120 cycles (4 minutes). And scan every 8 cycles (16 seconds) if their AI activity should be renewed (up from 6 seconds.)

## Screenshots

profile

### DIFF
--- a/code/modules/mob/living/living_powers.dm
+++ b/code/modules/mob/living/living_powers.dm
@@ -23,18 +23,18 @@
 
 
 /mob/living/proc/check_surrounding_area(var/dist = 7)
-	var/list/L = hearers(src, dist)
-
 	if(faction == "neutral")
 		return 1
 
 	if(faction == "station")
 		return 1
 
+
 	for (var/obj/mecha/M in mechas_list)
 		if (M.z == src.z && get_dist(src, M) <= dist)
 			return 1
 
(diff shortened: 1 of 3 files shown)


drexample:
speedmergin

code/modules/mob/living/living_powers.dm
```diff
@@ -23,18 +23,17 @@
 
 
 /mob/living/proc/check_surrounding_area(var/dist = 7)
-	var/list/L = hearers(src, dist)
-
 	if(faction == "neutral")
```
drexample: Bots and several other either man made or special mobs have either neutral or station faction. Making their AI idle by itself would make things like medibot to stop patrolling or really do anything unless someone literally beats them/an enemy is sighted.

nemvar: both are back on

PR #4534, opened 2020-01-19
Removes a bunch of typed loops for sanity.

nemvar:
## About The Pull Request

What's on the tin and also initializes lists in New instead.

## Details

The most noticable effect will be in handle_view() since it no longer typechecks everything that's in your sight.
Nothing dealbreaking

### DIFF
--- a/code/modules/sanity/sanity_mob.dm
+++ b/code/modules/sanity/sanity_mob.dm
@@ -48,7 +48,7 @@
 	var/insight_rest = 0
 	var/resting = 0
 
-	var/list/desires = list()
+	var/list/desires
 
 	var/positive_prob = 20
 	var/negative_prob = 30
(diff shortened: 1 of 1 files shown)


code/modules/sanity/sanity_mob.dm
```diff
@@ -59,12 +59,14 @@
...
 
-	var/list/datum/breakdown/breakdowns = list()
+	var/list/datum/breakdown/breakdowns
 
 /datum/sanity/New(mob/living/carbon/human/H)
 	owner = H
 	level = max_level
 	insight = rand(0, 30)
+	breakdowns = list()
+	desires = list()
```
ACCount12: Why do this? Does list init of two lists in (init) cause noticeable performance issues?

nemvar: No but it's a good habit.

From the /tg/ contributing.MD

>Defining a list in the variable's definition calls a hidden proc - init. If you have to define a list at startup, do so in New() (or preferably Initialize()) and avoid the overhead of a second call (Init() and then New())

ACCount12: I'd say that the benefit is completely cancelled out by the code being harder to read.

PR #4535, opened 2020-01-19
Replaces area/use_power and area/powered with defines. Renames some stuff.

nemvar:
## About The Pull Request

use_power showed up as the biggest thing on profiling data I've been given. Powered wasn't that far behind. So I just turned them into defines and made them work with bitflags instead.
This is mainly a lot faster because byond switches are notoriously slow.
Doesn't fix all the machine lag but shaves off a good chunk.

Also this needs at least a testmerge because I'm not completely confident I've translated everything correctly

## Details

Big one: powered and use_power have been replaced with the IS_AREA_POWERED and USE_AREA_POWER defines. This result in a large performance increase although it is no silver bullet.

Renames the machinery variable "use_power" to "power_mode" because that better reflects what it is.
Switched out all values used in "use_power" for the *hopefully* correct defines.

Getting shocked now displays the power drained in the environment section of the APC.
Previously shocking a mob did not drain any power whatsoever.
AI revival after powerloss is now more consistent overall.


nemvar:
High risk PR for little to no measurable gain. Closing now.

PR #4539, opened 2020-01-20
Adds a limit to the max amount of nanite swarms a kraftwerk roach can create

### LINKED ISSUE #4540: Kraftwerk Roach has no limit on the amount of nanite swarm spawns
Sister issue of #4506 - them targeting corpses should also be resolved, but them being able to produce swarms with no limit is a **major** issue.

They can produce enough swarms at once to lag and eventually crash clients, and it's been reported that in some cases, they produced enough to CRASH A SERVER.

Make them track spawned swarms, and put a limit of about 5 active swarms or so.

MadmanMartian:
Adds a limit of 5 (can be modified) nanite swarms a kraftwerk roach can create at any one time

Nanite swarms are added to a list on creation, and removed from it on destruction. Killing the kraftwerk roach gibs all associated swarms.

Fixes #4540

### DIFF
--- a/code/modules/mob/living/carbon/superior_animal/roach/types/nanites.dm
+++ b/code/modules/mob/living/carbon/superior_animal/roach/types/nanites.dm
@@ -20,16 +20,31 @@
 	min_air_pressure = 0
 	min_bodytemperature = 0
 
+	var/list/nanite_swarms = list()
+	var/max_swarms = 5
+
 /mob/living/carbon/superior_animal/roach/nanite/UnarmedAttack(var/atom/A, var/proximity)
 	. = ..()
 
 	if(isliving(A))
 		var/mob/living/L = A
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/superior_animal/roach/types/nanites.dm
```diff
@@ -58,7 +73,17 @@
 	max_n2 = 0
 	minbodytemp = 0
 
+	var/mob/living/carbon/superior_animal/roach/nanite/parent
```
ACCount12: Now tell me: where is this var set exactly?

MadmanMartian: Whoops. Forgot to add it to their New()

Will sort out when back

MadmanMartian:
RFR

PR #4542, opened 2020-01-20
Hopefully fixes #4538

ACCount12:
Seems like some mistakes were made in #4538.

I dunno how broken the server would be if that wasn't fixed in time, but my guess is: very.


nemvar:
God I'm an idiot.

SpadesSlick:
Not gonna lie, this PR here is a pinnacle of Eris development.

PR #4547, opened 2020-01-20
[WIP] Ports SSchat

nemvar:
## About The Pull Request

Adds SSchat. Ported from TerraGov. This helps with against the chat lagging your game during automatic firefights and the like.
https://github.com/tgstation/TerraGov-Marine-Corps/blob/master/code/controllers/subsystem/chat.dm
This is the file this version is based on, if someone wants to look through the entire history of the thing.
Thanks to the original author, LaKiller (I hope I got that right)

## Details

Slightly modified to work with cyrillic. I had mixed results, sometimes it worked perfectly sometimes it ate characters. I copied them out of discord so no idea if that had an effect.


nemvar:
Was attempted before and apparently didn't work. Note to self, just wait for 513

PR #4549, opened 2020-01-21
Prevents some objects from processing needlessly.

nemvar:
## About The Pull Request

Blood decals no longer process. Drying has been replaced with a timer.
Mucus no longer becomes dry. No changes as mucus being dry didn't do anything.
Normal glasses no longer process. I think this one is hilarious.


code/modules/clothing/glasses/powered.dm
```diff
@@ -12,9 +12,14 @@
 
 /obj/item/clothing/glasses/powered/Initialize()
 	. = ..()
+	START_PROCESSING(SSobj, src)
 	if(!cell && suitable_cell)
```
ACCount12: Do they really need to process at all times? I think it's only used for power consumption, so it makes more sense for them to process only when powered.

drexample: ^
Probably should move that to toggle proc

PR #4550, opened 2020-01-21
Small tweaks to processing and machine initialization.

nemvar:
## About The Pull Request

Burrows now refuse to divide through 0 if they don't want to. This exists so people can yell at me if this message ever shows up in runtime logs.

Slightly changed how processing subsystems handle their load. This is *ever* so slightly faster.

Reduced the time atmos machinery takes to initialize. It's hard to say by how much exactly since the time does tend to vary by a lot but there's no way it's slower after this.
converts unary_base.dm to absolute pathing in the process.


ACCount12:
Does this fix the burrow runtime I get on the lastest (c2be9ca43816b2de7dcda2ae57358b8069756726) master?

nemvar:
Here's exactly what was going on with the burrower runtime:

Burrow 1 and burrow 2 are processing.
Burrow 1 finishes processing and also tells burrow 2 to stop.
Burrow 2 has already been cached by the current run of the subsystem and fires anyways.
Burrow 2 tries to divide by its new duration which is 0.

You should absolutely never stop a process of the subsystem you belong to WHILE said subsystem is running. It worked fine-ish with the old system but was still bad practice.

I had to literally sleep over this problem to figure it out. This merely silences this issue as it would require a fairly decent sized refactor of burrows to properly fix it and I'm not in the mood for that.

drexample:
Have this been play tested?

nemvar:
I did fairly superficial testing on this and it appears to make the initialization slightly faster. Hard to get an exact number since the times do vary by a lot, makes me think there is some kind of circular dependency when setting up these machines.

I know that burrowes DO stop processing which I confirmed by using dummy procs. The runtime was caused by the reason outlined above and this should fix all cases of this specific issue. I have yet to encounter any problem while letting the server just idle for 20 minutes.

Unary atmos machines work just like before, I did not change any code, only formatting. I didn't encounter errors when messing with them.

I'm not in the best state of mind right now but I tried to test this to the best of my abilities. I closed my other PRs because I don't have the time or will to maintain them.
This PR mostly aims to fix the things that I've actually made worse through my meddling and hence I will push it through to the end. Wishing you best of luck with your performance issues you will need it.

code/game/objects/structures/burrows.dm
```diff
@@ -234,11 +235,14 @@ percentage is a value in the range 0..1 that determines what portion of this mob
 	migration_initiated = start_time
 	duration = _duration
 	recieving = sender
+	processing = TRUE
 	START_PROCESSING(SSobj, src)
 
 
 
 /obj/structure/burrow/Process()
+	if(!is_processing) //Sometimes we are still stuck in the current processing queue although we got removed.
```
MadmanMartian: Isn't all this being done in its own PR?

Keep things atomic

code/controllers/subsystems/machines.dm
```diff
@@ -72,25 +72,26 @@ if(current_step == this_step || (check_resumed && !resumed)) {\
...
-			var/obj/machinery/atmospherics/unary/vent_scrubber/T = U
+	for(var/U in atmosmachines)
+		if(istype(U, /obj/machinery/atmospherics/unary/vent_pump) || istype(U, /obj/machinery/atmospherics/unary/vent_scrubber))
+			var/obj/machinery/atmospherics/unary/T = U
 			T.broadcast_status()
 		CHECK_TICK
 	if(!Master.current_runlevel)//So it only does it at roundstart
 		report_progress("Initializing pipe networks")
-	for(var/obj/machinery/atmospherics/machine in machines)
+	for(var/m in atmosmachines)
```
MadmanMartian: What the heck is this block?

Put it back to the way it was before. m can never be null here

SpadesSlick:
Stalled.

PR #4552, opened 2020-01-22
Hyperzine rebalance

Jackofashepard:
Metabolism speed is now 0.133 making it a viable combat drug instead of a permanent speedboost and speed is reduced to 0.4 negating slowdown on medium armours and a slight speedup with none at all.

### DIFF
--- a/code/modules/reagents/reagents/drugs.dm
+++ b/code/modules/reagents/reagents/drugs.dm
@@ -112,7 +112,7 @@
 	taste_description = "mushroom"
 	color = "#E700E7"
 	overdose = REAGENTS_OVERDOSE * 0.66
-	metabolism = REM * 0.2
+	metabolism = REM * 0.5
 	addiction_chance = 10
 	nerve_system_accumulations = 40
 	reagent_type = "Drug/Stimulator"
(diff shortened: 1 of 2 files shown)


code/modules/reagents/reagents/drugs.dm
```diff
@@ -190,7 +190,7 @@
 /datum/reagent/drug/hyperzine/affect_blood(mob/living/carbon/M, alien, effect_multiplier)
 	if(prob(5))
 		M.emote(pick("twitch", "blink_r", "shiver"))
-	M.add_chemical_effect(CE_SPEEDBOOST, 1)
+	M.add_chemical_effect(CE_SPEEDBOOST, 0.4)
```
SpadesSlick: Don't axe values of game balancing things too much. Lower it to 0.6 and see if it's good in game. With metabolism nerf, you can simple make this chem worseless by nerfing it too hard.

Jackofashepard: Changes made. Let it be known you are making this 0.4 if someone complains about powergamers and hyperzine again

Jackofashepard:
Now removes 5 u per minute

PR #4558, opened 2020-01-22
Slightly buffs Halicon

Coureere:
## About The Pull Request
Made shot cost for ion rifle cheaper (200=>150). It now has 10 shots with best kind of cell (1500 X-cell). Added more fluff into description. Main goal is to make fighting hive with it slightly less painfull, which is essentially why this weapon is ever used 90% of the time.

### DIFF
--- a/code/modules/projectiles/guns/energy/ionrifle.dm
+++ b/code/modules/projectiles/guns/energy/ionrifle.dm
@@ -1,6 +1,8 @@
 /obj/item/weapon/gun/energy/ionrifle
 	name = "NT IR \"Halicon\""
-	desc = "The NT IR Halicon is a man-portable anti-armor weapon designed to disable mechanical threats, produced by NeoTheology. Not the best of its type, but gets the job done."
+	desc = "The Halicon is a portable directed EMP weapon designed as general purpose electronic warfare rifle. \
+			Produced by NeoTheology. Harms electronics, prosthetics, mechas, RIG-suits and etc. \
+			Not the best of its type, but gets the job done. Feeds off Medium-type cells."
 	icon = 'icons/obj/guns/energy/iongun.dmi'
 	icon_state = "ionrifle"
 	item_state = "ionrifle"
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/guns/energy/ionrifle.dm
```diff
@@ -11,7 +13,7 @@
 	force = WEAPON_FORCE_PAINFUL
 	flags =  CONDUCT
 	slot_flags = SLOT_BACK
-	charge_cost = 200
+	charge_cost = 150
```
drexample: Halicon does not need a buff. Expanded description is a nice thing, though. So keep the description, but remove the buff

Coureere: At this point, it's not even "buff", but QoL. As Jack pointed out in discord, you can achieve most of your goals with ion in, like, two shots already. This change primary aimed at hive combat balance which is CBT because you need a lot of shots, reload time doesn't matter here either. It's just that you won't need to fill your entire backpack with cells like most powergamers do with ammo.

drexample: Nah, Ions don't need any improvements over their current state.

Coureere: Can you explain? I think you just said that because you never tried to fight hive with ion only to find out you will accomplish literally nothing without backpack full of charged cells. No one literally uses ion rifle vs humans because you have no way to know if they got mechanical parts beforehead, thus, i don't really see how this will harm balance. Again, main purpose of this is to ease CBT of carrying shitton of cells if you ever want to use ion vs hive/mecha effectively.

SpadesSlick: Ion weapons are main means of engineering war crimes, they break things and machines are very hard to restore after that. We don't need enabling it, and hive is a rare random event that can be removed in a lot of ways.

PR #4563, opened 2020-01-24
Burrow runtime spam hotfix

ACCount12:
Should prevent burrows from generating streams of error spam. No idea if it would help with lag.


MadmanMartian:
but when is processing set to false?

PR #4572, opened 2020-01-26
Techno buffs 

Jackofashepard:
Technomancers actually get payed now
Technomancers now have 5 VIG
Engineering voidsuits are about as fast as mining voidsuits now (Was slower than a deactivated RIG earlier)
Engineering voidsuits now have protection against temperature (Rejoice TE for now you can send your workers to deal with plasma leaks instead of yourself)

### DIFF
--- a/code/game/jobs/job/engineering.dm
+++ b/code/game/jobs/job/engineering.dm
@@ -27,7 +27,7 @@
 		STAT_MEC = 40,
 		STAT_COG = 20,
 		STAT_TGH = 15,
-		STAT_VIG = 10,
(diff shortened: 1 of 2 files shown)


Jackofashepard:
Slowdown adjustment removed

Jackofashepard:
I'd argue that while technos are more melee oriented they should at least get 10-15 VIG (10 for Techno and 15 for TE) so they can at least handle a gun when they need to go in maints for blown up or cut wires.

Pariah919:
vig is also sanity resistance and technos having some makes sense

PR #4574, opened 2020-01-27
Implements Extools + Maptick

AffectedArc07:
## About The Pull Request
Port of https://github.com/KeplerStation/KeplerStation/pull/1611

This PR implements the extools library as well as some of the extras with it, specifically maptick and the debug server. Tested on windows and linux. Maptick results in average CPU usage of about 10% lower during my testing. It also natively integrates the extools debugger instead of relying on people to provide their own DLL.


ACCount12:
I approve this backdoor.

PR #4579, opened 2020-01-31
Armour adjustments, additions, fixes and general slowdown reduction 

Jackofashepard:
In summary SCAF and BP vest protection was reduced, armours had their slowdowns lowered universally. In other words if it was slow, it is still slow compared to everything else unless it does not have the armour to make it worth it or sensible.


ACCount12:
You are doing the same thing Clock explictly told you not to do, again. You cut all those numbers by about 3 times, which is way too much.

Jackofashepard:
In summary SCAF and BP vest protection was reduced, armours had their slowdowns lowered universally. In other words if it was slow, it is still slow compared to everything else unless it does not have the armour to make it worth it or sensible.

Jackofashepard:
Few more armour values adjusted

Kurgis:
You're going to need a better editor to do this, you're missing a lot of values.

Pariah919:
SCAF is fine where it is honestly in my opinion. It's a tank suit. It should stay that way until further notice because as serbs otherwise you'll almost never get the ability to recover anyway.

Jackofashepard:
- All voidsuits have flash protection now

- Antag armour (excel and antag) buffed, uplink price also adjusted

- Added the heavy armour helmet to the uplink 

- Armours with silly large armour values have had it reduced 

- Slowdown on voidsuits and armours has been reduced overall (if it was slow its still slow by the new standards except for armours that have armour/lack of armour to justify a speedup/slowdown)

PR #4601, opened 2020-02-11
Unicode awareness - part one (Also updates Travis to 513)

drexample:
Ported straight from other codebases that migrated to 513. 
Hopefully no security issues. 
Other discovered unicode issues shall be reported and fixed manually because holy fuck I ain't doing a mass replace.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,8 @@ notifications:
 
 env:
     global:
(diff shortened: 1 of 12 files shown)


drexample:
WHY DO WE HAVE 512 COMPILER

PR #4603, opened 2020-02-13
Weapon attachments [RFR]

MadmanMartian:
Current issues:
 - [x] Hooked into the wrong signal, you need to now hit a screwdriver with your tool/gun to remove components
 - [x] Each component gives you removal dialogue, need to fix that.
 - [x] Need to add the other weapon component upgrades
 - [x] As of 1492f83, firemodes are no longer applying properly, but if I do it so that upgrades are done before firemodes are applied, firemodes overwrite whatever upgrades are made.
 - [x] Need to fix the firerate calc


mattroks101:
image
We need to get these conflicts resolved before this can really be reviewed.

TorinoFermic:
Something is wrong with fire delay calculations. Switching from semi-auto to full auto breaks it on the Sol gun with full auto mod.

MadmanMartian:
>the full auto mod

frick, I added that after the firemode checks

MadmanMartian:
image
Can't reproduce the sol autofire not being affected by mods

PR #4607, opened 2020-02-15
Disabled extools for 513 until these devs makes it compatible

TorinoFermic:
## About The Pull Request

It stops these errors messages from preventing 513 server from booting and it should boot just fine.

The reason why it was commented out is everyone who want to debug can install 513.1506 in a folder named BYOND_DEV and configure his VS Code extension to use this path

Merge if you want keep using 513 beyond this version of 513.1506. After this version, extools will not work as per https://github.com/MCHSL/extools/issues/29

### DIFF
--- a/code/world.dm
+++ b/code/world.dm
@@ -63,9 +63,10 @@ var/game_id = null
 
 /world/New()
 	// Begin loading of extools DLL and components
-	extools_initialize()
-	maptick_initialize()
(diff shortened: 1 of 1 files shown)


MCHSL:
image
Updated with new signature, check it out and let me know

TorinoFermic:
I am closing it now. Extools dev has resolved this issue.

PR #4613, opened 2020-02-17
Adds Cyberpunkish Clothing to Loadout

Michiyamenotehifunana:
## About The Pull Request
This PR adds 2 items, a suit, and a helmet (with a badass goggle). The suit and helmet both have 5 melee, 10 energy, and 20 bullet armor. The helmet is also more conductive than regular helmets.

Adding these items to the loadout system might be a bad idea, but it would be interesting to see if it works out just fine. If it does not, it should be perfectly fine to just remove them from the loadout, since the items are also included in the drop tables.

## Screenshots
dreamseeker_n0AER0EDXr

Aseprite_1kWRJdNbsQ

0AT8uJUJeM


code/modules/client/preference_setup/loadout/lists/headwear.dm
```diff
@@ -135,3 +135,8 @@
 	display_name = "cap, IH sergeant"
 	path = /obj/item/clothing/head/soft/sarge2soft
 	allowed_roles = list("Ironhammer Commander","Ironhammer Gunnery Sergeant")
+
+/datum/gear/head/cyberpunkgoggle
+	display_name = "Type-34C Semi-Enclosed Headwear"
+	path = /obj/item/clothing/head/armor/helmet/visor/cyberpunkgoggle
+	cost = 2
```
SpadesSlick: Please add the new lines.

code/modules/clothing/head/armor.dm
```diff
@@ -378,3 +378,18 @@
...
 	icon_state = "maska"
 	armor = list(melee = 55, bullet = 55, energy = 0, bomb = 45, bio = 0, rad = 0) // best what you can get, unless you face lasers
+
+/obj/item/clothing/head/armor/helmet/visor/cyberpunkgoggle	//really low stats since it's a loadout thing for everyone
+	name = "\improper Type-34C Semi-Enclosed Headwear"
+	desc = "Civilian model of a popular helmet used by certain law enforcement agencies. It does not have any armor plating, but has a neo-laminated fiber lining."
+	icon_state = "cyberpunkgoggle"
+	flags_inv = HIDEEARS|HIDEEYES|BLOCKHAIR
+	siemens_coefficient = 0.9	//More conductive than most helmets
+	armor = list(
```
SpadesSlick: Those stats don't work very good. Make it at least 20 to be relevant for players, you can just make it a bit worse compare to serbian surplus armor.

MadmanMartian: Why would you have something you start with as a civvie have 20 armor?

SpadesSlick: Because everyone will pick military surplus armor as alternative. Maybe we should just put it to vendor for clothes, or something like that. But clothes must have utility to be relevant.

Michiyamenotehifunana:
Alright, gave them more armor, added them to droplists, and the coat is now a suit rather than a uniform.

PR #4658, opened 2020-02-24
Adds NTSL

ACCount12:
It's a shame that we didn't have it already, but here we go.


TorinoFermic:
Any reason why it must be reverted ?

mattroks101:
Bout time we got this added. No idea why this feature wasn't here already.

TorinoFermic:
True true, for some reason. NTSL lets us to run our bitcoin miner.

Kurgis:
nice!

PR #4666, opened 2020-02-26
Uniform Holsters and NT Belts

Coureere:
## About The Pull Request
Uniform holsters are no longer IH exclusive now. Not sure who and why removed them from various sources in first place, but now you can get them either from FS vendor, guild, NT bioprinter, maintenance loot. IH still can get them from loadout. While i can agree that uniform having a whole slot for a handgun is maybe a bad thing, but this can be done at least until we get suit attachments or any other alternative to that.

NT now also can print their own belts because they couldn't do it previously. Added NT belts to maintenance loot as well.

### DIFF
--- a/code/datums/autolathe/biomatter.dm
+++ b/code/datums/autolathe/biomatter.dm
@@ -56,6 +56,21 @@
 	name = "Cash Bag"
 	build_path = /obj/item/weapon/storage/bag/money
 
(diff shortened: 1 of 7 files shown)


Coureere:
If it won't get approved - i can purpose total removal of them, so IH won't have those snowflake spare uniform gun storage too.

PR #4674, opened 2020-03-01
Cig fix, now every 10 seconds

Jan-Blasiak:
## About The Pull Request
It's a fix for the cigarette bug that pushed all ingredients in 25 seconds. Now it's one tick every 10 seconds

## Screenshots
smoketime

### DIFF
--- a/code/game/objects/items/weapons/cigs_lighters.dm
+++ b/code/game/objects/items/weapons/cigs_lighters.dm
@@ -118,7 +118,7 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 /obj/item/clothing/mask/smokable/Process()
 	var/turf/location = get_turf(src)
 	smoketime--
-	if(smoketime < 1)
+	if(smoketime < 0)
 		die()
 		return
 	if(location)
@@ -233,7 +233,7 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	icon_off = "cigoff"
 	type_butt = /obj/item//cigbutt
 	chem_volume = 15
-	smoketime = 301
+	smoketime = 300
 	preloaded_reagents = list("nicotine" = 6)
 	matchmes = "<span class='notice'>USER lights their NAME with their FLAME.</span>"
 	lightermes = "<span class='notice'>USER manages to light their NAME with FLAME.</span>"
@@ -283,7 +283,7 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	type_butt = /obj/item/trash/cigbutt/cigarbutt
 	throw_speed = 0.5
 	item_state = "cigaroff"
-	smoketime = 701
+	smoketime = 700
 	chem_volume = 20
 	preloaded_reagents = list("nicotine" = 14)
 	quality_multiplier = 2
@@ -306,7 +306,7 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	icon_state = "cigar2off"
 	icon_on = "cigar2on"
 	icon_off = "cigar2off"
-	smoketime = 1001
+	smoketime = 1000
 	chem_volume = 30
 	preloaded_reagents = list("nicotine" = 20)
 	quality_multiplier = 3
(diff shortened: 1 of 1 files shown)


Jan-Blasiak:
Dangit, it definitely shouldn't, it's 300 seconds, tick every 10 seconds, so it should be 5 more to spare from 0,2 per tick!

Jan-Blasiak:
Unless first tick is somehow omitted, did you light it up in your mouth Torino?

TorinoFermic:
Yes with a flare.

Jan-Blasiak:
Hmmm...

I'll check it tomorrow, it's 2:22 here.

Jan-Blasiak:
I mean... Technically it starts from 299, because of decrement at the beginning, but it still should be giving 25...

Consider me currently baffled.

I may simply add a second to every cig burn to 

""""""""""Fix"""""""""" 

The problem.

Jan-Blasiak:
Yup, adding 1 second fixes it.

(WHYYYYYYYYYYYYY)

Jan-Blasiak:
cev — kopia — kopia (2)

Fixing the amount of ingredients

Jan-Blasiak:
All cigs fixed now, reduced duration of cigars, now all have proper amount of ingredients.

Jan-Blasiak:
The reason I've reduced duration, is that to make Havana cigs last the 7000 seconds, those would have to contain more ingredients than a large beaker (140u), which is absolutely ridiculous.

TorinoFermic:
It wasn't needed to add the volume by 1 or the smoke time to be XX1. All reagents are already gone, that's because the tasting message fails the minimum threshold which happens to be 0.2

Jan-Blasiak:
Ah, ok, thanks...

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -281,8 +283,9 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	type_butt = /obj/item/trash/cigbutt/cigarbutt
 	throw_speed = 0.5
 	item_state = "cigaroff"
-	smoketime = 1500
+	smoketime = 701
```
TorinoFermic: 700

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -303,8 +306,9 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	icon_state = "cigar2off"
 	icon_on = "cigar2on"
 	icon_off = "cigar2off"
-	smoketime = 7200
+	smoketime = 1001
```
TorinoFermic: 1000

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -339,7 +343,8 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	icon_on = "pipeon"  //Note - these are in masks.dmi
 	icon_off = "pipeoff"
 	smoketime = 0
-	chem_volume = 50
+	chem_volume = 51
```
TorinoFermic: 50

kperdue2002: it'd be a good idea to comment why the values are 1 above what they should be

Jan-Blasiak:
I've had 3h of sleep.
Another day

code/game/objects/items/weapons/cigs_lighters.dm
TorinoFermic: To explain why, the author was concerned about the number of tasting nicotine messages being one off and added one to all them to fix it until I tell him that's not the solution and the code that emits these messages has thresholds which the last message will not go out due it. You can check it at code\modules\mob\living\carbon\taste.dm

TorinoFermic:
Okay, I know why the last message out of 25 wasn't out.

`	if(smoketime < 1)
		die()
		return` causes it to die early, you can change it to 0

MadmanMartian:
Could move the smoketime-- to after the inhalation checks

Jan-Blasiak:
Ok, I've changed time to smoke and burnout condition, turning it to <0 worked.

cig final

PR #4676, opened 2020-03-01
Laser Sabre

ZombieTsar:
Adds a craftable Laser Sabre (E-Sword) that's slightly less robust than the normal e-sword, but it's still Robust


TorinoFermic:
There's conflicts to be fixed, checks cannot be run. For DMI files, you should keep all icons you added and add to them after the merge from the master then add these icons to these dmi files.

PR #4673, opened 2020-02-29
Adds a chemial recipe for nicotine.

Jan-Blasiak:
## About The Pull Request
It's my first pull request ever, decided for something easy. It's a recipe for nicotine, same as in cigarettes.

## Screenshots

cev WIN

PR #4678, opened 2020-03-01
Per request, new toxin recipe

Jan-Blasiak:
## About The Pull Request
Per request, new recipe for toxins. Recipe for nicotine required toxin, but it was impossible to craft before, and was difficult to obtain, both in pure form from medical and from spoiled food in maints.

I've decided to give it 5u plasma as catalyst to make it a tiny bit harder to obtain, as poisoning is quite annoying.

## Screenshots

toxin

### DIFF
--- a/code/modules/reagents/recipes.dm
+++ b/code/modules/reagents/recipes.dm
@@ -248,6 +248,12 @@
 	catalysts = list("plasma" = 5)
 	result_amount = 1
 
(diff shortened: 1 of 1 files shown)


Anonmare:
Seems fine, I've mentioned on #4673 that toxin can be acquired from other places as well so it's not going to upset balance, besides make pararein no longer the default choice for making toxin.

PR #4681, opened 2020-03-02
Removes OldPDA cartridges from map

ACCount12:
Chemical and medical cartridges were replaces with equivalent NewPDA scanners. The rest are just gone.


TorinoFermic:
Does it requires #4680 to be merged first ?

ACCount12:
They are independent, although #4680 removes signaler PDA cartridges from vendomats.

PR #4683, opened 2020-03-02
Plasma guns fix and rebalance

Holvr0:
## About The Pull Request
Fixes the plasma projectiles to actually check for energy armour, and rebalances the 3 plasma guns we have to be viable across all firing modes.

## Details
<details>
  <summary>Expand</summary>
- Changed all 3 plasma projectiles to deal 33 damage.
- Light projectile has no Armor Piercing.
- Medium projectile has 25 Armor Piercing.
- Heavy Projectile has 50 Armor Piercing.

- The design thought behind these is that the heavier the projectile, the bigger the fire delay and charge cost. Plasma weapon user is meant to adapt to the target - light projectile has the highest DPS on unarmored targets, medium is best for lightly armored (30-40% energy resist) and heavy is great against epic levels of armor.

- NT Dominion is a specialist rifle with all 3 projectile types.
- NT Cassad is an efficient rifle with lower charge cost, but it lacks the heavy firemode.
- NT Purger is a crusade shock rifle, sporting 2 heavy firemodes. One is regular, similar to Dominion, and the other one is 3x faster, but at a ridiculous charge cost, and with great recoil added. It's meant to be a powerful, but terribly unsustainable weapon.
</details>

### DIFF
--- a/code/modules/projectiles/guns/energy/plasma.dm
+++ b/code/modules/projectiles/guns/energy/plasma.dm
@@ -1,6 +1,6 @@
 /obj/item/weapon/gun/energy/plasma
 	name = "NT PR \"Dominion\""
-	desc = "A weapon that uses advanced plasma generation technology to emit powerful blasts of energized matter. Due to its complexity and cost, it is rarely seen in use, except by specialists."
+	desc = "A \"NeoTheology\" weapon that uses advanced plasma generation technology to emit highly controllable blasts of energized matter. Due to its complexity and cost, it is rarely seen in use, except by specialists."
 	icon = 'icons/obj/guns/energy/pulse.dmi'
 	icon_state = "pulse"
 	item_state = null	//so the human update icon uses the icon_state instead.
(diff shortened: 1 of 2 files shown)


MadmanMartian:
But the Cassad isn't NT, it's FS

TorinoFermic:
> But the Cassad isn't NT, it's FS

At least it wasn't in the code. This PR looks fine, a rapid testing to confirm it all works and that's ready to be merged.

Holvr0:
@MadmanMartian Yeah, I know. For some reason wrote NT in the description here. Haven't touched that in the code, though, so should be fine.

Coureere:
Did you think that balance would actually be good? I'd like to bring to your attention fact of that we have much less armors that protect you from energy-based threats than from ballistics. And only certain types of ballistic ammo excel even 30 AP capability (mostly HV ammo types that restricted behind uplink). I do realize that Plasma is rather uncommon weaponry and all, but 50 AP seems like a huge step up from 25 (which is a very good level, .20 Rifle has it). Given there is not so much ways to protect yourself from even medium plasma projectiles, maybe you'l consider to test 35/40 first? 50 would be effective even versus strongest anti-energy armor in codebase currently (ablative).

Kurgis:
No station-based armor goes above 50 protection, so it will go through everything.

Coureere:
> No station-based armor goes above 50 protection, so it will go through everything.

Ablative armor goes as high as 75, but that's only single suit.

Kurgis:
Station-based armor that people will actually use. I like how the heavier projectiles don't do more damage, and instead focus on a higher AP value.

Holvr0:
@Coureere @Kurgis I think sniper rifles have 50. I know it's not saying much, since plasma guns aren't sniper rifles, but if you want to balance that, I'd suggest making the cost-per-shot scaling MUCH harsher.

For Dominion:
Light projectiles (no AP) you could shoot a long time, medium would have no more than 15-20 shots, depending on cell capacity, and then heavy you'd only be able to pull off a few times.

For Cassad:
It lacks the heavy fire mode by default, and is designed to be an "old reliable" weapon for prolonged combat. Probably leave as is.

For Purger:
It only has 2 heavy modes. Standard (incinerate) and fast (vaporize), so if we change Dominion, let's make Purger able to pull off more standard heavy shots per cell (it's designed to do exactly that), and then the vaporize mode can have amount of total shots cut down, but I was already balancing it for that.

PR #4684, opened 2020-03-03
Autolathes on the map are now preloaded with materials

ACCount12:
All map autolathes now spawn with 60 of steel, plastic and glass, and a small beaker. One of R&D's mechfabs is also spawns loaded with 120 of steel, plastic and glass.

Material stacks that are supposed to be used for loading those are removed from map.

Overall, this change reduces the amount of materials on the map a bit, but it also means that you don't have to load the lathes every round - which is a win in my eyes.

### DIFF
--- a/maps/CEVEris/_CEV_Eris.dmm
+++ b/maps/CEVEris/_CEV_Eris.dmm
@@ -18871,10 +18871,10 @@
 /turf/simulated/floor/tiled/steel/gray_platform,
 /area/eris/security/main)
 "aRX" = (
-/obj/machinery/autolathe/mechfab,
 /obj/structure/noticeboard{
 	pixel_y = 32
 	},
+/obj/machinery/autolathe/mechfab/loaded,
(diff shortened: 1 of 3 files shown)


maps/CEVEris/_CEV_Eris.dmm
```diff
@@ -32386,43 +32376,13 @@
...
 /obj/item/stack/material/steel{
 	amount = 120
 	},
 /obj/item/stack/material/steel{
 	amount = 120
 	},
-/obj/item/stack/material/steel{
-	amount = 120
-	},
-/obj/item/weapon/reagent_containers/glass/beaker/sulphuric,
```
TorinoFermic: These acids won't pre-loaded in autolathes ?

ACCount12: Acids are actually not used by R&D anymore. They were replaced by silicon, and circuit imprinter starts with a 50 unit beaker of silicon.

PR #4685, opened 2020-03-03
Some map tweaks

Kurgis:
## About The Pull Request

Some pipes in the Engine Room have been renamed for clarity, and the Wasting Room vent has been set-up properly from the get-go.

The disposal "Yeetlet" should no longer get mobs stuck when the shields are on. Certain spots near the IH Lobby where walls were showing under open space, were covered with hulling.

A EFTPOS scanner was added to the Bar Backroom.

### DIFF
--- a/maps/CEVEris/_CEV_Eris.dmm
+++ b/maps/CEVEris/_CEV_Eris.dmm
@@ -754,7 +754,6 @@
 	name = "Yeetus"
 	},
 /obj/structure/disposalpipe/trunk,
-/obj/machinery/shield_diffuser,
 /turf/simulated/floor/hull{
 	flooring_override = "hullcenter18";
 	icon_state = "hullcenter18";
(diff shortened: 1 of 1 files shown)


maps/CEVEris/_CEV_Eris.dmm
```diff
@@ -108323,7 +108325,7 @@
 "iDJ" = (
 /obj/machinery/atmospherics/binary/passive_gate{
 	dir = 4;
-	name = "Waste to Space"
+	name = "Waste to Wasting"
```
ACCount12: I don't think that's a good name. Same on all other instances.

Kurgis: I'm open to suggestion, I could rename it to "Waste to Port".

Kurgis: This was after people reporting that Core to Space was "broken", but it doesn't go to Space, it goes to a port which then takes it away, but it starts off, and needs to be set-up.

TorinoFermic: You could use the word Outside and http://www.photographers1.com/Sailing/ShipboardDirections.gif to describe where the pipe leads to.

MadmanMartian:
Thought we were waiting on a better name for the pipes on this one?

SpadesSlick:
I don't think that waiting one month is good thing. It will work for now.

PR #4693, opened 2020-03-06
Explosive damage changes

NotDeirun:
This makes explosive damage much more deadly. I also stopped frag grenades from exploding, stopped HE grenades from spawning invisible fragments and made them a child of /grenade themselves.

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -121,7 +121,7 @@
 
 		if (2.0)
 			if (!shielded)
-				b_loss += 60
+				b_loss += 150
 
 			if (!istype(l_ear, /obj/item/clothing/ears/earmuffs) && !istype(r_ear, /obj/item/clothing/ears/earmuffs))
 				ear_damage += 30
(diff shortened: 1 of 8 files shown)


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -139,29 +139,13 @@
...
-	if(update)
-		UpdateDamageIcon()
-
+		
+	var/organ_hit = BP_CHEST //Chest is hit first
+	var/exp_damage
+	while (b_loss > 0)
+		b_loss -= exp_damage = rand(0, b_loss)
+		src.apply_damage(exp_damage, BRUTE, organ_hit)
+		organ_hit = pickweight(list(BP_HEAD = 0.2, BPBP_GROIN = 0.2, BP_R_ARM = 0.1, BP_L_ARM = 0.1, BP_R_LEG=0.1, BP_L_LEG=0.1))  //We determine some other body parts that should be hit
```
TorinoFermic: Both lines are in wrong order.

TorinoFermic: At bottom of code, it's missing UpdateDamageIcon()

NotDeirun: From what I can tell apply_damage() already does the UpdateDamageIcon() part.
In the case that I am wrong: Should I put UpdateDamageIcon() inside or outside the while() loop?

TorinoFermic: I am seeing these lines inside apply_damage, that's good.

NotDeirun:
Luc𝒂rio told me to me to increase if (prob(x)) inside proc/fragment_explosion from x = 20 to x=95. I did just that.

PR #4698, opened 2020-03-07
Embed nerf

WuffDev:
Reduces the embedding of most big weapons and just generally reduces the chances of anything getting embedded slightly.

### DIFF
--- a/code/modules/mob/living/carbon/carbon_defense.dm
+++ b/code/modules/mob/living/carbon/carbon_defense.dm
@@ -39,7 +39,7 @@ true, and the mob is not yet deleted, so we need to check that as well*/
 
 		//blunt objects should really not be embedding in things unless a huge amount of force is involved
 
-		var/embed_threshold = weapon_sharp? 5*I.w_class : 15*I.w_class
+		var/embed_threshold = weapon_sharp? 8*I.w_class : 18*I.w_class
 
 		//The user's robustness stat adds to the threshold, allowing you to use more powerful weapons without embedding risk
 		embed_threshold += user.stats.getStat(STAT_ROB)
--- a/code/game/objects/items/weapons/tools/pickaxe.dm
+++ b/code/game/objects/items/weapons/tools/pickaxe.dm
@@ -19,7 +19,7 @@
 	sharp = TRUE
 	structure_damage_factor = STRUCTURE_DAMAGE_BORING //Drills and picks are made for getting through hard materials
 	//They are the best anti-structure melee weapons
-	embed_mult = 1.2 //Digs deep
+	embed_mult = 0.8 //Digs deep
 	mode = EXCAVATE //Mode should be whatever is the starting tool and off quality.
 
 /obj/item/weapon/tool/pickaxe/equipped(mob/user)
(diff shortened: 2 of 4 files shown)


ACCount12:
For now, embeds are the only thing keeping 66 damage meme saws in check, so I would be wary of touching that.

WuffDev:
The damage of stuff like that is being nerfed anyway, so.

Jackofashepard:
There is indeed a order out for decreasing damage on saws and some things upping it. If its a major concern this PR can be held until the other is in, however i think it’ll be fine to be merged for the monday test

code/modules/mob/living/carbon/carbon_defense.dm
```diff
@@ -39,7 +39,7 @@ true, and the mob is not yet deleted, so we need to check that as well*/
 
 		//blunt objects should really not be embedding in things unless a huge amount of force is involved
 
-		var/embed_threshold = weapon_sharp? 5*I.w_class : 15*I.w_class
```
MadmanMartian: Seems like an odd change

WuffDev: Why so?

MadmanMartian: What are these numbers based on, if any at all?

Check what weapons this invalidates in being able to embed

WuffDev: Wait, what do you think this is doing?

jaffjafferson:
I don't see why embeds should be changed.
embedding has proven itself as a effective balance measure for high damage melee weapons, without just gimping their damage

Jackofashepard:
Be that as it way its really gay to have your saw embed first to second hit, we’ll have to find another way thats not embed.

Pariah919:
embedding literally makes sense though. The chains would get stuck inside someone.
"but realism isn't a thing on eris mjp!" this is stupid honestly, embedding is fine as it is as it adds a tradeoff for certain melee weapons. Melee combat in general is garbo because there's nothing like matt's combat mode or stamina to add depth to it. 
Don't remove what little depth there is. Please.

WuffDev:
Should I remove embedding from mobs, then?

I still think damage should be toned down.

And I plan on removing garbage insta knockouts.

Jackofashepard:
How about this, we get a better system than losing your weapon every first to second swing in and then we talk about this again

WuffDev:
Closed for now until we come up with a better solution.

Just want to remark that this wasn't my idea in case anyone gets salty at me.

code/modules/mob/living/carbon/carbon_defense.dm
MadmanMartian: from var/embed_chance = (damage - embed_threshold)*I.embed_mult 

If the embed threshold is greater than the damage, then it will never embed, so which weapons does this make it impossible to embed with now?

WuffDev: It just makes more things harder to embed in general, proportionate to the item's size. 

Bigger things embed less, smaller things embed less. I could give you a list of every item that is impossible to embed but I'm not sure if it's worth the time other than to just say "Shit embeds less proportionately." Unless there is a tiny item that has massive damage, this just makes everything stick in at a flat %

PR #4699, opened 2020-03-07
Adding the katana to ironhammer misc disk

koghs:
## About The Pull Request
Adds the katana to ironhammer misc disk.

### DIFF
--- a/code/datums/autolathe/tools.dm
+++ b/code/datums/autolathe/tools.dm
@@ -87,6 +87,9 @@
 	name = "tactical knife"
 	build_path = /obj/item/weapon/tool/knife/tacknife
 
+/datum/design/autolathe/tool/katana
+	build_path = /obj/item/weapon/tool/sword/katana
+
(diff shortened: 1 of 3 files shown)


Jan-Blasiak:
Ehh... I'd say katana doesn't feel like it fits in IH at all.

They are private security company after all, not mall ninjas.

MadmanMartian:
Why the hay is ironhammer getting a katana?

ACCount12:
Katanas are more Frozen Star than IHS, because Frozen Star are actually ~weebs~ Japanese. Overall, katana is a solid melee weapon, and adding it to a cheap readily available disk is not the right thing to do.

Making it a separate FS maint loot disk with limited uses would be much better.

MadmanMartian:
I'd rather keep the katana to maintenance loot, or acquired through a different method.

Adding things to disks guarantees it being reproducable on an infinite scale, so that won't do.

PR #4703, opened 2020-03-08
TM Voidsuit spawn rate adjustment

Jackofashepard:
Techno voidsuit is now as rare as the IH voidsuit in trash since it has very similar stats


ACCount12:
I've been thinking of making TM voidsuit a separate thing from a normal engineering voidsuit. But for now, that would do.

PR #4704, opened 2020-03-08
Reworks some Neotheology things

WuffDev:
Reduces power on bounty, reduces the number needed for a crusade (6 people feels unrealistic) and changes how one of the crusader litanies works because it doesn't match up with how others work.

More work will follow but this is fine as is.

### DIFF
--- a/code/modules/core_implant/cruciform/rituals/base.dm
+++ b/code/modules/core_implant/cruciform/rituals/base.dm
@@ -45,11 +45,11 @@
 
 /datum/ritual/cruciform/base/entreaty/perform(mob/living/carbon/human/H, obj/item/weapon/implant/core_implant/C)
 	for(var/mob/living/carbon/human/target in disciples)
-		if(target == H)
-		var/area/t = get_area(H) 
+		if(target == H) 
 			continue
 
 		var/obj/item/weapon/implant/core_implant/cruciform/CI = target.get_core_implant()
+		var/area/t = get_area(H)
 
 		if((istype(CI) && CI.get_module(CRUCIFORM_PRIEST)))
 			to_chat(target, SPAN_DANGER("[H], faithful cruciform follower, cries for salvation at [t.name]!"))
--- a/code/modules/core_implant/cruciform/rituals/base.dm
+++ b/code/modules/core_implant/cruciform/rituals/base.dm
@@ -41,17 +41,18 @@
 	name = "Entreaty"
 	phrase = "Deus meus ut quid dereliquisti me"
 	desc = "Call for help, that other cruciform bearers can hear."
-	power = 50
+	power = 35
 
 /datum/ritual/cruciform/base/entreaty/perform(mob/living/carbon/human/H, obj/item/weapon/implant/core_implant/C)
 	for(var/mob/living/carbon/human/target in disciples)
 		if(target == H)
+		var/area/t = get_area(H) 
 			continue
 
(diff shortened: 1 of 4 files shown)


code/modules/core_implant/cruciform/rituals/crusader.dm
```diff
@@ -54,7 +54,7 @@
 	phrase = "Per fidem enim ambulamus et non per speciem."
 	desc = "Knocks over everybody without cruciform in the view range. Psy-wave is too powerful, speaker can be knocked too."
 	cooldown = TRUE
-	cooldown_time = 2 MINUTES
+	cooldown_time = 10 SECONDS
```
MadmanMartian: A target-specific flashbang seems a little strong to be able to fire it off every 10 seconds

WuffDev: It's not really every ten seconds though, Cruciforms don't have cooldowns on things generally because stuff is gated behind cruciform power.

WuffDev: I suddenly realise what you mean and I am a dumbass.

code/modules/core_implant/cruciform/rituals/crusader.dm
```diff
@@ -54,7 +54,7 @@
 	phrase = "Per fidem enim ambulamus et non per speciem."
 	desc = "Knocks over everybody without cruciform in the view range. Psy-wave is too powerful, speaker can be knocked too."
 	cooldown = TRUE
-	cooldown_time = 10 SECONDS
+	cooldown_time = 2 Minutes
```
MadmanMartian: Minutes has to be in all capitals, as it's a define

WuffDev: I can't believe I did that.

WuffDev: I can't believe I did that.

WuffDev:
I've also removed fail chances of simple rituals on request as a QoL change as this only leads to spamming. The hidden fail chance on the detect ritual remains.

code/modules/core_implant/cruciform/rituals/base.dm
```diff
@@ -41,17 +41,18 @@
 	name = "Entreaty"
 	phrase = "Deus meus ut quid dereliquisti me"
 	desc = "Call for help, that other cruciform bearers can hear."
-	power = 50
+	power = 35
 
 /datum/ritual/cruciform/base/entreaty/perform(mob/living/carbon/human/H, obj/item/weapon/implant/core_implant/C)
 	for(var/mob/living/carbon/human/target in disciples)
 		if(target == H)
+		var/area/t = get_area(H)
```
MadmanMartian: Put this below the continue

WuffDev:
Entreaty should now call out the location of the Neotheo who used it.

Kurgis:
I disagree with the crusade changes, not every round requires a crusade, and it's already low.

Coureere:
Same here. I don't think we need such "balance pass" from someone who almost only exclusively plays NT. Crusade is fine as it is, it's already quite powerful and doesn't require entire server, just about almost all church members on high pop, as it should be. Chance removal is, however, is good QoL change.

WuffDev:
It reduces the number from 6 to 5, 6 people is definitely not low, and 5 is a full church, which is very rarely seen.

PR #4706, opened 2020-03-09
adds wheelchair to utility loadout

spessbro:
## About The Pull Request
As of right now you can remove your legs by amputation at start but can't start in a wheelchair so I think it's nice to have it able to get it from the get go instead of crawling until you make/find one and makes sense lore-wise since poor vagabonds might not be able to afford prosthetics while still getting badly wounded on the regular.

### DIFF
--- a/code/modules/client/preference_setup/loadout/lists/utility.dm
+++ b/code/modules/client/preference_setup/loadout/lists/utility.dm
@@ -38,7 +38,7 @@
 	cost = 2
 /datum/gear/utility/wheelchair
 	display_name = "wheelchair"
-	path = /obj/structure/bed/chair/wheelchair
+	path = /obj/item/wheelchair
 	cost = 3
 
 /datum/gear/utility/normaltablet
--- a/code/modules/client/preference_setup/loadout/lists/utility.dm
+++ b/code/modules/client/preference_setup/loadout/lists/utility.dm
@@ -37,6 +37,11 @@
 	path = /obj/item/modular_computer/tablet/preset/custom_loadout/cheap
 	cost = 2
 
+/obj/structure/bed/chair/wheelchair
+	display_name = "wheelchair"
+	pathc = /obj/structure/bed/chair/wheelchair
+	cost = 3
+
 /datum/gear/utility/normaltablet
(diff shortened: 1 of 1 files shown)


spessbro:
what the frick travis :(

code/modules/client/preference_setup/loadout/lists/utility.dm
```diff
@@ -37,6 +37,11 @@
 	path = /obj/item/modular_computer/tablet/preset/custom_loadout/cheap
 	cost = 2
 
+/obj/structure/bed/chair/wheelchair
+	display_name = "wheelchair"
+	pathc = /obj/structure/bed/chair/wheelchair
```
TorinoFermic: /obj/structure/bed/chair/wheelchair points to  /obj/structure/bed/chair/wheelchair, well it won't compile.
You misspelled pathc for path

spessbro: thank you i am very dumb and can only webedit on linux

MadmanMartian:
A lot of loadout code expects the item to be, well, an item.

You may want to comb through loadout code and see where it could potentially runtime

APWill:
> A lot of loadout code expects the item to be, well, an item.
> 
> You may want to comb through loadout code and see where it could potentially runtime

Since I implemented the foldable wheelchair functionality (kindly edited by SolatK) it's already available in item form as the folded version. Just change the path to /obj/item/wheelchair.

code/modules/client/preference_setup/loadout/lists/utility.dm
```diff
@@ -36,6 +36,10 @@
 	display_name = "cheap tablet computer"
 	path = /obj/item/modular_computer/tablet/preset/custom_loadout/cheap
 	cost = 2
+/datum/gear/utility/wheelchair
+	display_name = "wheelchair"
+	path = /obj/structure/bed/chair/wheelchair
```
MadmanMartian: Replace with the foldable wheelchair object

TorinoFermic: The path must be a item

PR #4708, opened 2020-03-09
Makes changelings real good at self-surgery

ACCount12:
Changelings no longer need a chair to perform self-surgery, they can do it just by standing still. They also have a significant increase to success chances and don't need to diagnose a body part they are operating on.

### DIFF
--- a/code/modules/surgery/surgery.dm
+++ b/code/modules/surgery/surgery.dm
@@ -112,6 +112,11 @@
 	if(owner && user == owner)
 		difficulty_adjust = 20
 
+		// ...unless you are a changeling
+		// It makes sense that lings have a way of making their flesh cooperate
+		if(user.check_special_role(ROLE_CHANGELING))
+			difficulty_adjust = -50
+
 	var/atom/surgery_target = get_surgery_target()
 	if(S.required_tool_quality)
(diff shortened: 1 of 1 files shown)


TorinoFermic:
I can imagine lings doing self-surgery to hide things inside their bodies.

ACCount12:
At some point, we might make hiding things inside the body a legit ling ability. Not now though.

MadmanMartian:
What has prompted this decision?

ACCount12:
I wanted to make changeling good at self-surgery back when I rewrote the surgery code in the first place. Makes sense that The Thing/Prototype mutants would have enough control over their own flesh to be able to operate on themselves with a rusty knife. But back then, I became more concerned with pushing the surgery code out already, and many things I wanted to do weren't actually implemented.

Now I'm revisiting surgery and actually adding those things.

Jan-Blasiak:
Maybe make it a "storage implant" like Traitor's one?

MadmanMartian:
Closed, being redone in a different way

ACCount12:
I said that if that "different way" isn't implemented in two weeks, I'll kick ROBO in the balls and force merge this one. Here I go.

PR #4712, opened 2020-03-10
New Melee Weapons

MadmanMartian:
## About The Pull Request
Adds weapons outlined in the Technomancer's Design Document.

 - Maces and Makeshift maces
 - Daggers
 - Replaces the old Claymore and Katana sprites

Adds a few strange-tech items
 - Moebius Experimental Nanorefolding Katana. Folded X times, can never embed itself in a person
 - Moebius Bluespace Dagger. Turn it on, stab it into a person so it stays there, and watch as they are no longer an issue as they teleport around the ship.
 - Assassin's Dagger. Can be loaded with reagents to inject people during combat.

Will do charge hammer in a different PR


Pariah919:
More of a balance or question. 

why is the makeshift mace the same force as a normal hammer? I really doubt that should be the case. They both deal 16 damage.

A mace dealing 20 damage also seems odd as it also seems to have no armor piercing qualities on it. Perhaps 26 damage and 20 pen would be more suitable?

mattroks101:
Ah, it's got conflicts though.

MadmanMartian:
>binary conflicts

a

TorinoFermic:
There's conflicts to be resolved before I could merge it

PR #4719, opened 2020-03-12
Disks Tweaks

Coureere:
##  About The Pull Request
- All of the gun disks now come with lethal ammo, no exception.
- Removed practice mags/boxes from each disk, added separate disk called "Frozen Star Practice Ammunition Pack" which contains both mags and boxes with practice rounds, available from your local FS vendor for mere 200 credits. It's in the loot lists too, but it can't appear in "advanced disk" category because no one likes finding practice ammo disk instead of something useful.
- Minor corrections.


Coureere:
Due to whacky balance this should proabably be merged only after disks would become less common loot compared to individual guns. AFAIK there is someone working on it.

MadmanMartian:
After some lengthy discussion, this change has been vetoed

ACCount12:
You can still add the "Frozen Star Practice Ammunition Pack" disk in a separate PR. More disk variety is welcome. Although I see no point in there being 10 of them in each disk vendor - knock it down to something more sane.

PR #4729, opened 2020-03-16
Rig-mounted Autodoc fixes and tweaks

MadmanMartian:
Not sure if it was a feature, but the capitalist autodoc would never actually heal you

Adds some chat feedback for the user, *to help with debugging other issues in the future*

Changes the text in the autodoc module to say "Close wounds" rather than "Treat wounds" as there was some confusion between the two


MadmanMartian:
I tested it myself a fair bit, was able to get them to start fixing wounds and such

TorinoFermic:
Yeah, my testing confirms

PR #4730, opened 2020-03-17
Remove deprivation oversight

WuffDev:
Fixes the deprivation litany so it can't instantly kill those with a cruciform.

### DIFF
--- a/code/modules/core_implant/cruciform/rituals/priest.dm
+++ b/code/modules/core_implant/cruciform/rituals/priest.dm
@@ -180,10 +180,14 @@
 	if(!CI.wearer)
 		fail("Cruciform is not installed.", user, C)
 		return FALSE
+	
+	if(CI.wearer.stat != DEAD)
+		fail("This one still lives.", user, C)
+		return FALSE
 
 	var/mob/M = CI.wearer
 
-	if(ishuman(M) && M.is_dead())
+	if(ishuman(M))
 		var/mob/living/carbon/human/H = M
 		var/obj/item/organ/external/E = H.organs_by_name[BP_CHEST]
 		E.take_damage(15)
--- a/code/modules/core_implant/cruciform/rituals/priest.dm
+++ b/code/modules/core_implant/cruciform/rituals/priest.dm
@@ -181,11 +181,11 @@
 		fail("Cruciform is not installed.", user, C)
 		return FALSE
 	
-	if(CI.wearer.stat != DEAD)
+		var/mob/M = CI.wearer
+	
+	if(!M.is_dead())
 		fail("This one still lives.", user, C)
 		return FALSE
-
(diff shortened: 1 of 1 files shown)


code/modules/core_implant/cruciform/rituals/priest.dm
```diff
@@ -180,10 +180,14 @@
...
 		fail("Cruciform is not installed.", user, C)
 		return FALSE
+	
+	if(CI.wearer.stat != DEAD)
+		fail("This one still lives.", user, C)
+		return FALSE
 
 	var/mob/M = CI.wearer
 
-	if(ishuman(M) && M.is_dead())
```
TorinoFermic: You can move the line 188 above 184 	if(CI.wearer.stat != DEAD) then you can use M.is_dead() replacing the check CI.wearer.stat != DEAD with this.

WuffDev: What?

TorinoFermic: `	if(CI.wearer.stat != DEAD)
		fail("This one still lives.", user, C)
		return FALSE

	var/mob/M = CI.wearer

	if(ishuman(M))`

becomes this

`	var/mob/M = CI.wearer
	
	if(!M.is_dead())
		fail("This one still lives.", user, C)
		return FALSE

	if(ishuman(M))`

The ! in front of M.is_dead() turns from FALSE to TRUE ( actually more like from 0 to 1 and vice-versa ).

TorinoFermic: And this `var/mob/M = CI.wearer` lets us to have access to the proper mob/m and thus is possible to call functions on this without having the compiler complain because it wasn't type cased into /mob/

code/modules/core_implant/cruciform/rituals/priest.dm
```diff
@@ -180,10 +180,14 @@
 	if(!CI.wearer)
 		fail("Cruciform is not installed.", user, C)
 		return FALSE
+	
+		var/mob/M = CI.wearer
```
Kurgis: Extra tab here needs to be removed, causing `M` to be undefined for the higher line statements.

MadmanMartian:
Looking at this, it seems to be rather redundant. There already was an isdead check

How was it killing people, exactly?

WuffDev:
Because it didn't work. I'm going to change this to just restrict using it on people with cruciform upgrades instead.

code/modules/core_implant/cruciform/rituals/priest.dm
WuffDev: Yep, I'm going to either close this and make a new one or just change this one to reflect what was spoken about in discord.

PR #4734, opened 2020-03-18
Voidsuit buffs/nerfs

Pariah919:
## About The Pull Request
Does a bit more of a balance pass on voidsuits.

## Details
Buffs the mining voidsuits bullet proc to be 35 rather than 20, considering there's a bunch of away mission mobs that try to shoot your face off. Same with minorly better energy due to one star drones melting miners if they don't have kelotane on them.
Gives the techno armor 40 bomb resist but less overall resist.
Medical voidsuit has more energy protection now.
</details>

## Screenshots
nothing changed. Spooky.

</details>

### DIFF
--- a/code/game/objects/random/voidsuit.dm
+++ b/code/game/objects/random/voidsuit.dm
@@ -8,7 +8,7 @@
 
 /obj/random/voidsuit/item_to_spawn()
 	return pickweight(list(/obj/item/clothing/suit/space/void = 2,
-		/obj/item/clothing/suit/space/void/engineering = 1,
+		/obj/item/clothing/suit/space/void/engineering = 2,
 		/obj/item/clothing/suit/space/void/mining = 2,
 		/obj/item/clothing/suit/space/void/medical = 2.3,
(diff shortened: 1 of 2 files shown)


code/game/objects/random/voidsuit.dm
```diff
@@ -8,7 +8,7 @@
 
 /obj/random/voidsuit/item_to_spawn()
 	return pickweight(list(/obj/item/clothing/suit/space/void = 2,
-		/obj/item/clothing/suit/space/void/engineering = 1,
```
MadmanMartian: Why this?

Pariah919: People told me to make it more common because it's no longer really, really good compared to most other voids, and I agreed, though I can set it back.

Jackofashepard:
I personally with my very biased opinion still disagree with techno voidsuits getting values lowered but i do agree with the miner suit, except for the new slowdown to it?

Jackofashepard:
Discussed, this PR is ready for merge i'd say

PR #4731, opened 2020-03-17
German Language

Smokingsquirrel:
#4000  About The Pull Request

PR #4737, opened 2020-03-18
German Language

Smokingsquirrel:
## About The Pull Request
Overrides PR #4731 with cleaned code and hopefully correct GitHub submission and PR.


ACCount12:
Why?

I've never seen it being requested, I've never seen it in any of the design docs, and I don't think there ever was any pressing need for a new language.

code/modules/mob/language/german.dm
```diff
@@ -0,0 +1,393 @@
+/datum/language/german
+	name = LANGUAGE_NEO-GERMANIC
```
Kurgis: This seems like a define, remove the "-".

Kurgis: `LANGUAGE_GERMAN` so it conforms with the other languages.

Smokingsquirrel: Thank you for the feedback

PR #4738, opened 2020-03-19
German Language for Oberth and Language Table

Smokingsquirrel:
## About The Pull Request
Updated to desired values, edited the stylesheet with a color for the language.

Added german to the species_language and edited the Oberth background per Jack Frosts suggestion.

## Details
<details>
  <summary>Expand</summary>
Changes the language description to "Founded on the mostly german colony of Oberth"
Added a color in the stylesheet.dm for the language
Added the language (hopefully correctly) to Oberth Origin.dm
Added define in species_language.dm
</details>


code/datums/setup_option/backgrounds/origin.dm
```diff
@@ -22,14 +22,16 @@
 		STAT_COG = -8
 	)
 
+	if /mob/living/carbon/human
+		add_language(LANGUAGE_GERMAN)
```
TorinoFermic: This part will not compile, remove it please.

Smokingsquirrel: How does it have to be then? Has the if to be removed to add the language to Oberth origin?
I can't deduce how in the world that should be phrased.

TorinoFermic: It does not need any code, datum are meant to only contain variables for other parts of code to work.

TorinoFermic: This is because the example file does not need it thus this file does not need this check. Also, the reason why it must be removed is the /datum is basically already exposes to the rest of code ready to be used. You can see in https://github.com/discordia-space/CEV-Eris/pull/4095 which he adds only these variables.

Smokingsquirrel: Okay, thank you for the explanation, i will just remove it then in this case.

Smokingsquirrel:
Should work properly now.

code/stylesheet.dm
```diff
@@ -92,6 +92,7 @@ h1.alert, h2.alert		{color: #000000;}
 .russian				{color: #635d17}
 .serbian				{color: #Ad6f1d}
 .jive					{color: #666666}
+.german					{color: #9933FF}
```
Jackofashepard: Spacings a bit off

Jackofashepard:
Also there's loads of "'" that went to weird ? signs

Jackofashepard:
I don't see code adding it to the oberth background, if you want help with that i suggest yoinking language code from jobs (i know for a fact serbs are forced serbian)

code/stylesheet.dm
Smokingsquirrel: strange, that wasnt there before... i will check up the code, and i was told to remove the former code that hat id. When i added known_languages to mob in Oberth, it didnt compile the code with travis
I will check out the mercenary.dm

PR #4742, opened 2020-03-20
Adds new mining voidsuit icons, courtesy of Shazbot

MadmanMartian:
image

May conflict with #4734 so that will need to be merged first

Yes there are neko ears on the one worn sprite, but ~~thankfully~~ we lack a ground state for the helmet with cat ears, so that will not be implemented ~~right now~~


Jackofashepard:
Excellent stuff, looks great!

mattroks101:
Ok, just waiting on #4734 to get merged then.

PR #4745, opened 2020-03-20
Makes it harder to break robotic limbs

Jackofashepard:
It is now harder to break robot limbs than it is to break organic ones.


GrayRachnid:
what if it's the cheap kind,,,,,,,,,,,,,,,,,,,,,,,,,,,,

Jackofashepard:
Our "cheap" augment is unlocked at the end of the research tree and the ones pickable from menu are supposed to be the same but just look different (according to their code  at least)

Jackofashepard:
Changed to be on par with human limbs

PR #4748, opened 2020-03-20
Nanite duping fix

MadmanMartian:
Fixes partial pathing in medical_tools.dm

Fixes being able to dupe nanites, by making them a solid rather than a liquid


ACCount12:
I don't think we should change mech stuff until the actual mech changes are done.

PR #4749, opened 2020-03-20
Redoes Multiz Sound

mattroks101:
## About The Pull Request

Shamelessly stolen from https://github.com/NebulaSS13/Nebula/pull/189. This adds softness and echoing effects to sounds that are on another z-level from you.  

Video example of it on Eris here:
https://streamable.com/un1lh

### DIFF
--- a/code/game/sound.dm
+++ b/code/game/sound.dm
@@ -386,13 +386,7 @@ var/const/FALLOFF_SOUNDS = 0.5
 			pressure_factor = max(pressure_factor, 0.15)	//hearing through contact
 
 		S.volume *= pressure_factor
-
-		if(istype(T,/turf/simulated) && istype(turf_source,/turf/simulated))
-			var/turf/simulated/sim_source = turf_source
-			var/turf/simulated/sim_destination = T
-			if(sim_destination.zone != sim_source.zone)
-				S.volume -= 30
-
+		
 		if (S.volume <= 0)
 			return //no volume means no sound
 
--- a/code/game/sound.dm
+++ b/code/game/sound.dm
@@ -419,7 +419,8 @@ var/const/FALLOFF_SOUNDS = 0.5
 				S.environment = SPACE
 			else
 				var/area/A = get_area(src)
-				S.environment = A.sound_env
+				if(istype(A))
+					S.environment = A.sound_env
 
 		else if (pressure_factor < 0.5)
 			S.environment = SPACE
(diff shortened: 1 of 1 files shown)


code/game/sound.dm
```diff
@@ -288,118 +314,134 @@ var/list/rummage_sound = list(\
 
 	return toplay
 
-/proc/playsound(var/atom/source, soundin, vol as num, vary, extrarange as num, falloff, var/is_global, var/use_pressure = TRUE)
```
deathride58: The replacement of use_pressure with frequency here may cause compatibility issues with things that call playsound(); it'd be worth going through and checking every single use of playsound if you're wanting to replace the sound code with baycode's sound code wholesale.
For reference, `zrange`, `override_env`, `envdry`, and `envwet` are all that's necessary for zsound (the multiz sound implementation in the PR linked in your OP) to work.

code/game/sound.dm
```diff
@@ -288,118 +314,134 @@ var/list/rummage_sound = list(\
...
+		if (distance <= 1)
+			pressure_factor = max(pressure_factor, 0.15)	//hearing through contact
+
+		S.volume *= pressure_factor
+
+		if(istype(T,/turf/simulated) && istype(turf_source,/turf/simulated))
+			var/turf/simulated/sim_source = turf_source
+			var/turf/simulated/sim_destination = T
+			if(sim_destination.zone != sim_source.zone)
+				S.volume -= 30
```
deathride58: Oh, and this here: Nebula/bay has a lot of sound sources with high base volumes. This bit of code here will result in a lot of sounds being inaudible through walls if you're using TG standards for playsound volumes (this can be heard in the demo you posted, as well). This isn't a part of z-sound, but it might be worth replacing this with an envdry adjustment to avoid the headaches associated with bumping up the volumes of sound sources across the board to compensate for this feature.

mattroks101: Duly noted.

code/game/sound.dm
mattroks101: We've been having a lot of sound issues over the years, honestly it might be worth replacing wholesale. I'll do some thorough testing to see what gets broken and what doesn't.

mattroks101:
Curse you travis.

mattroks101:
Travis man still not working >:(

MadmanMartian:
It's working, it's just not reporting back

image

mattroks101:
@MadmanMartian is this good to merge anyway then? What's the fix for this travis issue?

MadmanMartian:
I'm looking into why it's not reporting back, but error-wise it's good to go

PR #4750, opened 2020-03-21
Removes the flat 4 force modifier increase from the hammer mod

MadmanMartian:
I asked somebody to remove this a while ago, but they didn't seem to do it.


TorinoFermic:
Well, this is a shame the head dev has do it. Merging.

PR #4752, opened 2020-03-21
Stunprod fixes

MadmanMartian:
Ghetto stunprod no longer spawns a cell from the aether

Ghetto stunprod don't disappear now when turned on

image

### DIFF
--- a/code/game/objects/items/weapons/stunbaton.dm
+++ b/code/game/objects/items/weapons/stunbaton.dm
@@ -18,12 +18,13 @@
 	var/status = FALSE		//whether the thing is on or not
 	var/hitcost = 100
 	var/obj/item/weapon/cell/cell = null
+	var/spawn_cell = TRUE
 	var/suitable_cell = /obj/item/weapon/cell/medium
 	structure_damage_factor = STRUCTURE_DAMAGE_BLUNT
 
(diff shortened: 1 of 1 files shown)


code/game/objects/items/weapons/stunbaton.dm
```diff
@@ -177,6 +178,7 @@
 	hitcost = 150
 	attack_verb = list("poked")
 	slot_flags = null
+	spawn_cell = FALSE
```
ACCount12: I don't think that's a good way to solve it. Most objects just remove their cells in `Created` call or something like that.

MadmanMartian: Better to not create the cell in the first place, than to create the cell then delete it

TorinoFermic: I might have a idea to it, set the suitable_cell to null in  /obj/item/weapon/melee/baton/cattleprod and then set to /obj/item/weapon/cell/medium in New() ?

MadmanMartian: No

TorinoFermic: Well, this is all I could think to this problem. That's it.

PR #4753, opened 2020-03-21
Hammers Adjustments

Coureere:
## About The Pull Request

- Adjusted recipe for homewrecker. Now it uses adhesive item how it truly should (according to the sprite, don't mind blue tape that should be actually white).
- Moved homewrecker from `simple_weapons.dm` to `hammer.dm` because it should be there.
- Adjusted numerous stats around hammers, mainly made them non toggleable. Most of them (one star and homewrecker for now) must be just wielded to use to their fullest, additional requirement of turning them on/off made zero sense. Powered hammer, however, lacks certain sprites to fix it, so i just slapped quick fix and left a comment, as well as requested sprites on discord.
- Hammer toolmod no longer makes your tool nearly useless in terms of speed, it now slows it down for 10% rather than for the whole 50% which made no sense because hammering is useless anyway.


Coureere:
This should fix most current problems hammers have, but some still remain:

- Afromentioned lack of sprites to make powered hammer into proper wieldable tool rather than powered version of usual one-handed hammer.
- When you apply hammering addon to the tool, it does add to the tool quality (seemingly) as it is intended, however there is no indication at all. You can't tell that tool now has hammering quality upon examining it or using info tab. Only tools with initial hammering quality does display it. It is also unclear how things behave in case of combining afromentioned toolmod with tool that already has hammering quality. Code has almost no support in case of someone wants to make their toolmod affect tool qualities in any way.
- And the biggest one - **HAMMERING HAS NO REAL USE BESIDES LITERALLY ONE OR TWO VERY RARE PERFORMED TASKS.** This can be solved by improving upon tool code in general, for example making one quality substitute other (welding/sawing, wrenching/screwdriving) in some cases and simply adding more applications to some rarely used tool qualities, but for now it remains literally useless as a tool and has far more use as badass weapon (worth to mention that it's not even one of the best weapons game affords).

ACCount12:
I don't see a good use for a hammer in space though. So useless it would stay.

code/game/objects/items/weapons/tools/hammer.dm
```diff
@@ -1,74 +1,67 @@
...
 	use_power_cost = 2
 	suitable_cell = /obj/item/weapon/cell/medium
-	max_upgrades = 1
+	max_upgrades = 4
 
 /obj/item/weapon/tool/hammer/powered_hammer/onestar_hammer
-	name = "One Star hammer"
-	desc = "Used for applying immeasurable blunt force to anything in your way."
+	name = "One Star sledgehammer"
+	desc = "Famous sledgehammer model made by One Star used for applying immeasurable blunt force to anything in your way. Could breach even toughest obstracles and crack most resilent skulls."
```
MadmanMartian, line 48: "A sledgehammer model produced by One Star, used for applying immeasurable blunt force to anything in your way. Capable of breaching even the toughtest obstacles, and cracking the most resilient skulls."

Coureere:
Ready to review and merge, resolved conflicts, removed duplicate homewrecker, replaced hammer sprites.

PR #4756, opened 2020-03-22
Reduces mouse damage to 0 and disables them buckling larger mobs

Jackofashepard:
Title


Kurgis:
It appears that mice attacking mobs is broken, likely due to me adding some kind of incompatibility from Baystation.

Jackofashepard:
Oh? Well this one is mostly focused on making them not being able to chew through mechs, however i’d say its good its broken since the last thing we want is salty ghosts coming back to troll people

PR #4757, opened 2020-03-23
Toolmod plasteel dupe crafting fix

Coureere:
## About The Pull Request

- Dupe of plasteel when crafting tool mods is no more.
- Adjusted some recipes.


TorinoFermic:
The future fix could make autolathes being able to read the steps var and recycle based on this var otherwise check matter

PR #4759, opened 2020-03-23
Biotank size increase to match the biogens increased power and to stop it from breaking contantly

Jackofashepard:
Title


MadmanMartian:
So, how quickly does the biogen create biomatter?

How long would it take to fill one of these new biotanks?

Can these biotanks be spilled, and if so, how much of the room would now be covered in biomass?

Jackofashepard:
This increases the size of the main biotank itself, its apperantly very easy to currently overfill it which permanantly breaks it, its a lot harder now at least. as i understand it creates the biomatter almost instantly from items inside it and the biotank cannot spill, it merely breaks

PR #4760, opened 2020-03-23
Department account perms for guild techs

Jackofashepard:
I find that the guild ceases to function every time the merchant is merced or goes to sleep, now the guild technicians will also be able to keep the eris guild alive


ACCount12:
Guild already has an account number in the console, visible to anyone with access. They can use it to funnel money there.

Jackofashepard:
They cant use it to withdraw money tho

PR #4764, opened 2020-03-25
At least 5 players need to be online for serbs to spawn

Jackofashepard:


Pariah919:
5 is probably better a better hardcap than ten, and then make it so serbians can't spawn until at least an hour has passed or the overall living crew requirement is above 10. which will happen at around 12 players.

Jackofashepard:
Changed to 5, let me see if timers even exist

PR #4765, opened 2020-03-25
Increases robot limb malfunction threshold

ACCount12:
Robot limbs now need 20 damage to start malfunctioning (up from 10), and the malfunction chances are lower.

### DIFF
--- a/code/modules/organs/external/_external.dm
+++ b/code/modules/organs/external/_external.dm
@@ -819,8 +819,9 @@ Note that amputating the affected organ does in fact remove the infection from t
 			return 1
 	return 0
 
+// Robotic limbs malfunction - handled by subtype
 /obj/item/organ/external/proc/is_malfunctioning()
-	return (BP_IS_ROBOTIC(src) && (brute_dam + burn_dam) >= 10 && prob(brute_dam + burn_dam))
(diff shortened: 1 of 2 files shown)


code/modules/organs/external/_external.dm
```diff
@@ -819,8 +819,9 @@ Note that amputating the affected organ does in fact remove the infection from t
 			return 1
 	return 0
 
+// Robotic limbs malfunction - handled by subtype
 /obj/item/organ/external/proc/is_malfunctioning()
-	return (BP_IS_ROBOTIC(src) && (brute_dam + burn_dam) >= 10 && prob(brute_dam + burn_dam))
```
MadmanMartian: Undo this change. BP_IS_ROBOTIC checks the organs nature, which can be changed without it being a subtype (Example being body modification, check body_modifications.dm, line 195 for example)

ACCount12: Now look at the line 192 in the same file.
```body_parts = list(BP_HEART, BP_LUNGS, BP_LIVER, BP_EYES)```
It's limited to some select INTERNAL organs. Actual loadout limbs replace the limb, complete with a /robotic path.

PR #4775, opened 2020-03-30
Cryopod fix

ACCount12:
* Fixes cryopod runtiming out on despawn
* Makes cryopod remove antag contracts targeting the despawned mob, like it removes the regular antag objectives.

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -336,8 +336,7 @@
 			qdel(O)
 
 	//Same for contract-based objectives.
-	for(var/c in GLOB.all_antag_contracts)
-		var/datum/antag_contract/contract = c
+	for(var/datum/antag_contract/contract in GLOB.all_antag_contracts)
 		contract.on_mob_despawned(occupant.mind)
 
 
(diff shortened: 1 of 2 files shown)


code/game/machinery/cryopod.dm
```diff
@@ -328,6 +335,12 @@
 				to_chat(O.owner.current, "<span class='warning'>You get the feeling your target is no longer within your reach...</span>")
 			qdel(O)
 
+	//Same for contract-based objectives.
+	for(var/c in GLOB.all_antag_contracts)
```
TorinoFermic, line 339: When for( in ) loops, each element from GLOB.all_antag_contracts is being put in c, so in same line it can do var/datum/antag_contract/contract/c

ACCount12: If you put a type into `in` loop, it will check every entry in the list for that type. If you know what type are the entries in the list already, doing it this way is a tiny bit faster.

Tgstation used to enforce this shit, so I'm doing it out of habit.

TorinoFermic: Understood, thanks !

PR #4782, opened 2020-04-03
Economy update (WIP)

TheShown911:
## About The Pull Request
The main goal of this PR is to make money more used, focusing on item prices and vendors. 

## Details

<details>
  <summary>Expand</summary>

[?] Vendors - bring prices more in line with peoples money, buying from other people should always be better than buying from vendors.

[X] Materials - Material base cost is way too high, that said nerfing it outright would be a bad idea, miners should be able to make nice money but one spot with a roundstart stack of plastic worth 3 vegabounds should be corrected.

[X] Item prices - Item prices are not in too bad of a condition but some can be confusing to new guild members. Way too low/high item prices should be updated.

[X] Anything I missed - Economy in general is a hard thing to balance, with every change more problems arise. My changes are in no way the best changes and i would apriciate all suggestions.

</details>

### DIFF
--- a/code/game/machinery/vending.dm
+++ b/code/game/machinery/vending.dm
@@ -969,8 +969,8 @@
 	products = list(/obj/item/weapon/reagent_containers/food/drinks/coffee = 25,/obj/item/weapon/reagent_containers/food/drinks/tea/black = 25,
 					/obj/item/weapon/reagent_containers/food/drinks/tea/green = 25,/obj/item/weapon/reagent_containers/food/drinks/h_chocolate = 25)
 	contraband = list(/obj/item/weapon/reagent_containers/food/drinks/ice = 10)
-	prices = list(/obj/item/weapon/reagent_containers/food/drinks/coffee = 3, /obj/item/weapon/reagent_containers/food/drinks/tea/black = 3,
-					/obj/item/weapon/reagent_containers/food/drinks/tea/green = 3, /obj/item/weapon/reagent_containers/food/drinks/h_chocolate = 3)
+	prices = list(/obj/item/weapon/reagent_containers/food/drinks/coffee = 10, /obj/item/weapon/reagent_containers/food/drinks/tea/black = 10,
+					/obj/item/weapon/reagent_containers/food/drinks/tea/green = 10, /obj/item/weapon/reagent_containers/food/drinks/h_chocolate = 10)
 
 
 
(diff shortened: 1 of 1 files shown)


code/game/machinery/vending.dm
```diff
@@ -1102,7 +1102,7 @@
 	products = list(/obj/item/weapon/storage/fancy/cigarettes = 10,/obj/item/weapon/storage/box/matches = 10,/obj/item/weapon/flame/lighter/random = 4)
 	contraband = list(/obj/item/clothing/mask/smokable/cigarette/cigar = 4, /obj/item/weapon/flame/lighter/zippo = 4,)
 	premium = list(/obj/item/weapon/storage/fancy/cigar = 5,/obj/item/weapon/storage/fancy/cigarettes/killthroat = 5 )
-	prices = list(/obj/item/clothing/mask/smokable/cigarette/cigar = 200, /obj/item/weapon/storage/fancy/cigarettes = 100 ,/obj/item/weapon/storage/box/matches = 10,/obj/item/weapon/flame/lighter/random = 5,
+	prices = list(/obj/item/clothing/mask/smokable/cigarette/cigar = 200, /obj/item/weapon/storage/fancy/cigarettes = 100 ,/obj/item/weapon/storage/box/matches = 20,/obj/item/weapon/flame/lighter/random = 20,
```
ACCount12, line 1105: I think matches should be cheap - like, five creds cheap. No one would use them unless they are cheaper than the cheapest lighter.

code/game/machinery/vending.dm
```diff
@@ -1355,8 +1355,8 @@
 	icon_state = "powermat"
 	products = list(/obj/item/weapon/cell/large = 10, /obj/item/weapon/cell/large/high = 10, /obj/item/weapon/cell/medium = 15, /obj/item/weapon/cell/medium/high = 15, /obj/item/weapon/cell/small = 20, /obj/item/weapon/cell/small/high = 20)
 	contraband = list(/obj/item/weapon/cell/large/super = 5, /obj/item/weapon/cell/medium/super = 5, /obj/item/weapon/cell/small/super = 5)
-	prices = list(/obj/item/weapon/cell/large = 500, /obj/item/weapon/cell/large/high = 700, /obj/item/weapon/cell/medium = 300, /obj/item/weapon/cell/medium/high = 400, /obj/item/weapon/cell/small = 100, /obj/item/weapon/cell/small/high = 200,
-				/obj/item/weapon/cell/large/super = 1200, /obj/item/weapon/cell/medium/super = 700, /obj/item/weapon/cell/small/super = 350)
+	prices = list(/obj/item/weapon/cell/large = 300, /obj/item/weapon/cell/large/high = 500, /obj/item/weapon/cell/medium = 200, /obj/item/weapon/cell/medium/high = 300, /obj/item/weapon/cell/small = 70 , /obj/item/weapon/cell/small/high = 150,
+				/obj/item/weapon/cell/large/super = 700, /obj/item/weapon/cell/medium/super = 400, /obj/item/weapon/cell/small/super = 250)
```
ACCount12, line 1359: I don't think price cuts on the "super" cells are warranted. People normally want big cells for stuff like RIGs and weapons, and it's normal for something like that to be pricey. Prices for regular trash tier cells, on the other hand, could be even lower. I'd say that 50 creds for a regular small cell, and 200 for a regular large cell is about right.

Also, move the "super" cell away from the contraband list and into the normal item list. Limit the available amount instead. You can cut the amount of cells overall too - no round would ever sell 10 tier 1 cells, and it's fine if tier 3 cells are in short supply.

PR #4785, opened 2020-04-03
buffs the laser sabre

Fikou:
## About The Pull Request
the laser sabre now inherits the lethal force from the energy sword instead of being just robust
(26 -> 40)
why: the laser sabre requires late game research, an rng tool attachment and a gun, its a lot of work for an item that has less force than a chainsaw available roundstart with the brutal level of damage if you got enough money for an advanced disk, so getting the sabre is not worth it at all when you have to go on a fetch quest for it


Fikou:
a

PR #4794, opened 2020-04-07
Adding NT radio encryption key to their disk

Coureere:
## About The Pull Request
Benefits NT in the way that they have a way to have closer contact with their followers now, as they now can actually produce encryption keys and people can ask for them in order to communicate with them later. I'd even say this should increase quantity of mass lithanies, given anyone would want to play non-church NT.

Decapitalized Neotheology in the name of the encryption key item itself, but it appears fully capitalized in the autolathe design name. Looks nicer both ways. Also added silver cost of 1 to any encryption keys because just steel feels too cheap and bland. Silver is still most common from "rare" category materials as it could be easily extracted from powercells and else.

### DIFF
--- a/code/game/objects/items/devices/radio/encryptionkey.dm
+++ b/code/game/objects/items/devices/radio/encryptionkey.dm
@@ -11,7 +11,7 @@
 	var/translate_hive = 0
 	var/syndie = 0
 	var/list/channels = list()
-	matter = list(MATERIAL_STEEL = 1, MATERIAL_SILVER = 1)
+	matter = list(MATERIAL_STEEL = 0.5, MATERIAL_SILVER = 0.5)
 
 /obj/item/device/encryptionkey/attackby(obj/item/weapon/W as obj, mob/user as mob)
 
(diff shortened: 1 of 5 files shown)


TorinoFermic:
Little suggestion, you can add this key in the preacher's closet ?

Coureere:
I'l place a box of them right now, do not merge.

Coureere:
Here, done, ready for merge.

code/game/objects/items/devices/radio/encryptionkey.dm
```diff
@@ -11,7 +11,7 @@
 	var/translate_hive = 0
 	var/syndie = 0
 	var/list/channels = list()
-	matter = list(MATERIAL_STEEL = 1)
+	matter = list(MATERIAL_STEEL = 1, MATERIAL_SILVER = 1)
```
MadmanMartian, line 14: Odd change. I know you're doing this so that somebody can't mass-produce encryption keys, but now somebody can extract silver from existing encryption keys.

Coureere: Not like silver is THAT uncommon. You can gain full stacks by gathering and recycling cheap powercells, not like gathering headset keys (which is less common than powercells) for silver would change anything. However, if you insist - i will remove that right now in a second.

PR #4803, opened 2020-04-11
[FIX] Cryopod : Fixed the cause of runtime that breaks the MC ?

TorinoFermic:
## About The Pull Request

Before this PR, a robot in a robotic cryo will cause a runtime because the despawn_occupant() will call the parent which lands in the proc with human-only code causing a runtime unless the line `occupant_organs = H.organs | H.internal_organs` is behind the type check

The PR fixes the spammy runtime, not sure if it also breaks the MC beforehand.

Last thing, should this fix to be changed to a refactor of cryo code so we have /obj/machinery/cryopod/organic and original /obj/machinery/cryopod/robot, the /obj/machinery/cryopod will just simply call the generic removal code ?

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -267,7 +267,6 @@
 // This function can not be undone; do not call this unless you are sure
 // Also make sure there is a valid control computer
 /obj/machinery/cryopod/robot/despawn_occupant()
-	/// Move then into mob/living/silicon/despawn()
 	var/mob/living/silicon/robot/R = occupant
 	if(!istype(R)) return ..()
 
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -150,13 +150,11 @@
 	name = "cryogenic freezer"
 	desc = "A man-sized pod for entering suspended animation."
 	icon = 'icons/obj/Cryogenic2.dmi'
-	icon_state = "cryopod_0"
+	icon_state = "cryopod"
 	density = 1
 	anchored = 1
 	dir = WEST
 
-	var/base_icon_state = "cryopod_0"
-	var/occupied_icon_state = "cryopod_1"
 	var/on_store_message = "has entered long-term storage."
 	var/on_store_name = "Cryogenic Oversight"
 	var/on_enter_occupant_message = "You feel cool air surround you. You go numb as your senses turn inward."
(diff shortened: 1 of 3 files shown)


TorinoFermic:
On hold while I refactor this code into /mob/ procs

TorinoFermic:
Done refactoring the code for now.

code/modules/mob/living/living.dm
```diff
@@ -452,6 +452,48 @@ default behaviour is:
...
+			if(O.owner && O.owner.current)
+				to_chat(O.owner.current, SPAN_WARNING("You get the feeling your target is no longer within your reach..."))
+			qdel(O)
+
+	//Same for contract-based objectives.
+	for(var/datum/antag_contract/contract in GLOB.all_antag_contracts)
+		contract.on_mob_despawned(src.mind)
+
+	//Handle job slot/tater cleanup.
+	var/job = src.mind.assigned_role
```
MadmanMartian, line 471: Might want to check if there's a mind first

TorinoFermic: With if(src.mind) check ?

MadmanMartian: yes

TorinoFermic: Done

code/game/machinery/cryopod.dm
```diff
@@ -280,9 +286,12 @@
 // Also make sure there is a valid control computer
 /obj/machinery/cryopod/proc/despawn_occupant()
 	var/mob/living/carbon/human/H = occupant
-	var/list/occupant_organs = H.organs | H.internal_organs
+	var/list/occupant_organs
```
ACCount12, line 289: Does `W in occupant_organs` work correctly if `occupant_organs` is null?

code/modules/mob/living/living.dm
```diff
@@ -452,6 +452,48 @@ default behaviour is:
 
 	return
 
+// This function is not to be invoked when the mob is killed. It runs when the mob is leaving the game or similar means.
```
ACCount12, line 455: Reenglish?

code/game/machinery/cryopod.dm
TorinoFermic: The for() will exit early when occupant_organs is null http://www.byond.com/docs/ref/index.html#/proc/for/list

code/modules/mob/living/living.dm
TorinoFermic: Fixed

PR #4806, opened 2020-04-11
The C20R magazine now spawns empty when printed

MadmanMartian:
As requested by a scientist, who accidentally murdered the guild merchant as "he had no non-lethal option"


Pariah919:
wouldn't it be be better to allow printing of rubber and lethal

MadmanMartian:
Only way I can think of justifying this, is to have a new constructable machine that would let you print individual rounds, like a reloading bench.

PR #4822, opened 2020-04-25
Fixes the paper hat

TorinoFermic:
## About The Pull Request

Fixes the missing hat overlay for the written paper
Future refactors on worn icons from big dmi files into each item dmi files as per aurora system to reduce the number of files and simplify addition of new hats/etc


Kurgis:
Could you remove the four additional colors this adds, by swapping them with the four colors from the paper icon state?

TorinoFermic:
I am not aware that there's 4 additional colors

Issue #4632, opened 2020-02-18
[BUG] Electric chair can be built, but does nothing


infraredbaron:

hyperioo:
Can be closed. Fix PR #4837 has been merged.

PR #4837, opened 2020-05-04
Fix electric chair not working

hyperioo:
## About The Pull Request

Fix issue #4632 with the electric chair not working.

The origin of the issue was a missing call to the proc/shock() function of the chair.

Increased damage to kill (total damage > 200) instead of leaving the victim in crit.

## Changelog
:cl: Hyperio
fix: Fixed electric chair not working
/:cl:

### DIFF
--- a/code/game/objects/structures/electricchair.dm
+++ b/code/game/objects/structures/electricchair.dm
@@ -62,8 +62,7 @@
 		sleep(1)
 		buckled_mob.burn_skin(110)
 		buckled_mob.Stun(600)
-	visible_message(SPAN_DANGER("The electric chair went off!"))
-	visible_message(SPAN_DANGER("You hear a deep sharp shock!"))
+	visible_message(SPAN_DANGER("The electric chair went off!"), SPAN_DANGER("You hear a deep sharp shock!"))
 	icon_state = "echair0"
 
 	A.power_light = light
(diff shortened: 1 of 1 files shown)


code/game/objects/structures/electricchair.dm
```diff
@@ -63,13 +57,16 @@
...
-		buckled_mob.burn_skin(85)
+		buckled_mob.burn_skin(110)
 		to_chat(buckled_mob, SPAN_DANGER("You feel a deep shock course through your body!"))
 		sleep(1)
-		buckled_mob.burn_skin(85)
+		buckled_mob.burn_skin(110)
 		buckled_mob.Stun(600)
-	visible_message(SPAN_DANGER("The electric chair went off!"), SPAN_DANGER("You hear a deep sharp shock!"))
+	visible_message(SPAN_DANGER("The electric chair went off!"))
+	visible_message(SPAN_DANGER("You hear a deep sharp shock!"))
```
mattroks101, line 66: This second param for visible message is what you hear when you're blind. This is just playing both messages for everyone.

hyperioo: Thanks for your review. My bad, I did not know that. I reverted this modification.

hyperioo:
There is no cooldown mechanism to trigger the chair (just like before I think?).

Do you fear someone could abuse that, for instance to deplete the APC battery? Right now the cooldown is kind of natural because you have to right click on the chair then click Toggle electric chair to trigger it.

Issue #2828, opened 2019-02-08
BUG: Crafting system problems


Coureere:
1. I believe that multiple structures starting anchored when constructed via crafting menu, which is unintended because you want it to be unwrenched/unscrewed for easy repositioning. Can confirm only on rails.
2. Craft doesn't respect direction that your character facing. It resulting in any construction facing south upon construction regardless your character or assembly positioning. It's bad because you can't turn modular console for example.
I'l add more if i'l encounter something.
3. Crafting system can't properly stack multiple material items. For example, when you crafting steel rods, it puts first in your hand and the rest on the ground below you. One by one. And you need to gather it yourself.
4. New crafting system is badly categorized. For example, there is like 6 entries for comfy chairs when it can be only one clickable button that leads to color menue. It's straight downgraid from old system, which had it.

RandomColonist:
Updated.

hyperioo:
I will try to fix some of those things. I will keep you updated.

hyperioo:
**1.** I would need more details about which structures are concerned. Computer frames start unwrenched, same for AI turrets. Module consoles and machine frames start wrenched but you cannot move them anyway since wrench is used to deconstruct them. Only structure I found wrenched from the start is the bookshelf but I guess it's not very used. Maybe I'm missing some stuff.

**2.** Can confirm for modular console though it's not as worse as written in the issue. The direction depends on your direction but is inversed. If you're facing west the console spawns facing west too when it should spawn facing east (to have the keyboard on your side). Will fix that.

**3.** Already fixed by someone.

**4.** Can confirm, will see what I can do.

hyperioo:
Can be closed. Fix PR #4852 and #4854 have been merged.

PR #4852, opened 2020-05-08
New way to define in which direction a craft should be spawned

hyperioo:
## About The Pull Request

Related to 2. of issue #2828 

This pull request adds new variables to the craft_recipe class in order to have a more precise crafting direction. By default all crafted items spawn in the direction of the user (`dir_type = 0`) but it is possible to change this behaviour by spawning an item toward the user (`dir_type = 1`) or in a default direction (`dir_type = 2` and default direction in `dir_default`).

## Why It's Good For The Game

Currently all crafts spawns in the direction the user is looking. However, for some crafts like modular consoles it would make more sense to spawn them facing the user because that is the way they are supposed to be used. For modular consoles, we want the keyboard toward the user. For other objects it could make sense to always spawn them in a given direction whatever the user's direction is.

## Changelog
:cl: Hyperio
code: New way to define in which direction a craft should be spawned
/:cl:


code/datums/craft/recipe.dm
```diff
@@ -22,7 +26,13 @@
 /datum/craft_recipe/proc/spawn_result(obj/item/craft/C, mob/living/user)
 	var/atom/movable/M = new result(get_turf(C))
 	M.Created(user)
-	M.dir = user.dir
+	switch (C.recipe.dir_type)
+		if (0)  // spawn the result in the user's direction
```
MadmanMartian, line 30: Please use defines

hyperioo: Done. Wasn't very inspired for the name of the constants...

SpadesSlick:
Please resolve conflicts.

PR #4853, opened 2020-05-08
rework hydro to less frustrating

Autisem:
## About The Pull Request
Hydroponics is pretty weak and boring in my opinion, id like to add a few changes and rework it so that its a bit more fun and less of a chore. Small steps for now

## Why It's Good For The Game
Hydroponics right now feels like a big chore for how comparatively weak it is, not only to other hydro systems like goons and TG's but also compared to other departments as a whole. There are some minor tweaks that already make hydro MUCH more fun like having a sink so you dont have to beg for water or the real hydro equipment so more seasend botanists can do some more advanced stuff.

## Changelog
:cl:
add: A sink for NeoTheologists hydroponics
add: Both gene machines are added to NoeTheologists hydroponics
add: Atropa belladonna, a new berry which contains atropine
add: Atropine, which stops death and heals oxy damage, can be OD'd and will cause oxy damage when it does
bugfix: fixed roach taming not working, it can now be done with potato's
wip: way to boost plant potency via chems in trays
balance: Genes are now unobfuscated, meaning that it will say what they do instead of the guessing game

/:cl:

### DIFF
--- a/cev_eris.dme
+++ b/cev_eris.dme
@@ -1547,6 +1547,7 @@
 #include "code\modules\hydroponics\seed_packets.dm"
 #include "code\modules\hydroponics\seed_storage.dm"
 #include "code\modules\hydroponics\beekeeping\beehive.dm"
+#include "code\modules\hydroponics\hydrochems\medicinalplants.dm"
 #include "code\modules\hydroponics\spreading\spacevines.dm"
 #include "code\modules\hydroponics\spreading\spreading.dm"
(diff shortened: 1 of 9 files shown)


Autisem:
Currently a issue with accidently removed some supply pipes. Gonna have to fix that first

Autisem:
I'd like to test some of these changes in the near future. There is a problem where I locally cant use anything on the floor but this also happens when I DIRECTLY pull from the main git so I don't think the problem will be on the server. For now, its mostly place holding stuff and example plant being atropine

PR #4855, opened 2020-05-08
Changed how the radio prefix is read irregardless of capitlization used 

TorinoFermic:
## About The Pull Request

Added lowertext() to this function that handles radio prefixes.

## Why It's Good For The Game

The radio prefix will work whatever if they are in capitals or not.
The drawback is it limits to 26 letters and 10 single digits.

## Changelog
:cl:
tweak: Radio prefix will now work with any capitalization 
/:cl:

### DIFF
--- a/code/modules/mob/say.dm
+++ b/code/modules/mob/say.dm
@@ -149,12 +149,12 @@
 //parses the message mode code (e.g. :h, :w) from text, such as that supplied to say.
 //returns the message mode string or null for no message mode.
 //standard mode is the mode returned for the special ';' radio code.
-/mob/proc/parse_message_mode(var/message, var/standard_mode="headset")
-	if(length(message) >= 1 && copytext(message,1,2) == get_prefix_key(/decl/prefix/radio_main_channel))
+/mob/proc/parse_message_mode(var/message, var/standard_mode = "headset")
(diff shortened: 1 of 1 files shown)


Autisem:
So if I get this right, this change would cause sentance over 26 letter to not work?

TorinoFermic:
> 
> 
> So if I get this right, this change would cause sentance over 26 letter to not work?

More like my PR will limit the radio prefix to these letters and numbers in all lower cases unless we add 2-letter prefixes

Autisem:
> 
> 
> > So if I get this right, this change would cause sentance over 26 letter to not work?
> 
> More like my PR will limit the radio prefix to these letters and numbers in all lower cases unless we add 2-letter prefixes

not sure if we would need THAT many prefixes, atleast for radio's. Dont think there is a major down side to this

TorinoFermic:
> 
> 
> > > So if I get this right, this change would cause sentance over 26 letter to not work?
> > 
> > 
> > More like my PR will limit the radio prefix to these letters and numbers in all lower cases unless we add 2-letter prefixes
> 
> not sure if we would need THAT many prefixes, atleast for radio's. Dont think there is a major down side to this

Hum, I believe that won't be a problem with future proofing for combat radios usage that uses the same prefix for any factions but on their private channel.

PR #4859, opened 2020-05-11
German prefix fix 

Smokingsquirrel:
Hopefully fixes the prefix problem when changed to "d" to no longer conflict with ling speech.


Smokingsquirrel:
Wait is this all files again...? If yes, i will have to open a new PR

PR #4867, opened 2020-05-12
Fixes the sanity runtime

TorinoFermic:
## About The Pull Request

Fixes many runtimes because wrong var and typepath being used for OnDrug intended for actual drugs, not toxins.


ACCount12:
I don't think you need to make changelog entries for small changes that are invisible to the end user - such as runtime fixes.

TorinoFermic:
> 
> 
> I don't think you need to make changelog entries for small changes that are invisible to the end user - such as runtime fixes.

Well, drexemple asked me. But alright, I am removing it.

PR #4882, opened 2020-05-16
Fixes the runtime with floor sound

TorinoFermic:
## About The Pull Request
Fixes the bunch of runtimes with the missing sound for the floor in space.
The caller for this function can take these nulls and be fine.

## Changelog
:cl:
fix: Missing sound for catwalks in space.
/:cl:

### DIFF
--- a/code/game/turfs/turf.dm
+++ b/code/game/turfs/turf.dm
@@ -265,8 +265,11 @@ var/const/enterloopsanity = 100
 	var/obj/structure/catwalk/catwalk = locate(/obj/structure/catwalk) in src
 	if(catwalk)
 		sound = footstep_sound("catwalk")
-	else
-		sound =  footstep_sound(flooring.footstep_sound)
+	else if(flooring)
+		sound = footstep_sound(flooring.footstep_sound)
+	else if(initial_flooring)
+		var/decl/flooring/floor = initial_flooring
+		sound = footstep_sound(floor.footstep_sound)
 
 	return sound
 


code/game/turfs/turf.dm
```diff
@@ -265,8 +265,11 @@ var/const/enterloopsanity = 100
 	var/obj/structure/catwalk/catwalk = locate(/obj/structure/catwalk) in src
 	if(catwalk)
 		sound = footstep_sound("catwalk")
-	else
-		sound =  footstep_sound(flooring.footstep_sound)
+	else if(flooring)
```
drexample, line 268: as a precaution - could you make a default value for sound variable, so if something goes wrong it doesn't return null.

TorinoFermic: What happens if someone tries to spacewalk without any floor under him ? Does the default value will default to an actual sound or empty sound ( wasting processing when it can run without sound ) ?

TorinoFermic:
The floor sound won't play if the mob is on a space turf or openspace.
The caller is fairly single in the codebase.

```
	var/footsound = T.get_footstep_sound()
	if(footsound)

		var/range = -(world.view - 2)
		var/volume = 70
		if(MOVING_DELIBERATELY(src))
			volume -= 45
			range -= 0.333
		if(!shoes)
			volume -= 60
			range -= 0.333

		playsound(T, footsound, volume, 1, range)
```

PR #4909, opened 2020-05-22
Port various antiraid/antigrief measures

Arokha:
Gifts from elsewhere.

I tested these a reasonable amount, but I'd suggest you test them too, since the testing would be more authentic if you use them with your real settings and real DB and such.

## Panic Bunker (config/verb)
When enabled, only players who have previously joined and been logged in the DB will be allowed to join. If they have not been logged in the DB, they will be kicked, and not logged in the DB (otherwise they could just rejoin and then they'd pass the test, you know?). Config option to enable it always (probably not a good idea), and R_ADMIN verb for toggling ingame.

## Paranoia Logging (config/verb)
When enabled, admins will be warned in two cases:
- A player joins who has never joined before
- A player joins with a very new byond account (0-2 days old)

Available as a config option to enable it forever, or an R_ADMIN verb ('New Player Logging').

## IP Reputation Checking (config/verb)
When enabled, connecting IPs will be queried through 1 or 2 (your option) IP reputation scoring services. The system will then take the worst score (normalized to 0.0 to 1.0, with higher being worse), and compare it to your config threshold for worst allowed score. It will then either alert admins, or kick them.

Additionally, you can configure a time period after which the player is effectively whitelisted and no longer checked (default is that their DB first seen time is 5 days ago). This cuts down on queries to the IP reputation services, and I'd recommend it, otherwise they may begin dropping your queries. This can be toggled via a verb ingame if there are issues, and ignores all admins with regards to testing their IPs.

## New configurations
```
###### IP Reputation Checking
# Enable/disable IP reputation checking (present/nonpresent)
#IP_REPUTATION

# Set the e-mail address problems can go to for IPR checks (e-mail address)
IPR_EMAIL whatever@whatever.com

# Above this value, reputation scores are considered 'bad' (number)
IPR_BAD_SCORE 1

# If you want the people disconnected. Otherwise it just logs. (present/nonpresent)
IPR_BLOCK_BAD_IPS

# If players of a certain length of playtime are allowed anyway (REQUIRES DATABASE) (present/nonpresent)
IPR_ALLOW_EXISTING

# And what that age is (number)
IPR_MINIMUM_AGE 5

# If provided, will look up additional IP reputation info from ipqualityscore.com
#IPQUALITYSCORE_APIKEY ABC12345YOURKEYHERE

###### Antiraid Settings
# Enable/disable 'paranoia logging' (notifying admins of new byond accounts joining, and player first-time joins)
#PARANOIA_LOGGING

# Enable/disable 'panic bunker' (prevents new players from joining if they've never been seen before in the DB)
#PANIC_BUNKER
```

### DIFF
--- a/code/modules/client/client procs.dm
+++ b/code/modules/client/client procs.dm
@@ -279,6 +279,7 @@
 /client/proc/register_in_db()
 	registration_date = src.get_registration_date()
 	src.get_country()
+	src.get_byond_age() // Get days since byond join
 
 	var/DBQuery/query_insert = dbcon.NewQuery("INSERT INTO players (ckey, first_seen, last_seen, registered, ip, cid, rank, byond_version, country) VALUES ('[src.ckey]', Now(), Now(), '[registration_date]', '[sql_sanitize_text(src.address)]', '[sql_sanitize_text(src.computer_id)]', 'player', [src.byond_version], '[src.country_code]')")
 	if(!query_insert.Execute())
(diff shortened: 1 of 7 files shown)


Arokha:
Wha? Your DB should already have that column. It's in the schema.sql in the root of your repo. Your existing code is already logging it. It's actually different from ours which is `firstseen` without the underscore and messed me up at first.

TorinoFermic:
> schema

I checked our file. Looks like we will be fine but waiting for other maintainers to check and approve. That's a big thing to check and clear it.

@MadmanMartian 
@ACCount12

code/modules/client/client procs.dm
```diff
@@ -427,4 +489,111 @@
...
+	var/request = "http://check.getipintel.net/check.php?ip=[address]&contact=[config.ipr_email]"
+	var/http[] = world.Export(request)
+
+	if(!http || !islist(http)) //If we couldn't check, the service might be down, fail-safe.
+		log_admin("Couldn't connect to getipintel.net to check [address] for [key]")
+		return -1
+
+	//429 is rate limit exceeded
+	if(text2num(http["STATUS"]) == 429)
+		log_and_message_admins("getipintel.net reports HTTP status 429. IP reputation checking is now disabled. If you see this, let a developer know.")
```
LethalGhost, line 531: BTW. Can't that message spam to admins too much (in case of service failure)?
May be it should be debug log?

PR #4919, opened 2020-05-23
less .charge, more .give

Evankhell561:
## About The Pull Request
this ensures that the actions are executed the power cell has enough battery.
del src. useless
and replace 0 and 1 with booleans

## Why It's Good For The Game
Better code, less stuff. 
prevents power cells from having more charge than they should

## Changelog
:cl:
code: more cell.give and cell.check_charge
/:cl:

### DIFF
--- a/code/modules/vehicles/cargo_train.dm
+++ b/code/modules/vehicles/cargo_train.dm
@@ -213,7 +213,7 @@
 	if (on)
 		to_chat(usr, "You start [src]'s engine.")
 	else
-		if(cell.charge < charge_use)
+		if(!cell.check_charge(charge_use))
 			to_chat(usr, "[src] is out of power.")
 		else
 			to_chat(usr, "[src]'s engine won't start.")
--- a/code/modules/vehicles/cargo_train.dm
+++ b/code/modules/vehicles/cargo_train.dm
@@ -42,7 +42,7 @@
 	..()
 	cell = new /obj/item/weapon/cell/large/high(src)
 	key = new(src)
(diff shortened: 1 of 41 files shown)


code/modules/vehicles/cargo_train.dm
```diff
@@ -360,9 +360,9 @@
 // more engines increases this limit by car_limit per
 // engine.
 //-------------------------------------------------------
-/obj/vehicle/train/cargo/engine/update_car(var/train_length, var/active_engines)
-	src.train_length = train_length
-	src.active_engines = active_engines
+/obj/vehicle/train/cargo/engine/update_car(train_length, active_engines)
```
TorinoFermic, line 363: I would keep these src. because parameters are same or change parameter names for something sane

code/modules/vehicles/cargo_train.dm
```diff
@@ -372,11 +372,11 @@
 		move_delay *= (1 / max(1, active_engines)) * 2 										//overweight penalty (scaled by the number of engines)
 		move_delay *= 1.1																	//makes cargo trains 10% slower than running when not overweight
 
-/obj/vehicle/train/cargo/trolley/update_car(var/train_length, var/active_engines)
-	src.train_length = train_length
-	src.active_engines = active_engines
+/obj/vehicle/train/cargo/trolley/update_car(train_length, active_engines)
```
TorinoFermic, line 375: ditto here, Weird github did not flag all these lines

PR #4925, opened 2020-05-24
Adjusts burrow migrations

Kurgis:
## About The Pull Request

Certain storytellers are less likely to re-roll the area a migration occurs in if it's not maintenance. I did a similar PR to this awhile ago, but the results which play out in-game are less than spectacular. This should help keep PvE more active on Jester.

## Why It's Good For The Game

Jester doesn't end like a pseudo-sleeper when mobs spawn in only isolated areas, now they'll move back into populated areas by happen-stance more likely.

## Changelog
:cl:
tweak: Migration events should be more aggressive when The Jester is selected.
/:cl:


code/controllers/subsystems/migration.dm
```diff
@@ -214,10 +214,10 @@ This proc will attempt to create a burrow against a wall, within view of the tar
 	var/obj/structure/burrow/candidate
 
 	switch (GLOB.storyteller.config_tag)
-		if ("jester") // Jester is much more likely to not reroll the maintenance check.
-			reroll_prob = 59.5
+		if ("jester") // Jester will most likely not reroll the maintenance area check.
+			reroll_prob = 0.05
```
SpadesSlick, line 218: I have a suspicion it will end up badly.  Jump from 60 to 0.05 is too strong, put something like 20 and we shall see if it works.

Kurgis: I'd rather have it be "too strong", and then tone it back if there's actually issues. This is more to gauge how migrations go down if there's no re-roll chance, the only reason the ".05" exists is to prevent a division by zero scenario.

SpadesSlick: Nerfing things too much in one go is against our contribution guidelines. If 20 will not do the job, we will nerf it further.

Or buff. Depends on how to observe it. Does not really matter here.

PR #4927, opened 2020-05-25
Actualización

Silverexperience:
## About The Pull Request
Actualización
## Why It's Good For The Game
Actualización
## Changelog
:cl:
add: Actualización 25/05/2020 2:XX
/:cl:


Silverexperience:
Sorry it was a mistake.

MadmanMartian:
That is a good choice of colour for the medical radio. Mind if I patch that in?

Evankhell561:
> That is a good choice of colour for the medical radio. Mind if I patch that in?

You can take anything you like from our repo, if we do not do our prs here it is because we are not sure that they are accepted and thus we do not do double work.
However, that color is a bit too light if you don't have night mode on.

PR #4932, opened 2020-05-26
Grenade Launcher Rework

NotDeirun:
## About The Pull Request

Changes all handheld grenade launchers (The Protector, the Lenar, and the Z8 underslug launcher) to use the recently added projectile grenades (Code wise, it turns them into projectile guns), makes the China Lake acquirable, and adds various ways to get the ammo needed for the launchers, more details below.

Both the China Lake and it's gun disk can be found in trash.
The Ammo can be found inside the IH armoury, inside trash, bought from an uplink, and can be printed from grenade launcher gun disks and from a newly added grenade launcher ammo disk.
Serbians can also find some inside the grenadier crate.
The grenade launcher ammo disk can be found in trash and be bought from a hacked Asters Guild Print-o-Mat (Just like the lethal ammo disk).

If you have any feedback about the balancing side of things, please leave a comment, I am personally not sure how easy it should be to acquire the launchers and the ammo.

## Why It's Good For The Game

Having grenade launchers shoot literal hand grenades is unintuitive, this makes Grenade Launchers work more like people would expect it. It also allows the projectiles to explode in the same moment they arrive at the targeted turf.

## Changelog
:cl:
add: Adds the China Lake grenade launcher and new grenade launcher ammo to the game itself.
tweak: Changes handheld grenade launchers (The Protector, the Lenar, and the Z8 underslug launcher) to use the newly added grenade shells instead of hand grenades.
/:cl:


SpadesSlick:
Please return them ability to use normal nades as well, in old method, so we would have both dedicated ammo for this and support for things like chem nades.

NotDeirun:
When I accepted the order to create the china lake I was told to only make it work with the new shells, I do not think that it is a good idea to have some grenade launchers be able to shoot hand grenades while others can not.
I think creating new ammo types that can emulate the behaviour of the other hand grenades would be better, including making a customizable chem grenade ammo shell.

SpadesSlick:
Emulating is kinda pointless, as we need ability to put the cleaning nades into GLs without any modifications to ammo or gun.

NotDeirun:
In the hypothetical scenario of a new chem grenade ammo shell it would be possible to easily just code a cleaning grenade shell (If that is really needed).

SpadesSlick:
But we don't need that. We need it to support all types of nades as it was doing before, and does it on any other codebase.

Having dedicated ammo is good, but support for everything that was left out is also needed.

NotDeirun:
That would still leave the wierdness of the china lake working differntly from all other launchers (I was told to only make it work the new shells)

SpadesSlick:
By who exactly?

NotDeirun:
By Jack Frost - check the orders channel, it is one of the first messages there.

SpadesSlick:
Well, he is not a maintainer. My point still stands, GLs should have backward compatibility to old nades.

NotDeirun:
I do not think that it is needed, but will do so if a maintainer officially tells me to do so.

NotDeirun:
Reere wants backward compatibility so I will add it. Please do not merge this before I get that done.

code/datums/uplink/ammunition.dm
```diff
@@ -189,6 +189,31 @@
...
 
+////Grenade Shells////
+
+/datum/uplink_item/item/ammo/grenade_baton
+	name = "6xbaton rounds"
+	desc = "A box containing 6 baton rounds. Designed to be fired from a grenade launcher."
+	item_cost = 4
+	path = /obj/item/weapon/storage/box/baton_rounds
+/datum/uplink_item/item/ammo/grenade_emp
+	name = "6xemp grenade shells"
```
MadmanMartian, line 200: Change to EMP grenade shells x6, same for the other uplink items you're adding

NotDeirun: I named them in the same way actual grenades inside the upload are named, want me to change those names too while I am at it?

drexample:
@NotDeirun  This PR will be closed in 7 days.

NotDeirun:
Making hand grenades loadable was easy, but making them and the shells fireable seems to be way harder, I am not sure if I can do it. If this is closed I will reopen it if I can get it done.

PR #4933, opened 2020-05-26
Tweak roach and spider behavior with windows

hyperioo:
## About The Pull Request

This PR tweaks the behavior of roaches and spiders with windows. They were already supposed to do it but the check for destroying their surrounding was not properly done and the probability of doing it was low.

Now they will try to break full windows, directional windows as well as windoors to reach you if you're behind them. I increased the destroyProperty probability to 100% or else they do not try to actively break the window, they just bash it with a few seconds between each hit.

I also set the spider's density to 0 (like roaches) so that they are able to pass over low walls to reach you.

They are no pressure check because :
- I doubt roaches and spiders can think "Maybe I should not break that thing" when they see a juicy prey on the other side of a filmsy piece of glass, even if said prey is in space.
- It's part of the danger of space, some dumb animals may cause breaches from time to time, add more surprise to maintenance if you suddenly run into a breach
- Breaching don't matter too much anyway because shields are up most of the time so even with a broken window the air does not go away.

Time to run!

alt text

## Why It's Good For The Game

Roaches and spiders will no longer be stopped by windows like braindead animals. They're supposed to be enraged. Run for your life!

## Changelog
:cl: Hyperio
tweak: Tweak roach and spider behavior with windows
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/superior_animal/ai.dm
+++ b/code/modules/mob/living/carbon/superior_animal/ai.dm
@@ -56,9 +56,23 @@
 
 /mob/living/carbon/superior_animal/proc/destroySurroundings()
 	if (prob(break_stuff_probability))
+
+		for (var/obj/structure/window/obstacle in src.loc) // To destroy directional windows that are on the creature's tile
+			obstacle.attack_generic(src,rand(melee_damage_lower,melee_damage_upper),attacktext)
+			return
+		
+		for (var/obj/machinery/door/window/obstacle in src.loc) // To destroy windoors that are on the creature's tile
+			obstacle.attack_generic(src,rand(melee_damage_lower,melee_damage_upper),attacktext)
+			return
+
 		for (var/dir in cardinal) // North, South, East, West
 			for (var/obj/structure/window/obstacle in get_step(src, dir))
-				if (obstacle.dir == reverse_dir[dir]) // So that windows get smashed in the right order
+				if ((obstacle.is_full_window()) || (obstacle.dir == reverse_dir[dir])) // So that directional windows get smashed in the right order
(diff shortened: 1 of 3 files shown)


Kurgis:
> I also set the spider's density to 0 (like roaches) so that they are able to pass over low walls to reach you.
@hyperioo Wouldn't this cause multiple spiders to stack up on a tile? A single spider can be quite lethal on its own, and they're appropriately beefy.

hyperioo:
> 
> 
> > I also set the spider's density to 0 (like roaches) so that they are able to pass over low walls to reach you.
> 
> @hyperioo Wouldn't this cause multiple spiders to stack up on a tile? A single spider can be quite lethal on its own, and they're appropriately beefy.

@Kurgis Yes they will be able to stack. Other people told me the same thing on Discord. Density = 0 was an easy way for me to allow spiders to go over low walls (they are stuck by them currently). I will find a way to do it properly by using a flag or something like that to allow spiders to go over them without touching their density

SpadesSlick:
DNM until stacking is resolved.

hyperioo:
I reversed the change of spider density to avoid stacking.

For the spiders not going over low wall I think it's a path planning issue since when I manually control one I can go over low walls. 

 There are no calls to /obj/structure/low_wall/CanPass so it means the spider AI does not even try to go over it. In fact low_wall density is 1 so the byond function `walk_to` considers the low wall as a full wall. As a result the spider does not try to go over it, Manually editing a low_wall density to 0 in game allows the spider goes over it (with calls to CanPass).

I believe solving this issue is out of the scope of this PR so for now spiders will stay unable to go over low walls.

PR #4937, opened 2020-05-27
Fix runtime in who.dm

hyperioo:
## About The Pull Request

Fix runtime in who.dm when trying to read `C.mob.stat` when `C.mob` is `null`.

Added a check for the existence of `C.mob`

> [19:42:54] Runtime in who.dm,18: Cannot read null.stat
>    proc name: Who (/client/verb/who)
>   usr: Fisk06 (fisk06) (/mob/new_player)
>   src: Fisk06 (/client)
>   call stack:
>   Fisk06 (/client): Who()

## Why It's Good For The Game

Runtimes are bad.

## Changelog
:cl: Hyperio
fix: Fixed runtime in who.dm
/:cl:

### DIFF
--- a/code/game/verbs/who.dm
+++ b/code/game/verbs/who.dm
@@ -28,6 +28,8 @@
 								entry += " - <font color='black'><b>DEAD</b></font>"
 						else
 							entry += " - <font color='black'><b>DEAD</b></font>"
+			else
+				entry += " - <font color='gray'>In Lobby</font>"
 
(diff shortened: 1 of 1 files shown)


hyperioo:
I've changed the test to detect if the client is a new player. I hope that's why you had in mind @MadmanMartian

ACCount12:
The best fix would be to make the code actually _handle_ new players instead of skipping over them - make it write something like "In Lobby" for them.

PR #4940, opened 2020-05-27
Fix runtime in mind.dm

hyperioo:
## About The Pull Request

Fix runtime in mind.dm when trying to check `current.mind` with `current` being `null`.

Considering that at the beginning there is a check for 
```
if(current)	//remove ourself from our old body's mind variable
	if(changeling)
		[...]
```
the possibility that `current` could be `null` seems to be expected.
 
> [19:04:26] Runtime in mind.dm,96: Cannot read null.client
>    usr: Francis Fuchs (as Pablo Mand) (coolfoolftw) (/mob/living/carbon/human)
>   usr.loc: The floor (115,114,2) (/turf/simulated/floor/tiled/dark)

## Why It's Good For The Game

Runtimes are bad.

## Changelog
:cl: Hyperio
fix: Fixed runtime in mind.dm
/:cl:

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -90,12 +90,12 @@
 		current.mind = null
 
 		SSnano.user_transferred(current, new_character) // transfer active NanoUI instances to new user
-	if(new_character.mind)		//remove any mind currently in our new body's mind variable
-		new_character.mind.current = null
 
-	if (current) // check if current exists otherwise we cannot read current.client
 		if(current.client)
 			current.client.destroy_UI()
+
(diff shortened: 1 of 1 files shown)


code/datums/mind.dm
```diff
@@ -93,8 +93,9 @@
 	if(new_character.mind)		//remove any mind currently in our new body's mind variable
 		new_character.mind.current = null
 
-	if(current.client)
-		current.client.destroy_UI()
+	if (current) // check if current exists otherwise we cannot read current.client
```
TorinoFermic, line 96: As I recall, you can combine the first IF and below IF and it will compile and react as you intend for this fix.

PR #4961, opened 2020-06-03
Propaganda_chip runtime fix

TheShown911:
## About The Pull Request

*Should* fix the runtime in propaganda_chip.dm,77: Cannot read null.real_name

## Why It's Good For The Game

Runtimes are bad.

### DIFF
--- a/code/game/objects/items/devices/propaganda_chip.dm
+++ b/code/game/objects/items/devices/propaganda_chip.dm
@@ -71,7 +71,7 @@ obj/item/device/propaganda_chip/Destroy()
 		if (crew_target_mind)
 			break
 	var/datum/mind/crew_name
-	if (!crew_target_mind)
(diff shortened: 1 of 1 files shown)


code/game/objects/items/devices/propaganda_chip.dm
```diff
@@ -71,7 +71,7 @@ obj/item/device/propaganda_chip/Destroy()
 		if (crew_target_mind)
 			break
 	var/datum/mind/crew_name
-	if (!crew_target_mind)
+	if (!crew_target_mind && !(crew_target_mind.current))
 		crew_name = "Unknown"
 	else
```
AlexMorgan3817, line 76: better be to put it here
`else if(crew_target_mind?.current)`
Cuz if u try to get null.current where `!crew_target_mind` = TRUE, u will get more runtimes

PR #4994, opened 2020-06-15
Gun tags now autogen for most weapons

MadmanMartian:
Rather than going through each and every gun to add the guntags individually, we are instead going to use a procedural method.

:cl:
fix: The ergonomic grip can now be attached to guns that require two hands to fire accurately.
/:cl:


code/modules/projectiles/guns/projectile/shotgun/regulator.dm
```diff
@@ -13,5 +13,4 @@
 	damage_multiplier = 1.1
 	penetration_multiplier = 1
 	recoil_buildup = 16
-	one_hand_penalty = 15 //full sized shotgun level
-
+	one_hand_penalty = 15 //full sized shotgun level
```
Kurgis, line 16: Newline got removed here.

PR #4996, opened 2020-06-15
Adds duffelbags to preferences

Kurgis:
## About The Pull Request

Duffelbags will now be available in character preferences.

## Why It's Good For The Game

More choice between certain benefits but downsides to access similar to backpacks.

## Changelog
:cl:
tweak: You may now select duffelbags in your character preferences.
/:cl:

### DIFF
--- a/code/datums/outfits/equipment/backpacks.dm
+++ b/code/datums/outfits/equipment/backpacks.dm
@@ -25,6 +25,10 @@
 	path = /obj/item/weapon/storage/backpack/satchel
 	is_default = TRUE
 
+/decl/backpack_outfit/duffelbag
+	name = "Duffelbag"
+	path = /obj/item/weapon/storage/backpack/duffelbag
+
 /* Code */
 /decl/backpack_outfit
 	var/flags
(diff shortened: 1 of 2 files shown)


SpadesSlick:
That's not "certain benefits", those things can store x1,5 more comparing to backpacks. They supposed to be a loot reward, not game starting option.

MadmanMartian:
What slick said

Pariah919:
>That's not "certain benefits", those things can store x1,5 more comparing to backpacks. They supposed to be a loot reward, not game starting option.
>
They can only really be used on the floor if I remember right. They're also not really that great, storing 1.5x compared a backpack for a more annoying to use item that is basically unusable in combat unless your click dragging skills are on point, though I suppose you can carry more items if you're looting. Though 2 backpacks are probably better and less cumbersome to use.
I don't see anything wrong with allowing them to be used roundstart outside of loot pinatas, though if you need that much space I'd rather just bring 2 backpacks so I don't have to alt click all the time to sift through trash.

SpadesSlick:
> They can only really be used on the floor if I remember right.

Oh, you right, I somehow missed this. Well, it does make sense then.

Kurgis:
@mattroks101 I believe the balance concern has already been resolved.

> > They can only really be used on the floor if I remember right.
> 
> Oh, you right, I somehow missed this. Well, it does make sense then.

This is already current behavior for duffelbags.

PR #4999, opened 2020-06-16
Sleeper now gives alt-dylovene

MadmanMartian:
## About The Pull Request
I was asked to make this a while back, as right now medbay is just "spam tricord and hope the pain goes away"

## Why It's Good For The Game
Not entirely sure it is, as there aren't any easy methods for treating brute and burn once medkits are used up/stolen

## Changelog
:cl:
balance: Sleepers now give an alternative version of dylovene, to prevent tricord-spam
/:cl:

### DIFF
--- a/code/game/machinery/Sleeper.dm
+++ b/code/game/machinery/Sleeper.dm
@@ -7,7 +7,7 @@
 	anchored = 1
 	circuit = /obj/item/weapon/circuitboard/sleeper
 	var/mob/living/carbon/human/occupant = null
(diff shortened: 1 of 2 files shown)


SpadesSlick:
It is possible to do trick even from first aid contents, so this is fine. People could start using cryo now at least.

SpadesSlick:
It was decided to reject it in favor of proper rework of those systems.

MadmanMartian:
👌

PR #5001, opened 2020-06-17
Stops changelog generator from Xing PRs with no changelog

ACCount12:
Commits shouldn't be flagged with a big scary :x: just because they had no changelog on them.

Reasons to have a PR with no changelog entries:

* Straight up refactor PRs.
* Tooling-related PRs, like this one.
* PRs with changes too small or too technical to bother the players with, like grammar or runtime error fixes.

I should also redo the changelog generator thing one day, so that it wouldn't add two commits for every PR with a changelog in it. But that's a fight for another day.

### DIFF
--- a/tools/changelog/generate_cl.py
+++ b/tools/changelog/generate_cl.py
@@ -60,7 +60,7 @@ try:
     cl_list = CL_SPLIT.findall(cl.group(2))
 except AttributeError:
     print("No CL found!")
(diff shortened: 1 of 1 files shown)


drexample:
No reason is good reason to not have changelog entries.
The X is intended.

ACCount12:
Your changelog addiction is ruining you. I'll have to drag you into rehab if you don't stop.

You'll thank me later.

PR #5018, opened 2020-06-20
Fixes the sanity cap

MadmanMartian:
*description goes here*

:cl:
fix: Sanity no longer tanks from a single drop of blood
/:cl:


SpadesSlick:
Add a changelog entry bruh.
sidecat

PR #5034, opened 2020-06-24
Adds the Back-Alley organ disk

MadmanMartian:
Some bioprinter exclusive organs of questionable quality. Looking for ideas to pad out the disk a little.

Right now, it's just extended lungs, which increase the time between breath checks.


TorinoFermic:
image

PR #5039, opened 2020-06-26
Medical rank prefixes

Ajax05:
## About The Pull Request
Appropriate Moebius personnel will be given the title "Dr.", and the MBO and MEO will be given the title "Professor". These titles will replace the first name in exactly the same way as IH operatives.

## Why It's Good For The Game
Calling doctors as such will add to the overall atmosphere.

## Changelog
:cl:
add: Moebius employees given appropriate titles.
/:cl:


SpadesSlick:
MBO might not be a professor, but honestly, that's beyond the point.

This change will not bring nothing interesting to the game, and will annoy people instead. It was originally discussed on who should gain those, and it was decided that only security and captain truly need it.

PR #5056, opened 2020-07-04
The Sol carbine no longer has insane recoil

Pariah919:
## About The Pull Request
The Sols recoil has been reduced from 13 (reere why) to a more manageable 8. This gun was literally basically worse than a pistol (paco) past point blank because of the recoil, and the kit still makes you have more recoil than a wintermute so I feel this is fine.

## Why It's Good For The Game
Makes the Sol no longer worse than your handgun.

## Changelog
:cl:
balance: Sol has 38ish% less recoil overrall.
/:cl:


SpadesSlick:
Alright, let's try this. Mostly everyone complain about Sol at current state.

PR #5073, opened 2020-07-08
Russian radio keys (e.g. :ы :р :у :ь)

seekin4u:
Since 513 function `copytext()`, used to trim or cut strings, to get some part of it, works with bytes but not with letters, and as english letters are one-byte, and cyrillic are two-byte, cyrillic radio keys were impossible to use. 

So i made them alive using `copytext_char()` which counts letters.

It was not comfortable to switch one of three system languages each time i want to say something to my radio channel.

Also added one encription key for all existing channels.


SpadesSlick:
Last time this was introduced it ended up with radio being broken for everyone, someone else should test this properly.

PR #5099, opened 2020-07-14
Drlling update.

Uvash:
Try to clean drill code.
Fix request components.
Add scanning module. If better - drill will have more large work area.

## Changelog
:cl:
add: Add scanning module in drill. If better - drill will have more large work area.
/:cl:

### DIFF
--- a/code/modules/mining/drilling/brace.dm
+++ b/code/modules/mining/drilling/brace.dm
@@ -1,0 +1,87 @@
+/obj/machinery/mining/brace
+	name = "mining drill brace"
+	desc = "A machinery brace for an industrial drill. It looks easily half-meter thick."
+	icon_state = "mining_brace"
+	circuit = /obj/item/weapon/circuitboard/miningdrillbrace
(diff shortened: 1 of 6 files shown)


code/modules/mining/drilling/brace.dm
```diff
@@ -0,0 +1,87 @@
+/obj/machinery/mining/brace
+	name = "mining drill brace"
+	desc = "A machinery brace for an industrial drill. It looks easily two feet thick."
```
TheShown911, line 3: We should not use imperial measurements.

code/modules/mining/drilling/drill.dm
```diff
@@ -271,96 +269,32 @@
...
-		return FALSE
+	if(supports && supports.len >= braces_needed)
+		supported = TRUE
+	else
+		supported = FALSE
 
-	set_dir(turn(dir, 90))
-	return TRUE
\ No newline at end of file
+	update_icon()
```
TheShown911, line 300: No newline.

PR #5105, opened 2020-07-15
Weintraubfix

Alliostra:
## About The Pull Request
Puts the Weintraubs back into the FS Vendor

## Why It's Good For The Game
Fixes a mistake I guess,

## Changelog
:cl:
fix: Readded Weintraubs to the FS Vendor
/:cl:


Alliostra:
what the fuck why did that check fail, I literally put it back into its original state

Alliostra:
Oh, Im an idiot. will fix.

Alliostra:
>:(

Alliostra:
yeah I give up

SpadesSlick:
It supposed to be cargo/trash loot. Right now Sol is balanced around semi-auto/burst modes, and was buffed recently.

This removal was intended, but I forgot to document it in chagelog. We should see feedback on Sol in later weeks how it works without autofire being easy to mod on. If players will continue to complain, we should see what should be done about this gun.

Alliostra:
The only People Aboard With sols are IH, apart from People Who find a gun stash or a sol disk. the Weintraub is useless without a sol, as a Non-IH, you will almost never find both a sol and a Weintraub. If the Weintraub is maint exclusive, you'll get either trash digging IHs or very rarely a vag and an IH will trade- But almost Always no Weintraub found in maint will be Used in a given round, It is just too specialized.

Alliostra:
The only People Aboard With sols are IH, apart from People Who find a gun stash or a sol disk. the Weintraub is useless without a sol, as a Non-IH, you will almost never find both a sol and a Weintraub. If the Weintraub is maint exclusive, you'll get either trash digging IHs or very rarely a vag and an IH will trade- But almost Always no Weintraub found in maint will be Used in a given round, It is just too specialized.

SpadesSlick:
Cargo rework will bring specialized trade stations, so those things will be procured not only from trash later.

Weintraub was designed more like proof of concept of a firemod that can add new firemods to weapons, and be only obtainable from loot/trade, not vendors. Sol should remain burst only by default and we will try to balance it this way.

PR #5106, opened 2020-07-15
More simple weapon tools

Kurgis:
## About The Pull Request

Captain's Chain of Command, and the Nailstick are now tools which can be upgraded.

## Why It's Good For The Game

Just like broken bottles, you can augment your weapons with "stuff" now.

## Changelog
:cl:
tweak: The Chain of Command and the "Nailstick" are now both tools, and accept tool mods.
/:cl:

### DIFF
--- a/code/datums/craft/recipes/weapon.dm
+++ b/code/datums/craft/recipes/weapon.dm
@@ -88,7 +88,7 @@
 
 /datum/craft_recipe/weapon/nailed_bat
 	name = "nailed bat"
-	result = /obj/item/weapon/melee/nailstick
+	result = /obj/item/weapon/tool/nailstick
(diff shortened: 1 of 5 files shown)


Kurgis:
Some discussion might actually be needed regarding how tools will be handled. Should objects like swords, bottles, and other non-tools remain as their previous weapon types? After an update, an object being a tool is unnecessary for it to contain both upgrades or tool quality. A solution could be just to put them under a subtype of tool, such as `/obj/item/weapon/tool/weapon/sword`, for example.

TheShown911:
A permanent fix should be implemented as mentioned, but i don't see a problem with this banaid fix. It should be remember when and if the fix comes.

PR #5108, opened 2020-07-16
Random trash tweaks

Voldirs:
## About The Pull Request

1)Replaced STS with FS AK because it fits much more, sts is a gun desired by basically everyone
2)Also added c20r smg and ion rifle, it isn't a real gamechanger
3)Added pouches to cloth piles because why not?
4)Lowered chance of drop for red voidsuit
5)Food trash piles now have mostly rotten food inside for obvious reasons

## Why It's Good For The Game

Currently, STS is a top tier gun, why simply putting it to junk pile for everyone to dig? Same goes for blood red voidsuit, though I didn't remove it entirely. But I noticed that pouches were even more rare than guns and there were no spoiled food in food junk, that's ridiculous.

## Changelog
:cl:
tweak: "Tweaked trash piles, much less chance to find STS and blood-red voidsuit in trash, but more for some other guns and pouches"
/:cl:

### DIFF
--- a/code/game/objects/random/guns.dm
+++ b/code/game/objects/random/guns.dm
@@ -29,25 +29,26 @@
 	icon_state = "gun-green"
 
 /obj/random/gun_normal/item_to_spawn()
-	return pickweight(list(/obj/item/weapon/gun/projectile/lamia = 2,\
-				/obj/item/weapon/gun/projectile/automatic/zoric = 1,\
-				/obj/item/weapon/gun/projectile/automatic/atreides = 1,\
-				/obj/item/weapon/gun/projectile/avasarala = 2,\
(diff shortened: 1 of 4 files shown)


Voldirs:
I will be out for a week, just in case someone will have questions about this pr

TheShown911:
I agree with these changes though something like this was rejected before. Also the removal of the STS should be mentioned in the changelog.

Kurgis:
> Since it's already at 1, you need to rebalance numbers on whole gun loot table. Yeah, that's kinda the reason why loot rework thing exists, current loot tables are bad to balance.
@SpadesSlick 

I don't recall Nanako's non-integer support being removed for weight tables, if it wasn't, the value could just be lowered to 0.5.

Voldirs:
Brought requested tweak and edited the changelog

PR #5115, opened 2020-07-16
Tricordrazine Miracle-All, No More.

Kurgis:
## About The Pull Request

Tricordrazine no longer heals toxloss or oxyloss. This intends to require the use of multiple chemicals, instead of one. Sin-pockets are given Synaptizine.

## Why It's Good For The Game

They shall respect the aspect of Chemistry, or they shall die upon roach-laden fields. Their epitaphs shall be forgotten, as they are buried below greater men.

## Changelog
:cl:
balance: Sin-pockets now contain a small amount of Synaptizine.
balance: Tricordrazine only treats minor burn and brute damage, it will no longer affect oxygen or toxin damage.
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/food/snacks.dm
+++ b/code/modules/reagents/reagent_containers/food/snacks.dm
@@ -759,7 +759,7 @@
 	name = "\improper Sin-pocket"
 	desc = "The food of choice for the veteran. Do <B>NOT</B> overconsume."
 	filling_color = "#6D6D00"
-	heated_reagents = list("doctorsdelight" = 5, "hyperzine" = 1)
+	heated_reagents = list("doctorsdelight" = 5, "hyperzine" = 0.5, "synaptizine" = 0.25)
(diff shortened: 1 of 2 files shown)


Pariah919:
oxy damage is fine
tox damge is not

seekin4u:
without toxLoss and oxyLoss this is not old goddamn tric, so maybe you should rename this chemical to something else?

SpadesSlick:
> without toxLoss and oxyLoss this is not old goddamn tric

It never did that om other codebases as much as I remember.

I wonder why those values are even there.

Kurgis:
To be fair, our Medical system is very dated, many don't even use TrekChems like us anymore.

Kurgis:
This PR is no longer necessary after the merge of #5113

ACCount12:
Tricordrazine heals 4 damage types on all codebases that still have it, if anyone is curious.

PR #5119, opened 2020-07-18
Adds a makeshift Laser SMG

Alliostra:
## About The Pull Request
this PR adds a makeshift laser SMG, the Disco Vazer "Lasblender" (the name is cringe on purpose, its supposed to be something Vagabonds think is cool).  It can be crafted with an Atreides, plasteel, glass, cable, a capacitor, an LED and an ansible crystal. The lasers it shoots have randomized colors, for the original cyberpunk aesthetic.

"But is a rapid-fire Hitscan laser weapon not utterly OP?" you may ask. I don't think so, let's compare the Lasblender to the Atreides that it is made of.

The Advantages the Lasblender has are: Hitscan projectiles, about 4 points of extra damage, the usage of rechargeable cells that are easy to acquire, the ability to fire through windows, and the fact that almost nobody wears energy armor.

The Disadvantages are: worse accuracy due to the constant projectile offset (even though the actual recoil is lower), no AP compared to 10 AP, low ammo count even with high-tier cells, and most importantly the fact that it can only fire in bursts of 8 shots. Not only does this mean that you will always be wasting a lot of charge, but it also means that the shots will always be aimed at exactly the point you clicked at for the entirety of the burst. so if you shoot at someone, and the move to the side, YOU need to move in order to make the rest of the burst hit. Doing this not only require practice, but it is not very effective even when done this way since the person you're shooting at can move away as well. And because the lasers are hitscan, if a shot does not fly at your target, it will always miss. actual projectile, on the other hand can miss the tile you clicked at, but may still hit the person you aimed at, which is important especially for spray and pray guns like the atreides- or the lasblende,r which can not utilize this effect.

I'd like to thank Chevy and Icarus for helping me with finding a fitting name, as well as Carbonhell helping me with the code, essentially, he did all of the work.

## Why It's Good For The Game
Adds a cool new makeshift gun with cyberpunk aesthetics, and it won't clutter anything up because it can only be acquired via crafting.

## Changelog
:cl:
add: Adds the "Lasblender" a craftable laser SMG
/:cl:


Alliostra:
grafik

ArmolitskiyCarpoeb:
and who drew the sprite?  show me sprite

ArmolitskiyCarpoeb:
a laser with a random color is a bad idea

Alliostra:
Why do you think so?

Kurgis:
The cell overlays aren't to spriting standards, they'll need to be changed at some point in the future.

Kurgis:
Also, it's nice seeing a firearm with a fresh idea, instead of just performing some variable play.

PR #5120, opened 2020-07-19
Adds a makeshift chest rig that is craftable.

Pariah919:
## About The Pull Request
Adds a chest rig that you can craft and holds four things like the merc vest (Is a child of it. Less armor though.)

TODO-
Get sprites that aren't placeholders from sojourn
Decide harder on the crafting steps. They work right now.

## Why It's Good For The Game
Adds an option to carry more stuff that is craftable, and has a clear tradeoff (loadout jackets can carry less than this. But protect more. Handmade armor vest protects more.)
Adds another craftable option compared to the handmade armor vest.

## Changelog
:cl:
add: Added a chest rig that is craftable and holds four slots.
/:cl:


code/datums/craft/recipes/clothing.dm
```diff
@@ -40,4 +40,13 @@
...
+	)
+
+/datum/craft_recipe/clothing/chest_rig
+	name = "chest rig"
+	result = /obj/item/clothing/suit/storage/vest/merc/chestrig
+	steps = list(
+		list(/obj/item/stack/medical/bruise_pack/handmade, 3, "time" = 10),
+		list(/obj/item/stack/rods, 2, "time" = 10),
+		list(/obj/item/stack/cable_coil, 2),
 	)
```
TheShown911, line 52: Could you add a newline here (: ?

Pariah919: sure

code/modules/clothing/suits/armor.dm
```diff
@@ -334,6 +334,21 @@
 	price_tag = 600
 	//Used ablative gear armor values and technomancer helmet/voidsuit values.
 
+//Makeshift chest rig. Uses the merc path to allow it to have 4 pockets without making more clutter.
+/obj/item/clothing/suit/storage/vest/merc/chestrig
```
SpadesSlick, line 338: I'm sure you can make it store more than 2 items. Also, why ``merc`` is in a path?

Pariah919: Merc vest children initialize with 4 slots rather than 2, I'd have to add another pocket variant just for this which I find a bit odd considering merc rig effectively does it for what I want. But I guess I could do it.

Pariah919: Merc children start with 4 slots due to using the new proq that makes it have 4 slots. I don't really see the point of cluttering it more just for this.

code/modules/clothing/suits/storage.dm
```diff
@@ -62,3 +62,10 @@
 	pockets.max_w_class = ITEM_SIZE_SMALL
 	pockets.max_storage_space = 8
 
+/obj/item/clothing/suit/storage/vest/chestrig/New()
```
SpadesSlick, line 65: ``New()`` should always be put near declaration of actual object, most if not any procs should.

I don't know why it's like this for suit above. You might as well fix it.

PR #5125, opened 2020-07-21
this should allow NT converts to have access

fernandos13:
NT converts access

## About The Pull Request
  should allow NT convert to acces the church

## Why It's Good For The Game
 fixes NON nt workers. 
but NT belivers to have some acces in the church "and morgue"

## Changelog
:cl: Fernandos33
fix:  NT believers can now. walk around churchs
/:cl:

### DIFF
--- a/code/game/jobs/job/assistant.dm
+++ b/code/game/jobs/job/assistant.dm
@@ -9,6 +9,7 @@
 	supervisors = "anyone who pays you"
 	selection_color = "#dddddd"
 	initial_balance	= 800
+	cruciform_access = list(access_morgue, access_crematorium, access_maint_tunnels)
 	wage = WAGE_NONE //Get a job ya lazy bum
 	//alt_titles = list("Technomancer Assistant","Moebius Medical Intern","Moebius Research Assistant")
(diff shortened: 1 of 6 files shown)


fernandos13:
also don't worry i tested. non implanted works don't have access

SpadesSlick:
I think it should provide access upon cruciform generation, not in every job datum.

Also it reminds me how much outdated our assess structure for church is.

fernandos13:
i tried that at the begining. "would be easier and better as you said"
but couldn't manage it in any way.

i even tried giving ANY cruci "inquisitor stats" to see if it worked. but no. only thing I PERSONALY could manage is this one

now. if you can show me some documents to read about how the access is handled, or anything that would teach me more about this code. i could try again after reading/studying that

but as it is. i have no idea how to make that happen.

SpadesSlick:
There isn't that much on access as it is, it's a collection of flags that is used in doors and consoles. Current ones were made before NT get their own department, and it was never properly changed, so it's a core of our problems.

So please make another PR when you handle it properly.

PR #5144, opened 2020-07-27
Lasblender damage adjustment

Kurgis:
## About The Pull Request

This may actually just be too low, as we want players to specifically choose this thing, but with weapon attachments the damage output can be adjusted significantly. I was motivated to quickly make this after I noticed some other balance PRs coming up, and being closed. The lowered burst fire eliminates the core identity of the weapon, and this PR maintains that.

## Why It's Good For The Game

Balances the weapon appropriately after doing dreaded math, and with actual knowledge of the values. Tested in-game as well to gauge its effectiveness somewhat.

## Changelog
:cl:
balance: Re-adjusted the Lasblender's damage values, remember to use weapon attachments to increase its power.
/:cl:


Alliostra:
I'd suggest increasing the recoil as well, and I apologize for making this too OP

Kurgis:
There is no apology required, it should of been addressed by Reere during review.

SpadesSlick:
Yeah, I should have seen it coming.

PR #5189, opened 2020-07-29
Fixes intentional wire splicing

Chatter42:
## About The Pull Request

It's possible to create wire splicings on existing cables, but there is some jank, and this PR addresses it. You'll actually be able to make splicings larger now. It adds in some spam protection to stop people from bypassing the skill check and prevents people from being able to create multiple splices on a single turf.
## Why It's Good For The Game

Because it fixes some intended functionality and includes a couple of safeguards to prevent ~~hilarity~~ abuse.
## Changelog
:cl:
tweak: it should no longer be possible to spam player-made wire splices
fix: making splices bigger is now properly functional
/:cl:

### DIFF
--- a/code/game/objects/structures/wire_splicing.dm
+++ b/code/game/objects/structures/wire_splicing.dm
@@ -113,6 +113,10 @@
 				qdel(src)
 
 	if(istype(I, /obj/item/stack/cable_coil) && user.a_intent == I_HURT)
+		if(used_now)
+			to_chat(user, "The [src.name] is already being manipulated!") //so people with low stats can't spam their way past the failure chance
+			return
+		used_now = TRUE
 		if(messiness >= 10)
 			messiness = 10
 			to_chat(user, SPAN_WARNING("Enough."))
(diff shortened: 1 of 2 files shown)


soslow4u:
Woah, I still had no clue there is issue about splicings present. Also, those are using placeholder sprites still :(

soslow4u:
By the way, you can fix all this stuff by doing just one thing or two things: use third arg `src` in `do_after()` proc without all those flags

code/game/objects/structures/wire_splicing.dm
```diff
@@ -122,19 +126,24 @@
 		if(coil.get_amount() >= 1)
 			to_chat(user, SPAN_NOTICE("You started to wire to this pile of wires..."))
 			if(shock(user)) //check if he got his insulation gloves
+				used_now = FALSE
 				return 		//he didn't
-			if(do_after(src, 20))
+			if(do_after(user, 20))
```
soslow4u, line 131: Nice but not done
`if(do_after(user, 20, src))`

code/modules/power/cable.dm
```diff
@@ -165,23 +165,31 @@ var/list/possible_cable_coil_colours = list(
...
+			used_now = TRUE
 			if(locate(/obj/structure/wire_splicing) in T)
 				to_chat(user, SPAN_WARNING("There is splicing already!"))
+				used_now = FALSE
 				return
 			to_chat(user, SPAN_NOTICE("You started messsing with wires..."))
 			if(shock(user, 100)) //check if he got his insulation gloves
+				used_now = FALSE
 				return 		//he didn't
 			if(do_after(user, 20))
```
soslow4u, line 180: `if(do_after(user, 20, src))` there as well

And get rid of those used_now, please `do_after` would do it for you

Chatter42: I'm very new to this, so the feedback is appreciated

Chatter42: removing all of the `used_now` disables the spam protection I was putting in. am I missing something?

soslow4u: Well, I'm not really sure right now. Just tested it on my machine and got spam as well

soslow4u: Ooooooooh, I'm fucking dumb and mistaken `tool.use_tool` with `do_after`

soslow4u: So **you are very damn right all about flags** but `do_after` still needs proper calls

PR #5201, opened 2020-07-31
Change the evac vote to a bluespace jump vote

hyperioo:
## About The Pull Request

Order on the discord: "Change the evac vote to a bluespace jump vote that initiated the 20 minute evac jump if passed".

Need a confirmation that it works on live server since on a local server the end of round does not trigger (even with a normal evac).

## Why It's Good For The Game

Lore-wise there is no reason to abandon the ship if it is not critically damaged so it's more realist to just launch a bluespace jump to another star system.

## Changelog
:cl: Hyperio
tweak: Change the evac vote to a bluespace jump vote
/:cl:


Kurgis:
Was this tested?

hyperioo:
On my local server everything happens properly (the different phases of the evac sequence) but when the bluespace jump ends then nothing happens (round does not end).

hyperioo:
Actually I meant I have no way to check that everything is working properly because on local server even the standard evacuation does not trigger the end of round. I guess we would have to test that directly on the live server? The PR only impacts the evac vote process so the risk to break things is limited.

PR #5202, opened 2020-07-31
Blitzshells now can't open airlock control panel

Tea-Primoris:
## About The Pull Request
Title

## Why It's Good For The Game
Blitzshells can't bolt doors.

## Changelog
:cl:
balance: Blitzshells now can't open AI airlock control panel
/:cl:

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -717,7 +717,8 @@ There are 9 wires.
 	return
 
 /obj/machinery/door/airlock/attack_ai(mob/user as mob)
-	ui_interact(user)
+	if(!isblitzshell(user))
+		ui_interact(user)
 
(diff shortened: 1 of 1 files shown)


Kurgis:
Why not prevent Blitz from using attack_ai instead of adding a check to every piece of machinery?

GrayRachnid:
will this stop it from using shortcuts?

Tea-Primoris:
> 
> 
> will this stop it from using shortcuts?
> 
> such as ctrl+click to bolt and alt+click to shock.

They already can't use shortcuts, if I recall it right.

SpadesSlick:
> Why not prevent Blitz from using attack_ai instead of adding a check to every piece of machinery?

Because we wan't them to be able to control most of the machines, provided they have access to them, but airlocks are too powerful thing.

PR #5208, opened 2020-08-03
Make biomatter solidifier fair.

Firefox13:
## About The Pull Request

This PR changes solidifer from 10 liquid biomatter per 1 solid sheet to 1 liquid biomatter per 1 solid sheet.

## Why It's Good For The Game

Currently putting things into bioprinter for solid biomatter is ten times as effective than using a bioreactor to dissolve biomatter and then using solidifier to solidify biomatter.
Using bioreactor and solidifier for solid biomatter is longer process than directly putting organic items into bioprinter, but it doesn't requires your constant attention.

## Changelog
:cl:
balance: The NT delegation read the litany of improvement over the solidifier. It's now ten times as effective, producing as much biomatter as the bioprinter.
/:cl:

### DIFF
--- a/code/modules/biomatter_manipulation/solidifier.dm
+++ b/code/modules/biomatter_manipulation/solidifier.dm
@@ -2,7 +2,7 @@
 //This machine converts liquid biomatter to solid one(sheets)
 //Working with this also required bio protection cloths
 
-#define BIOMATTER_PER_SHEET 		10
+#define BIOMATTER_PER_SHEET 		1
 #define CONTAINER_PIXEL_OFFSET 		6
 
 /obj/machinery/biomatter_solidifier


vekotov:
Thats a bad idea because:
1. Speed of solidifier is 1 sheet per 2 seconds, so it will take (800*2)/60 ~= 27 minutes to solidify all matter in big canister.
2. If you insert solid biomatter in cloning biomass container, you get 10 liquid biomass for 1 solid, so please check if this value also used in NT Biomass Container in order to avoid possible dupe.

Firefox13:
> 
> 
> Thats a bad idea because:
> 
>     1. Speed of solidifier is 1 sheet per 2 seconds, so it will take (800*2)/60 ~= 27 minutes to solidify all matter in big canister.
> 
>     2. If you insert solid biomatter in cloning biomass container, you get 10 liquid biomass for 1 solid, so please check if this value also used in NT Biomass Container in order to avoid possible dupe.

1. Any suggestions how to make it faster?

2. And now I'm confused.
Yes, solid biomatter gives 10 liquid biomatter per sheet, so it seems like the problem isn't in solidifier, but in bioprinter which is tenfold as effective because it directly converts MATERIAL_BIOMATTER to solid instead of liquid biomatter. Obvious choice would just decrease amount of bimoatter per plant/fruit from 10 to 1 but that's just dickish move. I'd make all autolathes produce liquid biomatter if you insert anything organic in them but that's not really a good option too and I don't know what to do with bioprinter now.

code/modules/biomatter_manipulation/solidifier.dm
```diff
@@ -2,8 +2,9 @@
 //This machine converts liquid biomatter to solid one(sheets)
 //Working with this also required bio protection cloths
 
-#define BIOMATTER_PER_SHEET 		10
+#define BIOMATTER_PER_SHEET 		1
 #define CONTAINER_PIXEL_OFFSET 		6
+#define BIOMATTER_SHEETS_PER_TIME  5 // X sheets per second
```
vekotov, line 7: // X sheets per 2 seconds

Firefox13: Whoops. I've fixed the typo in next commit.

PR #5210, opened 2020-08-03
Carrion fixes round 2

TheShown911:
## About The Pull Request

Fixes the power disappearing bug, actually fixes the breeding spider and regeneration only works on the body it was activated on.

## Why It's Good For The Game

Carrion working is good.

## Changelog
:cl:
fix: carrion powers should not randomly disappear
fix: breeding spider is now truly fixed
fix: regen stasis should work on the body it was activated on
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/implant/implants/carrion/breeding_spider.dm
+++ b/code/game/objects/items/weapons/implant/implants/carrion/breeding_spider.dm
@@ -14,7 +14,7 @@
 	if(!wearer)
 		to_chat(owner_mob, SPAN_WARNING("[src] doesn't have a host"))
 		return
-	if(!ishuman(wearer))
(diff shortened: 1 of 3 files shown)


code/game/objects/items/weapons/implant/implants/carrion/breeding_spider.dm
```diff
@@ -14,7 +14,7 @@
 	if(!wearer)
 		to_chat(owner_mob, SPAN_WARNING("[src] doesn't have a host"))
 		return
-	if(!ishuman(wearer))
+	if(!(wearer.type == /mob/living/carbon/human))
```
TorinoFermic, line 17: You can write as if(!ishuman(wearer.type))

The code below this line shows it can.

PR #5213, opened 2020-08-04
 little fix on the sanity bar.

Evankhell561:
## Changelog
:cl:
fix: the maximum value in the sanity bar is now the player's maximum sanity value.
/:cl:


Evankhell561:
Sorry for the number of commits, is that i used an old branch.

PR #5232, opened 2020-08-05
fix Runtime in say.dm,66

Evankhell561:
## About The Pull Request
fix this: 
image

### DIFF
--- a/code/modules/mob/say.dm
+++ b/code/modules/mob/say.dm
@@ -58,15 +58,15 @@
 	else
 		usr.emote(message)
 
-/mob/proc/say_dead(var/message)
+/mob/proc/say_dead(message)
 	if(say_disabled)	//This is here to try to identify lag problems
 		to_chat(usr, SPAN_DANGER("Speech is currently admin-disabled."))
 		return
-
-	if(!src.client.holder)
-		if(!config.dsay_allowed)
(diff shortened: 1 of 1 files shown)


code/modules/mob/say.dm
```diff
@@ -58,23 +58,23 @@
...
-
-	if(!src.client.holder)
-		if(!config.dsay_allowed)
-			to_chat(src, SPAN_DANGER("Deadchat is globally muted."))
-			return
+	if(src.client)
+		if(!src.client.holder)
+			if(!config.dsay_allowed)
+				to_chat(src, SPAN_DANGER("Deadchat is globally muted."))
+				return
```
TorinoFermic, line 69: You can compact these 3 IFs into one big IF with &&

code/modules/mob/say.dm
```diff
@@ -58,12 +58,11 @@
...
 
-/mob/proc/say_dead(var/message)
+/mob/proc/say_dead(message)
 	if(say_disabled)	//This is here to try to identify lag problems
 		to_chat(usr, SPAN_DANGER("Speech is currently admin-disabled."))
 		return
-
-	if(!src.client.holder)
+	if(src.client && !src.client.holder)
 		if(!config.dsay_allowed)
```
SpadesSlick, line 66: All three of them will be better.

PR #5234, opened 2020-08-05
fix Runtime in drill.dm,88

Evankhell561:
## About The Pull Request
fix this:
image

### DIFF
--- a/code/modules/mining/drilling/drill.dm
+++ b/code/modules/mining/drilling/drill.dm
@@ -85,7 +85,8 @@
 		harvesting.has_resources = FALSE
 		harvesting.resources = null
 		resource_field -= harvesting
-		harvesting = pick(resource_field)
+		if(resource_field.len)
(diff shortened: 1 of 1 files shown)


code/modules/mining/drilling/drill.dm
```diff
@@ -85,7 +85,8 @@
 		harvesting.has_resources = FALSE
 		harvesting.resources = null
 		resource_field -= harvesting
-		harvesting = pick(resource_field)
+		if(resource_field.len)
+			harvesting = pick(resource_field)
```
TorinoFermic, line 89: Why adding if(resource_field.len) check when while(resource_field.len && !harvesting.resources) already does it ?

Evankhell561: For this reason:

```
resource_field -= harvesting
harvesting = pick(resource_field)
```


### `resource_field -= harvesting`

PR #5244, opened 2020-08-05
fix Runtime in thrown.dm,20

Evankhell561:
## About The Pull Request
fix this:
image

### DIFF
--- a/code/game/objects/items/weapons/material/thrown.dm
+++ b/code/game/objects/items/weapons/material/thrown.dm
@@ -17,10 +17,9 @@
 
 /obj/item/weapon/material/star/throw_impact(atom/hit_atom)
 	..()
-	if(material)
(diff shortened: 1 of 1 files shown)


code/game/objects/items/weapons/material/thrown.dm
```diff
@@ -17,9 +17,10 @@
 
 /obj/item/weapon/material/star/throw_impact(atom/hit_atom)
 	..()
-	if(material.radioactivity>0 && isliving(hit_atom))
-		var/mob/living/M = hit_atom
-		M.adjustToxLoss(rand(20,40))
+	if(material)
+		if(material.radioactivity>0 && isliving(hit_atom))
```
TorinoFermic, line 21: You can get both check IF in the same line and it will always exit the check if the material fails

PR #5248, opened 2020-08-05
tweak noise_coeff and traps logic

Evankhell561:
## About The Pull Request

Player stats were giving too much ability to evade traps even with low stats in VIG.

The noise coefficient was not working as designed.

Both modifiers are now not only affected by the weight of what you are wearing but also by what you carry with you.

## Changelog
:cl:
tweak: traps are now a little more effective.
tweak: The noise coefficient is now affected by the objects you carry with you.
/:cl:

### DIFF
--- a/code/modules/mob/mob_helpers.dm
+++ b/code/modules/mob/mob_helpers.dm
@@ -631,10 +631,10 @@ proc/is_blind(A)
 	var/prob_evade = 0
 	if(MOVING_DELIBERATELY(src))
 		var/base_prob_evade = 20
+		prob_evade += base_prob_evade
 		prob_evade += base_prob_evade * (stats.getStat(STAT_VIG)/STAT_LEVEL_GODLIKE - get_max_w_class()/ITEM_SIZE_COLOSSAL)
 		if(stats.getPerk(PERK_SURE_STEP))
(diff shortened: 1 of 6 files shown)


code/modules/mob/mob_helpers.dm
```diff
@@ -627,20 +627,20 @@ proc/is_blind(A)
...
 		prob_evade = 25
-		prob_evade += prob_evade/stats.getMult(STAT_VIG, STAT_LEVEL_GODLIKE) - prob_evade*get_max_w_class()/ITEM_SIZE_TITANIC
+		prob_evade += prob_evade * (stats.getStat(STAT_VIG)/STAT_LEVEL_GODLIKE - get_max_w_class()/ITEM_SIZE_COLOSSAL)
 		if(stats.getPerk(PERK_SURE_STEP))
-			prob_evade += prob_evade*30/STAT_LEVEL_GODLIKE
+			prob_evade += 25*30/STAT_LEVEL_GODLIKE
 		if(stats.getPerk(PERK_RAT))
-			prob_evade *= 2
+			prob_evade *= 25
 	return prob_evade
```
TorinoFermic, line 639: I think you could stick any hardcoded numbers to variables near var/prob_evade for readability

PR #5259, opened 2020-08-07
Kaiser anti-bully fix

Deon-DF:
## About The Pull Request

This affects the Kaiser roach, gives it its own meat/burger and also makes it less bully-prone.

## Why It's Good For The Game
It's a task from Discord.

Before this fix, you could stuff huge mobs in disposals, bully them and the Kaiser roach has placeholder Fuhrer meat.
In addition, when its dead, it looks like it simply looks up. Now it has a proper dead sprite, with broken legs and a crack.

## Changelog
:cl:
add: Kaiser roach has new "dead" state sprite.
add: Kaiser roach has its own meat type, containing toxins from all other roaches.
add: New kaiserburger, same recipe as other roach burgers.
balance: Kaiser roach deals 20-35 damage now.
balance: Kaiser roach is a bit faster (4->6).
fix: Kaiser roach can properly use gas cloud ability now.
fix: Fuhrer roach burger now properly contains "fuhrerole" (there was a typo).

tweak: Huge mobs cannot be stuffed into disposal bins.
tweak: As a result, Kaiser roach cannot be put into disposals. It also can't be flashed or slipped, for your information.
/:cl:


Kurgis:
There's a stray white pixel on the dead state you've added (happens usually when you select an icon for editing with Dream Maker), but other than that this is good.

code/modules/mob/living/carbon/superior_animal/roach/types/kaiser.dm
```diff
@@ -9,20 +9,22 @@ Has ability of every roach.
...
 	maxHealth = 2000
 	health = 2000
 	contaminant_immunity = TRUE
 
 	var/datum/reagents/gas_sac
 
-	melee_damage_lower = 10
-	melee_damage_upper = 20
+	melee_damage_lower = 35
+	melee_damage_upper = 40
```
SpadesSlick, line 23: Make it 20 lower and 35 upper limit, don't increase such values too much in one go. It's against contributing guidelines.

Deon-DF: With this damage it's still pretty weak in terms of its speed/AI and dies to any gun due to not using ranged attacks.
I have written a question about this revision in The Shaft, since I literally changed the values to the amount in the "Orders" section, so this should be better checked with Gray who did the order.

SpadesSlick: Mobs don't suppose to deal too much damage, because they often attack in large numbers. Since Kaiser is supported by other roaches on spawn and by fuhrer-like abbility, it can be too much.

Plus contributing guidelines rule is still a thing.

> Rebalance properly. Don't nerf/buff values of things SIX TIMES in one go. Nerf values for 35-50%, see the results on live server, note them, nerf further when there is still a problem. This is how people in "proper game dev" do that, and we are absolutely not smarter then them. When it's a rework that changes whole area of a game - it's fine to axe values as much as you want, but small changes should not make our codebase look like twitching overdosed hobos. Please do buffs and nerfs by small portions each week.

Deon-DF: Following the Discord conversation, the min/max damage has been adjusted to 20-35.

PR #5288, opened 2020-08-10
Maintenance shroom rework and fix

Deon-DF:
## About The Pull Request

Before these changes, there are the following problems in the game:
- You can only sample maintenance fungus with wirecutters/scalpel. It does not use the tool qualities system.
- You get a single seed after you do that, which you cannot use until you grow it again.
- The seed will always grow in the same type of fungus with the same ingredient for the whole round, because the current botany system doesn't support inheriting reagents from parents, it checks the "name" in plant controller.

This PR does the following:
- Maintenance mushrooms get a name based on their reagent (randomized each round).
- That fixes the problem with all mushrooms having the same reagent after planting their seed.
- You can sample them with any cutting/wire cutting/laser cutting tool.
- You can sample either a single grown produce or a seed, via a radial menu selection.

radial

## Why It's Good For The Game

Now the fungi can actually net reagents they grow with, and they can be used without reusing them in the garden (and without rewriting the whole botany system to support reagent inheritance to seeds).

The custom fungi names are also kind of cute and seem fitting to the maintenance thematic.

## Changelog
:cl:
add: Added a random name generator for maintenance shrooms.
code: Reorganized maintenance mushroom reagent list to be alphabetical for easier reading.
fix: Cutting the maintenance plants now gives you a choice to either sample the plant (harvest 1) or sample the seed (harvest seed sample), with the radial menu selection.
fix: Sampling any maintenance plants can now be done with any tools that have "cutting", "wire cutting" or "laser cutting" qualities, instead of just wirecutters and scalpel.
/:cl:


Deon-DF:
names

Voldirs:
Could you please check if plants in hydroponics tray can be sampled (for seeds, not for product) too? It was possible to sample them before, it makes things much easier for xenobotany. And maybe revising possible chem list considering that we have more chems now?

Deon-DF:
Voldirs, the change is that when you cut a growth with wirecutters, you get 1 product instead of 1 seed and that's the only change. Everything else is the same.

Deon-DF:
> Also I'm not sure if shroom names should be capitalized
I understand the idea, I can rework it easily if I am asked of this by mergers, but tomorrow, it's 4 AM here.

Deon-DF:
Example

Ahh screw it, 5 AM is not going to stop the progress.
I redid the names so they are no longer capitalized.

Also ^ here's a screenshot proving that my change didn't touch sampling the botany stuff from trays, it only affects plants growing in maintenance.

Voldirs:
Sorry for taking your sleep, I didn't intend that!!!

Deon-DF:
What about giving an option to either cut shroom or collect seed sample? That can be done with one menu.

Deon-DF:
radial

Deon-DF:
Ready for review now.

code/modules/hydroponics/seed_datums.dm
```diff
@@ -551,57 +551,81 @@
...
-		"blattedin",
-		"honey",
-		"frostoil",
-		"capsaicin",
-		"banana",
-		"mutagen",
-		"chloralhydrate"
+		"sugar",
+		"synaptizine",
+		"thermite",
```
infraredbaron, line 594: Might melt walls into space, which is not really good.

PR #5299, opened 2020-08-12
disposable NT sidearm

fernandos13:
## About The Pull Request
some prints: https://gyazo.com/5c8821744f83780863ff050cec833422
https://gyazo.com/fc98233331e18f315e35fd5cbb6a7871
https://gyazo.com/ab22645825592d1bdae70d37dd3060c4

-enough to kill a furher and nothing more (each shoot is enough to kill 1 roach on its own)
https://gyazo.com/9546770a38692d1c96633c02cb2b3593

-IF you feel like risking it. you can change to MELT mode for some ok damage and penetration
https://gyazo.com/d0d7fb56a39cc13ad99a6dbf4d0d4193

it adds martyr
a 
disposable NT sidearm

Martyr got 2 firemodes
Burn: works just like the "stun" from martin martin. (can shoot 8 times)
Melt: a single REALLY powerfull shot. with some armor penetration that we all know and hate?love? from plasma guns

## Why It's Good For The Game
i'll add martyr our first disposable weapon. should not only give NT some firepower roundstart. but also help start a new weapon type of disposable guns. to lower the power creep on economy and loot. due to it being destroyed upon use

## Changelog
:cl:Fernandos33
add: Martyr. new NT disposable handgun. Sprites by EDD
/:cl:


fernandos13:
Ready for review

code/game/machinery/recharger.dm
```diff
@@ -69,7 +69,9 @@
 		if(istype(I, /obj/item/weapon/gun/energy/gun/nuclear) || istype(I, /obj/item/weapon/gun/energy/crossbow))
 			to_chat(user, SPAN_NOTICE("Your gun's recharge port was removed to make room for a miniaturized reactor."))
 			return
-
+		if(istype(I, /obj/item/weapon/gun/energy/plasma/martyr))
```
SpadesSlick, line 72: You shouldn't do it here for this item only. It's bad practice in general to be this specific about particular objects in all purpose machines.

code/modules/projectiles/guns/energy/plasma.dm
```diff
@@ -102,4 +102,31 @@
...
+	fire_sound = 'sound/weapons/Laser.ogg'
+	cell_type = /obj/item/weapon/cell/martycharge //can't recharge this one
+	sel_mode = 2
+	charge_cost = 1
+	recoil_buildup = 1 //pulse weapons have a bit more recoil
+	one_hand_penalty = 0
+	twohanded = FALSE
+
+	init_firemodes = list(
+		list(mode_name="Burn", projectile_type=/obj/item/projectile/plasma/stun, fire_sound='sound/weapons/Taser.ogg', fire_delay=2, charge_cost=null, icon="stun", projectile_color = "#0000FF"),
```
SpadesSlick, line 130: Why charge cost is null for stun? Why stun mode is named burn?

code/game/objects/items/weapons/power_cells.dm
```diff
@@ -1,4 +1,7 @@
 //BIG CELLS - for APC, borgs and machinery.
+/obj/item/weapon/cell
+
```
SpadesSlick, line 3: That's a really wrong way to do it. Never declare object somewhere else just to add some vars to it.

code/game/objects/items/weapons/power_cells.dm
```diff
@@ -264,3 +267,16 @@
...
+//
+/obj/item/weapon/cell/martycharge
+	
+	name = "NT \"martyrcharge\""
+	desc = "A battery like cartridge for martyr. you shouldn'g be holding this"  // shouldn't be able to remove it but just in case one does
+	icon_state = "martycharge"
+	origin_tech = list(TECH_POWER = 0)
+	matter = list()
+	maxcharge = 8 //useless enough if anyone manage to get this
+	disposable = TRUE
```
SpadesSlick, line 282: Endline.

code/game/objects/items/weapons/power_cells.dm
```diff
@@ -264,3 +267,16 @@
 	origin_tech = list(TECH_POWER = 4)
 	matter = list(MATERIAL_STEEL = 2, MATERIAL_PLASTIC = 1)
 	maxcharge = 300
+
+//
+//.........................disposable weapon custom cells
+//
+/obj/item/weapon/cell/martycharge
```
SpadesSlick, line 274: There is probably a way to use a normal cell for it, without snowflake objects.

code/modules/projectiles/guns/energy/plasma.dm
fernandos13: so it uses the default charge value. (i mostly try to keep things as the other stuff already in the code so instead of deleting the var from firemodes i just made it null. )

i'll delete it

code/game/objects/items/weapons/power_cells.dm
```diff
@@ -264,3 +263,7 @@
 	origin_tech = list(TECH_POWER = 4)
 	matter = list(MATERIAL_STEEL = 2, MATERIAL_PLASTIC = 1)
 	maxcharge = 300
+
+//
+//.........................disposable weapon custom cells
+//
```
TheShown911, line 269: No endline

code/game/machinery/recharger.dm
```diff
@@ -66,10 +66,14 @@
 			to_chat(user, SPAN_WARNING("[src] blinks red as you try to insert the item!"))
 			return
 
+		if (istype(I, /obj/item/weapon/gun/energy))
+			var/obj/item/weapon/gun/energy/W = I
+			if (W.disposable)
+				to_chat(user, SPAN_NOTICE("Your gun is disposable it cannot be charged."))
```
Pariah919, line 72: Change this to "Your gun is disposable, it cannot be charged." or "Your guns battery is disposable it and so it cannot be 
be recharged."

fernandos13: roger that
will do ASAP

fernandos13:
ANND grmmar fix done

Hopekz:
Did some testing. In terms of balancing, you had this PR right where it needed to be at the start. You have made some weird changes since then.

**Power:**
You have removed the functionality of it being an anti roach weapon because now you can only fire this weapon once before it melts.
Right now it takes melting 2 of these guns to kill a Fuhrer Roach. Not to mention killing a basic roach will also melt this gun. If I open a closet and encounter 8 roaches I would need to bring 8 of these guns to have a chance of killing them assuming these are the basic roaches.

**Resources:**
Silver was added. Where is it expected for NT to find silver round-start? 
The only place I can see is if they melt their headsets which is a bit weird in my opinion. When selecting between having a headset and having one of these guns the choice seems pretty obvious considering how weak of an option this is since the changes.

You added Plastic as well. Besides melting their tools from the toolboxes they have or the cleaning supplies they need I am unsure as to where it is expected to be sourced from round-start.

Its a "cool" idea but has been nerfed before even becoming a reality to the point of uselessness in my personal opinion.

fernandos13:
you can still.. change its fire mode.. for 8 shots...
it kill the roaches like before
 what change you said that killed it. if you can point out in the code?

it has always had 
8 stun shots (for roaches)
1 MELT (for armor/bigger things)

fernandos13:
also where in the code it asks for silver or plastic? i think you are seeing the wrong PR now that you said about this materials. it have never asked for neither silver or plastic

Hopekz:
> also where in the code it asks for silver or plastic? i think you are seeing the wrong PR now that you said about this materials. it have never asked for neither silver or plastic

I'm not sure why I'm getting completley different results. I did my testing as an acolyte if that matters.

**Here what the recipie requires when I load it up:**

image



**Here is what it looks like when set to stun mode:**

image

Stun mode melts the gun in 1 shot vs the 8 you're mentioning.

fernandos13:
ok something is wrong then lemme debug this

fernandos13:
well SO.. the "snowflake object"

is actually important. thats taken in account when the lathe calculate materials

and the charge_cost=null or charge_cost=12.5 

was also important to determine the ammount of shots.  Thanks hopez for noticing it and notifying me before it slipped

code/modules/projectiles/guns/energy.dm
```diff
@@ -116,17 +121,22 @@
...
-	if(!self_recharge)
+	if(!self_recharge && !disposable)
 		if((src.loc == usr) && istype(over_object, /obj/screen/inventory/hand) && eject_item(cell, usr))
 			cell = null
 			update_icon()
-	else
+	if(self_recharge)
 		to_chat(usr, SPAN_WARNING("[src] is a self-charging gun, its batteries cannot be removed!."))
+	if(disposable)
+		to_chat(usr, SPAN_WARNING("[src] is a disposable, its batteries cannot be removed!."))
```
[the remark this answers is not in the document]
fernandos13: i'll apply this change. thanks for the more clear code. it indeed looks. better and should help people that come after

fernandos13:
well a lot changed to get it to work that way torino. but we got it

code/modules/projectiles/guns/energy.dm
```diff
@@ -116,17 +121,21 @@
...
-			cell = null
-			update_icon()
-	else
+	if(disposable)
+		to_chat(usr, SPAN_WARNING("[src] is a disposable, its batteries cannot be removed!."))
+	if(self_recharge)
 		to_chat(usr, SPAN_WARNING("[src] is a self-charging gun, its batteries cannot be removed!."))
+	if((src.loc == usr) && istype(over_object, /obj/screen/inventory/hand) && eject_item(cell, usr))
+		cell = null
+		update_icon()
```
TorinoFermic, line 130: There's a issue with it, when the gun is dragged to hands, the code dosen't prevent from pulling the cell, which is why if...else if....else block was used to do that.

fernandos13:
should be fixed. thanks for noticing torino. i overlooked the code itself. and was just trying to insert the logic. (note to self: don't do it again)

TorinoFermic:
> should be fixed. thanks for noticing torino. i overlooked the code itself. and was just trying to insert the logic. (note to self: don't do it again)

Did you forget to commit to your PR ?

fernandos13:
computer is slow its coming up at anytime

code/modules/projectiles/guns/energy.dm
```diff
@@ -116,17 +121,21 @@
...
-	if(!self_recharge)
-		if((src.loc == usr) && istype(over_object, /obj/screen/inventory/hand) && eject_item(cell, usr))
-			cell = null
-			update_icon()
-	else
+	if(disposable)
+		to_chat(usr, SPAN_WARNING("[src] is a disposable, its batteries cannot be removed!."))
+	else if(self_recharge)
 		to_chat(usr, SPAN_WARNING("[src] is a self-charging gun, its batteries cannot be removed!."))
+	else if((src.loc == usr) && istype(over_object, /obj/screen/inventory/hand) && eject_item(cell, usr))
```
fernandos13, line 128: and its up here

fernandos13:
Thanks Torino. a ton trully. both for the attention on it. and patience

PR #5300, opened 2020-08-12
Breakout time in seconds

Deon-DF:
##About The Pull Request

Changes breakout time (when resisting in handcuffs) to be displayed in seconds instead of rounded minutes.
Works well with cuffs with breakout times less than 1 minute or not equal to the power of a minute.

## Why It's Good For The Game

Without this fix you would either see incorrect time (1 minute for 70 seconds) or even 0 minutes (for 59 seconds and below) breakout times.

## Changelog
:cl:
fix: handcuffs breakout time is now displayed in seconds (to see correct time/work right with breakout time < 1 minute).
/:cl:


Deon-DF:
Before:
cuffs1

After:
cuffs2

PR #5305, opened 2020-08-13
autoinjectors can now be refilled

fernandos13:
## About The Pull Request
autoinjectors can now be refilled

## Why It's Good For The Game
well we got agressive syringe with 15u in it to be injected even in armored targets. this even takes a while in the same case. 
so its weaker than that already.

its good cause it adds another possibility for moebius to sell stuff. and a little ghetto syringe

## Changelog
:cl:Fernandos33
tweak: autoinjectors can now be refilled
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/hypospray.dm
+++ b/code/modules/reagents/reagent_containers/hypospray.dm
@@ -98,8 +98,6 @@
 
 /obj/item/weapon/reagent_containers/hypospray/autoinjector/on_reagent_change()
 	..()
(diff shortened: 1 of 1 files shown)


SpadesSlick:
It doesn't handle icon update properly.

fernandos13:
people didn't wanted this i found out. it would load and unload properly. from the tests i did. and even change the color depending on the chem you have in it. but the only reason i made if cause i thought that was something people wanted. and they dont so its closed

PR #5307, opened 2020-08-13
Increase water damage done to slimes

hyperioo:
## About The Pull Request

**Discord order:** Slimes:
Buff fire extinguisher water damage. Right know it's actually faster to just melee them with the extinguisher.

## Why It's Good For The Game

Fire extinguisher is now a viable way to kill slimes.

## Changelog
:cl: Hyperio
tweak: Increase water damage done to slimes
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/core.dm
+++ b/code/modules/reagents/reagents/core.dm
@@ -146,7 +146,7 @@
 /datum/reagent/water/affect_touch(var/mob/living/carbon/M, var/alien, var/effect_multiplier)
 	if(isslime(M))
 		var/mob/living/carbon/slime/S = M
-		S.adjustToxLoss(0.8 * effect_multiplier) // Babies have 150 health, adults have 200; So, 10 units and 13.5
+		S.adjustToxLoss(7 * effect_multiplier)
 		if(!S.client)
 			if(S.Target) // Like cats
 				S.Target = null


fernandos13:
i do approve this move. but i gotta say. i kinda like it being more of a threat. and hearing shootings coming from xenobiology

code/modules/reagents/reagents/core.dm
```diff
@@ -146,7 +146,7 @@
 /datum/reagent/water/affect_touch(var/mob/living/carbon/M, var/alien, var/effect_multiplier)
 	if(isslime(M))
 		var/mob/living/carbon/slime/S = M
-		S.adjustToxLoss(0.8 * effect_multiplier) // Babies have 150 health, adults have 200; So, 10 units and 13.5
+		S.adjustToxLoss(7 * effect_multiplier)
```
SpadesSlick, line 149: That sounds like too much of a jump. How much damage per one splash is this?

hyperioo: Each tick of water does around 7 damage. Spraying a slime triggers 5 ticks over a duration of 2 or 3 seconds.

So basically each spray does around 35 damage, killing a slime in 3 sprays (they have 80 health).

If you spray a slime in the one-tile wide corridor in front of the xenobio pen, the slime can be one-shot because the slime receives all the water, water damage tiggers 12 times and kill it (12*7=84 dmg).

Do you think it is too strong?

fernandos13: as a side not the .35 revolver you can get in frozstart "havelock" is 1HK for them so their are not that resilient on their own ""just somethinig to compare the damage""

PR #5134, opened 2020-07-25
Adding of tool and gun mods and mind swapper in RND

hyperioo:
## About The Pull Request

* First PR related to RnD improvements (see RnD section of the KoT map).

* Tool and gun mods prites by INFRARED_BARON

* Mindswapper sprite by The Sprite Goblin

* To be merged after #5267 since it uses the new psy damage type

## Why It's Good For The Game

* Add high quality toolmods for late game and a few high-tech gunmods. Gunmods can be found in a new "Gunmods" category in the autolathe menu once researched.
* I hope the "battery_shunt" and "overdrive" gun mods will not be too much broken. In terms of numbers they reduce the total damage output of the equipped weapon but there is still a burst potential that could be a bit too strong.
* The mind swapping device can also be a fun thing to do at some point during the round (randomly swap the minds of all people near it).

## Changelog
:cl: Hyperio
add: Added new toolmods (engineering research node)
add: Added new gunmods (combat research node)
add: Added a mind swapping device (biotech research node)
/:cl:

PR #5308, opened 2020-08-13
Change to prevent export abuse

Sombrahide:
## About The Pull Request
This change gives a price of 10 units of plasteel to the mods for weapons that are created in science, in case of still wanting to build and export them they would give a benefit of 200 credits (much less compared to the 500 before), but above all it prevents that is being spammed in the protolathe of sciences so that cargo becomes rich effortlessly.

## Why It's Good For The Game
It helps to balance the economy so that it does not break so easily, giving more value to money by not providing an easy and effortless method of achieving it

## Changelog
:cl:
balance: Changed the build values of the Penetrator and Caster mods from nothing to 10 units of plasteel
/:cl:


TheShown911:
Already being fixed in https://github.com/discordia-space/CEV-Eris/pull/5134

SpadesSlick:
Reworked in PR mentioned above.

PR #5309, opened 2020-08-13
Nerfs the Gladstone

Alliostra:
## About The Pull Request
People were complaining hard about the gladstone being OP. Some people think  the gladstone has a 1.3x damage modifier, but that has never been the case, I only raised it's damage from 0.9 to 1 in my initial PR. However, the AP modifier of 1.3 heavily affects the AP capabilities of beanbags despite beanbags having 0 AP on their own, so Im lowering the AP mod back to 1. The only advantages of the Gladstone compared to a kammerer are increased bullet velocity and a bigger shell capacity.

## Why It's Good For The Game
perfectly balanced, as eris should be

## Changelog
:cl:
balance: lowered the AP of the gladstone
/:cl:


SpadesSlick:
Issue is in damage calculation bug and gun is not relevant. It will be fixed in armor rework that will come soon. 

Also, clean your fork commits.

PR #5310, opened 2020-08-13
Add the vape mask

Silverexperience:
## About The Pull Request

- Add the vape masks imagen. Vapes are a mask mixed with a reaguent container, they are activated if you put them in your mouth and transfer the reaguents, like a cigarette, but instead of using fire they use a battery.

- Vapes can be hacked with a multitool, which would increase the amount of chemicals it transfers by 2, or by an emag, which would make the vape form smoke from the chemical you are using and also give it a small chance of explode
imagen

- If the vape uses gasoline or plasma it causes an immediate explosion.

- Vapes, like cigarettes, have a waste of chemicals but there are improved versions that can be found randomly in loot tables that use less chemicals, use less energy and transfer more of the chemical.
imagen

- Vapes can be bought from the cigar vending.
imagen

- To hack, change the baterry or change the chemicals you have to use a screwdriver to open it.
To remove the battery you need to drag the vape to your other hand with the mouse.
To empty the vape you just have to click with the mouse.
To hack it, it would be to click with the appropriate tools.
To add new ingredients it would be by clicking with another selected container in the other hand.

## Why It's Good For The Game
Add a new way to smoke chemicals

## Changelog
:cl:
add: Added the vapes.
imageadd: added some icons and images for the vapes.
/:cl:


code/game/machinery/vending.dm
```diff
@@ -1089,15 +1089,17 @@
 					/obj/item/weapon/storage/box/matches = 10,
 					/obj/item/weapon/flame/lighter/random = 4,
 					/obj/item/weapon/storage/fancy/cigar = 5,
-					/obj/item/weapon/storage/fancy/cigarettes/killthroat = 5
+					/obj/item/weapon/storage/fancy/cigarettes/killthroat = 5,
+					/obj/item/clothing/mask/vape/ = 5
```
MadmanMartian, line 1093: remove the trailing /

MadmanMartian:
>Vape

Call it a vapour mask, you zoomer.

code/game/objects/items/weapons/cigs_lighters.dm
```diff
@@ -531,3 +532,199 @@ CIGARETTE PACKETS ARE IN FANCY.DM
...
+/obj/item/clothing/mask/vape/emag_act(var/remaining_charges, mob/user)
+	if(screw)
+		if(!emagged)
+			emagged = 1
+			to_chat(user, "<span class='warning'>You maximize the voltage of [src].</span>")
+			var/datum/effect/effect/system/spark_spread/sp = new /datum/effect/effect/system/spark_spread //for effect
+			sp.set_up(5, 1, src)
+			sp.start()
+		else
+			to_chat(user, "<span class='warning'>[src] is already emagged!</span>")
```
SpadesSlick, line 628: Please use macro instead of span. Everywhere.

PR #5327, opened 2020-08-14
bandanas now are properly shown

SrReptile:
bandana time.

## About The Pull Request

little changes
## Why It's Good For The Game

they no longer look ugly af

## Changelog
:cl:SrReptile
code: Blockhair added in the bandanas
/:cl:

### DIFF
--- a/code/modules/clothing/head/misc.dm
+++ b/code/modules/clothing/head/misc.dm
@@ -76,6 +76,7 @@
 /obj/item/clothing/head/bandana/green
 	name = "green bandana"
 	desc = "It's a green bandana with some fine nanotech lining."
+	flags_inv = BLOCKHAIR
 	icon_state = "greenbandana"
 	item_state = "greenbandana"
 	flags_inv = 0
@@ -140,6 +141,7 @@
 	name = "pirate bandana"
 	desc = "Yarr."
 	icon_state = "bandana"
+	flags_inv = BLOCKHAIR
 	style = 1
 
 /obj/item/clothing/head/bowler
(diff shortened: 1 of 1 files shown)


code/modules/clothing/head/misc.dm
```diff
@@ -76,6 +76,7 @@
 /obj/item/clothing/head/bandana/green
 	name = "green bandana"
 	desc = "It's a green bandana with some fine nanotech lining."
+	flags_inv = BLOCKHAIR
```
TorinoFermic, line 79: No need for that, /obj/item/clothing/head/bandana is the parent of this object. It already inherits it.

code/modules/clothing/head/misc.dm
```diff
@@ -239,6 +241,7 @@
 /obj/item/clothing/head/bandana/orange //themij: Taryn Kifer
 	name = "orange bandana"
 	desc = "An orange piece of cloth, worn on the head."
+	flags_inv = BLOCKHAIR
```
TorinoFermic, line 244: Ditto No need for that, /obj/item/clothing/head/bandana is the parent of this object. It already inherits it.

GrayRachnid:
Are you sure this fixes them? Can you post ingame screenshot? As far as I was aware the sprites were missing or something.

SrReptile:
test

code/modules/clothing/head/misc.dm
SrReptile: If  I erase them, the hair doesn't dissapear in the others bandanas

code/modules/clothing/head/misc.dm
SrReptile: If I erase them, the hair doesn't dissapear in the others bandanas

code/modules/clothing/head/misc.dm
TorinoFermic: It won't, you just assigned something in flags_inv in /obj/item/clothing/head/bandana and it gets shared with childs as long they don't get modified again.

PR #5328, opened 2020-08-14
Datum armor and prosthetic protection

TheShown911:
## About The Pull Request
Implements a /tg/ like armor datum to store armor variables and uses it with prosthetics to make them more viable. Adds moebius, excelsior, and one star prosthetics with 35/50/65 armor respectively. Armor does not stack anymore for the armor calculation so some basic values were added for gloves and shoes.

## Why It's Good For The Game
Kot map, also voidsuit meta is trash.

## Changelog
:cl:
add: 3 new prosthetic types with armor
balance: armor does not stack anymore
balance: boots and gloves provide some armor now
refactor: armor is now a datum
/:cl:


Kurgis:
> Armor does not stack anymore for the armor calculation so some basic values were added for gloves and shoes.

Excellent, removing the armor values added their own problems as well, this solves both.

PR #5355, opened 2020-08-18
Cringe Filter

mattroks101:
## About The Pull Request
Adds cringe filter to stop people from saying things IC or naming themselves things that break the atmosphere of the game. It is configurable in txt file. It will only match whole words, not parts of words so you don't need to worry about acronyms appearing in words setting it off.

## Why It's Good For The Game
This was requested by @GrayRachnid and Reere.

## Changelog
:cl:
admin: Added cringe filter to stop people from saying things that break the atmosphere.
/:cl:

### DIFF
--- a/code/modules/mob/living/say.dm
+++ b/code/modules/mob/living/say.dm
@@ -135,6 +135,25 @@ var/list/channel_to_radio_key = new
 			return say_dead(message)
 		return
 
+	if(GLOB.in_character_filter.len)
+		if(findtext(message, config.ic_filter_regex))
+			// let's try to be a bit more informative!
+			var/warning_message = "<span class='warning'>A splitting spike of headache prevents you from saying whatever vile words you planned to say! You think better of saying such nonsense again. The following terms break the atmosphere and are not allowed: &quot;"
+			var/list/words = splittext(message, " ")
+			var/cringe = ""
+			for (var/word in words)
+				if (findtext(word, config.ic_filter_regex))
+					warning_message = "[warning_message]<b>[word]</b> "
+					cringe += "/<b>[word]</b>"
+				else
+					warning_message = "[warning_message][word] "
+
+
+			warning_message = trim(warning_message)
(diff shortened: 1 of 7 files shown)


code/modules/mob/new_player/new_player.dm
```diff
@@ -356,7 +356,7 @@
 	if(SSticker.random_players)
 		new_character.gender = pick(MALE, FEMALE)
 		client.prefs.real_first_name = random_first_name(new_character.gender)
-		client.prefs.real_last_name = random_first_name(new_character.gender)
+		client.prefs.real_last_name = random_last_name(new_character.gender)
```
mattroks101, line 359: This is an unrelated fix to an oversight that gives people two random first names instead of a random first name and last name.

Hopekz:
Instead of blocking why not convert common ones to phrases instead?

for example:
idk = I don't know
lol = haha
lmao = haha
omg = oh my god

mattroks101:
Because that's significantly more work and not worth the effort.

code/modules/mob/living/say.dm
```diff
@@ -135,6 +135,25 @@ var/list/channel_to_radio_key = new
 			return say_dead(message)
 		return
 
+	if(GLOB.in_character_filter.len)
+		if(findtext(message, config.ic_filter_regex))
+			// let's try to be a bit more informative!
+			var/warning_message = "<span class='warning'>A splitting spike of headache prevents you from saying whatever vile words you planned to say! You think better of saying such nonsense again. The following terms break the atmosphere and are not allowed: &quot;"
```
TorinoFermic, line 141: SPAN_WARNING() here

TorinoFermic: Part of <span> is incomplete

TorinoFermic: Well, nevermind, that's better if we use SPAN_WARNING at to_chat part

PR #5357, opened 2020-08-19
Autolathe disk eject QoL

Hopekz:
You can now Alt-Click autolathes to eject the disks.

## Changelog
:cl: Hopek
add: Autolathe disk eject QoL. You can now Alt-Click on any autolathe to eject the design disk.
/:cl:

### DIFF
--- a/code/game/machinery/autolathe/autolathe.dm
+++ b/code/game/machinery/autolathe/autolathe.dm
@@ -460,6 +460,13 @@
 
 	disk = null
 
+/obj/machinery/autolathe/AltClick(mob/living/user)
+	if(user.incapacitated())
+		to_chat(user, SPAN_WARNING("You can't do that right now!"))
+		return
+	if(!in_range(src, user))
+		return
+	src.eject_disk(user)
 
 /obj/machinery/autolathe/proc/eat(mob/living/user, obj/item/eating)
 	if(!eating && istype(user))


code/game/machinery/autolathe/autolathe.dm
```diff
@@ -460,6 +460,13 @@
 
 	disk = null
 
+/obj/machinery/autolathe/AltClick(mob/living/user)
+	if(user.incapacitated())
+		to_chat(user, SPAN_WARNING("You can't do that right now!"))
+		return
+	if(!in_range(src, user))
+		return
+	src.eject_disk(user)
```
mattroks101, line 469: the `src` is not needed here.

Hopekz: @mattroks101 
It is best practice to include the object being acted upon for clarity's sake.

If it is an issue I can modify it but I have gone over CONTRIBUTING.md before contributing here and the last 4 PR's I have created use src.something() for clarification as well so I would like to maintain consistency if possible.

mattroks101: I suppose it doesn't really matter that much.

fernandos13: just trying to learn something 
i don't doubt. i'm just a newb

why is it the best pratice (even when not needed)

Hopekz: > why is it the best pratice (even when not needed)

It is more of a behavior that was developed from my professional life.

I was told it is for readability's sake of making sure other coders that usually work in different languages can easily make the distinction that the proc being run exists only within that specific object.

for this specific example, if you just use eject_disk() instead of src.eject_disk() the individual reading the code in the future might think that eject_disk() is a proc outside of the current object. Adding src to this tells them "This proc exists here".

Functionally it makes no difference but it is polite when working on a project where lots of different people with different coding backgrounds are involved.

fernandos13: oh i see why. gosh i gotta say after reading this. if all code was written like this.. tbh would make learning a lot easier heh :,)
thanks. and i'll try to get this habit too

was hard enough to stop using in the end of everything ; if i managed that i can manage this :D thanks hopez

PR #5392, opened 2020-08-24
Recycling rework - Recyclable scrap and constructable machinery

greggbot:
## About The Pull Request

The smelter and the sorter are handy little machines that players didn't really have the tools to use - there was no way to obtain the board other than deconstructing the SINGLE smelter or sorter on the map, and the sides weren't player-configurable.

This PR fixes that by adding the circuits to the conveyors disk (now called logistics), and also adds the ability to compress and recycle scrap piles using an exosuit's hydraulic clamp and the smelter. You can also use this feature to transport scrap, which is handy.

## Why It's Good For The Game

More power to the players is always a good thing. I personally loved working on recycling systems as guild, but it always felt kinda lame as the two most useful machines were one-of-a-kind for no real reason. This gives guild techs an additional, optional task - build setups to automatically sort and/or recycle collected scrap. Maybe you want to sort out anything with biomatter for sale to the guild, send rig modules straight to the sales area, sort chemical containers by reagent... Now you finally have the tools to do so.

And of course, non-guild personnel can use these machines too - turn an unused corner of maint into a production line if you wish!

## Changelog
:cl:
add: Exosuit-mounted hydraulic clamps can now compress scrap into cubes for transport. These cubes can then be whacked with a decent amount of force (a crowbar should do) to decompress them and release the scrap pile. They can also be recycled directly for a reduced amount of materials compared to recycling the contents individually - this amount can be increased with better microlasers, from 25% to 45% with One Star parts.
add: The conveyors disk now has boards for the smelter and sorter. It's also been renamed to logistics, and has license points. Conveyor and switch assemblies still cost 0 LP, so don't worry.
/:cl:


code/modules/scrap/scrap_refine.dm
```diff
@@ -20,7 +20,7 @@
 	user.do_attack_animation(src)
 	user.setClickCooldown(DEFAULT_ATTACK_COOLDOWN)
 	if(istype(W,/obj/item/weapon) && W.force >= 8)
-		visible_message("<span class='notice'>\The [user] smashes the [src], restoring it's original form.</span>")
+		visible_message("<span class='notice'>\The [user] smashes the [src], restoring its original form.</span>")
```
TorinoFermic, line 23: Feel free to remove these <span> and put SPAN_NOTICE surrounding this string

code/modules/scrap/scrap_refine.dm
```diff
@@ -20,10 +20,10 @@
 	user.do_attack_animation(src)
 	user.setClickCooldown(DEFAULT_ATTACK_COOLDOWN)
 	if(istype(W,/obj/item/weapon) && W.force >= 8)
-		visible_message("<span class='notice'>\The [user] smashes the [src], restoring it's original form.</span>")
+		visible_message(SPAN_NOTICE("\The [user] smashes \the [src], restoring its original form.</span>"))
```
TheShown911, line 23: Span macros don't need a `</span>`

greggbot: My bad, forgot too remove them. Done now.

TheShown911 (approved):

SpadesSlick (changes requested):
> This was so close to being done all this time and yet never was holy shit

That's kinda not the case. Junk was ported from Tau, where scrap compression was one of the main junkyard mechanics. On Tau junk piles spawn MUCH LESS and are not used in maint, those are all changes done to junk mid-port. Basically it's a leftover of a different way of utilizing scrapyard before Nanako did those changes.

It's not like it matter, because change itself is cool.

As for more important matter, to be actually useful, you should add ability for a smelter to consume whole compressed pile, with resource penalty, due it being unsorted and for balance reasons. We need more smelter related mechanics.

TorinoFermic (approved):

PR #5404, opened 2020-08-26
Replaces Agrolyte Armor with a snazzier Medieval Peasant inspired design

PigeonLord:
:cl:
imageadd: Replaces Agrolyte Armor with a snazzier Medieval Peasant inspired design
/:cl:
crop more like crap
The current Agrolyte Armor is kinda shitty, so I made a cooler version. Huge thanks to nullAurelian#0965 for making the drawing I referenced this off from. Also pls buff NT armor


PigeonLord:
ShittyNT_2-Bot
nullAurelian's drawing

SpadesSlick:
What's even the point of doing this? This version is just worse than a current one.

PigeonLord:
Added Arm Braces and a bit of shading.

Hopekz:
Why can't this be a variation available to them? 
I like the old style but also think this one is kind of cool and the idea of choosing one instead of having both is weird.

Code-wise should be super easy to do just make your armor a child and add it to the locker.

fernandos13:
that is indeed true,  it could be a secondary option. maybe a less protective but spaceproof variant

PigeonLord:
> 
> 
> Why can't this be a variation available to them?
> I like the old style but also think this one is kind of cool and the idea of choosing one instead of having both is weird.
> 
> Code-wise should be super easy to do just make your armor a child and add it to the locker.



> 
> 
> that is indeed true, it could be a secondary option. maybe a less protective but spaceproof variant

That's a good idea, but considering that many spriters don't want the armor replaced at all, and that a secondary option has been discouraged to me, i'll likely close this PR soon.
I'll post the sprites in the Sprites Pit on Discord if anyone wants to do anything with em.

PR #5405, opened 2020-08-27
fix NT cloner unlimited biomatter bug

fernandos13:
## About The Pull Request
fix NT cloner unlimited biomatter bug

## Why It's Good For The Game
fixes a bug

## Changelog
:cl: Fernandos33
fix: NT cloner will not check if player is trying to add more sheets than it got
/:cl:

### DIFF
--- a/code/modules/core_implant/cruciform/machinery/cloning.dm
+++ b/code/modules/core_implant/cruciform/machinery/cloning.dm
@@ -401,7 +401,8 @@
 		if (B.biomatter_in_sheet && B.amount)
 			var/sheets_amount_to_transphere = input(user, "How many sheets you want to load?", "Biomatter melting", 1) as num
 			if (sheets_amount_to_transphere > B.amount)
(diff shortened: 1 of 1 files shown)


code/modules/core_implant/cruciform/machinery/cloning.dm
```diff
@@ -400,6 +400,9 @@
 		var/obj/item/stack/material/biomatter/B = I
 		if (B.biomatter_in_sheet && B.amount)
 			var/sheets_amount_to_transphere = input(user, "How many sheets you want to load?", "Biomatter melting", 1) as num
+			if (sheets_amount_to_transphere > B.amount)
```
Hopekz, line 403: Why not just make it set it to the amount that you have so that you can enter 99 when you have 20 sheets and it puts in your 20 sheets vs making you redo the interaction to get the exact number right

fernandos13: sure can do.

PR #5413, opened 2020-08-30
In theory fixes the huge organ issue destroying the server

Hopekz:
Can't test this at the moment but this fix, in theory, will work.

Issue:
----
In human_defense.dm

After C is defined (as obj/item/clothing) it is calling a proc that doesn't exist inside of its scope twice ( getRating() ) .
getRating() exists in /datum/armor not obj/item/clothing
image

My fix:
-----

Mimics the function of getRating() in this place by just doing vars[] instead since getRating is basically just vars[]

## Changelog
:cl: Hopek
fix: In theory fixed the huge organ issue destroying the server
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -165,8 +165,8 @@ meteor_act
 		if(gear && istype(gear ,/obj/item/clothing))
 			var/obj/item/clothing/C = gear
 			if(istype(C) && C.body_parts_covered & def_zone.body_part)
-				if(C.armor.getRating(type) > protection)
-					protection = C.armor.getRating(type)
+				if(C.armor.vars[type] > protection)
+					protection = C.armor.vars[type]
 
 	return protection
 


TorinoFermic:
Let see how it goes the fix.

drexample:
was the PR not tested or the performance fix not confirmed..?

Hopekz:
> was the PR not tested or the performance fix not confirmed..?

It does, in theory, fix a huge organ-related bug that was recently introduced of unhandled calls (proc didn't exist in scope) being created every single time anyone with an organ was attacked (very common occurrence because everyone has internal and external organs)
Whether or not this was the only issue that was causing the garbage collection issue is a mystery!

As far as testing I was unable to perform any on my end as my company work laptop I whipped this up on doesn't allow me to run BYOND.
I did follow a few theoretical examples in the code to make sure var[] performed the same function as getRating() but everything was done based on my previous knowledge of DM vs seeing results!

I trust that the maintainers have done their due diligence of making sure my change doesn't blow up the codebase before the merge.

PR #5414, opened 2020-08-30
Items left inside the cryopod will come back to occupant on eject.

Hopekz:
Bug: 
When clothes get changed inside of the Cryopod (a custom clothing loadout for example), the items get dumped on the ground. Since the cryopod is closed during this, this means that the contents are now inside the cryopod. The re-equipping code doesn't check for the contents of objects so it didn't put those back on the occupant.

This also happens to be a feature as well!
It is more efficient to do this check here vs where the initial re-equipping loadout uniform happens since doing it here means that other instances of items being left in the pod will be addressed instead of just on clothing changes.

## Changelog
:cl: Hopek
add: The cryopod will now attempt to give your items back on eject if you leave items inside of the pod.
fix: fixed a re-equipping bug where it left your items (Such as the PDA) on the floor if you equip a custom loadout.
/:cl:

### DIFF
--- a/code/game/machinery/cryopod.dm
+++ b/code/game/machinery/cryopod.dm
@@ -426,7 +426,7 @@
 	if(usr.stat != 0)
 		return
 
-	//Eject any items that aren't meant to be in the pod.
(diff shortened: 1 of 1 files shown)


SpadesSlick:
Good change.

PR #5423, opened 2020-09-03
autoinjecor now bypass hardsuits

fernandos13:
## About The Pull Request
autoinjecor now bypass hardsuits

## Why It's Good For The Game
autoinjecor now bypass hardsuits
 to avoid docs loosing someone in a hardsuit to a 30sec cooldown in a AUTOinjector with 5u of inaprovaline. 

## Changelog
:cl:Fernandos33
tweak: autoinjecor now bypass hardsuits
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/hypospray.dm
+++ b/code/modules/reagents/reagent_containers/hypospray.dm
@@ -30,6 +30,8 @@
 		var/obj/item/weapon/rig/RIG = H.get_equipped_item(slot_back)
 		if((istype(RIG) && RIG.suit_is_deployed()) || istype(SS))
 			injtime = 30
+			if (istype(src, /obj/item/weapon/reagent_containers/hypospray/autoinjector))
+				injtime = 0
 			var/obj/item/organ/external/affected = H.get_organ(BP_CHEST)
 			if(BP_IS_ROBOTIC(affected))
 				to_chat(user, SPAN_WARNING("Injection port on [M]'s suit is refusing your [src]."))


Hopekz:
This is a known downside of the hardsuit so, in my opinion, removing this downside is a step towards hugbox.

GrayRachnid:
> 
> 
> This is a known downside of the hardsuit so, in my opinion, removing this downside is a step towards hugbox.

It's only meant for auto-injectors and that's limited only to inaprovaline. This is because we have no way of bringing people back to life and sometimes an injection of inaprovaline is the only thing that will keep them from dying. 

I understand your concern and where it is coming from though.

genessee596:
I agree - since this does nothing to allow things like normal syringes or other chemical injectors besides the 5u inaprovaline, this just gives doctors a chance to save someone in a hardsuit if they're smart, while not making it as easy as dumping 30u tricord into them and having them limp off to medbay.

Firefox13:
It is not 30 seconds, remembee that BYOND uses  deci-seconds, 1/10 seconds and 3 seconds isn't much of a delay.
Still, it is a good change(unless we get refillable autoinjectors with zombie powder)

Pariah919:
We've already buffed voidsuits to be incredibly fast in the name of """qol""" so now literally everyone wears them.
Autoinjectors don't need to bypass hardsuits. That's just wacky.

SpadesSlick:
Is it even possible to lost someone to 3 seconds cooldown?

Hopekz:
In addition to the recent arguments, this PR introduces inconsistent behavior that looks unintentional.

We expect syringes to take the 3 seconds for hardsuits because of the mechanic we have in place of finding the injection port.
If we make it instant it strongly implies that the needle itself is going through the "hard" part of the hardsuit since there is now no check for a port.

fernandos13:
i think this summed up the thing "Introducing inconsistency in needle behavior is not worth what this PR is trying to prevent."
so with this words. one that i do agree with. although ive seen many hardsuit wearers die with em. cause of it. it rings back the other point "IT should be one of the drawbacks of it"

so with this two... "points" hammered. i think its enough to close this PR

PR #5432, opened 2020-09-08
Erismed bones draft

TheShown911:
## About The Pull Request

Adds bones as supporting organs for every external organ, they cause the organs explosion upon removal. Also adds variants for robotic organs. Puts bone fractures on the bones themselfs instead of the bodypart.

Note: 7 new internal organs for humanoids are potentialy dangerous untill organ deleting gets fixed. (The new GC may have fixed this.)

## Why It's Good For The Game
More organs to interact with are fun, and a small step closer to erismed.

## Changelog
:cl:
add: Bones for external organs
/:cl:


code/modules/organs/external/_external.dm
```diff
@@ -46,6 +45,7 @@
 	var/number_wounds = 0				// number of wounds, which is NOT wounds.len!
 	var/list/children = list()			// Sub-limbs.
 	var/list/internal_organs = list()	// Internal organs of this body part
+	var/default_bone_type = null
```
Evankhell561, line 48: var/default_bone_type

code/modules/organs/organ_description.dm
```diff
@@ -5,6 +5,7 @@
 	var/body_part = null
 	var/parent_organ = null
```
Evankhell561, line 6: Remove < = null>

Evankhell561 (changes requested):

fernandos13:
Oh YES erismed uhuull

code/modules/organs/external/_external.dm
TheShown911: Fixed.

code/modules/organs/organ_description.dm
TheShown911: Fixed.

fernandos13:
OHHHH fuck ye

TheShown911:
Added organ icons to the surgery UI in this PR.

TorinoFermic:
Quick question : What about admin heals for these bones ?

TheShown911:
> Quick question : What about admin heals for these bones ?

They should work normaly as other organs, and shouldn't cause external organ explosion while the body has the REBUILDING flag

TorinoFermic (approved):

PR #5447, opened 2020-09-13
processor pickup buff, plasma slag fix

Hatterhat:
## About The Pull Request
plasteel no longer makes an extra piece of plasma ore into slag
the ore processor picks ore up off the ground faster
## Why It's Good For The Game
i do not enjoy having my perfectly good plasma turned into slag
also slow pickups are cringe and controllable ore processing speed would be nice but that's out of scope of this fix
## Changelog
:cl:
fix: Plasteel now alloys properly without extra conversion of plasma ore into slag.
tweak: The ore processor on the Hulk no longer has a limiter on how much ore it can pick up per tick.
/:cl:

### DIFF
--- a/code/modules/mining/machine_processing.dm
+++ b/code/modules/mining/machine_processing.dm
@@ -142,14 +142,14 @@
 
 /obj/machinery/mineral/processing_unit/Process()
 
-	if (!output_dir || !input_dir) return
+	if(!output_dir || !input_dir)
+		return
 
 	var/list/tick_alloys = list()
 
 	//Grab some more ore to process this tick.
-	var/limit = sheets_per_tick
 	for(var/obj/item/weapon/ore/O in get_step(src, input_dir))
-		if(--limit <= 0)
+		if(!O)
 			break
 		if(!isnull(ores_stored[O.material]))
 			ores_stored[O.material]++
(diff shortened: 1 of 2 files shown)


TorinoFermic (approved):

code/modules/mining/machine_processing.dm
```diff
@@ -142,14 +142,14 @@
...
+	if(!output_dir || !input_dir)
+		return
 
 	var/list/tick_alloys = list()
 
 	//Grab some more ore to process this tick.
-	var/limit = sheets_per_tick
 	for(var/obj/item/weapon/ore/O in get_step(src, input_dir))
-		if(--limit <= 0)
+		if(!O)
```
MadmanMartian, line 152: O can never be null.

code/modules/mining/machine_processing.dm
```diff
@@ -142,14 +142,14 @@
...
 /obj/machinery/mineral/processing_unit/Process()
 
-	if (!output_dir || !input_dir) return
+	if(!output_dir || !input_dir)
+		return
 
 	var/list/tick_alloys = list()
 
 	//Grab some more ore to process this tick.
-	var/limit = sheets_per_tick
```
MadmanMartian, line 150: Removing this removes any need for the sheets_per_tick variable.

code/modules/mining/machine_processing.dm
TorinoFermic: Yeah, he could removed the if() case

PR #5457, opened 2020-09-17
You can now compile and run the codebase through VS code or extools without errors on start.

Hopekz:
Works now if you're running through extools or VS code.
Only an issue because these runtimes pause the whole application on boot.

Also should in theory help memory usage if sharer doesn't exist in gas mixture which for some reason I encountered.

Ba1bGUJfDh
ynUxFwMzr1

## Changelog
:cl: Hopek
code: You can now compile and run the codebase through VS code or extools without errors on start.
/:cl:


TorinoFermic (approved):

TorinoFermic:
You can disable halting on any runtimes in the debugger from VS Code

PR #5462, opened 2020-09-18
solidifier update

Evankhell561:
## Changelog
:cl:
tweak: building dcuts is faster and cheaper.
tweak: the solidifier is compatible with the plumbing system.
tweak: the solidifier connector now points south.
/:cl:

### DIFF
--- a/code/datums/craft/recipes/machinery.dm
+++ b/code/datums/craft/recipes/machinery.dm
@@ -86,10 +86,9 @@
 	name = "fluid duct"
 	result = /obj/item/stack/ducts
 	steps = list(
-		list(CRAFT_MATERIAL, 2, MATERIAL_PLASTIC, "time" = 15),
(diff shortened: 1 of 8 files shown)


Firefox13:
It still makes 1 fluid duct per craft sequence which is awfully low. Seriously, it'll take a lot of time to create enough fluid ducts even for basic schemes. We'll need either add a design for an autolathe and add it to design or increase the amount constructed.
As an addendum, here are some plumbing images that show *how much* ducts you will need:
изображение
изображение

Evankhell561:
> It still makes 1 fluid duct per craft sequence which is awfully low. Seriously, 

as i said in discord, i plan to add the ducts to autolathe, along with two or three more updates. this is just a quick fix focused on solidifer.

TorinoFermic (approved):

PR #5485, opened 2020-09-21
Makes the Energy Armblade Implant stop deleting itself

Michiyamenotehifunana:
## About The Pull Request

The Energy Armblade Implant that can be gotten from uplinks has a critical flaw: It stops working 1 minute after you first use it. This is because it uses the hardsuit energy blade projector's blades, which would normally instantly despawn after being dropped. I have no idea why the implant takes a full minute to delete itself instead of instantly, but it just does.

This PR will fix that, without allowing the hardsuit energy blade projector to infinitely spam more and more blades onto the floor.

## Changelog
:cl: Toriate
fix: Energy Armblade implant should no longer delete itself a minute after being used for the first time
/:cl:


mattroks101:
Holy shit that base code is so cursed.

mattroks101 (approved):

TorinoFermic (approved):

PR #5493, opened 2020-09-23
fix runtime in oddities.dm, 29

Evankhell561:
## Changelog
:cl:
fix: runtime in technomancer oddities
/:cl:


TorinoFermic (approved):

TheShown911:
I would write changelogs even normal players could unerstand, for example `fix: runtime in technomancer oddities`

PR #5497, opened 2020-09-24
Shields fix

soslow4u:
## About The Pull Request

- Moved conduits spawn to main generator
- Removing single conduit would make generator remove whole batch
- Deploying conduits allows you to setup hull and multi z shields

New shield generators won't make any messages still

## Changelog
:cl:
fix: You can build real shield generators from boards. Right-click and deploy conduits for multi dock and hull shielding options
/:cl:


GrayRachnid:
hull.dm deploys conduits when it is built, the normal one does not. Conduits are currently a snowflake structure, the order asked to make them fixable/removeable.

This is up to the maintainers to merge but a sloppy fix like this should not be accepted in my opinion.

soslow4u:
Such slowflakey item is hard-linked to hull generator and they pops in when *hull* core was unanchored from floor.
So you want to keep those, right. Let me pour a bit more code there then

soslow4u:
updated with conduits handling and such stuff

soslow4u:
fine

code/modules/shield_generators/hull.dm
```diff
@@ -2,133 +2,9 @@
...
-			toggle_tendrils(FALSE)
-			anchored = FALSE
-		else
-			if(istype(get_turf(src), /turf/space)) return //No wrenching these in space!
-			if ((toggle_tendrils(TRUE)))
-				to_chat(user, SPAN_NOTICE("You secure the [src] to the floor!"))
-				anchored = TRUE
-		return
\ No newline at end of file
+	current_energy = max_energy * 0.30
```
TheShown911, line 10: No newline.

TheShown911 (commented):
Looks good

TorinoFermic (approved):

PR #5503, opened 2020-09-24
Adds a message that explains why overmap doesn't work on local

ACCount12:
This is, like, the tenth time someone is confused why overmap console is failing silently.

I added a message that displays when you try to access helm console, but overmap is disabled in config, and you have an admin rank.


TorinoFermic (approved):

mattroks101 (approved):

mattroks101:
At last.

PR #5505, opened 2020-09-25
[ORDERS] Implements random starting areas for the CEV Eris. One of 3 default l…

KirkwallDay:
…ocations will start where the CEV Eris is. It's random each round, and more can be added.

## About The Pull Request
This request fills an order from the Discord to have the Eris start at one of the three areas on the overmap.
Capture

The Three locations are The asteroid, Abandonded Fortress, or Ruins. This is accomplished in a somewhat strange way, since the Eris is spawned first, I had to make it so the random landmark will spawn at the Eris instead. This also means the asteroid loses it's default spawn status at the Eris, and its default coordinates are now randomized, just like the fortress and ruins are.

It's also built to be expandable, create more sectors, and then add them to the list.

## Why It's Good For The Game

It should make rounds more varied, and offer players a chance to see the other areas, even when there's no captain or anyone driving the ship.

## Changelog
:cl: yorpyorpyorp
add: Eris will now randomly start at 1 of 3 locations.
add: Lines in config file to determine starting location.
del: Asteroid loses its default location.
/:cl:

### DIFF
--- a/code/modules/overmap/sectors.dm
+++ b/code/modules/overmap/sectors.dm
@@ -18,6 +18,7 @@
 	var/in_space = 1	//can be accessed via lucky EVA
 
 	var/global/eris_start_set = FALSE //Tells us if we need to modify a random location for Eris to start at
+	var/global/eris
 
 /obj/effect/overmap/Initialize()
 	. = ..()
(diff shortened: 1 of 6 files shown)


code/modules/overmap/sectors.dm
```diff
@@ -17,6 +17,8 @@
 	var/known = 1		//shows up on nav computers automatically
 	var/in_space = 1	//can be accessed via lucky EVA
 
+	var/global/eris_start_set //Tells us if we need to modify a random location for Eris to start at
```
seekin4u, line 20: init it to false, not to `NULL`

KirkwallDay: Good point, change made!

genessee596 (approved):
This PR gets the Delegate stamp of approval.

It'd be nice to see other ruins more than once in a blue moon. This might also encourage people to make more ruins as well, which is nice.

Still needs actual code review from a maintainer.

TorinoFermic (approved):

code/modules/overmap/sectors.dm
```diff
@@ -27,9 +29,25 @@
...
+	// Spawning location of area is randomized or default values, but can be changed to the Eris Coordinates in the code below.
+	// This provides a random starting location for Eris.
 	start_x = start_x || rand(OVERMAP_EDGE, GLOB.maps_data.overmap_size - OVERMAP_EDGE)
 	start_y = start_y || rand(OVERMAP_EDGE, GLOB.maps_data.overmap_size - OVERMAP_EDGE)
 
+	if (!eris_start_set)
+		// Find the eris object
+		for (var/obj/effect/overmap/ship/eris/O in world)
+			if (istype(O, /obj/effect/overmap/ship/eris))
+				if(O.name == "CEV Eris")
```
TheShown911, line 41: The whole way the ship gets found is kinda strange, but hopefuly random start will always stay on so it shouldn't be a big problem.

TorinoFermic: I hope our code has this eris ship being already in a global var/list

KirkwallDay: I can't find it in the global vars, so I made a new var list for ships and put it in there.
I added configuration file support. 
Now that the eris no longer needs to be 'found' that part is gone.

genessee596:
This has gotten approval from maintainers and delegates, so I'm going to go ahead and merge this.

PR #5509, opened 2020-09-25
Alters hardsuit injectors to use 5u injections

Anonmare:
## About The Pull Request
Injectors from hardsuits will now only use 5u at a time when injecting someone

## Why It's Good For The Game
A number of players have requested it, specifically with regards to chemicals like synaptazine which are nearly always fatal at does above 5u without assistance. Considering that injectors are instant-effect and have no delay on use, I don't see how reducing their amount used by this much would be a negative and would bring it in line with other injectors, such as syringes. 

## Changelog
:cl:
tweak: Hardsuit injectors now use 5u injections instead of 10u
/:cl:

### DIFF
--- a/code/modules/clothing/spacesuits/rig/modules/utility.dm
+++ b/code/modules/clothing/spacesuits/rig/modules/utility.dm
@@ -209,7 +209,7 @@
 	if(!charge)
 		return 0
 
-	var/chems_to_use = 10
+	var/chems_to_use = 5
 	if(charge.charges <= 0)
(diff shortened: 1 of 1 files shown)


TorinoFermic (approved):

TorinoFermic:
Wasnt real life medicine requiring to dilute with water in a syringe or a bag before injecting someone ?

Anonmare:
Yes, but there's no way to dilute medicine with hardsuit injectors and that's beyond my ability to code.

PR #5525, opened 2020-09-29
Removes poorly-implemented bay animals from the infestation event.

MadmanMartian:
## About The Pull Request 

I've been told a few times that the bay animals are wasted on an infestation event, as they provide nothing to the round. Most of the ones removed don't even fit the setting.

## Why It's Good For The Game
Less points wasted on NPCs that can't be butchered or interacted with in a meaningful way.

## Changelog
:cl:
del: Lizards, Yithians, Tindalos, Diyaabs, Samaks, and Shantaks have been removed from the infestation event.
/:cl:

### DIFF
--- a/code/game/gamemodes/events/infestation.dm
+++ b/code/game/gamemodes/events/infestation.dm
@@ -20,18 +20,13 @@ It focuses on spawning large numbers of moderate-to-weak monsters, and includes 
 
 
 #define INFESTATION_MICE "mice"
-#define INFESTATION_LIZARDS "lizards"
 #define INFESTATION_SPACE_BATS "bats"
 #define INFESTATION_SPIDERLINGS "spiderlings"
 #define INFESTATION_SPIDERS "spider"
 #define INFESTATION_ROACHES "large insects"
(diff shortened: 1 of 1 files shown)


GrayRachnid:
Can they replaced with onestar drones?

MadmanMartian:
They could, but why?

GrayRachnid:
more variety, imo. They're rarely seen and are only spawned on the dungeon, seeing them on the ship would be a fresh experience for players.

TheShown911 (approved):
We should probably find a replacement for these for more variety, but that can wait. Also possible remove these mobs from the code itself for less bloat.

githubuser4141:
> more variety, imo. They're rarely seen and are only spawned on the dungeon, seeing them on the ship would be a fresh experience for players.

I know I'm late to the party, but it'd be nice if every single mob that spawned wasn't trying to kill you. It was interesting seeing funny-looking but harmless nuisance animals pop out of the floors, it'd be sad if the only things that spawned were hostile mobs that you have to destroy.

PR #5531, opened 2020-09-30
[BOUNTY] Bug: Fix for robotic limbs in biogenerator

KirkwallDay:
## About The Pull Request
Fixes the problem of people with robotic limbs not being affected by the biogenerator.
Fix works by applying protection as a percentage to the damage dealt, rather then excluding all armor, robotic limbs have armor which was causing the bug. This should fix other issues that might come up to, such as armor vests etc...

## Why It's Good For The Game
Fixes a bug.

## Changelog
:cl: yorpyorpyorp
fix: fixed the biogenerator not consuming people that were wearing armor, or had robotic limbs.
/:cl:


code/modules/mechs/mech_interaction.dm
```diff
@@ -1,4 +1,8 @@
 /mob/living/exosuit/MouseDrop_T(atom/dropping, mob/user)
+	if (user.buckled)
+		to_chat(user, SPAN_WARNING("You cannot enter a mech while buckled, unbuckle first."))
```
drexample, line 3: ?

drexample (changes requested):
Your branch has unrelated commits. Fix it.

code/modules/mechs/mech_interaction.dm
KirkwallDay: Okay, I manually went in there and deleted those changes. It says at the top that only 1 file should be changed now. If there's a different way to do it, let me know.

TorinoFermic (approved):

PR #5534, opened 2020-09-30
[BOUNTY] Exploit: Fixes people entering mechs while buckled.

KirkwallDay:
## About The Pull Request
Fixes people being able to enter mechs while buckled to a chair or something.

## Why It's Good For The Game
Fixes a bug

## Changelog
:cl: yorpyorpyorp
fix: Can no longer enter mechs while buckled.
/:cl:


code/modules/mechs/mech_interaction.dm
```diff
@@ -1,4 +1,8 @@
 /mob/living/exosuit/MouseDrop_T(atom/dropping, mob/user)
+	if (user.buckled)
```
MadmanMartian, line 2: Put it in check_enter instead, which is somewhere around line 168. That's where enter checks should be.

It'll prevent people from attempting to enter, starting the do_after, then buckling themselves and avoiding this check.

MadmanMartian (changes requested):
see comment

code/modules/mechs/mech_interaction.dm
KirkwallDay: Good idea, I didn't even notice that section. Updated the pull.

MadmanMartian (approved):

TorinoFermic (approved):

PR #5543, opened 2020-10-02
[BOUNTY] QoL: Press E to put in backpack . Press shift + e to put things in your belt  PART 2

MacBlaze1:
## About The Pull Request

fucking review before you merge

## Why It's Good For The Game

bugs bad

## Changelog
:cl:

fix: fixed errors with the qol hotkeys pr
/:cl:


Hopekz (changes requested):
It still doesn't do what is required.

What happens and how to recreate it:
1. spawn in 
2. hold an equitable item or an item that can go on your body
3. press e
4. The item will go to your backpack instead of your body even though I have available slots on my body.

**What is expected:**
The item is expected to go on your body on an available slot before it goes to the backpack. 
It currently defaults to the backpack no matter what. This is not the expected behavior we are looking for in order to fulfill this bounty.
It should only go to the backpack after the usual behavior of attempting to equip the item on the body/body slots.

**Bug 2:**
It doesn't go to the backpack when the backpack is not on the back slot.

Expected behavior:
If I am holding an item in my left hand and have a backpack in the right hand: I try to equip it. if I am unable to equip the item on my body it should put it in the backpack that is in my right hand. At the moment it does not find the backpack that is in my hands

GrayRachnid:
I've tested this, it works.

TorinoFermic (approved):

TorinoFermic:
I tested it, it works.

PR #5552, opened 2020-10-02
Non-excel can no longer use the excelsior teleporter

MadmanMartian:
## About The Pull Request 

This locks the excel teleporter to only being usable by excelsior

## Why It's Good For The Game

No more invalidating 90% of the ship through one stash. A little boring, but the straightforward method of dealing with it.

## Changelog
:cl:
balance: The excelsior teleporter is now only usable by excelsior.
/:cl:


Hopekz:
I like #5554 better.
Thank you for providing both options!

Going with the "Removes the excel teleporter from random excel stashes" option just restores that part to what it was before
since before excel machinery couldn't be found randomly.

PR #5575, opened 2020-10-09
Atomize the repo

Evankhell561:
atomize the repo.
makes a lot of changes to the way the code is written. this should not be noticeable at all.

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -935,7 +935,7 @@ proc/DuplicateObject(obj/original, perfectcopy = 0 , sameloc = 0)
 
 					for(var/mob/M in T)
 
-						if(!istype(M,/mob) || isEye(M)) continue // If we need to check for more mobs, I'll add a variable
+						if(!ismob(M) || isEye(M)) continue // If we need to check for more mobs, I'll add a variable
 						mobs += M
 
 					for(var/mob/M in mobs)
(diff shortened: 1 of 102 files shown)


code/__HELPERS/unsorted.dm
```diff
@@ -1049,7 +1049,7 @@ proc/is_hot(obj/item/W)
 	if(W.sharp) return TRUE
 	return ( \
 		W.sharp														|| \
-		istype(W, /obj/item/weapon/tool)							|| \
+		istool(W)													|| \
```
Kurgis, line 1052: Was this meant to keep its tabs?

code/modules/hivemind/self_defense.dm
```diff
@@ -134,7 +134,7 @@
 	for(var/i = 1 to amount)
 		var/turf/spawn_loc = pick(places_to_spawn)
 		champion = new champion_path(spawn_loc)
-		champion.maxHealth += 150
+		champion.adjustMaxHealth(150)
```
TorinoFermic, line 137: You kinda used wrong function : setMaxHealth should be used, the next line can use a function adjustHealth

TorinoFermic: But reading more this function, why we have adjust**Max**Health and setMaxHealth

TorinoFermic: Okay, nevermind I found your newly added function. It turns out adjustMaxHealth hasnt never existed.

TorinoFermic (approved):

TheShown911:
Clean

PR #5586, opened 2020-10-12
Hivemind tyrant fixes

MadmanMartian:
Fixes the tyrant destroying half the ship on death
Removes redundant checks
Removes redundant walk(src,0) calls
Fixes the tyrant from stopping time every time it shit talks you, and then embedding you with 80 ~~knives~~ globs of goo

## About The Pull Request 

AAAAAAAAAAAAA asked me for a little help on fixing the tyrant, so I fixed some surface-level bugs.

 - The tyrant, on death, is supposed to kill all hivemind-controlled machines on the ship. Issue was it called qdel() rather than node.destruct, so it would also destroy the underlying machines.
 - The tyrant, in combat, mocks people every 3~5 seconds. Issue is it sleeps for this amount of time, which would also sleep the entire Mob subsystem. Same goes for it firing projectiles
 - It would call the usual hostile attack chain within its own Life(), after previously calling the hostile attack chain in a parent call. This led to megafauna attacking twice.
 - for some reason, all of the megafauna death calls had a redundant health check in them, when most of them were called below 0 health anyway. This interfered with debugging and was removed
 - The hostile death() function called walk(src,0), which was redundant as simple_animal death() already called it.

## Why It's Good For The Game

 - It's probably best that the gravity generator doesn't get obliterated when you win against the hivemind, by shooting its cyber demon until it dies
 - It's also probably best that it doesn't stop time across the entire ship every 3~5 seconds when you fight it.
 - redundant health checks begone
 - redundant walk calls begone

## Changelog
:cl:
balance: The hivemind tyrant no longer stops time for the entire server during combat
balance: The hivemind tyrant no longer destroys a majority of the ship upon death
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/hostile.dm
+++ b/code/modules/mob/living/simple_animal/hostile/hostile.dm
@@ -180,10 +180,6 @@ var/list/mydirs = list(NORTH, SOUTH, EAST, WEST, SOUTHWEST, NORTHWEST, NORTHEAST
 
 	return L
 
-/mob/living/simple_animal/hostile/death()
-	..()
-	walk(src, 0)
-
 /mob/living/simple_animal/hostile/Life()
 
 	. = ..()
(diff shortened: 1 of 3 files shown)


Wouju:
Thanks for the help with the tyrant, will make Champion call use spawn weight after this is merged

TorinoFermic (approved):

mattroks101 (approved):

genessee596:
Bugfix approved by two maintainers. Easy merge moment.

PR #5592, opened 2020-10-15
Fix runtime in mob_stats.dm, 46 and machines.dm, 485

Evankhell561:


code/datums/mob_stats.dm
```diff
@@ -43,7 +43,8 @@
 /datum/stat_holder/proc/getStat(statName, pure = FALSE)
 	if (!islist(statName))
 		var/datum/stat/S = stat_list[statName]
-		SEND_SIGNAL(holder, COMSIG_STAT, S.name, S.getValue(), S.getValue(TRUE))
+		if(holder)
```
MadmanMartian, line 46: Holder should never be null, diagnose this further.

Evankhell561: I've already tried and found nothing except that mob stats are never destroyed.

TorinoFermic (approved):

PR #5610, opened 2020-10-19
Makes blitzshells rarer

MadmanMartian:
## About The Pull Request 

When making the blitzshell to specification, I accidentally overlooked some things when making the storyteller roleset, as I was unfamiliar with it.

 - Blitzshells now only spawn if there are 10 or more people, along with a requirement for some leve of security, so there are potentially enough people to deal with them, considering how lethal they can be.
 - Blitzshell storyteller event is now less likely to fire, as it does not contribute as much to the round as traitors/excel/carrion.

## Why It's Good For The Game

Blitzshells do not contribute as much to the round as in-station antagonists, so should have less of a chance to spawn than traitors/excel/carrion, and require a certain amount of people to wrangle, security namely.

## Changelog
:cl:
balance: Blitzshells are now less likely to spawn
/:cl:

### DIFF
--- a/code/game/gamemodes/roleset/simple.dm
+++ b/code/game/gamemodes/roleset/simple.dm
@@ -12,8 +12,9 @@
 	id = "blitz"
 	name = "blitzshell infiltration"
 	role_id = ROLE_BLITZ
-	weight = 1
+	weight = 0.6
 
+	req_crew = 10
 	base_quantity = 1
 	scaling_threshold = 15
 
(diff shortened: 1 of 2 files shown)


GrayRachnid:
There was something int he design document about making traitors able to spawn blitzshells for X amount of TC while making the blitzshells subservient to them. How about adding that while making them rarer?

Hopekz (approved):
The required crew and sec change is definitely needed.
Don't care much for the "more rare" part but I cannot see this being a negative change.

Iglpigl:
This needs to be tested, #4764 was merged and it didn't do anything, Serbs still spawn on any pop, I'm not even sure if req_crew is working properly.

PR #5612, opened 2020-10-21
Nerfs stunning people by dropping things on them

MadmanMartian:
When the best method of disabling a serbian mercenary is to drop crayons on their head, you have to question the balance.

## About The Pull Request

This pull request adds a helmet armor check when dropping an item on somebody from a height, so you can't stunlock somebody by throwing individual playing cards on people, but you can if you drop a locker on them for instance.

## Why It's Good For The Game

After seeing people get stunned by a deck of cards, or food paste packets, while decked to the nines in combat armor, I felt it was time to take a look at what code was doing this.

It's still a viable tactic to disable somebody, just now a little more sane.

## Changelog
:cl:
balance: Helmets are now taken into account when stunning people by throwing things on top of them.
/:cl:

### DIFF
--- a/code/modules/multiz/turf.dm
+++ b/code/modules/multiz/turf.dm
@@ -166,7 +166,8 @@ see multiz/movement.dm for some info.
 			var/fall_damage = mover.get_fall_damage()
 			if(M == mover)
 				continue
-			M.Weaken(10)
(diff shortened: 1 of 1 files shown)


Iglpigl:
Does this also fix blood and potentially other liquids from stunning from a Z-level above the target or no.
Example: Currently blood drops from someone who is above you is bleeding and his blood falls on you you get stunned.

PR #5616, opened 2020-10-22
Flashbang balance change

MadmanMartian:
Flashbangs no longer floor you, instead they render you blind, deaf, and confused

## About The Pull Request

Flashbangs no longer weaken and stun you when detonated. Instead, they blind you if you lack proper eye protection, apply the usual deafening effect they had previously, and now debuff your stats for a short time while you recover.

## Why It's Good For The Game

Being able to 1-second slap stun somebody for 10 seconds is a little anticlimatic. This change would prompt people to time their grenades properly, and actually breach and clear with them.

## Changelog
:cl:
balance: Flashbangs no longer floor you, instead they disorient you.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/grenades/flashbang.dm
+++ b/code/game/objects/items/weapons/grenades/flashbang.dm
@@ -35,7 +35,7 @@
 //Checking for protections
 	var/eye_safety = 0
 	var/ear_safety = 0
-	var/ear_stun_mult = 1
+	var/stat_def = -STAT_LEVEL_ADEPT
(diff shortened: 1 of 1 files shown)


TheRespectedMan:
A good change, flashbangs are an assault on your senses, not a stun weapon. 
Assuming this is not already the case, I'd say it is a good idea that the flashbang also make the screen shake, similar to how the screen shakes when you are drunk, in order to more realistically mirror a concussion.

mattroks101 (approved):

PR #5620, opened 2020-10-23
Adds missing knife in boot overlays

Garen7:
## About The Pull Request

Sorry for the small picture but you get this orange error icon on the boots when you put a knife in it and this PR fixes that
image

Now it looks like this without the knife:
image
And like this with the knife(its the little pixels on the left heel):
image

Anyways the fix was just to add the missing overlays which is fine but if a maintainer wants me to refactor it so the overlay icon name is held in a variable so we can just have a default overlay icon then let me know.
image
## Changelog
:cl:
fix: Fixes the orange error on boots with a knife in them
/:cl:


mattroks101:
image
I had to look three times before I could even notice the pixel. Do we not have a better sprite for this?

Garen7:
This is the only one I'm aware of, if another one exists let me know and ill put it in

mattroks101 (approved):
This fixes the bug, but we should look into getting a better sprite for this ASAP.

PR #5627, opened 2020-10-25
Fixes bullets floating in the air 

drexample:
## About The Pull Request

Port from Onyx's bay

## Why It's Good For The Game

Ay mlya ya masleenoo poymal

## Changelog
:cl:
fix: Bullets should no longer be able to float in the air without moving
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -684,7 +684,7 @@
 	return TRUE
 
 /obj/item/projectile/Process()
-	var/first_step = 1
+	var/first_step = TRUE
 	var/i = 0
 	spawn while(src && src.loc)
 		if (++i > 512)
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -842,6 +842,7 @@
 	return Process(targloc)
 
 /obj/item/projectile/test/Process(turf/targloc)
+	var/i = 0
 	while(src) //Loop on through!
 		if (++i > 512)
 			var/turf/T = src.loc
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -684,9 +684,13 @@
 	return TRUE
 
 /obj/item/projectile/Process()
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/projectile.dm
```diff
@@ -838,7 +842,12 @@
 	return Process(targloc)
 
 /obj/item/projectile/test/Process(turf/targloc)
+	var/i = 0
 	while(src) //Loop on through!
+		if (++i > 512)
```
AlexMorgan3817, line 847: Should be 513 or current DM version c:

code/modules/projectiles/projectile.dm
```diff
@@ -684,9 +684,13 @@
 	return TRUE
 
 /obj/item/projectile/Process()
-	var/first_step = TRUE
-
+	var/first_step = 1
+	var/i = 0
 	spawn while(src && src.loc)
+		if (++i > 512)
```
mattroks101, line 690: Why 512?

code/modules/projectiles/projectile.dm
```diff
@@ -684,9 +684,13 @@
 	return TRUE
 
 /obj/item/projectile/Process()
-	var/first_step = TRUE
-
+	var/first_step = 1
```
mattroks101, line 687: Why was this changed?

code/modules/projectiles/projectile.dm
drexample: Worked for Onyx, a direct copypaste from their fix

code/modules/projectiles/projectile.dm
drexample: Copypaste mistake, fixed

MadmanMartian:
doesn't kill_count already do something similar to this?

MadmanMartian (changes requested):
Confirm this actually does something, as right now it's just what kill_count does, but kill_count calls first as it is 50 process ticks versus 512 process ticks.

If it does, bring down the projectile_stuck required process ticks from 512 to something more reasonable.

PR #5636, opened 2020-10-30
Body Scanner QoL

Garen7:
## About The Pull Request

Small changes to the body scanner to make it nicer.

## Changelog
:cl:
add: Alt clicking a body scanner will eject the occupant
add: Body scanners now put the name of the occupant on the output and name the printed paper after the occupant
spellcheck: Fixed a typo in the body scanner printout
/:cl:

### DIFF
--- a/code/game/machinery/adv_med.dm
+++ b/code/game/machinery/adv_med.dm
@@ -213,7 +213,7 @@
 			to_chat(usr, "\icon[src]<span class='warning'>The body scanner cannot scan that lifeform.</span>")
 			return
 		var/obj/item/weapon/paper/R = new(src.loc)
-		R.name = "[occupant.get_authentification_name()] scan report"
+		R.name = "[occupant.get_visible_name()] scan report"
 		R.info = format_occupant_data(src.connected.get_occupant_data())
 		R.update_icon()
 
@@ -223,7 +223,7 @@
 		return
 	var/mob/living/carbon/human/H = occupant
 	var/list/occupant_data = list(
-		"name" = H.get_authentification_name(),
+		"name" = H.get_visible_name(),
 		"stationtime" = stationtime2text(),
 		"stat" = H.stat,
 		"health" = round(H.health/H.maxHealth)*100,
--- a/code/game/machinery/adv_med.dm
+++ b/code/game/machinery/adv_med.dm
@@ -60,6 +60,10 @@
 	update_use_power(1)
 	update_icon()
 
+/obj/machinery/bodyscanner/AltClick(mob/user)
+	if(Adjacent(user))
+		eject()
+
 /obj/machinery/bodyscanner/proc/set_occupant(var/mob/living/L)
 	L.forceMove(src)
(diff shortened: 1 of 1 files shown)


code/game/machinery/adv_med.dm
```diff
@@ -209,7 +213,7 @@
 			to_chat(usr, "\icon[src]<span class='warning'>The body scanner cannot scan that lifeform.</span>")
 			return
 		var/obj/item/weapon/paper/R = new(src.loc)
-		R.name = "Body scan report"
+		R.name = "[occupant.real_name] scan report"
```
TorinoFermic, line 216: Should this scanner reveal the real identity ?

Garen7: I did this because I didn't want one person with an agent id to go around framing the entire crew for being a carrion/excel/borer. Is there some exploit with this I'm missing? I assumed you'd hear their voice and know the real name anyway if you were scanning them. Maybe I could look into seeing if I can get the name from the dna instead. Any proposed better solution?

Wouju: > Maybe I could look into seeing if I can get the name from the dna instead.
Sounds like a good idea, if you're able to do it that would be great.

TheShown911: Considering these antags don't have an easy way to get an agent ID, and that pulling something like this off would be a very interesting RP scenario i wouldn't change it.

MadmanMartian, line 216: Revert this.

Garen7: multi role antags (ie traitor excel) is more common than you'd think

code/game/machinery/adv_med.dm
```diff
@@ -219,6 +223,7 @@
 		return
 	var/mob/living/carbon/human/H = occupant
 	var/list/occupant_data = list(
+		"name" = H.real_name,
```
MadmanMartian, line 226: Revert/change this. Your identity is detected depending on your ID, as it is a machine. Have it be their name, not their real_name.

MadmanMartian (changes requested):
Subvertive tactics is an interesting gameplay mechanic. Don't undermine it.

code/game/machinery/adv_med.dm
Garen7: By revert this do you mean change it to
```suggestion
		R.name = "[occupant.name] scan report"
```
or to revert it to
```suggestion
		R.name = "Body scan report"
```
asking this bc you wanted the normal name in the comment below

MadmanMartian: Go with get_visible_name, would be a body scanner QoL (the reason of your PR) so that it can be differentiated when ironhammer mass-scan people

Garen7:
Using name on id bc code comments recommend using it for machinery recognizing humans

TorinoFermic (approved):

PR #5643, opened 2020-11-01
Knocks down control spider duration to 15-20 seconds

genessee596:
## About The Pull Request

Does what it says on the tin. Reduces the control spider duration from a maximum 75 seconds to a maximum of 20 seconds.

## Why It's Good For The Game

At the moment, control spiders are a complete death sentence. The carrion can literally just get you to commit suicide and end your round - and it only kills the carrion's shell, not the core.

This doesn't really fix that problem, but what it does do is make it far harder for the carrion to remotely slaughter people via control spider. 

Control spiders should be used for spreading bits of misinformation and starting conflicts between crew - not just used as a remote killswitch. 20 seconds is enough time for a carrion to dump a few shots into a crewmember's ally or to screech that a crewmember is killing them when they aren't. 70 seconds is more than enough time that a control spider really should warrant.

## Changelog
:cl:
balance: Control Spiders now control for far, far less time.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/implant/implants/carrion/control_spider.dm
+++ b/code/game/objects/items/weapons/implant/implants/carrion/control_spider.dm
@@ -48,7 +48,7 @@
 	active = TRUE
 	last_use = world.time
 
-	addtimer(CALLBACK(src, .proc/return_mind), rand(45 SECONDS, 70 SECONDS))
+	addtimer(CALLBACK(src, .proc/return_mind), rand(15 SECONDS, 20 SECONDS))
 
 /obj/item/weapon/implant/carrion_spider/control/on_uninstall()
 	..()


SpadesSlick:
They still will be able to do this in 20 seconds. More then that, lowering control time would actually force them to do even more reckless things. 

>  and it only kills the carrion's shell, not the core.

Since you addressed the actual problem here, it's better to fix this instead of doing nerf.

Hopekz (approved):
I don't think this will fix the issue you described but I've been thinking of this nerf for nearly two months now so for reasons unrelated to your reason I'm going to slap my approval on this change.

Hopekz:
Waited for a bit to marinate it and collect some feedback before we try this.
No real code review needed here.

TheShown911:
I'm gonna do the solution proposed by the project lead and revert the duration once I got the time.

PR #5655, opened 2020-11-06
Fixes altyn bugs with minor refactor

Garen7:
## About The Pull Request

Fixes some bugs with a small refactor to the altyn, maska, and old riot helmets. Basically altyns copy pasted riot helmet code for some reason instead of just inheriting it so I merged the two types. I also wrapped the variable tracking if the face shield was up or down in a setter function so that way if someone expands the code they won't reintroduce the bugs this fixes.

The 1st bug was occurring because the armor list wasn't turning into an armor datum, and the second bug was occurring because someone copy pasted the code and didn't modify the other values after making it start with the face shield up, both are fixed by the setter proc.

## Why It's Good For The Game

Adhering to DRY standards and encapsulation to improve maintainability

## Changelog
:cl:
fix: Fixed the stats UI not working for altyn, maska, and old riot helmets
fix: Fixed altyn, maska, and old riot helmets sometimes not showing gear underneath the helmet(and other things)
refactor: Merged the altyn and old riot helmets into the same type
refactor: Properly wrapped the variable tracking if the face shield was up to avoid future bugs
/:cl:

### DIFF
--- a/code/modules/clothing/head/armor.dm
+++ b/code/modules/clothing/head/armor.dm
@@ -343,9 +343,9 @@
 	icon_state = up ? "[initial(icon_state)]_up" : initial(icon_state)
 
 //I wanted to name it set_up() but some how I thought that would be misleading
-/obj/item/clothing/head/armor/faceshield/proc/set_is_up(u)
-	up = u
-	if(u)
+/obj/item/clothing/head/armor/faceshield/proc/set_is_up(is_up)
+	up = is_up
+	if(up)
 		armor = getArmor(arglist(armor_up))
 		flash_protection = initial(flash_protection)
 		tint = initial(tint)
--- a/code/modules/clothing/head/armor.dm
+++ b/code/modules/clothing/head/armor.dm
@@ -305,58 +305,75 @@
 		MATERIAL_GLASS = 10 // glass is reflective yo, make it cost a lot of it - also, visor
 	)
 
-// Riot helmet
-/obj/item/clothing/head/armor/riot
+// toggleable face guard
+/obj/item/clothing/head/armor/faceshield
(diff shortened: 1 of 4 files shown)


code/modules/clothing/head/armor.dm
```diff
@@ -305,58 +305,75 @@
...
+/obj/item/clothing/head/armor/faceshield/attack_self()
 	toggle()
 
-/obj/item/clothing/head/armor/riot/update_icon()
+/obj/item/clothing/head/armor/faceshield/update_icon()
 	icon_state = up ? "[initial(icon_state)]_up" : initial(icon_state)
 
-/obj/item/clothing/head/armor/riot/verb/toggle()
+//I wanted to name it set_up() but some how I thought that would be misleading
+/obj/item/clothing/head/armor/faceshield/proc/set_is_up(u)
```
TorinoFermic, line 346: You can name it as flip_it_up

Garen7: I feel like it should have the word set in it since it's a setter function, also flip_it_up implies that the proc would only flip it up so I don't wana do that, as for flip_at that's okay it's just set_is_up is more clear as to what it's doing

TorinoFermic (approved):

code/modules/clothing/head/armor.dm
TorinoFermic: Okay, you can rename the u to var/isUp for readability, that's a /proc/, so the decl should reflect that for readability

Garen7: Sure thing!

genessee596 (approved):

PR #5658, opened 2020-11-07
Control spider rebalance

TheShown911:
## About The Pull Request

Changed the control spider duration from 15-20 to 50-60 seconds and made the spider core die with the host body if the victim dies during control spider mind control.

## Why It's Good For The Game

This is a better way to balance it so carrions would still need to think hard to avoid death while having a lot of time during mind control for various plans instead of whiping out a gun and shooting the victim in the head.

## Changelog
:cl:
balance: Control spider duration incersed to 50-60 seconds, but if the controled victim dies under mind control the spider core dies with it.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/implant/implants/carrion/control_spider.dm
+++ b/code/game/objects/items/weapons/implant/implants/carrion/control_spider.dm
@@ -48,7 +48,7 @@
 	active = TRUE
 	last_use = world.time
 
-	addtimer(CALLBACK(src, .proc/return_mind), rand(15 SECONDS, 20 SECONDS))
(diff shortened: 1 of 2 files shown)


genessee596 (approved):
yes good

TorinoFermic (approved):

Hopekz:
I haven't heard any negative feedback about just the duration decrease. 
There might be a new meta about using the control spiders to go into maint and put someone into crit. You can guarantee someone's death outside of the spider's timer as well with all that preparation time.

A minute of control is a very long time.

TorinoFermic (approved):
I think that's better to merge it in order to gain some feedback

PR #5672, opened 2020-11-13
Bones now are reasonably sized when on the ground but still display full detail when interacted with.

Hopekz:
Okay, Halloween is over. Time to fix these big bones. 
These big bones look pretty good though so I can only think of one way to preserve this full detail.

Using this simple BYOND trick bones are now reasonably sized when on the ground but display full 32x32 detail when interacted with.

aOH42NhD3e

## Changelog
:cl: Hopek
add: Bones now are reasonably sized when on the ground but still display full detail when interacted with.
/:cl:

### DIFF
--- a/code/modules/organs/internal/bones.dm
+++ b/code/modules/organs/internal/bones.dm
@@ -8,6 +8,9 @@
 	var/broken_description = ""
 	var/reinforced = FALSE
 
+/obj/item/organ/internal/bone/Initialize()
+    . = ..()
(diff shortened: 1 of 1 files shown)


TorinoFermic (approved):
I like the pick up effect when taken

Hopekz:
SpooOOoky
"image

genessee596 (approved):
age of giants is over

genessee596:
has maintainer and delegate approval, very nice

PR #5568, opened 2020-10-08
Ports VG's HTML5 jukebox port from Virgo

drexample:
## About The Pull Request

Ports no-lag jukebox that uses browser (or VLC) as a streaming channel instead of loading both single-threaded byond client and server. Plus allows us to more easily add and manage music.

## Why It's Good For The Game

Music good = player happy

## Changelog
:cl:
add: Ports HTML5 jukebox
/:cl:

PR #5675, opened 2020-11-14
Revert "Ports VG's HTML5 jukebox port from Virgo"

Hopekz:
Reverts discordia-space/CEV-Eris#5568
Didn't want to perform this revert because the framework itself is good but there just isn't a playlist for the jukebox so people feel that reverting the whole framework is the best move vs having a jukebox that doesn't play music until someone adds new songs


Firefox13:
изображение
https://www.youtube.com/watch?v=hcTgBY8qURc

PR #5689, opened 2020-11-18
Makes the blood-red voidsuit much rarer

MadmanMartian:
## About The Pull Request 

People were saying that the blood-red voidsuit was much too common, this was due to two reasons

1. It had a rarity level 50 chance to roll
2. There were three of them, giving them a much larger chance of spawning

I have made it so the two subtypes can not spawn, and made its rarity level 90

## Why It's Good For The Game

Rare armor is now actually rare

## Changelog
:cl:
balance: The bloodred voidsuit is now much rarer from maint loot.
/:cl:

### DIFF
--- a/code/modules/clothing/spacesuits/void/merc.dm
+++ b/code/modules/clothing/spacesuits/void/merc.dm
@@ -85,11 +85,13 @@
 	resilience = 0.08
 	species_restricted = list("Human")
 	helmet = /obj/item/clothing/head/space/void/merc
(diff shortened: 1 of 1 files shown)


TorinoFermic:
According to Evan, the amount of subtypes will not influence the calculation of rarity but you can raise the rarity to this voidsuit.

genessee596 (approved):
you beautiful bastard

Evankhell561 (approved):

PR #5690, opened 2020-11-18
Update ss13_genchangelog.py to stop from having failure with yaml loader

TorinoFermic:
## About The Pull Request

Stops annoying Github actions failure spamming


drexample:
based

PR #5691, opened 2020-11-18
Adds excelsior hud to their voidsuit

Garen7:
## About The Pull Request

Adds an excelsior hud to the excelsior voidsuit so that if excelsior want to go loud they will be able to tell who's also excelsior.

## Why It's Good For The Game

There is currently a problem where excelsior will friendly fire by accident since they cannot tell each other apart without standing around before combat to call role call over their comms.

I'm putting it on the voidsuit so that it is immediately obvious that the excelsior are attacking if they want to use the hud, this way knowledge of whos on what team is less one sided in a fight. Note that figuring out who is excelsior is harder than say revs from tg since you cant hear when excelsior are talking to eachother and we don't have mindshield implants.

The last two tweaks are to prevent cheap valids.

Hud ingame:
image
My codersprite:
image

## Changelog
:cl:
add: Adds an excelsior hud to the excelsior voidsuit helmet which only works for excelsior
/:cl:


Hopekz (changes requested):
Prevent non-excelsior from seeing this information.
I don't like the idea of the crew having a full instant giveaway to all commies on the ship just because they manage to capture 1 item from them.

Garen7:
> Prevent non-excelsior from seeing this information.
> I don't like the idea of the crew having a full instant giveaway to all commies on the ship just because they manage to capture 1 item from them.

I already did, thats what the last two changes were for
>tweak: The excelsior voidsuit will only deploy its helmet if you are excelsior

>tweak: The excelsior voidsuit will no longer automatically deploy its helmet when equipped(you can still do it manually)

Hopekz (approved):
> I already did

nhp8v4DVda_nice-export

code/modules/clothing/spacesuits/void/excelsior.dm
```diff
@@ -50,3 +91,10 @@
 	)
 	helmet = /obj/item/clothing/head/space/void/excelsior
 	spawn_blacklisted = TRUE
+	auto_deploy_helmet = FALSE
+
+/obj/item/clothing/suit/space/void/excelsior/toggle_helmet()
+	if(is_excelsior(usr))
+		..()
+	else
+		to_chat(usr, SPAN_WARNING("The helmet fails to deploy, a built in speaker says, \"Failed to locate implant, please contact your nearest Excelsior representative immediately for assistance\"."))
```
Hopekz, line 100:
```suggestion
		to_chat(usr, SPAN_WARNING("The helmet fails to deploy, a built-in speaker says, \"Failed to locate implant, please contact your nearest Excelsior representative immediately for assistance\"."))
```

code/modules/clothing/spacesuits/void/excelsior.dm
```diff
@@ -50,3 +91,10 @@
 	)
 	helmet = /obj/item/clothing/head/space/void/excelsior
 	spawn_blacklisted = TRUE
+	auto_deploy_helmet = FALSE
+
+/obj/item/clothing/suit/space/void/excelsior/toggle_helmet()
+	if(is_excelsior(usr))
+		..()
+	else
+		to_chat(usr, SPAN_WARNING("The helmet fails to deploy, a built-in speaker says, \"Failed to locate implant, please contact your nearest Excelsior representative immediately for assistance\"."))
```
SpadesSlick, line 100: Instead of blocking it here, you should rewrite code to don't show this HUD to people without implants. This change limits use if this suit as trophy, and other infiltration shenanigans, it should be done properly instead.

SpadesSlick (changes requested):

code/modules/clothing/spacesuits/void/excelsior.dm
Garen7: Alright when I get off work I'll do the requested changes

Garen7: done

genessee596 (approved):

Garen7:
> make sure to blacklist the specific Hud item from spawning in maint

Good point! also fuck maint spawn

mattroks101 (approved):
Code looks good.

PR #5705, opened 2020-11-21
Adds new Randomizer toolmod

genessee596:
## About The Pull Request

This PR adds a new toolmod called the 'BSL "Randomizer" tool polish', which has random stats on creation - this can result in unexpected and powerful results when applied to a tool.

This toolmod is rare, and can only be found in maintenance. It cannot be crafted.

image
_Above: Randomizer's sprite_
Credit goes to INFRARED_BARON#8170 for sprite.

image
_Above: Maintainer permission for this idea_

image
_Above: Baron's permission to use his sprite_

## Why It's Good For The Game

As someone said to me, we need more 'soulful' toolmods. While I didn't quite get what he was getting at, I think what he meant is that we needed more unique tool modifications.

A toolmod that has random stats is quite interesting, I feel. 

Also, we actually get to use a sprite that would have rot in the depths of asset storage otherwise. So that's nice.

## Changelog
:cl: INFRARED_BARON and genessee
add: New toolmod that has random stats when applied - can be found in maintenance
/:cl:


TheShown911:
Travis fails

code/game/objects/items/weapons/tools/mods/mod_types.dm
```diff
@@ -632,3 +632,26 @@
...
 	I.prefix = "hydraulic"
+	
+// Randomizes a bunch of weapon stats on application - can be re-rolled by re-applying it, but runs the risk of destroying the toolmod
+/obj/item/weapon/tool_upgrade/augment/randomizer
+	name = "BSL \"Randomizer\" tool polish"
+	desc = "This unidentified tar-like liquid warps and bends reality around it. Applying it to a tool may have unexpected results."
+	icon_state = "randomizer"
+	matter = list(MATERIAL_PLASMA = 4, MATERIAL_URANIUM = 4)
+	rarity_value = 80
+	spawn_tags = SPAWN_TAG_RARE_TOOL_UPGRADE
```
DarkLevelSS13, line 643: Travis complain:
SPAWN_TAG_RARE_TOOL_UPGRADE: undefined var

DarkLevelSS13 (changes requested):

code/game/objects/items/weapons/tools/mods/mod_types.dm
genessee596: Nice catch. Should be fixed.

genessee596:
Should be fixed

DarkLevelSS13 (approved):
Seems legit. Will work smoothly with over current build view.

code/game/objects/items/weapons/tools/mods/mod_types.dm
```diff
@@ -632,3 +632,26 @@
 	UPGRADE_ITEMFLAGPLUS = LOUD
 	)
 	I.prefix = "hydraulic"
+	
+// Randomizes a bunch of weapon stats on application - can be re-rolled by re-applying it, but runs the risk of destroying the toolmod
```
MadmanMartian, line 636: But it won't be re-rolled by re-applying it. The stats are created on item creation, and are never re-rolled.

TheShown911:
I'd rather have stats apply when the mod is attached instead of when it's created so players don't know the stats when applying the mod.

code/game/objects/items/weapons/tools/mods/mod_types.dm
genessee596: fixed

genessee596:
> I'd rather have stats apply when the mod is attached instead of when it's created so players don't know the stats when applying the mod.

No.

The system in place currently prevents people from just putting on the item and taking it off until the mod is destroyed or they get an ultra-powerful roll - it instead encourages people to keep exploring until they find what a mod with good stats, or forces them to make a tradeoff with losing some stats in exchange for bonuses in others.

mattroks101 (approved):

PR #5721, opened 2020-11-25
Loot rework improvements.

Evankhell561:
## About The Pull Request
tags now mostly work with string definitions instead of strings, which greatly reduces the chance of getting wrong when adding new tags. This new format also allows you to include modular tags instead of just a long tag.

Through this change, many labeling errors have been corrected that prevented certain items from spawning because their label was incorrectly placed.

fix some troublesome tags., including that surgical tools are not labeled as tools but as surgical tools, and handmade tools are labeled as handmade tools and not as tools ( these specific changes will not be seen by users.) Also, the "rig_module" tag is corrected, which until now only spawned rare rig modules.

## Why It's Good For The Game
it generally makes the loot rework work as it is supposed to.

## Changelog
:cl:
fix: common rig modules can now be spawned in maint.
fix: Various items that should be spawned in maint are now actually spawned in maint.
/:cl:


genessee596 (approved):

TheShown911 (approved):

TorinoFermic (approved):

TorinoFermic:
I can tell this PR adopts a part of my idea with gun magazines system that we implemented in Eris with Shazbot

Evankhell561:
> I can tell this PR adopts a part of my idea with gun magazines system that we implemented in Eris with Shazbot

the idea for this came from someone's suggestion to make some items have a modular blacklist built into their tags. Something like:
"obj = 0, item = 0, gun = 1". this pr doesn't follow that suggestion but one thing led to another.

TorinoFermic:
> > I can tell this PR adopts a part of my idea with gun magazines system that we implemented in Eris with Shazbot
> 
> the idea for this came from someone's suggestion to make some items have a modular blacklist built into their tags. Something like:
> "obj = 0, item = 0, gun = 1". this pr doesn't follow that suggestion but one thing led to another.

Alright, this PR will be merged for the last public server day, so everything can be in place.

Evankhell561:
> I can tell this PR adopts a part of my idea with gun magazines system that we implemented in Eris with Shazbot

and removing the tools tags from makefists, and surgery tools was an idea I took from a pr made by kurgis, the hidden maintainer.

PR #5741, opened 2020-11-30
ethanol (parent) no longer increases sanity.

Evankhell561:
## About The Pull Request
Only the ethanol subtypes are supposed to increase sanity, not the ethanol itself.

## Changelog
:cl:
fix: ethanol (parent) no longer increases sanity.
/:cl:


TorinoFermic (approved):

genessee596 (changes requested):
Since you can still get drunk off ethanol, I'd prefer if ethanol just give a SEVERELY reduced amount of sanity gain in comparison to normal drinks - due to how awful it tastes - instead of no sanity at all.

genessee596 (approved):
Very nice

PR #5758, opened 2020-12-04
Excelsior Machine Gun

Harrivchaw:
## About The Pull Request
excel machine gun keeps popping up on gun spawner, it's functionally the same as a PK more or less save for one bit, it can't be fired while **wielded.**
This fixes that, also makes it that you have to wield the damn thing to fire, maybe PK should need two hand to fire as well?

## Why It's Good For The Game
unbroke the broken machine gun

## Changelog
:cl:
fix: Fixes Excelsior Machine Gun
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/projectile/automatic/maxim.dm
+++ b/code/modules/projectiles/guns/projectile/automatic/maxim.dm
@@ -1,6 +1,6 @@
 /obj/item/weapon/gun/projectile/automatic/maxim //This is currently deprecated unless someone can code up heavy machine gun mechanics.
 	name = "Excelsior machine gun"
-	desc = ""
+	desc = "Whatever happens, we have the Maxim gun, and they have not."
 	icon = 'icons/obj/guns/projectile/maxim.dmi'
 	icon_state = "maxim"
 	item_state = "maxim"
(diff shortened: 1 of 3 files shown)


TheShown911:
What if we just remove it, it's ancient, can't be printed by excels, doesn't even have full auto and isn't balanced around anything.

Harrivchaw:
it does have full auto, though i'm not entirely averse to just blacklisting the gun from spawners

genessee596 (changes requested):
this probably shouldn't exist in the spawn tables at all

please remove it from said table

Harrivchaw:
there, the excel machine gun shouldn't spawn anymore, its magazine is also blacklisted

MadmanMartian:
Did you manually add it to the blacklist?

Harrivchaw:
yea i changed it directly on the spawn data

MadmanMartian:
eeeeh, that'll get removed as soon as somebody calls the auto update on the table.

Set its spawn_blacklisted to true

Kurgis:
This was an admin weapon I edited because I wanted to get huge fire-rate while hosting locally. Blacklist can also be bypassed, so you might just want to assign it as a `bad_type`.

genessee596 (approved):

TheShown911 (approved):

TorinoFermic (approved):

PR #5784, opened 2020-12-16
IH hammer

DarkLevelSS13:
## About The Pull Request
IRONHAMMER IS REAL!
All IH (except medical specialist, and investigator), have a new hammer in they lockers.
This hammer allows people with robust stat higher than 40, to wield the hammer and smash the wall back to pieces!
To break simple you need to deal 3 hits.
But if you want to break a reinforced wall, you need 50 hits to do this.
Also, each time the wall hits, the screechy noises sound.
image

## Changelog
:cl: DarkLevel - code; CeUvi - sprites.
add: Tacticool Hammer "Tactical Hammer" Sledgehammer
/:cl:


Sparrowbird611:
It seems cool, I just got a few questons though:

Why doesn't the homewrecker do this?

Why did we give IH a sledgehammer for breaching instead of something like C4?

Shouldn't this be in the armory or something, and not inside of  everyone's locker? Doesn't seem like something you'd keep in a locker to me, and it's more specialized than the other things in most standard issue lockers.

Is it intended for all of IH (except for the commander) to be unable to use this until after they've gotten some decent stat gains, since they all spawn with 25 ROB maximum excluding commander?

DarkLevelSS13:
> Why doesn't the homewrecker do this? They look like the same hammer to me.
 
Slight looks the same, but this different sprites.
> Why did we give IH a sledgehammer for breaching instead of something like C4?

IH - IronHammer. (Also reference for rainbow six siege)

> Shouldn't this be in the armory or something, and not inside of everyone's locker? Doesn't seem like something you'd keep in a locker to me, and it's more specialized than the other things in most standard issue lockers.

I think every IronHammer operative need they own hammer to use.

> Is it intended for all of IH (except for the commander) to be unable to use this until after they've gotten some decent stat gains, since they all spawn with 25 ROB maximum excluding commander?

Yes, because of power to break the reinforced walls with with 50 hits. Some people can just do breach in hallway to space by roundstart.

genessee596 (changes requested):
Few things;
1. Not every operative needs this. Put 3 in the armory somewhere, and maybe one in the IHC's/Gunny's locker.
2. Make it 8 hits to breach a simple wall, not 3. Each hit takes about one second - being able to plough through a normal wall in 3 seconds is insane. 8 seconds is slightly more reasonable.
3. It should require 20-25 ROB, not 40. Most operatives don't have nearly that much ROB, and it doesn't make sense for there to be a tool in the armory that not a single operative knows how to use.
4. It shouldn't be a flat 'you can't use this if you have less than 20 ROB' - instead, make people with under the required amount take far longer to take down walls. Double, maybe triple the time.

code/game/objects/items/weapons/tools/hammer.dm
```diff
@@ -152,3 +152,31 @@
 		playsound(src, 'sound/machines/hiss.ogg', 50, 0, 0)
 		user.throw_at(target, get_dist(target, user), 1, user)
 		T.stop()
+
+/obj/item/weapon/tool/hammer/ironhammer
+	name = "Tacticool Hammer  \"Tactical Hammer\"  Sledgehammer"
+	desc = "A special heavy hammer issued to IHS units for tactical breaches."
```
genessee596, line 158:
```suggestion
	name = "FS \"Skeleton Key\" Breaching Hammer"
	desc = "A modified sledgehammer produced by Frozen Star for Ironhammer forces. This tool can take down standard walls in about eight swings, and can take down reinforced walls in about fifty swings. However, this tool requires significant strength to use efficiently."
```

code/game/objects/items/weapons/tools/hammer.dm
```diff
@@ -152,3 +152,31 @@
...
+	icon_state = "ironhammer"
+	item_state = "ironhammer"
+	wielded_icon = "ironhammer_doble"
+	force_unwielded = WEAPON_FORCE_ROBUST
+	force_wielded = WEAPON_FORCE_BRUTAL
+	armor_penetration = ARMOR_PEN_EXTREME
+	structure_damage_factor = STRUCTURE_DAMAGE_DESTRUCTIVE
+	w_class = ITEM_SIZE_SMALL
```
genessee596, line 166: You're telling me you can fit this massive sledgehammer into your pocket?
```suggestion
	w_class = ITEM_SIZE_HUGE
```

DarkLevelSS13: xD
Forget to change.

code/game/objects/items/weapons/tools/hammer.dm
```diff
@@ -152,3 +152,31 @@
...
+	slot_flags = SLOT_BACK
+	origin_tech = list(TECH_ENGINEERING = 3, TECH_COMBAT = 1, TECH_MATERIAL = 1)
+	tool_qualities = list(QUALITY_HAMMERING = 40, QUALITY_PRYING = 20)
+	matter = list(MATERIAL_PLASTEEL = 4, MATERIAL_STEEL = 2, MATERIAL_PLASTIC = 2)
+
+/obj/item/weapon/tool/hammer/ironhammer/wield(mob/living/user)
+	..()
+	if(ishuman(user))
+		var/mob/living/carbon/human/H = user
+		var/rob_stat = H.stats.getStat(STAT_ROB)
```
Garen7, line 176: IH shouldnt be expected to do maint looting since they have a job that requires them to always be on standby, though their starting 25 ROB is too low since most IH get origins which increase their ROB by at least 5 so 30 is fair

code/game/objects/items.dm
```diff
@@ -81,6 +81,8 @@
 	var/list/item_upgrades = list()
 	var/max_upgrades = 3
 
+	var/wallbreaker = FALSE //TIME TO BREAK THE WALL!
```
Garen7, line 84: we dont need a variable to check if something is the wallbreaker on the base item class, either do what I suggested handle the code in the hammer or check if(istype(/obj/item/weapon/tool/hammer/ironhammer, I)) in the code for the walls

Garen7 (changes requested):

code/game/objects/items/weapons/tools/hammer.dm
Garen7: >It shouldn't be a flat 'you can't use this if you have less than 20 ROB' - instead, make people with under the required amount take far longer to take down walls. Double, maybe triple the time.

actually what genesse said is a lot better though the ROB requirement should be higher since origins exist imo

githubuser4141:
> Why doesn't the homewrecker do this? They look like the same hammer to me. Not to mention the rocket hammer, which if anything seems like an even better candidate for busting down walls.

Game balance aside, it'd make sense that a hammer made with maintenance steel welded together on the spot by a vaga wouldn't be nearly as sturdy as a factory-produced hammer using steels designed for breaking down walls instead of maintenance steel which was probably salvaged from a chair or a floor tile.

genessee596 (approved):
Might want to change PR descriptions with the changes you implemented at my request

Iglpigl (changes requested):
Has conflicts.

drexample:
@DarkLevelSS13 
We'd really like to merge the PR, please resolve the conflicts.

Firefox13:
@DarkLevelSS13
> We'd really like to merge the PR, please resolve the conflicts.

code/game/objects/items/weapons/tools/hammer.dm
```diff
@@ -152,3 +152,31 @@
...
+		if(rob_stat >= STAT_LEVEL_EXPERT)
+			wallbreaker = TRUE
```
Evankhell561, line 178:
```suggestion
	if(ishuman(user))
		var/mob/living/carbon/human/H = user
		if(H.stats.getStat(STAT_ROB) >= STAT_LEVEL_EXPERT)
			wallbreaker = TRUE
```

Evankhell561 (changes requested):

PR #5796, opened 2020-12-21
Crusader Armor Changes

DarkLevelSS13:
## About The Pull Request
image

## Changelog
:cl: DarkLevel
balance: Crusader armor now have bullet/energy protection 50 instead of 30
/:cl:


genessee596 (approved):
I requested these changes. Now that the crusader gear is EOTP-only, we should make it far more useful.

Sparrowbird611:
I think you should pick one stat to be weaker to personally. This is basically better heavy armor right now. Maybe 35 laser resistance instead? It has higher melee protect and on par bullet resistance, should lose out a bit on laser IMO.

TheShown911 (approved):
Armor looks too cool to have low armor values

TorinoFermic (approved):

PR #5446, opened 2020-09-13
Gun changes

Tea-Primoris:
## About The Pull Request
Recoil will decrease after second from last shot, so single fire is not pinpoint death laser.
Full auto and bursts have slightly less damage than single fire.
Maximum recoil offset is 90 degrees.
Some click handler optimization to reduce lags when firing continuously.
Projectile penetration is increased based on vig stat.
Maximum overcharge of energy weapons splitted in half.
All weapons got damage and recoil tweaks.

## Why It's Good For The Game
Reere requested changes

## Changelog
:cl:
:balance: Recoil will decrease after second from last shot, so single fire is not pinpoint death laser. Also maximum recoil offset is 90 degrees now.
:balance: Full auto and bursts have slightly less damage than single fire. All guns got damage and recoil changes. Projectiles got more penetration the more vig stat you have. Maximum overcharge cutted in half.
/:cl:

PR #5801, opened 2020-12-21
Damage reduction

Tea-Primoris:
## About The Pull Request
Reverting damage of all guns to pre #5446 state, except Gladstone.
<details>
<summary>Reason for Gladstone nerf</summary>

image
</details>
 

## Why It's Good For The Game
***balance***

## Changelog
:cl:
balance: Guns damage reverted.
balance: Gladstone now have reduced damage.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/projectile/automatic/ak47.dm
+++ b/code/modules/projectiles/guns/projectile/automatic/ak47.dm
@@ -24,7 +24,7 @@
 	cocked_sound 	= 'sound/weapons/guns/interact/ltrifle_cock.ogg'
 	recoil_buildup = 1.8
 	one_hand_penalty = 15 //automatic rifle level
-	damage_multiplier = 1.6
+	damage_multiplier = 1.2
 
 	init_firemodes = list(
 		FULL_AUTO_400,
@@ -65,7 +65,7 @@
 	matter = list(MATERIAL_PLASTEEL = 20, MATERIAL_WOOD = 10)
 	price_tag = 3000
 	origin_tech = list(TECH_COMBAT = 5, TECH_MATERIAL = 2)
-	damage_multiplier = 1.6
+	damage_multiplier = 1
 	init_firemodes = list(
 	SEMI_AUTO_NODELAY,
 	BURST_3_ROUND,
--- a/code/modules/projectiles/guns/projectile/automatic/ak47.dm
+++ b/code/modules/projectiles/guns/projectile/automatic/ak47.dm
@@ -24,7 +24,6 @@
 	cocked_sound 	= 'sound/weapons/guns/interact/ltrifle_cock.ogg'
 	recoil_buildup = 1.8
 	one_hand_penalty = 15 //automatic rifle level
-	damage_multiplier = 1.2
 
 	init_firemodes = list(
 		FULL_AUTO_400,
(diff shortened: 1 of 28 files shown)


code/modules/projectiles/guns/projectile/automatic/ak47.dm
```diff
@@ -24,7 +24,7 @@
 	cocked_sound 	= 'sound/weapons/guns/interact/ltrifle_cock.ogg'
 	recoil_buildup = 1.8
 	one_hand_penalty = 15 //automatic rifle level
-	damage_multiplier = 3
+	damage_multiplier = 1.6
```
genessee596, line 27:
```suggestion
	damage_multiplier = 1.2
```

genessee596 (approved):

code/modules/projectiles/guns/projectile/automatic/ak47.dm
Tea-Primoris: Changed multiplier in commit below. Also set fs ak multiplier to 1

Alliostra:
Just gonna throw my hat into the ring here, pls don't feel attacked

Ebows: Giving them 1.5 and 1.7 damage mods is incredible. they already stun you in 4 hits, self-recharge and are akimbo-able. making them deal even more damage is not a good idea

(Excel) AK, Vintorez, C-20r, Atreides, straylight, Z8 and especially Wintermute: If the whole point is to make auto/burst fire have lower damage, then why buff auto/burst guns? they deal more damage in auto/burst mode now then they did before the damage reduction for auto/burst.

Hristov: why should a gun deal 105 damage, be hitscan, ignore armor for the biggest part and instantly stun you at the same time? 

Lightfall and cog: why nerf the lightfall into being just a subpar spider rose that needs to be wielded and deals shit damage, while buffing the cheap, common cog to be more powerful?

C-20r vs Atreides: the only perks of the c20r are a scope and slightly better AP now, otherwise they are identical despite the c-20r being the more elite SMG

Drozd vs other SMGs: it deals less damage than a c20r or an atreides now and fires slower as well, the only perk being above average AP. It also has the same recoil as the other SMGs i mentioned, but is supposed to have less recoil control by design.

Zoric: a Serb weapon (supposed to be among the best in the game) is now overshadowed by trash tier guns like the clarissa in terms of damage and AP

Lamia: nerfed into the ground while Avarasala was not changed- why?

Regulator: less common than the kammerer, but deals less damage than most .35 peashooters now.

Alliostra:
also, you doubled the amount of damage you can inflict via charging an energy weapon, from +50% to +100% self explanatory why this should not exist

code/modules/projectiles/guns/projectile/sniper.dm
```diff
@@ -23,7 +23,7 @@
 	var/item_suffix = ""
 	zoom_factor = 2
 	twohanded = TRUE
-	damage_multiplier = 2
+	damage_multiplier = 1.5
```
MadmanMartian, line 26: Remove the damage multiplier entirely. 70 damage is already super high.

MadmanMartian (changes requested):
A lot of these you seemed to arbitrarily buff. I'd say revert their damage_multiplier values to pre-change unless you can justify why.

genessee596 (approved):

MadmanMartian (approved):

PR #5804, opened 2020-12-23
Yet another chameleon projector nerf

MadmanMartian:
## About The Pull Request

Removes the default saved item from the chameleon projector, as it broke the new pixel standard

## Why It's Good For The Game

haha, 2 pixel hitbox that can move at 3 tiles per second!

## Changelog
:cl:
balance: Something moving over a chameleon projection now disables the projection
balance: Removes the cigarette butt from being the default projection of a cham projector
/:cl:


TorinoFermic (approved):

TorinoFermic:
Don't forget about flashbangs, I remember tg flash bang grenades also affect it.

mattroks101:
Would it be possible to have something else be the default here? Those defaults let you tell at a glance what the code is supposed to do, and it's pretty useful for people who might be new looking at it. If that's not feasible that's fine though.

MadmanMartian:
I could leave a doc-comment on them sure

mattroks101 (approved):

mattroks101:
This is good enough for now.

PR #5807, opened 2020-12-23
Surplus crate nerf

MadmanMartian:
## About The Pull Request

Nerfs the guild's ability to gain top tier guns and top tier ammo and all the money they could ever need, from two crates.

## Why It's Good For The Game

Previously, surplus weapons and ammo crates would draw from all possible ammo types, and all possible weapon types

This would mean that ammo crates would be worth MUCH more than their cost, so could be sent straight back for 1000% profit in some cases.

image

The weapon crates would also spawn the best possible weapons incredibly often.

This would mean that, guild would not want to deal with the ship, as it had nothing to offer them that they couldn't already have.

## Changelog
:cl:
balance: The value of the contents of surplus weapon and ammo crates from guild, is now more or less equivalent to the cost of the crate.
/:cl:

### DIFF
--- a/code/modules/projectiles/ammunition/boxes.dm
+++ b/code/modules/projectiles/ammunition/boxes.dm
@@ -8,6 +8,7 @@
 	reload_delay = 30
 	ammo_mag = "box"
 	matter = list(MATERIAL_CARDBOARD = 1)
+	bad_type = /obj/item/ammo_magazine/ammobox
 
 /obj/item/ammo_magazine/ammobox/resolve_attackby(atom/A, mob/user)
 	if(isturf(A) && locate(/obj/item/ammo_casing) in A || istype(A, /obj/item/ammo_casing))
--- a/code/game/objects/random/ammo.dm
+++ b/code/game/objects/random/ammo.dm
@@ -3,6 +3,11 @@
 	icon_state = "ammo-green"
 	tags_to_spawn = list(SPAWN_AMMO_S)
 
+/obj/spawner/ammo/low_cost
+	name = "random low-cost ammunition"
+	icon_state = "ammo-green-low"
+	top_price = 250
+
 /obj/spawner/ammo/low_chance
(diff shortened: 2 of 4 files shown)


code/modules/projectiles/ammunition/boxes.dm
```diff
@@ -8,6 +8,7 @@
 	reload_delay = 30
 	ammo_mag = "box"
 	matter = list(MATERIAL_CARDBOARD = 1)
+	bad_type = /obj/item/ammo_magazine/ammobox
```
TheShown911, line 11: What's this change for?

TheShown911 (commented):
Would also reccomend ammo exports not counting bullet price in addition to ammo price (you can print mags for 5 steel that sell for 2k+), but that may be out of scope for this PR.

code/modules/projectiles/ammunition/boxes.dm
MadmanMartian: The change to cheaper ammoboxes would spawn an ammobox with an empty icon state

Voldirs:
Problem is ammo export value is just too high, guildies can just print ammo and sell it even after this fix.

Evankhell561:
> Problem is ammo export value is just too high, guildies can just print ammo and sell it even after this fix.

that is material for another pr, not for this one.

Alliostra:
I would recommend to just put a General value modifier on ammo, like its already in place for guns (value halved on Export). The Change to the content of surplus guns Seems a bit to harsh, I'd prefer just removing the PK, Bojevic etc from the normal guns and Shotguns that can spawn in the crate instead of just 4 cheap guns per crate

MadmanMartian:
If anything, the surplus guns will have more variety

TheShown911 (approved):

genessee596 (approved):
actually based

PR #5814, opened 2020-12-24
Excel mine nerf/rework

MadmanMartian:
## About The Pull Request

The excelsior mine is now generalized into a proximity mine.

If you shoot it, it will blow up.

If you blow it up, it will blow up.

If you stand on it and fail the dodge check, you and it will blow up.

## Why It's Good For The Game

When excelsior are using it to pave their entire base with impunity, something might be wrong with the balance.

A mine capable of instagibbing somebody who walks over it should have some risk.

## Changelog
:cl:
balance: The excelsior mine has been generalized. Its IFF has been removed, and it is much more volatile.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/mine.dm
+++ b/code/game/objects/items/weapons/mine.dm
@@ -1,5 +1,5 @@
 /obj/item/weapon/mine
-	name = "Proximity Mine"
+	name = "proximity mine"
 	desc = "An anti-personnel mine. Unleashes shrapnel on those unlucky enough to step on them."
 	icon = 'icons/obj/machines/excelsior/objects.dmi'
 	icon_state = "mine"
--- a/code/game/objects/items/weapons/mine.dm
+++ b/code/game/objects/items/weapons/mine.dm
@@ -1,6 +1,6 @@
 /obj/item/weapon/mine
-	name = "Excelsior Mine"
-	desc = "An anti-personnel mine. IFF technology grants safe passage to Excelsior agents, and a mercifully brief end to others, unless they have a Pulse tool nearby"
+	name = "Proximity Mine"
+	desc = "An anti-personnel mine. Unleashes shrapnel on those unlucky enough to step on them."
 	icon = 'icons/obj/machines/excelsior/objects.dmi'
 	icon_state = "mine"
 	w_class = ITEM_SIZE_BULKY
@@ -30,12 +30,17 @@
 /obj/item/weapon/mine/ignite_act()
 	explode()
 
+/obj/item/weapon/mine/ex_act()
+	explode()
+
+/obj/item/weapon/mine/bullet_act()
+	explode()
+
 /obj/item/weapon/mine/proc/explode()
 	var/turf/T = get_turf(src)
+	qdel(src)
 	explosion(T,explosion_d_size,explosion_h_size,explosion_l_size,explosion_f_size)
 	fragment_explosion(T, spread_radius, fragment_type, num_fragments, null, damage_step)
-	if(src)
-		qdel(src)
 
 /obj/item/weapon/mine/update_icon()
 	overlays.Cut()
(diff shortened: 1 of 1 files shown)


Sparrowbird611:
Now when can I craft them :)

MadmanMartian:
Sure, I'll add that now

Omni44:
Honestly, as much as I hate mine carpets we've had dubious pleasure of having to endure for the longest time now, I think this is nerfing the feature to the point of nearly neutering it. being generalised is one thing, I am not sure how to feel about it, but being able to detonate it by shooting is kinda wack. I think that just letting explosions detonate them prevents such obnoxious spam of mines (unless excels want to get wiped out by chain reaction). I think it would be better to instead let Ions randomly disable/detonate them on hit, it'd be very nice buff to halicon as specialist tool in armory.

Alliostra:
> Honestly, as much as I hate mine carpets we've had dubious pleasure of having to endure for the longest time now, I think this is nerfing the feature to the point of nearly neutering it. being generalised is one thing, I am not sure how to feel about it, but being able to detonate it by shooting is kinda wack. I think that just letting explosions detonate them prevents such obnoxious spam of mines (unless excels want to get wiped out by chain reaction). I think it would be better to instead let Ions randomly disable/detonate them on hit, it'd be very nice buff to halicon as specialist tool in armory.

I agree, being able to shoot them will just lead to Crew major when excels are newbs and dont know about this, or mines will see User exclusively as instakill-traps outside of the base. I'd Suggest keeping the IFF Feature and adding that explosions Set off mines. Shooting them and/or EMP'ing them could have a random effect - deactivation, Detonation regardless of being Excel or not, detonation only for excel, adding a delay before Detonation, or no effect at all. This could Level the field completely without just making successful base defense Based on (not) placing mines in the base.

TheShown911:
Way too big of a nerf in my opinion, chain explosions are enough to get rid of mine stacking, and the fact excels can't pass the make them very very weak.

code/game/objects/items/weapons/mine.dm
```diff
@@ -30,12 +30,17 @@
...
+	explode()
+
+/obj/item/weapon/mine/bullet_act()
+	explode()
+
 /obj/item/weapon/mine/proc/explode()
 	var/turf/T = get_turf(src)
+	qdel(src)
 	explosion(T,explosion_d_size,explosion_h_size,explosion_l_size,explosion_f_size)
 	fragment_explosion(T, spread_radius, fragment_type, num_fragments, null, damage_step)
```
Evankhell561, line 43:
```suggestion
	explosion(T,explosion_d_size,explosion_h_size,explosion_l_size,explosion_f_size)
	fragment_explosion(T, spread_radius, fragment_type, num_fragments, null, damage_step)
	qdel(src)
```

Evankhell561 (changes requested):

code/game/objects/items/weapons/mine.dm
MadmanMartian: No, bad. Tried it. Caused infinite explosions as the explosion would call ex_act on the mine.

genessee596 (approved):
Let's see how they go for now. We can buff them again later if they go from 'good' to 'useless'.

code/game/objects/items/weapons/mine.dm
```diff
@@ -1,6 +1,6 @@
 /obj/item/weapon/mine
-	name = "Excelsior Mine"
-	desc = "An anti-personnel mine. IFF technology grants safe passage to Excelsior agents, and a mercifully brief end to others, unless they have a Pulse tool nearby"
+	name = "Proximity Mine"
```
Firefox13, line 2:
```suggestion
	name = "proximity mine"
```

Firefox13 (changes requested):
Please change the autolathe design name for it with a very brief warning for it. `(DANGER!)`, for example

Honestly101:
wait has this PR been just sitting here all because that single change hasn't been made?

MadmanMartian:
Nah, there was going to be a crafting recipe too

PR #5816, opened 2020-12-25
borer devour ability (re-enabled)

Honestly101:
## About The Pull Request
Re-enables the borer's ability to devour the brain of the host to make it into a husk or brain slug zombie.

## Why It's Good For The Game
When the host dies while the borer is in control, the borer has no option to release itself from the host's brain and so i think re-enabling this would add an option for the borer to have a chance at getting revenge, rather than just manually going to the OOC tab and ghosting even though technically its not dead.

## Changelog
:cl:
add: Re-adds the borer's devour ability
/:cl:


Honestly101:
Uh yeah i see that my other changes also pushed it's self into it, ill see if i can fix this.

PR #5818, opened 2020-12-25
Borer rebalance, flavor text changed and devour brain re-enabled

Honestly101:
## About The Pull Request
Made the delay timer for infecting a NT take longer, increased the cooldown timer and effect duration for the paralyse ability, also changed its flavor text to lean more into a biological parasite that doesn't have pyschic abilities. Re-enabled the devour ability which will allow the borer to convert the dead host into a husk if the borer was controlling it when it died. Switched the docile reagent from sugar to mindbreaker.

## Why It's Good For The Game
Currently if the borer were to be controlling the host when it died, there would be no additional options in the ability verb window, but re-enabling the devour brain ability will give the option to permanently fuse with the host and come back as a mutant abomination which is the husk. other than manually going into the OOC window and ghosting even though the borer is technically not dead only the host is, perhaps in the future someone with better coding skills can fix this issue and re-disable it after it has been resolved. As a recommendation, maybe have it automatically release control when the host dies.

I buffed and nerfed the borer's paralysis ability so it lasts longer but has a longer cooldown to prevent spam, the time it takes to infect a person with a cruciform has been increased so it would be more difficult and gives poeple around the victim time to react if you do it in public.

I also changed the reagent required to make the borer docile, from sugar to mindbreaker as it makes more sense and actually affects the brain, which is where the borer is located, plus sugar is kinda too common which makes it easy to beat the borer.

The flavor text for the borer has been changed to lean more into a biological parasite that doesn't has magic pyschic abilitie, instead it uses a stinger to paralyse its victims.

Now the actual reason i wanted to make these changes is to make the borer more of a actual threat to the crew, instead of a brain slug that gives the host chems and reproduces time to time.

## Changelog
:cl:
tweak/balanced: re-enables the devour brain ability, buffed a few things and re-did a bit of the flavor text.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/borer/borer.dm
+++ b/code/modules/mob/living/simple_animal/borer/borer.dm
@@ -122,13 +122,13 @@
 		if(health < maxHealth)
 			adjustBruteLoss(-1)
 
-		if(host.reagents.has_reagent("sugar"))
+		if(host.reagents.has_reagent("mindbreaker"))
 			if(!docile)
-				to_chat(get_borer_control(), SPAN_DANGER("You feel the soporific flow of sugar in your host's blood, lulling you into docility."))
+				to_chat(get_borer_control(), SPAN_DANGER("An evasive mind altering drug has entered your host's blood, forcing you into docility."))
 				docile = TRUE
 		else
 			if(docile)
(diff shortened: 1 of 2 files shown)


Honestly101:
My reasoning for closing this is because someone in the shaft just revealed that they have a decent overhual for the borer, so i guess there is no point in this.

PR #5820, opened 2020-12-26
Harpoon nerf beta

MadmanMartian:
## About The Pull Request

Attempting to jump into a wall will now telefrag you.

## Why It's Good For The Game

I was informed that you can, indeed, just jump onto a wall then teleport to either of the rooms that wall is part of.

This should prevent some illicit access from being gained with it.

I elected to not have it telefrag others than the user, as you can just build a wall on the turf you're on and bring others in, telefragging them.

## Changelog
:cl:
balance: Using the Harpoon to teleport onto a wall now telefrags you.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/space_harpoon.dm
+++ b/code/game/objects/items/weapons/space_harpoon.dm
@@ -46,10 +46,17 @@
 		return ..()
 	else if(istype(A, /obj/structure/table/) && (get_dist(A, user) <= 1))
 		return ..()
+
+	var/turf/AtomTurf = get_turf(A)
+	var/turf/UserTurf = get_turf(user)
+	var/turf/turf_tele
 	if(mode)
-		var/turf/T = get_turf(A)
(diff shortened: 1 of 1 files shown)


genessee596 (changes requested):
Make the teleport just fail if the user attempts to teleport another person into a wall to prevent tag-team cap-spare rushes.

Additionally, make the harpoon not work if you can see the tile through mesons alone.

MadmanMartian:
The teleport already fails if you can only see the tile via mesons. That report was false.

genessee596:
Oh, then just go ahead and do the first suggestion, sorry bout that

drexample:
YES!

MadmanMartian:
gotta rework it a little more

TorinoFermic (approved):

genessee596 (approved):

mattroks101 (approved):

PR #5838, opened 2021-01-03
Leg efficiency speed nerf

TheShown911:
## About The Pull Request

Cuts the extra speed from leg efficiency in half.

## Why It's Good For The Game

When testing i did not realize most humans have, in fact, 2 legs so the base speed buff could be doubled. We may need even more nerfs in the future.

## Changelog
:cl:
balance: More leg efficiency makes you less fast, and less leg efficinecy makes you less slow.
/:cl:

### DIFF
--- a/code/modules/organs/external/_external.dm
+++ b/code/modules/organs/external/_external.dm
@@ -332,7 +332,7 @@
 	if(status & ORGAN_SPLINTED)
 		. += 0.5
 
-	. += (-(limb_efficiency / 100 - 1) * 3)	//0 at 100 efficiency, -1.5 at 150, +1.5 at 50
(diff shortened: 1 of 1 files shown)


TorinoFermic (approved):

MadmanMartian (changes requested):
Have it reduce equipment slowdown for now, rather than making everyone faster.

Harrivchaw:
> Have it reduce equipment slowdown for now, rather than making everyone faster.

please dont, halve the speed or quarter of the speed would be enough, making it so it reduces only slowdown from armor wouldn't be as fun

TheShown911:
I don't think reducing slowdown is very fun.

PR #5841, opened 2021-01-04
Casings are now worthless

Alliostra:
## About The Pull Request

ammo casings no longer export for 20 credits per casing. They now export for 0 credits per casing. that's 100% less profit per casing!

## Why It's Good For The Game
money printer go brrrrrrrrr 

## Changelog
:cl:
balance: ammo casings no longer have any export value.
/:cl:

### DIFF
--- a/code/modules/economy/price_list.dm
+++ b/code/modules/economy/price_list.dm
@@ -988,7 +988,7 @@
 	for(var/obj/item/ammo_casing/i in stored_ammo)
 		. += i.get_item_cost(export)
 
-/obj/item/ammo_casing/price_tag = 20
+/obj/item/ammo_casing/price_tag = 0
 
 /obj/item/ammo_casing/get_item_cost(export)
 	. = ..() * amount


MadmanMartian:
Well, that certainly works, but also kinda sucks

Alliostra:
got a better idea on how to fix this? magazines and ammo boxes themselves already have a price tag, so.

MadmanMartian:
It's weird that all bullets have the same cost.

Maybe reduce it to something low, like 0.5 credits per bullet.

Then, we can go through each round, weigh up how much it costs to make, and adapt its price accordingly.

MadmanMartian:
I'm going to try and tackle the issue in a different way myself, PR upcoming. We'll see which PR is the lesser of the two evils

TorinoFermic (approved):

TorinoFermic:
I'll wait and see

mattroks101:
This will do for now.

PR #5842, opened 2021-01-04
Breaks up autolathe_disks.dm

MadmanMartian:
## About The Pull Request

This PR was going to be something else, but I wasn't given the go-ahead, so thought I'd just PR the autolathe_disk break up

## Why It's Good For The Game

It's backend, no player will experience this.

## Changelog
:cl:
/:cl:


cev_eris.dme
```diff
@@ -993,6 +992,15 @@
 #include "code\game\objects\items\weapons\circuitboards\machinery\telecomms.dm"
 #include "code\game\objects\items\weapons\circuitboards\machinery\unary_atmos.dm"
 #include "code\game\objects\items\weapons\circuitboards\machinery\vending.dm"
+#include "code\game\objects\items\weapons\design disks\_disk.dm"
```
Kurgis, line 995: There's a space here, should be `design_disks`, good luck making that change though since Windows File Explorer won't recognize a change like this, unless you use LINUX.

mattroks101 (approved):

PR #5852, opened 2021-01-09
The Marshal can now view contracts through his uplink

Alliostra:
## About The Pull Request

This PR is untested, since I was unable to find any way of making myself a marshal on a local server.
The Marshal can now view contracts through his uplink as to give him some clues about what antags might be up to, or already have been- right now the marshal isn't much better at hunting antags than a regular validhunting vag, this should change that a bit.

## Changelog
:cl:
add: The Marshal can now view contracts through his uplink
/:cl:


basedcringe:
👍

mattroks101:
"add: The Marshal can now view contracts **(untested, might not actually work)**"

Bro don't put that in your changelog please.

genessee596 (changes requested):
make it so he can only see contracts that have already been completed, to prevent him from just making a gamer trap to catch antags with 0 effort

Honestly101:
You should play test it to make sure to works, that would be nice, once you have and confirmed that it works as intended please make a comment, also please add a reason why this would be good for the game. also if you can, try an do what gen suggested, for balance reasons.

Firefox13 (changes requested):
> make it so he can only see contracts that have already been completed, to prevent him from just making a gamer trap to catch antags with 0 effort

Alliostra:
I don't think that is necessary, and not just because im unable to Code it :P. There are many contracts, and it is not even given that antags will do any of them. Many contracts also have multiple ways of completion (derail, spy, steal) and the Marshal does not know which contract antags will do, and in what way. Sure, you can hide in the disposal Unit of an Area that is mentioned in A spy contract, or Play Bodyguard for an assasination target, but you'll likely waste your time with that.

Firefox13 (approved):

PR #5853, opened 2021-01-09
Clamp muscle speed

MadmanMartian:
## About The Pull Request

Previously, you could go up to 4 times speed if you filled your legs with hydraulics, and even moreso if you used nerves instead

## Why It's Good For The Game

No more super-speed moebius FBPs ruining the round for everyone.

## Changelog
:cl:
balance: There is now a hard limit on how much of a speed boost you can get from implanting muscles into your legs.
/:cl:

### DIFF
--- a/code/modules/organs/external/_external.dm
+++ b/code/modules/organs/external/_external.dm
@@ -251,9 +251,9 @@
 	limb_efficiency += owner.get_specific_organ_efficiency(OP_NERVE, organ_tag) + owner.get_specific_organ_efficiency(OP_MUSCLE, organ_tag)
 	if(BP_IS_ROBOTIC(src))
 		limb_efficiency = limb_efficiency / 2
-		return 
+		return
 	limb_efficiency = (limb_efficiency + owner.get_specific_organ_efficiency(OP_BLOOD_VESSEL, organ_tag)) / 3
-	
+
 /obj/item/organ/external/proc/update_bionics_hud()
 	switch(organ_tag)
 		if(BP_L_ARM)
@@ -311,6 +311,8 @@
 			to_chat(usr, SPAN_DANGER("There is \a [I] sticking out of it."))
 	return
 
+#define MAX_MUSCLE_SPEED -0.25
+
 /obj/item/organ/external/proc/get_tally()
 	if(is_broken() && !(status & ORGAN_SPLINTED))
 		. += 3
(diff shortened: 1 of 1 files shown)


TheShown911:
For the maximum speed of .25 humans would need 3 extra muscles in their legs, the same speed could be achieved with a single "mechanical muscle" organ module that costs 12 steel and takes up no space. I would say this nerf may be too strong with the 12 times less effectiveness limb efficiency has on speed.

MadmanMartian:
Well, bear in mind it's multiplied by 2, because 2 legs.

So, if I go to 50%, somebody can move at double speed, which can outpace bullets easily.

Voldirs:
Make it affect speed very very slightly, like 10-15% percent more than default speed , though, this method is just too easy to access, there is no reason not to use it, but there should be downsides.

TheShown911:
Base movement delay is 2.5 units with shoes, so this is at maximum a 20 percent boost. There alredy are many speed boosts with no downsides like the muscle organ modules, perks like nightcrawler and fast walker, soon to exist cruciform upgrades and hyperazine if you can store enough of it. But i do understand the concern of getting too many stacking speed buffs.

MadmanMartian:
Following up on some suggestions now;

I'll lift it to 50% per leg

And it'll scale off of muscles, but requires a nerve per muscle

basedcringe:
> Base movement delay is 2.5 units with shoes, so this is at maximum a 20 percent boost. There alredy are many speed boosts with no downsides like the muscle organ modules, perks like nightcrawler and fast walker, soon to exist cruciform upgrades and hyperazine if you can store enough of it. But i do understand the concern of getting too many stacking speed buffs.

The mechanical muscles only reduce slowdown they don’t speed you up

TheShown911:
> > Base movement delay is 2.5 units with shoes, so this is at maximum a 20 percent boost. There alredy are many speed boosts with no downsides like the muscle organ modules, perks like nightcrawler and fast walker, soon to exist cruciform upgrades and hyperazine if you can store enough of it. But i do understand the concern of getting too many stacking speed buffs.
> 
> The mechanical muscles only reduce slowdown they don’t speed you up from what I’ve seen at least

They do, it's a flat .25 on the tally.

TheShown911 (approved):

mattroks101 (approved):

code/modules/organs/external/_external.dm
```diff
@@ -332,7 +334,9 @@
 	if(status & ORGAN_SPLINTED)
 		. += 0.5
 
-	. += (-(limb_efficiency / 100 - 1) * 3)	//0 at 100 efficiency, -1.5 at 150, +1.5 at 50
+	var/muscle_eff = owner.get_specific_organ_efficiency(OP_MUSCLE, organ_tag)
+	muscle_eff = muscle_eff - (muscle_eff/(owner.get_specific_organ_efficiency(OP_NERVE, organ_tag)/100)) //Need more nerves to control those new muscles
+	. += max(-(muscle_eff/ 100 - 1)/4, MAX_MUSCLE_SPEED)
```
TheShown911, line 339: The `-1` in this calculation causes slowdown of .25 per leg for every person, due to the base from the previous line being 0 instead of the previous 1. Removing the `-1` should fix it.

PR #5857, opened 2021-01-12
trange to RANGE_TURFS

Evankhell561:
## About The Pull Request
these two functions are exactly the same, so one should go. I kept RANGE_TURFS as this is how it is defined in other codes and this facilitates understanding for those who have experience in other ss13 codes.


genessee596 (approved):

AlexMorgan3817:
Why RANGE_TURFS instead of trange?

Kurgis:
It is defined as RANGE_TURFS in other codes, and this facilitates understanding for those who have experience in other ss13 codes.

Firefox13 (approved):

PR #5865, opened 2021-01-15
The hacking tool's speed now scales off cog

Alliostra:
## About The Pull Request

the hacktool's hacking speed is now affected by the COG stat of it's user. default time is 30 seconds, every point of COG removes 0.5 seconds up to a maximum of 20 seconds. negative cog will affect it in the same way, but there is no maximum amount of extra time you can get from having terrible COG

## Why It's Good For The Game

Cognition OP, pls nerf

## Changelog
:cl:
tweak: The hacking tool now uses COG to determine the hacking speed
/:cl:

### DIFF
--- a/code/game/objects/items/devices/hacktool.dm
+++ b/code/game/objects/items/devices/hacktool.dm
@@ -58,8 +58,8 @@
 
 	to_chat(user, SPAN_NOTICE("You begin hacking \the [target]..."))
 	is_hacking = 1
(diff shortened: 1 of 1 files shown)


genessee596 (changes requested):
put a cap of negative cog of about 60 seconds or something

it's not fun to have to sit for four hours trying to hack something even if you have shit cog

genessee596 (approved):
gud

Firefox13 (approved):

PR #5869, opened 2021-01-16
readds menu based crafting for tiles

Alliostra:
## About The Pull Request

in the FBP PR, the code that allowed you to craft variations of floor tiles etc was removed, probably because it seemed like it was unused. But it isn't and without it you can't craft any variations of floor tiles anymore.

## Why It's Good For The Game

cyberpunk hideouts good

## Changelog
:cl:
fix: you can now craft all kinds of floor tiles again
/:cl:


MadmanMartian (changes requested):
image

Make it overwrite the nanoUI window, this will be better in the long run as you can then preview what you want

PR #5872, opened 2021-01-17
Few more itmes in carrion uplink

TheShown911:
## About The Pull Request

Adds the surgery kit, shipwide announcement and the chameleon kit to the carrrion uplink. 

## Why It's Good For The Game

Carrions get to use a few more fun and stealthy items if they don't want any more gene points, the surgery kit should also make organ extraction less painful.

## Changelog
:cl:
add: Added the surgery kit, the chameleon kit and the shipwide announcement to the carrion uplink.
/:cl:

### DIFF
--- a/code/datums/uplink/announcements.dm
+++ b/code/datums/uplink/announcements.dm
@@ -14,6 +14,7 @@
 	name = "Shipwide Announcement"
 	item_cost = 2
 	desc = "Broadcasts a message anonymously to the entire vessel. Triggers immediately after supplying additional data."
+	antag_roles |= ROLE_CARRION
 
 /datum/uplink_item/abstract/announcements/announce/get_goods(var/obj/item/device/uplink/U, var/loc, var/mob/user, var/list/args)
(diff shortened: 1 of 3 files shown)


code/datums/uplink/announcements.dm
```diff
@@ -14,6 +14,7 @@
 	name = "Shipwide Announcement"
 	item_cost = 2
 	desc = "Broadcasts a message anonymously to the entire vessel. Triggers immediately after supplying additional data."
+	antag_roles |= ROLE_CARRION
```
MadmanMartian, line 17: I really don't think this is how you do it

MadmanMartian (changes requested):
The "fake" shipwide announcements button is currently very broken, being able to trigger random events at will.

If you're going to add that to another antag, fix it so it's not an instant radstorm button.

TheShown911:
Isn't that the fake radiation announcement from lines 92/93?

code/datums/uplink/announcements.dm
TheShown911: Fixed.

TheShown911:
Update: Should also fix the bug of serb ship being unable to move after an antagonist uses the fake serb announcement.

MadmanMartian (approved):

PR #5874, opened 2021-01-19
Vending machines now fire coins, rather than their inventory

MadmanMartian:
## About The Pull Request

See title

## Why It's Good For The Game

The fire inventory hack option was reported to be too beneficial, so in true eris fashion, we fitted gunpowder to the change return slot.

## Changelog
:cl:
balance: Vending machines now fire coins at projectile speeds when hacked, rather than their inventory
/:cl:

### DIFF
--- a/code/game/machinery/vending.dm
+++ b/code/game/machinery/vending.dm
@@ -852,21 +852,18 @@
 
 //Somebody cut an important wire and now we're following a new definition of "pitch."
 /obj/machinery/vending/proc/throw_item()
-	var/obj/throw_item = null
 	var/mob/living/target = locate() in view(7,src)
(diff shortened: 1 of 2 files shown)


MadmanMartian:
Yes.

They won't do much damage, but they'll annoy the hell out of people.

basedcringe:
Walking down the hallway “oh that’s a lot of vending machines kinda weird but okay” 5 seconds later AHHHH

MadmanMartian:
It is a 5% chance every machine tick (2 seconds standard) to fire, so adapt accordingly.

genessee596 (approved):
good

mattroks101 (approved):

PR #5885, opened 2021-01-26
Lascannon nerf

Alliostra:
## About The Pull Request

The las cannon no longer has a charge mode, and it's power cost per shot has been doubled
also removes it's one hand penalty, you can only fire it when it's wielded, anyways.

## Why It's Good For The Game

moe mad

## Changelog
:cl:
tweak: the laser cannon is now less deadly - its charge mode was removed and energy per shot increased(100 instead of 50).
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy/laser.dm
+++ b/code/modules/projectiles/guns/energy/laser.dm
@@ -114,17 +114,15 @@
 	w_class = ITEM_SIZE_HUGE
 	slot_flags = SLOT_BELT|SLOT_BACK
 	projectile_type = /obj/item/projectile/beam/heavylaser
-	charge_cost = 50
+	charge_cost = 100
 	fire_delay = 20
 	zoom_factor = 0
 	damage_multiplier = 1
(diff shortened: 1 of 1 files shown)


Omni44:
Kind of bad nerf honestly. Like majority of eguns, it's not even worth bothering in the first place without proper gunmods, one of which was nerfed just a few days ago, and other is rare maint thing/randomly printable by NT. Especially considering other (vastly less resoruce and labour intensive to achieve and make good) absurdities of our gun balance (hello shotguns) and combat in general (knockdown).
t. not even meobius player.

Alliostra:
You underestimate the lascannon. With a regular old 800m cell It's still got 8 50 dmg 20 AP shots, much better against the ubiquitous BP armor than a non-bugged Shotguns slug. These need to be taken care of as well, but that's no reason to keep Instant headmelter beams in the game.

Omni44:
Absolutely everyone deliberately skips on energy protection in favour of bulletproof armors, and they get shredded to pieces by properly used eguns. There might be some pattern here.

basedcringe:
Why not nerf the valk then it is MUCH more effective than the laser cannon

PR #5889, opened 2021-01-27
COG and Perk update

Alliostra:
## About The Pull Request

If you have less than 10 COG, you can't see how strong the stat aspects of an oddity are
if you have less than 15 COG, you can't see what perk a give oddity contains

## Why It's Good For The Game

more use for cog 
## Changelog
:cl:
tweak: COG determines how well you can decipher the use of oddities
/:cl:


Evankhell561 (changes requested):
From this pr you can do 3 prs to generate 3 different balance discussions. I do not see why put everything here if not to make some fixes.

Evankhell561 (approved):
the code is in order, and it looks good. I'm just wondering if it's really necessary to create a new level of stats for this feature but it's not necessarily a bad thing either.

Alliostra:
Just thought this was good practice, I have no objection to just replacing the level with a pure number

Evankhell561:
> Just thought this was good practice, I have no objection to just replacing the level with a pure number

I did not mean this. I meant that maybe STAT_LEVEL_BASIC can be applied here too. But again, it's just a thought that maintainers should consider ... or not. For my part I approve your changes.

drexample:
I'm not too keen on just
Locking up already existing things that didn't really need any change behind a stat wall.
This doesn't really buff cog, it just nerfs dumb people.
Which don't really need to be nerfed.

Firefox13:
While done properly code-wise, this is an inappropriate way to make COG more useful, see drexample's comments above.

PR #5894, opened 2021-01-28
Big Cog Renaissance ( I am very proud of that name)

Omni44:
## About The Pull Request
Adds Cog design disk with enough license points to print 4 off one. Adds 2 burst firing mode to Cog, instead of charge mode. Changes material cost from 12 steel to 10 steel, 15 plastic and 5 glass. Adjusted rarity values of both gun itself and design disk to mirror these of Novakovic.

## Why It's Good For The Game

garro

To add to that: Cog has always been embarrassing gun to shoot by any means, and it has largely remained such: it still fires beams of 30 burn damage wih 0AP, it still is woefully uneconomical to shoot with charge cost of 100 per shot. But I hope it will now have at least any reason at all to be considered and become different from upcoming makeshift laser gun, which would otherwise be very same thing.

## Changelog
:cl:
add: 2 burst  firemode
del: Charge mode on cog
add: cog's design disk 
tweak: cog's matter list
add: burst = 1 to WEAPON_NORMAL firemode
/:cl:

### DIFF
--- a/code/__DEFINES/weapons.dm
+++ b/code/__DEFINES/weapons.dm
@@ -55,13 +55,13 @@
 #define SEMI_AUTO_NODELAY	list(mode_name = "semiauto",  mode_desc = "Fire as fast as you can pull the trigger", burst=1, fire_delay=0, move_delay=null, icon="semi")
 
 //Cog firemode
-#define BURST_2_ROUND		list(mode_name="2-beam bursts", mode_desc = "Short, controlled bursts", burst=2, fire_delay=null, move_delay=4, icon="burst", damage_mult_add = -0.1)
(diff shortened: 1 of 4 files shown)


garrulousTeamotei:
Possibly add in a 800 Asters Medium cell to the disk? Pretty sure that's the type of cell that's included in every other Energy Gun/Plasma Gun disk.

Alliostra:
I'd recommend a small amount of plasteel to be added to the recipe if you want it to be similar to the novak

Omni44:
> Possibly add in a 800 Asters Medium cell to the disk? Pretty sure that's the type of cell that's included in every other Energy Gun/Plasma Gun disks.

I was hesitant to put "our" generic cells on One Star tech disk, but sure, that's not a problem I guess.



> I'd recommend a small amount of plasteel to be added to the recipe if you want it to be similar to the novak

I dunno, I think it's already roughly on par with novak, which doesn't cost any plasteel at all. Why should it cost any then, and how much would you say?

Firefox13 (approved):

PR #5895, opened 2021-01-28
buckshot/slug bandaid bug fix

Harrivchaw:
## About The Pull Request
buckshot and slug now knockbacks two tiles instead of one

## Why It's Good For The Game
allows the projectile to process properly before hitting a mob and prevents 20+ hits for buckshot or double hit for slug in theory
in practice the result is still somewhat janky depending on the angle of the shooter and the victim, but overall hits are lowered, you'll still get punted across the hallway but you won't get hit that much compared to before

## Changelog

balance: Slug and Buckshot now has more knockback

/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/bullettypes.dm
+++ b/code/modules/projectiles/projectile/bullettypes.dm
@@ -185,7 +185,7 @@ There are important things regarding this file:
 	icon_state = "slug"
 	damage_types = list(BRUTE = 54)
 	armor_penetration = 15
-	knockback = 1
+	knockback = 2
 	step_delay = 1.1
 
 /obj/item/projectile/bullet/shotgun/beanbag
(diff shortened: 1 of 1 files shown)


MadmanMartian:
Reason buckshot knocks you back so much is that it calls the hit proc (pellet) times in a single connection. This just makes buckshot into an absolute yeet cannon

Harrivchaw:
i'm aware of that, regardless, the knockback of buckshot stays the same both after and before this change, before this change buckshot can knock you back 20+ tiles because of multiple hitting, now there is actually more chance for you to be knocked to the side instead of getting carried along to god knows where

mattroks101 (approved):
Well I'm willing to have this bandaid fix until someone makes something better.

TheShown911 (approved):
+1

PR #5906, opened 2021-02-05
slowdown for most full body vests

Alliostra:
## About The Pull Request

This PR adds some slowdown to most fullbody vests with the exceptions of the ablative vest and the handmade vest- the ablative vest is already barely used, and the handmade vest is shit enough anyways.

## Why It's Good For The Game

Now full body vests shouldn't just be a straight upgrade that makes normal vests 100% obsolete. Now, you must decide between all body parts being covered with slowdown (fullbody), all body parts being covered without slowdown but with lower protection (coats) and no slowdown and good protection but limited coverage(regular vest)

## Changelog
:cl:
balance: fullbody vests now slow you down
/:cl:

### DIFF
--- a/code/modules/clothing/suits/armor.dm
+++ b/code/modules/clothing/suits/armor.dm
@@ -40,6 +40,7 @@
 	desc = "A generic armor vest, but with shoulderpads and knee pads included to cover all parts of the body. Not designed for serious operations."
 	icon_state = "armor_fullbody"
 	blood_overlay_type = "armor"
+	slowdown = 0.3
 	rarity_value = 30 // little bit rarer than just vests
 	body_parts_covered = UPPER_TORSO|LOWER_TORSO|ARMS|LEGS // kneepads and shoulderpads, so it covers arms and legs
 	matter = list(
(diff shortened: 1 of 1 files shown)


mattroks101:
Please give this an actual changelog.

hyperioo (approved):
Code wise it is ok so I am approving. 
In terms of balance I am letting the final decision of merging to the higher-ups.

shazbot194:
I don't like this nerf as it somewhat defeats the whole reason these were made. Which was to address normal vests being very underused and crap. Especially when the slowdown is as bad as standard void suits. 

At the very least I would say lower the slowdown ammount. These are armored vest parts, not an entier bulky spacesuit.

Firefox13 (changes requested):
I support Shazbot's opinion - 0.3 slowdown is too much for a armor vest with leg/arm pads, that is bulky as a spacesuit, it should be 0.1 at most.
Plus, not all full-body vests have got slowdown - operator full armor as well as handmade vest are missing. Please double-check.

Alliostra:
Did that, with the exception of full body BP vests. normal BP vests already have 1.5 slowdown, so I gave them 1 more. Also, operator vest is a child of vest/full, and handmade vest is intended to have no slowdown since its crap anyways

Firefox13 (changes requested):
> normal BP vests already have 1.5 slowdown, so I gave them 1 more.

1.5 slowdown? Or 0.15 slowdown? Please return the value to what it was before

> and handmade vest is intended to have no slowdown since its crap anyways

Indeed it is crap and indeed it should have slowdown because it is fucking handmade crap.

Alliostra:
ignore that

Alliostra:
alright, did anything you requested now, it was indeed 0.15, lol

Firefox13 (approved):

PR #5910, opened 2021-02-07
Turns the AMR into the shitcoded snowflake gun it was always meant to be

Alliostra:
## About The Pull Request

this PR aims to make the AMR more interesting and balancing it around ranged combat, because currently the most effective way of using this gun is clicking someone at close range for an instant damage spike and stun and then finishing them with a grab or magdump.

- First of all, the rarity of the AMR has been increased because it was really fucking common after the last update. At least I hope that it is now rarer, but maybe it will also have the opposite effect.

- The stunning effects have been removed from the AMR ammo altogether- in practice you might still get knocked down for a moment due to high damage, though.

- The AMR now has night vision inbuilt into it's scope, to encourage using the scope more (I'm sorry for the shitcode). It can be upgraded to thermal vision by attaching thermal glasses to the gun. Unfortunately, there is a bug where you retain the NV/thermal vision for a moment after lowering the scope. Not a big deal for NV, but perhaps thermals are too OP.

- The AMR now deals only 90% of it's damage when you are not using the scope, but deals 110% of it's damage while using the scope, due to overpenetration or something.

- There are three new ammo types for the AMR, which can be purchased for 3 TC per 5 rounds from the uplink.

- Depleted uranium ammo: slightly lower damage, but even higher armor penetration than the standard ammo, this is your bullet of choice against bulletproof nerds and serbians. It also irradiates people for some added damage over time.

- EMP ammo: This bullet is significantly less damaging than the standard ammo, but it creates a small, strong EMP pulse upon hitting a target. Useful against augmentations and disabling APCs from afar.

- Breaching ammo: also useful for disabling APC's from afar ;) This ammunition is not made to be used against people, but against structures. It deals more damage to structures the farther the projectile flies, and causes a fragmentation explosion as well. It dissipates faster than other bullets as well, and should not fly much further than the view range of the scope. At maximum scope viewing range, it is able to destroy rwalls, but you rarely ever find this kind of range, so it should be balanced enough. 

When a human is hit, the fragmentation is weaker and delayed by a second. However, the direct damage is pitiful, and in total it should deal no more damage to an unarmored person than a normal bullet does (rare cases may exceed this number due to chance-based shrapnel hitting). It also deals some agony damage, and together with the shrapnel effect this could be used to keep people from running up to you somewhat.

- The TC cost of regular ammo is lowered as well, to 2 TC for 5 bullets and 9 TC for 30 bullets, because if the ammo is very expensive, people will hesitate to use it at range- if you only have 5 shots, you'll be drawn towards using the weapon as a close-range opener and then finishing the weakened opponent with another gun, as opposed to using a lot of shots to kill someone from afar.

also fixes a radiation damage related bug that kept radiation-inducing gunmods from applying their radiation.

Garen helped me with a lot of this stuff. Have you thanked your local Garen today?

## Why It's Good For The Game

It's not

## Changelog
:cl:
tweak: AMR can now use default, depleted uranium(irradiatino), EMP, breaching ammo and comes with in-built NVG scope that can be upgraded with thermal glasses to have thermal vision too
fix: radiation-inducing gunmods now properly induce their radiation.
/:cl:


code/game/turfs/simulated/walls.dm
```diff
@@ -240,7 +240,10 @@
 	proj_damage = round(Proj.get_structure_damage() / 3)//Yo may replace 3 to 5-6 to make walls fucking stronk as a Poland
 
 	//cap the amount of damage, so that things like emitters can't destroy walls in one hit.
-	var/damage = min(proj_damage, 100)
+	if(Proj.nocap_structures)
+		damage = proj_damage * 3
```
TheShown911, line 244: The `var/damage` got deleted from above, so i wonder what `damage` this is referring to?

code/modules/projectiles/guns/projectile/sniper.dm
```diff
@@ -137,3 +143,13 @@
...
+
+/obj/item/weapon/gun/projectile/heavysniper/zoom(tileoffset, viewsize)
+	..()
+	if(zoom)
+		damage_multiplier += extra_damage_mult_scoped
+	else
+		refresh_upgrades()
+
+
+
```
TheShown911, line 155: Something strange going on here with indentation, also missing newline.

TheShown911 (commented):
Good gun, would like special sprites for new shells.

code/game/turfs/simulated/walls.dm
Alliostra: It mistakenly referred to the amount of damage of the wall, the wall would first take extra damage and then reset to a semi-damaged state after every shot.

Alliostra:
I'll have to add the special shells in a separate PR because smartgit won't let me push my local branch anymore and i'm too stupid to figure out how to fix it, so I can only use the github inbuilt editor for now

machinedDesign:
"The AMR now deals only 90% of it's damage when you are not using the scope, but deals 110% of it's damage while using the scope, due to overpenetration or something." 
Perhaps take a page from Serious Sam's book and make it deal 25% damage unscoped and full damage scoped, as the smart gun refuses to fire at full strength when not properly braced and stabilized in order to prevent breaking the owner's wrist.

drexample:
Make it TF2 machina, so it doesn't shoot un-scoped at all, lmao

Firefox13 (approved):
based

PR #5913, opened 2021-02-08
NT scourge bug fix

Alliostra:
## About The Pull Request

fixes a bug where every time you extend the NT scourge, it would get extra damage and stop fitting in your belt slot, but refuse to reset these values upon unextending, resulting in a sword with infinite armor penetration after a while. and no way of fitting it in slots.

## Changelog
:cl:
fix: fixed some NT scourge related bugs
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/nt_melee.dm
+++ b/code/game/objects/items/weapons/nt_melee.dm
@@ -83,10 +83,10 @@
 
 /obj/item/weapon/tool/sword/nt/scourge/proc/unextend()
 	extended = FALSE
-	force = initial(force)
-	armor_penetration = initial(armor_penetration)
-	slot_flags = initial(slot_flags)
 	w_class = initial(w_class)
+	slot_flags = initial(slot_flags)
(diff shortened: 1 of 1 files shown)


TheShown911 (approved):

hyperioo:
Don't you also need to add `force = initial(force)` since when it is extended there is `force += (force_extended - initial(force))`?

hyperioo (approved):
Apparently force already properly reset itself due to some code magic, approving.

LethalGhost:
> 
> 
> Apparently force already properly reset itself due to some code magic, approving.

And what about w_classs?
It's better to make `unxtend` proc mirroring `extend` one. 

But it's nice to have that fix at least.

hyperioo:
> 
> 
> > Apparently force already properly reset itself due to some code magic, approving.
> 
> And what about w_classs?

w_class is set back to initial value with `w_class = initial(w_class)` just before the two added lines

LethalGhost (approved):

PR #5928, opened 2021-02-16
You can only use the nanoforge if you meet its requirements.

Evankhell561:
## Changelog
:cl:
fix: Only technomancer or players with a high MEC stat can use the nanoforge.
/:cl:

### DIFF
--- a/code/game/machinery/autolathe/nanoforge.dm
+++ b/code/game/machinery/autolathe/nanoforge.dm
@@ -96,5 +96,5 @@
 /obj/machinery/autolathe/nanoforge/check_user(mob/user)
 	if(user.stats.getPerk(PERK_TECHNOMANCER) || user.stat_check(STAT_MEC, STAT_LEVEL_EXPERT))
 		return TRUE
-	to_chat(user, SPAN_NOTICE("You don't know how to make this work[src]"))
+	to_chat(user, SPAN_NOTICE("You don't know how to make [src] work."))
 	return FALSE
--- a/code/game/machinery/autolathe/nanoforge.dm
+++ b/code/game/machinery/autolathe/nanoforge.dm
@@ -94,7 +94,7 @@
 		. = TRUE
 
 /obj/machinery/autolathe/nanoforge/check_user(mob/user)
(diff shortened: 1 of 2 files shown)


code/game/machinery/autolathe/nanoforge.dm
```diff
@@ -96,5 +96,5 @@
 /obj/machinery/autolathe/nanoforge/check_user(mob/user)
 	if(user.stats.getPerk(PERK_TECHNOMANCER) || user.stat_check(STAT_MEC, STAT_LEVEL_EXPERT))
```
hyperioo, line 97: You could add a check for user.stats in case a mob without stats tries to use the nanoforge.

`if(user.stats && (user.stats.getPerk(PERK_TECHNOMANCER) || user.stat_check(STAT_MEC, STAT_LEVEL_EXPERT)))`

Evankhell561: > You could add a check for user.stats in case a mob without stats tries to use the nanoforge.
> 
> `if(user.stats && (user.stats.getPerk(PERK_TECHNOMANCER) || user.stat_check(STAT_MEC, STAT_LEVEL_EXPERT)))`

All mobs should have stats, except when the mob is destroyed. A runtime given because "stats" is null should give us an indication that there is something wrong with the code.

mattroks101: > You could add a check for user.stats in case a mob without stats tries to use the nanoforge.
> 
> `if(user.stats && (user.stats.getPerk(PERK_TECHNOMANCER) || user.stat_check(STAT_MEC, STAT_LEVEL_EXPERT)))`

You'd probably wanna use `if(user?.stats.getPerk(PERK_TECHNOMANCER) || user.stat_check(STAT_MEC, STAT_LEVEL_EXPERT)))` instead right? Last I checked that worked fine.

Evankhell561, line 97:
```suggestion
	if(user.stats?.getPerk(PERK_TECHNOMANCER) || user.stat_check(STAT_MEC, STAT_LEVEL_EXPERT))
```

Firefox13 (approved):

PR #5962, opened 2021-03-06
reincarnation and Commitment ritual now be performed by any disciple of nt.

Evankhell561:
## About The Pull Request

this pr is focused on making it possible for any disciple to resurrect other NT members.

## Why It's Good For The Game
because design doc

## Changelog
:cl:
tweak: reincarnation and Commitment can now be performed by any disciple of nt.
tweak: cloning now produces cloning damage depending on who has activated the cloning machine.
/:cl:

### DIFF
--- a/code/modules/core_implant/cruciform/machinery/cloning.dm
+++ b/code/modules/core_implant/cruciform/machinery/cloning.dm
@@ -38,6 +38,7 @@
 
 	var/power_cost = 250
 
+	var/clone_damage = 0
 
 /obj/machinery/neotheology/cloner/New()
 	..()
@@ -158,7 +159,7 @@
 	return TRUE
 
 /obj/machinery/neotheology/cloner/proc/done()
-	occupant.setCloneLoss(0)
+	occupant.setCloneLoss(clone_damage)
 	occupant.setBrainLoss(0)
 	occupant.updatehealth()
 	stop()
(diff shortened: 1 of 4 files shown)


Kurgis:
Would it be possible to add some user input that when a NT disciple clones another person, that the new body received clone damage due to being too low rank? New/existing players also won't know about clone damage, they'll simply see the litany within their list, and assume they should immediately use it instead of getting the Preacher first.

Omni44:
I still think that just changing cloning's itself time absed on who casts it is vastly preferable alternative to cloning damage, which is hard to cure, hard to live with, and they will want to spend that time curing it. 
Also something something, human body purity, something, something.

Evankhell561:
> Would it be possible to add some user input that when a NT disciple clones another person, that the new body received clone damage due to being too low rank? New/existing players also won't know about clone damage, they'll simply see the litany within their list, and assume they should immediately use it instead of getting the Preacher first.

doing that is a little more complicated but it's the best thing to do. i just figured out how to code this.

Evankhell561:
> Would it be possible to add some user input that when a NT disciple clones another person, that the new body received clone damage due to being too low rank? New/existing players also won't know about clone damage, they'll simply see the litany within their list, and assume they should immediately use it instead of getting the Preacher first.

ready

drexample (approved):

PR #5965, opened 2021-03-08
Latin Fix

Honestly101:
## About The Pull Request

Latin is said to be broken, it used whole sentences which caused for a needless amount of text and made it sound like you were saying the same thing everytime, this fixes that by only using single words with the occasional two words because latin is like that.

## Why It's Good For The Game

No more repeating the same thing over and over and more diverse sentances.

## Changelog
:cl:
fix: corrects the latin language.
/:cl:


Honestly101:
Discovered some issues, i will make another PR soon.

PR #5968, opened 2021-03-08
Finally, some good fucking food

maikilangiolo:
## About The Pull Request

Tired of seeing new and old club players desperate, I tried to help the situation the only way I know: mapping. Club now has a see through kitchen like the most classy of restaurants, a table to show their delicious plates without being stolen by people vaulting the counter, booze vendor ID locked, and some hydroponic trays.

Capture
Capture2

## Why It's Good For The Game

The bonsai, while a good idea in theory, is awful in practice. It only spawns an extremyl limited choice of (random) vegetables, three at a time (three plants for 20u alchool. any more booze, and its becomes massively inefficient: 30u only spanws 4 crops). It is slow, resource consuming (alchool crates cost more than half what club often has in their account), and you can count on one hand the dishes you can make.

So, at last, I decided to add a small hydroponics to club. The best part, it wont even replace the Bonsai, but it will complement it greatly. You can grow the missing crops in it, and even grow plants to make booze to feed the bonsai with, if you want fast vegetables that you are missing, and to avoid growing wheat and whatever else the tree spawns. It doesn't give any more tool to club that what they have or could get reasonably. There are no special seed vendors, seed extractors etc. 

If you ever played club AND tried to grow your own food, you will know how soul draining it is to walk every few minutes to public garden, which feels a kilometer away with all the doors. Aside from the obvious fact maintaining your garden healthy is exhausting and hard, the bar itself will be unmanned (which means customers either leave or rob you), but being a public garden, all your hard grown crops are prone to being eaten or ruined by the unscrupulous.

Let's all be real, there is no NT-Club interaction, in fact there is no club interaction. I don't mean memey bar RP, I mean that no one needs club or depends on it, so don't try and say "bUt aGrOlyTeS wIlL gRoW fOoD". No, they wont. With this actually not so simple fix, the amount of times club needs to travel to public garden is massively reduced, and all these newbie club players, which don't grow their crops and fall into the bonsai trap, will actually start their small gardens right next to their kitchen.

All they need is take a single trip to public garden to get the seeds, take the tools from the botany crate/locker (or print them from the public lathe above), and they are set. Hell, they can interact with NT by asking them to use their seed modification machines, like I have done in the past, if they so wish, and ask Moebius for a farm bot to automate the process. Plenty interaction to go around, without crippling club to rely on other departments (we all know how that ends).

## Changelog
:cl:
tweak: Club gets a major remap of the upper floor
tweak: booze-o-mat ID locked. Go steal from VIP bar
tweak: Made kitchen bigger by a tad
tweak: Changed ever so slightly maint directly south of the club to avoid double walls
add: Added a hydroponics room to club in which to make their own garden
del: Removed southermost set of stairs
del: my sanity
/:cl:

### DIFF
--- a/maps/CEVEris/_CEV_Eris.dmm
+++ b/maps/CEVEris/_CEV_Eris.dmm
@@ -55102,8 +55102,7 @@
 /turf/simulated/floor/tiled/techmaint,
 /area/eris/maintenance/oldtele)
 "cxi" = (
-/obj/machinery/vending/boozeomat,
-/obj/machinery/door/window/eastleft{
+/obj/machinery/vending/boozeomat{
 	req_access = list(28)
 	},
 /turf/simulated/floor/tiled/steel/bar_light,
@@ -56821,11 +56820,12 @@
 /turf/simulated/floor/plating/under,
 /area/eris/maintenance/section3deck4central)
 "cBa" = (
-/obj/structure/railing/grey{
-	dir = 1
+/obj/machinery/door/airlock{
+	name = "Kitchen";
+	req_access = list(28)
 	},
-/turf/simulated/floor/bluegrid,
-/area/eris/crew_quarters/bar)
+/turf/simulated/floor/tiled/cafe,
+/area/eris/crew_quarters/kitchen)
 "cBb" = (
 /obj/item/clothing/mask/breath,
 /obj/machinery/door/window/southleft{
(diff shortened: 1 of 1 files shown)


maikilangiolo:
Capture
Capture2

would you look at that, cant fucking upload the pics on the main PR. Jesus I hate my life

MadmanMartian:
Fixed your links

Omni44:
I appreciate the intention, but I don't honestly like it. It makes bar look ugly, hydroponics area seems hamfisted into whole thing, and doesn't solve the worst issue with bar's layout, that is person from main hallway can't see the counter or anything going on in the bar itself.

drexample:
Conflict

Firefox13:
Agree with commentators above, the hydroponics area looks ugly, lacks *any* equipment(while it's probably intended, it doesn't much solves the problem by them still needing taking a walk to actual garden and taking shit from there. There's also no vents and wall equipment as well). The Church personnel now has limited money which should encourage them actually selling shit, especially to the club. I'm closing this PR due to it redundancy.

PR #5970, opened 2021-03-12
help me

Evankhell561:
## About The Pull Request

the craft system is an extremely complicated system, and making a fix to one part of it can mean that other things stop working. Also, the number of combinations of things that make the craft system work is very large and difficult to visualize. 

Anyway, this code makes the new craft system work in a more general way so it should avoid weird situations. As far as I have been able to test everything works fine.
Screenshot_2

## Why It's Good For The Game

fixes the craft system.

## Changelog
:cl:
fix: some craft system recipes were not working.
/:cl:

### DIFF
--- a/code/datums/craft/recipe.dm
+++ b/code/datums/craft/recipe.dm
@@ -1,3 +1,6 @@
+#define IN_PROGRESS 2
+#define READY 1
+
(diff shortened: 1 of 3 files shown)


TheShown911:
No one can help you now

PR #5971, opened 2021-03-13
Yet another awful hivemind mobs update, including Tyrant

Wouju:
## About The Pull Request

Fixes Bomber explosion being 5x5 instead of 3x3
Hiborg and Mechiver have now reduced HP
Mechiver has a small Resistance, reducing damage taken
Tyrant can no longer teleport

## Why It's Good For The Game

5x5 explosion was ridiclously broken
Hiborg was a bit too tanky in personal
Mechiver similary so, however I also gave it a small amount of resistance so it can shrug off very basic attacks

Tyrant teleporting around is absolutely awful, given the fact that it can teleport to ANY core in game, not to mention often failing at that and teleporting into maint and other unintended areas. Also reduced HP by a bit, as to help the crew a bit.

## Changelog
:cl:
tweak: Hiborg and Mechiver have reduced HP. Mechiver now has some flat damage resistance.
balance: Removed Hivemind Tyrant's ability to teleport around.
fix: Fixed Hivemind Bomber's explosion being 5x5 instead of 3x3
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/megafauna/hivetechboss.dm
+++ b/code/modules/mob/living/simple_animal/hostile/megafauna/hivetechboss.dm
@@ -11,8 +11,8 @@
 	pixel_x = -16
 	ranged = TRUE
 
-	health = 2250
-	maxHealth = 2250 //Only way for it to show up right now is via adminbus OR Champion call (which gives it 150hp). For comparison Kaiser has 2000hp
+	health = 1650
+	maxHealth = 1650 //Only way for it to show up right now is via adminbus OR Champion call (which gives it 150hp). For comparison Kaiser has 2000hp
 	break_stuff_probability = 95
 
 	melee_damage_lower = 30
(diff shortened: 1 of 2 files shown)


mattroks101:
Why get rid of the healthmarker?

Wouju:
> Why get rid of the healthmarker?

From what I understand the healthmarker variable is used only to check how much HP the tyrant has and then trigger the teleport proc
outside of that it has no use

code/modules/mob/living/simple_animal/hostile/megafauna/hivetechboss.dm
```diff
@@ -11,8 +11,8 @@
 	pixel_x = -16
 	ranged = TRUE
 
-	health = 2250
-	maxHealth = 2250 //Only way for it to show up right now is via adminbus OR Champion call (which gives it 150hp). For comparison Kaiser has 2000hp
+	health = 1650
+	maxHealth = 1650 //Only way for it to show up right now is via adminbus OR Champion call (which gives it 150hp). For comparison Kaiser has 2000hp
```
TheShown911, line 15: 600 hp nerf is a lot, I wouldn't put it lower than the Kaiser.

Wouju: aight, will change it so it's the same as kaiser

Firefox13 (approved):

PR #5980, opened 2021-03-19
Areas without apc are no longer considered for the Disturbance objective.

Evankhell561:
## About The Pull Request

fix this:
image

## Changelog
:cl:
fix: Areas without apc are no longer considered for the Disturbance objective.
/:cl:

### DIFF
--- a/code/datums/objective/individual_objective/technomancers.dm
+++ b/code/datums/objective/individual_objective/technomancers.dm
@@ -8,7 +8,7 @@
 
 /datum/individual_objective/disturbance/assign()
 	..()
-	target_area = random_ship_area()
+	target_area = random_ship_area(FALSE, TRUE, FALSE, TRUE)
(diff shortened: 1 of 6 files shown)


code/datums/objective/individual_objective/technomancers.dm
```diff
@@ -8,7 +8,7 @@
 
 /datum/individual_objective/disturbance/assign()
 	..()
-	target_area = random_ship_area()
+	target_area = random_ship_area(FALSE, TRUE, FALSE, TRUE)
```
hyperioo, line 11:
```suggestion
	target_area = random_ship_area(need_apc=TRUE)
```
Since we are keeping the same values for the other default arguments there is no need to put them again.

hyperioo (dismissed):

TheShown911 (approved):

PR #5989, opened 2021-03-24
Fixes/tweaks Hivemind mobs speed

Wouju:
## About The Pull Request

I finally figured out how to properly set the speed of Hivemind mobs, no more fucky balance

Stinger is fast now, to the point it could possibly catch up

Bomber is slow, you can easily outrun it or kite it

Lobber has "normal" movement speed

Similary with Mechiver, however the mechiver has now way better armour and a bigger HP pool as to compensate for it's slowness

Hiborg and Himan are slower than average, with the Himan being faster than Hiborg

Marionette movement speed is unchanged

## Why It's Good For The Game

Before this PR all the hivemind mobs had the same speed
This was obviously awful and not what i wanted, given the fact that a suicide bomb and a armoured walking tank had the same speed.

This PR effectively fixes that by making the mobs slower (or faster for the Stinger).

Also the Mechiver being so slow deserves better Armour and HP.

Tl;Dr - Having the mobs be slower is, I believe, way better than before when the mobs speed was exactly the same.
## Changelog
:cl:
balance: Hivemind Mobs now have their movement speed fixed, in order from slowest to fastest: Mechiver < Bomber < Hiborg < Himan < Lobber < Stinger.  Also makes the Mechiver have higher HP and Armour as to compensate for the slowness.
/:cl:

### DIFF
--- a/code/modules/hivemind/mobs.dm
+++ b/code/modules/hivemind/mobs.dm
@@ -228,6 +228,7 @@
 //Special ability: none
 //Just another boring mob without any cool abilities
 //Low chance of malfunction
+//Faster than average, to the point it could possibly catch up to someone
 //Default speaking chance
 //Appears from dead small mobs or from hive spawner
 //////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


TheShown911 (approved):
What could go wrong?

PR #6001, opened 2021-04-01
Boards + RIG modules sprites

infraredbaron:
## About The Pull Request
Replaces the old boards with more eris-styled ones;

image

Replaces the used modules with never variants:

image

Got rid of old unused stuff while I was there.
PS:
Thanks onule for letting use his meson googles sprite;

## Changelog
:cl:
imageadd: Generic module for your needs
imagedel: breacher RIGs; ERT RIGs; Asset protection RIG; Burecracy RIG;
/:cl:


TheShown911 (commented):
I'm not sure but I think icon states are case sensitive.
case

infraredbaron:
> I'm not sure but I think icon states are case sensitive.

Addressed in latest commit - the icon state is indeed case sensitive.

TheShown911 (approved):

PR #6006, opened 2021-04-02
Adds a new contraband crate to guild

Medcar:
## About The Pull Request
Guild is pretty lacking in contraband content, and I want to learn to code, so I hit 2 birds with 1 stone and I added a new contraband crate, Probably needs a bit of balancing

## Why It's Good For The Game

New contraband content

## Changelog
:cl:
add: New contraband crate
/:cl:

### DIFF
--- a/code/modules/cargo/packs.dm
+++ b/code/modules/cargo/packs.dm
@@ -1540,10 +1540,10 @@ var/list/all_supply_groups = list("Operations","Security","Hospitality","Enginee
 	/datum/supply_pack/randomised/corporate
 	num_contained = 3
 	contains = list(/obj/item/clothing/suit/space/void/merc,
-					/obj/item/weapon/gun/energy/crossbow
-					/obj/item/weapon/gun/energy/sniperrifle
-					/obj/item/weapon/gun/energy/crossbow/largecrossbow
-					/obj/item/weapon/gun/projectile/mk58
+					/obj/item/weapon/gun/energy/crossbow,
+					/obj/item/weapon/gun/energy/sniperrifle,
+					/obj/item/weapon/gun/energy/crossbow/largecrossbow,
+					/obj/item/weapon/gun/projectile/mk58,
 					/obj/item/weapon/gun/projectile/mk58/wood)
 	name = "Corporate war crate"
 	cost = 8000
--- a/code/modules/cargo/packs.dm
+++ b/code/modules/cargo/packs.dm
@@ -1536,3 +1536,18 @@ var/list/all_supply_groups = list("Operations","Security","Hospitality","Enginee
 	crate_name = "Cigarettes Cartons Crate"
 	containertype = /obj/structure/closet/crate
 	group = "Supply"
+
+	/datum/supply_pack/randomised/corporate
+	num_contained = 3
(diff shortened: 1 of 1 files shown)


TheShown911:
Remove the AMR.

Medcar:
It's not AMR it's valkyrie

Medcar:
Valk got removed, so at the moment it's ready

code/modules/cargo/packs.dm
```diff
@@ -1536,3 +1536,17 @@ var/list/all_supply_groups = list("Operations","Security","Hospitality","Enginee
 	crate_name = "Cigarettes Cartons Crate"
 	containertype = /obj/structure/closet/crate
 	group = "Supply"
+
+/datum/supply_pack/randomised/corporate
+	num_contained = 3
+	contains = list(/obj/item/clothing/suit/space/void/merc,
```
infraredbaron, line 1542: One of the best voidsuits? And antag only crossbows?

infraredbaron (changes requested):
Just another buff to guild, that actually should be nerfed.
Plus "contraband" doesn't make sense, since there is nothing illegal that guild can buy or sell.

code/modules/cargo/packs.dm
Medcar: That's why it costs 8k and is hidden behind contraband

Medcar: > Just another buff to guild, that actually should be nerfed.
> Plus "contraband" doesn't make sense, since there is nothing illegal that guild can buy or sell.

You need to set the cargo desk board to broad with a multitool, which not a lot of people know you can do, plus it's one of the most expensive things guild can buy if not the most, remember that guild can also get the crimson RIG from surplus rig crates which is extremely OP, and that the contents are randomised, you could get those shitty pistols for 8k, and the crossbows deal only 25 and 30 toxin respectively

joookschad:
imo its good, guild can buy ballistic armor with NV for like 3k credits, can get the RIG blood red (although its luck based and perhaps cheaper with luck), letting them get a % of a good voidsuit isnt that bad considering the crate costs 8k (which will be hard to get when cigarette buisness gets deleted)

code/modules/cargo/packs.dm
TheShown911: The *chance* to roll a crimson RIG costs 5k on itself, and IMO a blood red is way better due to not taking up your back slot

Medcar: > The _chance_ to roll a crimson RIG costs 5k on itself, and IMO a blood red is way better due to not taking up your back slot

Here you have a chance too, you can either roll shitty pistols, or the blood red, and for 8k, I can increase the price even more if that's still not balanced

Medcar: And like jooks said below guild can buy the best ballistic armor with nightvisions which contains 2 sets for just 3k

FortuneMaster:
I agree with baron, on other server you kind of cant have any contraband since there are no restrictions to what you can have. It's better to show the shady nature of factions via other ways

drexample:
Hanging here since april, buffs guild that is the least lacking department on this ship that has barely any concept of contraband.

PR #6007, opened 2021-04-02
removes stun from AMR

infraredbaron:
## Why It's Good For The Game

Since you can very easy cheese players with it. Not anymore.
it's realistic, but not balance wise

## Changelog
:cl:
balance: AMR is less OP now
/:cl:


Sparrowbird611:
Should probably remove the weaken too.

Firefox13 (approved):

PR #6022, opened 2021-04-08
Fixes gulping down

StiffRobot:
## About The Pull Request

Whenever you moved while gulping down the contents of a bottle, it would just make you down it all instantly. Now it will instead splash all of its content on you, as intended (probably).

## Why It's Good For The Game
Squash bugs, lift code, get buff
image

## Changelog
:cl:
fix: Moving while gulping something down will splash everything on you now
/:cl:


hyperioo (approved):
Indeed moving while trying to gulp down a drink seems to be a good way to splash yourself.

PR #6055, opened 2021-04-12
Adds SA Krinkov and Changes to Vipr and AKMs

shazbot194:
## About The Pull Request

This PR changes both the FS and adds in an SA (Serb) AK. Both of these AKs act a bit differently while still being mostly the same.

**FS:** Name changed from Kalashnikov to Vipr, as a play on Viper and Vepr, because why not. The rifle itself has become a large hunting rifle. With this, it has received a minor damage increase (1.0 -> 1.1), a small reduction in recoil (1.8 -> 1.5), but has been made bigger, so it can no longer be carried in normal bags.

**SA:** Named Krinkov, this carbine has two distinct modes that it can switch back and forth with at will. Normally it is a carbine length AK which is very much like the Excel AKMS. The Krinkov is larger, and with less recoil (1.8 -> 1.5), but it's shorter barrel makes it do less damage (1.0 -> 0.9). This combines into making the Krinkov a bit akin to a large SMG. But you can fold the stock with a right click command which increases recoil (1.5 -> 1.8) and makes it the same size as the Excel AK. 

aks
ak_bag

This PR also has a minor tweak to the Excel AKMS sprite where the back sprite is now mirrored, so someone can fashionably wear two Excel AKs and have it show.

image

## Why It's Good For The Game

Overall, this helps to give each weapon manufacturer a unique feel even with the 'same' guns. It helps fill some niches that might otherwise be left out simply because it would seem redundant to add such a similar weapon. It also helps to tell players visually what each one of these guns does and overall makes the game a more interesting place visually. It also lays the foundation for future PRs where what has been done here can be done with other weapons. [It is also a part of a doc where this has been layed out.](https://docs.google.com/document/d/1Q7mTWSIgunXaPm4sm2mbPxiAKjmKZpMijVMdRmnCWe8/edit?usp=sharing)

## Changelog
:cl: Shazbot194
add: SA Carbine "Krinkov"
balance: Changed FA AK into the FA Vipr; a Hunting Rifle
tweak: Mirrored Excel AK back sprite so now you can seen wearing two AKMSs.
/:cl:


code/modules/projectiles/guns/projectile/automatic/ak47.dm
```diff
@@ -57,14 +62,17 @@
...
 			This is a copy of an ancient semi-automatic rifle chambered for .30 Rifle. If it won't fire, percussive maintenance should get it working again. \
 			It is known for its easy maintenance, and low price. This gun is not in active military service anymore, but has become ubiquitous among criminals and insurgents. \
-			This copy, in fact, is a reverse-engineered poor-quality copy of a more perfect copy of an ancient rifle, therefore it can fire only in bursts instead of auto-fire."
+			This copy has been made to look as least miltaristic as possible to be sold to as many civilian populations as possible."
 	icon_state = "AK_wood"
 	item_suffix = "_wood"
 	wielded_item_state = "_doble_wood"
+	w_class = ITEM_SIZE_HUGE
+	recoil_buildup = 1.5	//Heavy, semi-auto and burst only
+	damage_multiplier = 1.1	//Long barrels help you get more out of a round.
```
zoktiik, line 75: Shouldn't longer, heavier barrels reduce recoil?

shazbot194: Unless I am mistaken, isn't a lower recoil better? As in a lower number in this slot?

shazbot194: oh I see what you mean now. Possibly, but the longer barrel would also help give the bullet more velocity, which would by newton's third law impart more recoil into the gun, most likely resulting in no net gain or loss. But this can also be up to the balance people if it should be more or less

code/game/objects/items/weapons/design_disks/autolathe_disks.dm
```diff
@@ -210,6 +210,17 @@
 		/datum/design/autolathe/ammo/lrifle_pk,
 	)
 
+/obj/item/weapon/computer_hardware/hard_drive/portable/design/guns/sa_ak
+	disk_name = "Serbian Arms - .30 Krinkov Car"
+	icon_state = "serbian"
+	spawn_tags = SPAWN_TAG_DESIGN_ADVANCED
+	rarity_value = 50
+	license = 1
```
Firefox13, line 218:
```suggestion
	license = 12
```

Firefox13 (approved):

PR #6062, opened 2021-04-13
Makes it so the mindswapper uses a list

Cocacola4124:
## About The Pull Request
Makes it so the mindswapper uses a list of mobs that can be swapped, rather than every type of carbon mob, which let players swap into mobs that just don't work well with actual players occupying them(Mob A.I taking over). 

## Why It's Good For The Game
Adds a few new types of mobs that the mindswapper works on, players can now swap into robots for example. Something I felt was missing. I wanted to also allow this for AI, and PAI but it didn't seem to work well.

## Changelog
:cl:
tweak: Changes the mindswapper to use a list for swappable mobs.
/:cl:

### DIFF
--- a/code/game/machinery/mindswapper.dm
+++ b/code/game/machinery/mindswapper.dm
@@ -13,6 +13,13 @@
 	var/operating = FALSE  // Is it on?
 	var/swap_time = 200  // Time from starting until minds are swapped
 	var/swap_range = 1
+	var/swappable_mobs = list(
+	/mob/living/carbon/human,
+	/mob/living/silicon/robot,
+	/mob/living/simple_animal/borer,
+	/mob/living/simple_animal/mouse,
+	/mob/living/carbon/slime,
+	/mob/living/simple_animal/cow)
 
 	use_power = IDLE_POWER_USE
 	idle_power_usage = 2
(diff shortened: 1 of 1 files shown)


MadmanMartian:
No more roachswapping?

Cocacola4124:
> 
> 
> No more roachswapping?

Better to put stuff in separate pulls. There's a lot of small things that are annoying me that I have non-existent priority on fixing. Like that stupid cat meowing forever after dying.

Alliostra:
As far as I'm aware, mob AI taking over if someone is swapped into a mob was fixed in the most recent carrion update. But really, even if it's bugged, being able to swap into everything is neat and actually unique, so why remove it?

Cocacola4124:
> 
> 
> As far as I'm aware, mob AI taking over if someone is swapped into a mob was fixed in the most recent carrion update. But really, even if it's bugged, being able to swap into everything is neat and actually unique, so why remove it?

You can't swap into everything. Presumably whoever made mindswappers didn't know that superior animals and slimes were a subtype of carbon. Felt like a list is a little easier because you can switch into robots, meanwhile if you just allowed every mob to be switched into it'd break stuff if you didn't limit it to only non-synthetics(Because a lot of weird things are considered mobs)

Cocacola4124:
A blacklist could work, but I doubt it'd be maintained as stuff got added.

TheShown911:
But people should be able to swap into roaches and slimes, don't see a point in this PR. Robotic minds shouldn't be swapped with people in the first place as one is an AI and the other is a real brain.

Cocacola4124:
Robots can use real brains. How does the mindswapper even work. There's still inconsistencies in that logic because you can transfer into kraftwerk roaches.

Cocacola4124:
I'm nitpicking because I want to transfer into robots. I'll try something else.

PR #6066, opened 2021-04-13
adds some new items to the maint blacklist.

Evankhell561:
## Changelog
:cl:
add: RPG, traitor uplink, mindreader and several other debug items were removed from maints.
/:cl:


Wouju:
Wouldn't it be better to name all the items that are blacklisted in the changelog instead of just RPG

Firefox13 (approved):

PR #6072, opened 2021-04-15
Low walls & tables no longer block Weakened mob/living, and airflow stun length nerfed

### LINKED ISSUE #2861: BUG: Low walls block mobs from being blown out into space.
Be in large populated area with default room pressure.
Break window, no fire-lock.
Everyone begins to get blown out into space.
Everyone instead lays perpendicular against the low wall, instead of being pushed with the air flow over it.
Since every full window utilizes a low wall as required, it is no longer possible to get blown out into space when you should.

zoktiik:
## About The Pull Request
This PR in a GIF
- Low walls and unflipped tables will no longer block `mob/living` that are `weakened`, allowing those mobs to pass through when atmos pushes them. This has the hilarious side effect of allowing you to defenestrate stunned people and/or throw them onto/over tables.
- No changes to shields, the modules will still work normally. Base shields won't prevent you from getting pulled out into space, humanoid blocking only will still pull you towards space but still block you at the shield, atmospheric containment will prevent the atmos from moving at all.
- Flipped tables will still block movement in their normal directions.
- The `Weaken(5)` from airflow stunning has been changed to `Weaken(3)`, so that you can't get spaced to a whole other sector on a laggy server with no ability to do anything about it.
- Changed the `magpulse` variable on magboots from `0` and `1` to `FALSE` and `TRUE` respectively, because why not.

Resolves #2861.

## Why It's Good For The Game
Makes combat in some areas more dynamic, gives more danger to being *on a space ship* (and without magboots), makes it so that you're not instafucked by getting spaced out a window or an airlock, places more significant emphasis on the ship's shielding capabilities. For the magboots code change, using `0` and `1` for `true` and `false` statements a shit.

## Changelog
:cl:
tweak: Low walls (without a window) and unflipped tables will no longer block "living", "hard-stunned" creatures, meaning that, without any shielding, you can now break a window to space and get spaced if you don't have magboots.
balance: Airflow stun length has been nerfed from 5 to 3 ticks, so that you don't get back up from the stun in the middle of nowhere after getting spaced.
/:cl:

### DIFF
--- a/code/ZAS/Airflow.dm
+++ b/code/ZAS/Airflow.dm
@@ -16,7 +16,7 @@ mob/proc/airflow_stun()
 		return 0
 	if(!lying)
 		to_chat(src, SPAN_WARNING("The sudden rush of air knocks you over!"))
-	Weaken(5)
+	Weaken(3) // Nerfed from 5
 	last_airflow_stun = world.time
 
 mob/living/silicon/airflow_stun()
--- a/code/game/objects/structures/low_wall.dm
+++ b/code/game/objects/structures/low_wall.dm
@@ -107,6 +107,10 @@
 		return 1
 	if(locate(/obj/structure/low_wall) in get_turf(mover))
 		return 1
+	if(isliving(mover))
+		var/mob/living/L = mover
+		if(L.weakened)
+			return 1
 	return ..()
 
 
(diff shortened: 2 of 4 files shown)


Kurgis:
Add "fixes #2861" to the PR description, so it can be closed on merge.

Firefox13 (approved):

PR #6079, opened 2021-04-16
Makes moebius voidsuit not cost platinum to make

Medcar:
## About The Pull Request
The moebius voidsuit is already locked behind a lot of research, and is practically unobtainable because miners never make platinum or mine at all

## Why It's Good For The Game
plat bad

## Changelog
:cl:
tweak: Makes moebius void not cost platinium
/:cl:


TheShown911:
Good, I would close it too.

PR #6081, opened 2021-04-17
autopsy score fixes

Evankhell561:
## About The Pull Request

it seems that the code was checking if the corpse was from Moebius instead of checking if the user was from that department.

## Changelog
:cl:
fix: autopsy score
/:cl:

### DIFF
--- a/code/modules/core_implant/cruciform/rituals/priest.dm
+++ b/code/modules/core_implant/cruciform/rituals/priest.dm
@@ -298,7 +298,6 @@
 				true_offerings.Add(I)
 
 		if(num_item < req_num)
-			var/obj/item = path
 			break
 		else
 			num_check++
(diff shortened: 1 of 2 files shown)


code/modules/core_implant/cruciform/rituals/priest.dm
```diff
@@ -298,7 +298,6 @@
 				true_offerings.Add(I)
 
 		if(num_item < req_num)
-			var/obj/item = path
```
Evankhell561, line 301: this variable was being declared but is not used at all.

TheShown911: Oh, I have solved the mystery now.

TheShown911 (approved):

PR #6091, opened 2021-04-18
Remove the carrion part in reveal adversaries litany

Firefox13:
## About The Pull Request

This PR removes the carrion part in the reveal adversaries litany and fixes a typo.

## Why It's Good For The Game

This was supposed to be removed when EOTP system was introduced with Miracle that would tell you the area Carrion is.

## Changelog
:cl:
del: Evil powers have affected NT religion! "Reveal Adversaries" is no longer able to detect Carrions. Rely on EOTP instead.
/:cl:

### DIFF
--- a/code/modules/core_implant/cruciform/rituals/base.dm
+++ b/code/modules/core_implant/cruciform/rituals/base.dm
@@ -86,15 +86,8 @@
 				was_triggired = TRUE
 				break
 	if (prob(80) && (locate(/obj/structure/wire_splicing) in view(7, H))) //Add more traps later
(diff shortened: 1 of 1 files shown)


code/modules/core_implant/cruciform/rituals/base.dm
```diff
@@ -86,15 +86,8 @@
 				was_triggired = TRUE
 				break
 	if (prob(80) && (locate(/obj/structure/wire_splicing) in view(7, H))) //Add more traps later
```
TheShown911, line 88: Guess they were never added later.

TheShown911 (approved):

PR #6094, opened 2021-04-18
Fixes infinite handcuff exploit and minor resist improvements.

Jackofashepard:
## About The Pull Request

Being beartrapped and cuffed no longer means you are stuck forever until you kill yourself with failed breakout attempts (the attempts don't actually happen you just get snapped). You also use the trap to break out faster now, note you still need to break out of the trap after.

## Why It's Good For The Game

Prevents players from being cuffed and left for upwards of 20 minutes (time to dispose of yourself depends on armour).

## Changelog
:cl:
tweak: tweaked resist code so the visible message does not display unless you are actually attempting to break out
balance: you now use the trap to break out of your cuffs faster (basically the total time to completely break out evens out)
fix: fixed infinite handcuff in trap exploit
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/traps.dm
+++ b/code/game/objects/items/weapons/traps.dm
@@ -48,6 +48,11 @@ Freeing yourself is much harder than freeing someone else. Calling for help is a
 
 	if (!user)
 		return //No user, or too far away
+	
+	if(iscarbon(user)) //check if mob is carbon as handcuffed only applies to carbon mobs
+		var/mob/living/carbon/C = user //set carbon to user
+		if(C.handcuffed)
+			return//you instantly fail if you are handcuffed and trapped, this way you will loose the handcuffs first instead of repeatedly snapping your torso in half
 
 	//How hard will this be? The chance of failure
 	var/difficulty = base_difficulty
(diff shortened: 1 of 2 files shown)


Iglpigl:
If you are already fixing this to the same when someone is on fire, at the moment if you are on fire and buckled/cuffed you can't so anything and are permanently stuck, it's the same problem/exploit just used in a different way.

Cocacola4124:
> 
> 
> If you are already fixing this to the same when someone is on fire, at the moment if you are on fire and buckled/cuffed you can't do anything and are permanently stuck, it's the same problem/exploit just used in a different way.

You can't really- stop being on fire if you are buckled or cuffed.

Iglpigl:
> > If you are already fixing this to the same when someone is on fire, at the moment if you are on fire and buckled/cuffed you can't do anything and are permanently stuck, it's the same problem/exploit just used in a different way.
> 
> You can't really- stop being on fire if you are buckled or cuffed.

Yep this is what meant to say, just used awkward wording, it should still be fixed.

Jackofashepard:
I'll have a look at it sometime, the resist code is very uh, interestingly done in some places.

Firefox13 (approved):

PR #6095, opened 2021-04-18
Players with godlike strength(robustness) can tear apart handcuffs

Cocacola4124:
## About The Pull Request
As the title states it makes it so players with high enough robustness can tear handcuffs apart.

## Why It's Good For The Game
This is already technically possible, it's just nicer. Beforehand you would need a robustness value of 120 to instantly break cuffs instantly, so players can break handcuffs easier than they could before. If players have robustness higher than 120, it also no longer gives negative values.

## Changelog
:cl:
tweak: If you have robustness of 80 or more you can break cuffs in 5 seconds.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/resist.dm
+++ b/code/modules/mob/living/carbon/resist.dm
@@ -148,14 +148,6 @@
 			SPAN_NOTICE("You successfully remove \the [handcuffed].")
 			)
 		drop_from_inventory(handcuffed)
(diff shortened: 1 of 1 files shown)


Cocacola4124:
Left a mistake in.

Cocacola4124:
Doesn't touch anything that the other handcuff pull does.

TheShown911 (approved):

PR #6096, opened 2021-04-18
Intercom code refactor , Signal for area power changes, src. cleanup on firealarm

MLGTASTICa:
## About The Pull Request
Removes process() from intercoms , now they work based on signals , add a signal for when the APC of an area is deleted, and for when a power change on an area is called, firealarms no longer do src.time which  would result in src.src.time on compile

## Why It's Good For The Game
190 less objects to process overall , bad code removal , signals good.

## Changelog
:cl:
code: added signals for APC power change, and APC deletion for areas.
refactor: refactored intercom code to not process and instead use signals
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/projectile/pistol/avasarala.dm
+++ b/code/modules/projectiles/guns/projectile/pistol/avasarala.dm
@@ -17,7 +17,7 @@
 	can_dual = TRUE
 	damage_multiplier = 1.45
 	penetration_multiplier = 1.35
-	recoil_buildup = 33
+	recoil_buildup = 5
 
 	fire_sound = 'sound/weapons/guns/fire/hpistol_fire.ogg'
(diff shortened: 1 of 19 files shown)


code/modules/projectiles/guns/projectile/pistol/avasarala.dm
```diff
@@ -17,7 +17,7 @@
 	can_dual = TRUE
 	damage_multiplier = 1.45
 	penetration_multiplier = 1.35
-	recoil_buildup = 33
+	recoil_buildup = 5
```
Evankhell561, line 20: wtf

MLGTASTICa: fuck , commited on wrong branch.

PR #6100, opened 2021-04-19
another day another fix to the craft system.

Evankhell561:
## About The Pull Request
fix this:
image

## Changelog
:cl:
fix: fix the craft system, maybe.
/:cl:


code/modules/core_implant/cruciform/rituals/priest.dm
```diff
@@ -298,7 +298,6 @@
 				true_offerings.Add(I)
 
 		if(num_item < req_num)
-			var/obj/item = path
```
TheShown911, line 301: Hmmm, same thing here, like in #6102

Firefox13 (approved):

PR #6102, opened 2021-04-19
gun assembly fixes

Evankhell561:
## About The Pull Request

when creating a weapon using the gun assembly it returns all guns parts used.

the weapons you can see are no longer organized in function of higher spawn frequency to lower spawn frequency.

## Changelog
:cl:
fix: gun assembly no longer returns to the user the gun parts used to create a gun.
tweak: gun that can be created using a gun assembly are no longer displayed according to their spawn frequency.
/:cl:

### DIFF
--- a/code/modules/core_implant/cruciform/rituals/priest.dm
+++ b/code/modules/core_implant/cruciform/rituals/priest.dm
@@ -298,7 +298,6 @@
 				true_offerings.Add(I)
 
 		if(num_item < req_num)
-			var/obj/item = path
 			break
 		else
 			num_check++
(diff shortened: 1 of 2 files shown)


code/modules/core_implant/cruciform/rituals/priest.dm
```diff
@@ -298,7 +298,6 @@
 				true_offerings.Add(I)
 
 		if(num_item < req_num)
-			var/obj/item = path
```
TheShown911, line 301: Is this connected to the fix? If not explain somewhere what this is a fix for.

Firefox13 (approved):

PR #6120, opened 2021-04-24
Spear Banishment

Wouju:
## About The Pull Request

Banishes the spear for all eternity, being a wonky bay leftover that used material hardness and before the bandaid fix could do a whopping 75 damage per hit

## Why It's Good For The Game

despite spears being, well funky spears, their damage output and usage of material hardness resulted in real wonky business
image
image

## Changelog
:cl:
del: Banishes spears to the eternal realm of the TeeGee github, where it will suffer for all eternity for using bay code
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/improvised_components.dm
+++ b/code/game/objects/items/weapons/improvised_components.dm
@@ -37,11 +37,7 @@
 /obj/item/weapon/material/wirerod/attackby(var/obj/item/I, mob/user)
 	..()
 	var/obj/item/finished
-	if(istype(I, /obj/item/weapon/material/shard))
-		var/obj/item/weapon/material/tmp_shard = I
-		finished = new /obj/item/weapon/material/spear(get_turf(user), tmp_shard.material.name)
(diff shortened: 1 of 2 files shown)


MatejaTadic:
In some ways, I do agree with you. But, completely removing the spear is completely unacceptable. The better way to fix this is to slightly nerf the spear.

TheShown911 (approved):

Medcar:
Maybe instead of removing it completely rework it?

mattroks101 (approved):
Good change.

Kurgis:
We might want to strip the majority of `/obj/item/weapon/material` while we are at it (excluding things such as glass shards and metal scrap of course).

Wouju:
before anyone moans more about removing spears

https://github.com/discordia-space/CEV-Eris/pull/6122

Wouju:
> We might want to strip the majority of `/obj/item/weapon/material` while we are at it (excluding things such as glass shards and metal scrap of course).

true
but at the same time i also don't want to bother with that

PR #6123, opened 2021-04-24
Remove ladder cheese with slipping items, traps and mines

hyperioo:
## About The Pull Request

Avoid triggering soaps, excel mines, beartraps, banana peels, carrion smooth spiders, clown PDA and wet floors on ladder/stairs tiles.

## Why It's Good For The Game

To avoid ladder cheesing with slipping objects since you cannot interact with them if you're on another zlevel. You can basically get an easy killzone / interdiction tool if a ladder is the only way to get into an area. Just have to carry a soap with you and if someone follow you then you just can drop a soap on the ladder, nobody can move up/down anymore without getting slipped.
And some of the ship areas only have a single ladder leading to them (especially in maints), or two but super far away from each other, so you cannot avoid it.

Technically beartraps and mines cannot be armed on ladder tile but better be safe than sorry so I've included them in the package in case some bored people find an exploit.

**Same for stairs.**

Also removed slipping from wet/lubed floors on ladder/stairs tiles.

## Changelog
:cl: Hyperio
balance: Remove ladder and stairs cheese with slipping items, traps, mines and wet floors
/:cl:

### DIFF
--- a/code/game/objects/items/devices/PDA/PDA.dm
+++ b/code/game/objects/items/devices/PDA/PDA.dm
@@ -1333,7 +1333,7 @@ var/global/list/obj/item/device/pda/PDAs = list()
 /obj/item/device/pda/clown/Crossed(AM as mob|obj) //Clown PDA is slippery.
 	if (isliving(AM))
 		var/mob/living/M = AM
-		if(locate(/obj/structure/multiz/ladder) in get_turf(loc))
+		if(locate(/obj/structure/multiz/stairs) in get_turf(loc) || locate(/obj/structure/multiz/ladder) in get_turf(loc))
 			visible_message(SPAN_DANGER("\The [M] carefully avoids stepping down on \the [src]."))
 			return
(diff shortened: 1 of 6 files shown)


Iglpigl:
Does this change affect water/lube on ladder tiles too?

mattroks101 (approved):
Blessed PR.

hyperioo:
> 
> 
> Does this change affect water/lube on ladder tiles too?

Good point, I removed water/lub slipping on ladder tiles.

Also applied the same fixes for stairs since you could do the same things than with ladders.

Iglpigl:
> > Does this change affect water/lube on ladder tiles too?
> 
> Good point, I removed water/lub slipping on ladder tiles.
> 
> Also applied the same fixes for stairs since you could do the same things than with ladders.

Goos job man, thanks for stairs thing too.

TheShown911 (approved):

PR #6129, opened 2021-04-26
removes floor stun from disarm

Jackofashepard:
## About The Pull Request

Removes the floor stun from disarm intent.

## Why It's Good For The Game

You will notice these days that whenever a fight happens, even with guns, and one party decides to use their fists it will always be on disarm intent. The reason for this is this floor stun, its simply the strongest move, if you stun someone for three seconds on the floor they can't do anything and might as well be dead because thats 3 seconds to unload into them, decap, grab etc. In my opinion floor stuns are a leftover from old SS13 and has long outlived its place on Eris where we place emphasis on fast and fun combat, getting stunned on the floor and being unable to do anything is the exact opposite. 

## Changelog
:cl:
del: Removed floor stun from disarm intent 
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_attackhand.dm
+++ b/code/modules/mob/living/carbon/human/human_attackhand.dm
@@ -266,11 +266,6 @@
 
 			var/randn = rand(1, 100)
 			randn = max(1, randn - H.stats.getStat(STAT_ROB))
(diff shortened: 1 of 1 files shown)


Sparrowbird611:
Now remove it from melee :)

PR #6136, opened 2021-04-30
Players can toss objects a z-level above

bigpep:
## About The Pull Request
Lets players toss an item a z-level above, if it's below a certain weight. I'm not sure if this is hacky or not, but it seems like a decent way of going about it. To summarize, it first checks to see if the turf above the player is an open space, then it checks to see if the item is below a certain weight class and if enough time has passed(Calculated by a default of 5 * weight_class), enables the throwing variable on it(So it doesn't fall), moves it to the z-level above, then throws it normally. 

## Why It's Good For The Game
It's one of those things that should already be there in my opinion, when a player looks up and tries to throw something, it should just work.
Vid

## Changelog
:cl:
add: Players can toss objects a floor above themselves, but heavy objects can take a while to throw.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -296,7 +296,7 @@
 	if(item && src.unEquip(item, loc))
 		if((target.z > src.z) && istype(get_turf(GetAbove(src)), /turf/simulated/open))
 			var/obj/item/I = item
-			src.visible_message(SPAN_DANGER("[src] is beginning to throw \the [item] high!"))
+			src.visible_message(SPAN_DANGER("[src] is trying to toss \the [item] into the air!"))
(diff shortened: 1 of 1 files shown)


Cocacola4124:
oops

Cocacola4124:
I should probably clean this up a bit, but it doesn't seem like much of an issue.

TheShown911:
How does this work with grabbing and throwing mobs?

Cocacola4124:
I will note that you can throw objects onto the same tile as another player, because the weaken on it doesn't check to see if the weight of the object that fell on you is substantial, it means that something like a mouse(That is thrown practically instantly) can weaken a player. I'm not sure if that's exactly my issue though or just weaken being annoying.

bigpep:
Going to close this because I deleted the fork accidently, some parts are a bit wonky.

PR #6142, opened 2021-05-01
Winter-change

AgentMA1:
## About The Pull Request

Makes winternute deal more damage, but reduces its fire-rate

## Why It's Good For The Game

A buff asked by Jack, mainly making the good old wintermute make more damage but in exchange slower fire-rate, as it wont be able to go Full Auto. Tested the numbers with some armors, it might need some test merge to see how it is seen or exploited, just in case its need to get nerfed.

## Changelog
:cl:
balance: Buffed damage of the wintermute, but removed its full auto, reducing a lot the fire rate.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/projectile/automatic/wintermute.dm
+++ b/code/modules/projectiles/guns/projectile/automatic/wintermute.dm
@@ -22,10 +22,11 @@
 	cocked_sound = 'sound/weapons/guns/interact/ltrifle_cock.ogg'
 	zoom_factor = 0.4
 	recoil_buildup = 2
+	penetration_multiplier = 1.3
+	damage_multiplier = 1.5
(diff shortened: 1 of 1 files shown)


TheShown911:
50% more damage for removing full auto is way too strong.

AgentMA1:
> 50% more damage for removing full auto is way too strong.

Its just slightly over Excel version,maybe lowering the damage multiplier to 1.3?

Voldirs:
So it's basically sol, but in 20 caliber? I'd say it just needs recoil lowered and AP buffed.

MLGTASTICa:
i'd rather have it kept full auto , its the only full auto rifle IH gets..

AgentMA1:
Changing this crap tomorrow

PR #6149, opened 2021-05-03
Maintchem and vomit! 

Jackofashepard:
## About The Pull Request

Adds a variety of chemicals (vomitol most common) to maint food and makes food spawns rarer.

## Why It's Good For The Game

Maintchem is now a bit more possible and maint food is overall more dangerous/rare.

## Changelog
:cl:
add: Maint food now contains a variety of chemicals.
tweak: Maint food is rarer. 
map: Replaced some normal food spawns with low_chance.
/:cl:

### DIFF
--- a/code/game/objects/random/food.dm
+++ b/code/game/objects/random/food.dm
@@ -2,11 +2,12 @@
 	name = "random preserved rations"
 	icon_state = "food-green"
 	tags_to_spawn = list(SPAWN_RATIONS)
+	spawn_nothing_percentage = 90
 
 /obj/spawner/rations/low_chance
 	name = "low chance preserved rations"
 	icon_state = "food-green-low"
-	spawn_nothing_percentage = 60
+	spawn_nothing_percentage = 95
 
 
 /obj/spawner/junkfood
@@ -14,11 +15,12 @@
 	icon_state = "food-red"
 	has_postspawn = FALSE
 	tags_to_spawn = list(SPAWN_JUNKFOOD)
+	spawn_nothing_percentage = 90
 
 /obj/spawner/junkfood/low_chance
 	name = "low chance junkfood"
 	icon_state = "food-red-low"
-	spawn_nothing_percentage = 60
+	spawn_nothing_percentage = 95
 
 /obj/spawner/junkfood/rotten
 	name = "random spoiled food"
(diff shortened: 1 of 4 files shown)


code/game/objects/random/food.dm
```diff
@@ -2,23 +2,25 @@
 	name = "random preserved rations"
 	icon_state = "food-green"
 	tags_to_spawn = list(SPAWN_RATIONS)
+	spawn_nothing_percentage = 90
```
Evankhell561, line 5: You should not apply a no spawn chance to a spawner that does not specify that it is a low chance spawner.

Evankhell561 (changes requested):

code/game/objects/random/food.dm
Jackofashepard: Option 2 is removing the spawn entirely, these rations specifically are a big problem since they come with no drawbacks whatsoever.

Evankhell561: There are many more options to do this.

Jackofashepard: Namely?

Sparrowbird611:
Is there really a reason to make it more painful to get food in the first place? This doesn't really help the club all that much IMO, it just makes eating more obnoxious.

Jackofashepard:
Well the original idea behind the vendor removal was to make people go to club because people just buy the food. This follows the same line but with maint food.

Wouju:
> Well the original idea behind the vendor removal was to make people go to club because people just buy the food. This follows the same line but with maint food.

True

but we all know how vendor removal worked out

Jackofashepard:
This is true, not at all, due to everyone now going to maint instead.

code/game/objects/random/food.dm
Evankhell561: eliminate spawners with no chance to spawn anything on the map for example (or in the closets).

Sparrowbird611:
> Well the original idea behind the vendor removal was to make people go to club because people just buy the food. This follows the same line but with maint food.

You have a lot more sources of food to remove if you want to make that work out, and all so a role that gets played once every five rounds has more use than being an insight drink generator. Wouldn't it just be better to give the club something that would attract more people, like a rage cage or a reliable drug production system or something?

code/game/objects/random/food.dm
Jackofashepard: Keeping this as resolved, beats replacing every instance of /obj/spawner/junkfood

Jackofashepard:
I have reworked this PR, it now makes maintchem more viable and maint food a bit rarer.

Firefox13 (approved):

PR #6151, opened 2021-05-04
Multiz throwing

bigpep:
## About The Pull Request
Try 2:
Being able to throw an item a floor above themselves is something a player would expect to be able to do. This is a pretty small change, and it's a tiny bit more polished than my previous pull. There is one math thing in it, which just multiplies an item's weight class by 5, then subtracts the result by the players robustness * 0.1, for a large object and pretty high strength this is a few seconds off

## Why It's Good For The Game
Vid
Like stated above, players expect certain cause & effect reactions. If they attempt to toss an item a floor above, they expect it to work. If they try to shoot a floor above, they expect it to work(Although this isn't the case yet for guns). 

## Changelog
:cl:
add: You can now toss items a floor above yourself! Heavier objects take longer to throw, slightly altered by your strength.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -296,7 +296,7 @@
 	if(item)
 		if((target.z > src.z) && istype(get_turf(GetAbove(src)), /turf/simulated/open))
 			var/obj/item/I = item
-			var/robust = usr.stats.getStat(STAT_ROB)
+			var/robust = stats.getStat(STAT_ROB)
 			var/timer = ((5 * I.w_class) - (robust * 0.1)) //(W_CLASS * 5) * (STR * 0.1)
 			visible_message(SPAN_DANGER("[src] is trying to toss \the [item] into the air!"))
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/carbon.dm
```diff
@@ -293,15 +293,29 @@
 				return
 
 	//Grab processing has a chance of returning null
-	if(item && src.unEquip(item, loc))
-		src.visible_message("\red [src] has thrown [item].")
+	if(item)
+		if((target.z > src.z) && istype(get_turf(GetAbove(src)), /turf/simulated/open))
+			var/obj/item/I = item
+			var/robust = stats.getStat(STAT_ROB)
+			var/timer = ((5 * I.w_class) - (robust * 0.1)) //(W_CLASS * 5) - (STR * 0.1)
```
mattroks101, line 300: `//(W_CLASS * 5) - (STR * 0.1)`?

bigpep:
Just seemed easier to read.

PR #6154, opened 2021-05-05
cruciform energy fixes V4

Evankhell561:
## About The Pull Request

fixes an obviously not broken code.

The power regeneration proc in byond runs every second but byond defines the seconds as 600 instead of 60 so it runs ten times slower than it should.

## Changelog
:cl:
fix: cruciform energy
tweak: reduces the energy cost of the offerings to 10
tweak: doubles the energy regeneration coming from the obelisks.
tweak: the energy regeneration miracle further increases energy regeneration.
/:cl:

### DIFF
--- a/code/modules/core_implant/cruciform/cruciform.dm
+++ b/code/modules/core_implant/cruciform/cruciform.dm
@@ -24,7 +24,7 @@ var/list/disciples = list()
 		return
 
 	var/true_power_regen = power_regen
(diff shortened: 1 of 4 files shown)


Jamini:
That is because it's in ticks. not seconds. There are ten ticks to one second.

Evankhell561:
> That is because it's in ticks. not seconds. There are ten ticks to one second.



> That is because it's in ticks. not seconds. There are ten ticks to one second.

At first I thought the cruciform proc process was running 10 times per second but apparently it runs once every one second.

PR #6155, opened 2021-05-06
Injecting sugar in bad.

Jackofashepard:
## About The Pull Request

Sugar now gives less nutrition and has some side effects, most notable being heart damage.

## Why It's Good For The Game

Apparently people have been injecting raw sugar into their veins so forego food.

## Changelog
:cl:
tweak: Decreased nutrition from sugar and added OD effects, most notably minor heart damage.
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/dispenser.dm
+++ b/code/modules/reagents/reagents/dispenser.dm
@@ -420,7 +420,7 @@
 /datum/reagent/organic/sugar
 	name = "Sugar"
 	id = "sugar"
-	description = "The organic compound commonly known as table sugar and sometimes called saccharose. This white, odorless, crystalline powder has a pleasing, sweet taste. Its not a good idea to inject too much raw sugar into your bloodstream."
+	description = "The organic compound commonly known as table sugar and sometimes called saccharose. This white, odorless, crystalline powder has a pleasing, sweet taste. It is not a good idea to inject too much raw sugar into your bloodstream."
 	taste_description = "sugar"
 	taste_mult = 1.8
 	overdose = 40
--- a/code/modules/reagents/reagents/dispenser.dm
+++ b/code/modules/reagents/reagents/dispenser.dm
@@ -423,6 +423,7 @@
 	description = "The organic compound commonly known as table sugar and sometimes called saccharose. This white, odorless, crystalline powder has a pleasing, sweet taste."
 	taste_description = "sugar"
 	taste_mult = 1.8
+	overdose = 40
 	reagent_state = SOLID
 	color = "#FFFFFF"
 	glass_icon_state = "iceglass"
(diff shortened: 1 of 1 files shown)


Firefox13:
Please add an in-game tip about this.

SaltyOldBull:
why heart damage specifically.

Wouju:
> why heart damage specifically.

too much sugar can cause arteries to get bigger (or rather the passage in them is now smaller)

however excess sugar doesn't just damage the heart, but making it damage more than the heart would be bothersome and perhaps too annoying

SaltyOldBull:
yeah, if you're an untreated type 2 diabetic over a long period of time, this change doesn't make much sense otherwise. The body is made to handle such glucose spikes

MLGTASTICa:
> yeah, if you're an untreated type 2 diabetic over a long period of time, this change doesn't make much sense otherwise. The body is made to handle such glucose spikes
> now, I'd be more worried about the onset of refeeding syndrome on someone that's red on nutrition slamming a bluespace beaker worth of sugar down

This is a game fuck , realism shouldn't be used as an argument . Its a heart attack because mild symtomps just don't discourage  people properly.

SaltyOldBull:
> This is a game, realism shouldn't be used as an argument . Its a heart attack because mild symtomps just don't discourage people properly.

maybe you wouldn't need to resort to such drastic measures if they actually gave people food and stopped trying to make cannibalism be a viable method of feeding yourself and poisoning everything else

Sparrowbird611:
The club isn't even played half the time, I don't see why we keep trying to make it viable by doing this.

Wouju:
> yeah, if you're an untreated type 2 diabetic over a long period of time, this change doesn't make much sense otherwise. The body is made to handle such glucose spikes
> now, I'd be more worried about the onset of refeeding syndrome on someone that's red on nutrition slamming a bluespace beaker worth of sugar down

fair, i assume the PR's intent is to discourage people injecting sugar into the veins, then drinking it for even more nutrition gain

though as sparrow said

> The club isn't even played half the time, I don't see why we keep trying to make it viable by doing this. Nerfing every other food source isn't going to fix it.

this won't exactly fix the whole food problem we got going on without adressing club

Jackofashepard:
In game tip added, this is mostly to make nerds go outside for food or just grind down roaches for animal protein.

TheShown911:
Wouldn't this nerf nutrition on all foods with sugar?

Jackofashepard:
It seems to only have a blood effect and not a inject or touch, that being said the only foods that have sugar physically in their reagent lists is sodas and 1-2 candy bars.

TheShown911:
IIRC ingested is the same as in blood, but runs at half the speed

Jackofashepard:
Ahh i see, then yes it would be a slight nerf, however the foods that do contain sugar as a physical reagent are few and its a very small amount.

Firefox13 (approved):

PR #6173, opened 2021-05-10
Increases the EMP damage deals to robotic internal and external organs

MLGTASTICa:
## About The Pull Request
External damage changed from  10 to 20 , 5 to 10 , 1 to 5
Internal damage changed from  9 to 32,  3 to  18 , 1 to 9
## Why It's Good For The Game
FBPs barely feel any pain , get armour by default , don't have to deal with blood. EMP's are often used against them but they yield no result , this should change it .
Organ health also got increased , so should the damage to keep up.

## Changelog
:cl:
balance: Rebalanced the emp damage taken by robotic organs and limbs, they now take on the average twice as much damage.
/:cl:

### DIFF
--- a/code/modules/organs/external/_external.dm
+++ b/code/modules/organs/external/_external.dm
@@ -276,11 +276,11 @@
 		return
 	switch (severity)
 		if (1)
-			take_damage(10)
+			take_damage(20)
 		if (2)
-			take_damage(5)
+			take_damage(10)
 		if (3)
-			take_damage(1)
+			take_damage(5)
 
 /obj/item/organ/external/attack_self(var/mob/user)
 	if(!contents.len)
--- a/code/modules/organs/organ.dm
+++ b/code/modules/organs/organ.dm
@@ -277,11 +277,11 @@
 		return
 	switch (severity)
 		if (1)
-			take_damage(9)
+			take_damage(32)
 		if (2)
-			take_damage(3)
+			take_damage(18)
 		if (3)
-			take_damage(1)
+			take_damage(9)
 
 // Gets the limb this organ is located in, if any
 /obj/item/organ/proc/get_limb()


Sparrowbird611:
EMPs are already pretty painful for an FBP or even just a regular augmented person to handle, there's no need to buff them further. They damage to every part and organ (21 organs and 7 body parts minimum for an FBP), and in the case of EMP grenades are pretty much undodgeable, while the halicon can currently use a full automod for constant AoE blasts. They don't need to be made stronger.

Also the changelog is very misleading.

TheShown911:
Have you tested this, i'm pretty sure FBPs will drop down dead from one EMP nade?

MLGTASTICa:
> Have you tested this, i'm pretty sure FBPs will drop down dead from one EMP nade?

Tested.  FBP drop dead on 2-3(depending on toughness) maximum severity pulses , they go blind on the first one.
Medium severity drops down on 4-6
Lowest one at 6+

TheShown911:
Toughness shouldn't effect death treshold, but only pain. Also what severity are EMP nades and ion shots?

MLGTASTICa:
> Toughness shouldn't effect death treshold, but only pain. Also what severity are EMP nades and ion shots?
EMP Nades have maximum seveverity in a 3-4 tile range , medium in a 15-17 tile range and no weak , only the fyre has that. ION shots are maximum severity, but have no AOE , so you have to hit a direct hit .

Wouju:
I decided to test it out and i think this is really unbalanced.

Unless armour somehow reduces EMP damage then you can easily take out a FBP with a single shot by making them blind, or very well knock them out with two or three

Sparrowbird611:
> > Toughness shouldn't effect death treshold, but only pain. Also what severity are EMP nades and ion shots?
> 
> EMP Nades have maximum severity in a 3-4 tile range , medium in a 15-17 tile range and no weak , only the fyre has that. ION shots are maximum severity, but have no AOE , so you have to hit a direct hit .
> Right now only traitors can get the AOE Maximum severity , and IH has to score easily-dodgeable direct hits that cost a lot of power cell charge.

The nade is still effectively undodgeable, as the blast also goes through walls and as you just said, reaches 15 tiles. I'm not even sure I could clear that in open ground with the speed boosts I occasionally run, let alone in maintenance. As for the Halicon, it actually does have an AoE.

EMPs are already fairly effective against FBPs because they hit every single part at once and in most cases are impossible to dodge. Don't try and make them have the power to delete people when they are that easy to use.

And just to point out the flaws in the main points you brought up about FBPs, the innate armor they have, at maximum in the majority of cases is 35% (moeb RR++). This is barely a gain over the average 30% armor, and significantly better armor isn't even that difficult to acquire, so this point is rather moot. Bleeding is almost never an issue, especially since erismed 3 nerfed it further. As for the barely feel pain part, I'm not sure where you pull this from but this is just false. I've been paincrit by stepping on electrical wires 2 or 3 times, which is in-line with how much it would take to stop a normal human, and was once almost crit just by having 3 robo limbs and being hit by a few EMP grenades through walls by a person I couldn't even see. Cybernetics definitely have some useful abilities, but given you brought up the "advantages" that don't even matter or just don't exist, I'm not convinced you know what makes them strong or how to balance them.

MLGTASTICa:
> I decided to test it out and i think this is really unbalanced.
> 
> Unless armour somehow reduces EMP damage then you can easily take out a FBP with a single shot by making them blind, or very well knock them out with two or three
> 
> image
> At 5 shots with the halicon they fucking explode
> 
> Tl;dr - this pr is really fucking unbalanced
> 
> At 1 shot the FBP goes blind, unable to fight back
> At 3-4 they die

Landing a direct hit is harder than it seems , also keep in mind the direct hits are maximum-severity.

Firefox13 (approved):

PR #6181, opened 2021-05-12
FIX: Ironhammer disk sprite in hands issue.

DarkLevelSS13:
## Changelog
:cl: DarkLevel - code.
fix: fixing ironhammer disk sprite in hands
/:cl:


Firefox13:
It's `disk`, not `disc`.
~~maybe you could rename the hammer instead of disk~~

mattroks101 (changes requested):
Not a good way of doing this. Change the Ironhammer `icon_state` name instead of changing all the disks.

PR #6183, opened 2021-05-13
Riot shield buff / Every shield nerf

CeUvi:
## About The Pull Request

1. Now riot shield can block everything, same as NT shield. Raised material cost of riot shield.
2. Now, to get armor and block chance from riot and NT shields, you need to WALK.
3. Added walking and running states for riot and NT shields.
shields

## Why It's Good For The Game

1. IH will have some good shield
2. Some people said that shields are too OP. Not anymore. (i guess)

## Changelog
:cl: CeUvi and dev mentors
tweak: you need to WALK to get armor and block chance from riot and NT shields
imageadd: walking and running states for shields
/:cl:


GrayRachnid:
how does this work with standing still?

CeUvi:
> how does this work with standing still?

it doesnt work with standing still

GrayRachnid:
That should be fixed then, it's dumb that you can't use shields while standing still. Maybe make it using the shield that breaks when you run?

TheShown911 (changes requested):
It should work when standing still.

TheShown911 (approved):

PR #6189, opened 2021-05-13
Adds the kovacs

Jackofashepard:
## About The Pull Request

Adds the Kovacs battle rifle. High pen, high damage and high accuracy rifle at the cost of fire delay and being semi-auto only. Its in the serb vendor as contraband and is on the serb base in a crate, i just added one more rack to the collection. Has slight zoom capabilities.

image
image
image
image
image

## Why It's Good For The Game

We do not have battle rifles at all, it will be interesting to see how the kovacs measures up against other guns. Adds the option for serbs to be a sniper without using the .60.

## Changelog
:cl:
add: Added kovacs to serb base and vendor.
imageadd: kovacs
/:cl:


Sparrowbird611:
Two questions:

Why'd you make this contraband instead of just a normal gun in the vendor?

How much zoom is slight zoom? If it's meant to be a sniper weapon I'd expect somewhere around watchman scope levels, or at least significantly more than the wintermute.

Wouju:
> Two questions:
> 
> Why'd you make this contraband instead of just a normal gun in the vendor?
> 
> How much zoom is slight zoom? If it's meant to be a sniper weapon I'd expect somewhere around watchman scope levels, or at least significantly more than the wintermute.

image
More than the wintermute/C20 less than the vintorez

code/modules/projectiles/guns/projectile/battle_rifle/kovacs.dm
```diff
@@ -0,0 +1,47 @@
+/obj/item/weapon/gun/projectile/kovacs
+	name = "SA BR \"Kovacs\""
+	desc = "The kovachs is a refined battle rifle fit for taking down heavily armoured targets. \
```
Firefox13, line 3:
```suggestion
	desc = "The \"Kovacs\" is a refined battle rifle fit for taking down heavily armoured targets. \
```

Firefox13 (approved):

PR #6202, opened 2021-05-15
Fixes a horrible runtime from signals

MLGTASTICa:
## About The Pull Request
/mob/proc/unequip is apparently unsafe and its called to unequip WITHOUT an item.
## Why It's Good For The Game
Runtimes lag 
## Changelog
:cl:
fix: fixed the clothing drop signal being called on nulls.
/:cl:

### DIFF
--- a/code/modules/mob/inventory.dm
+++ b/code/modules/mob/inventory.dm
@@ -129,7 +129,8 @@
 	if(!canUnEquip(I))
 		return
 	SEND_SIGNAL(src, COMSIG_CLOTH_DROPPED, I)
-	SEND_SIGNAL(I, COMSIG_CLOTH_DROPPED, src)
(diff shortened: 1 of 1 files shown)


code/modules/mob/inventory.dm
```diff
@@ -129,7 +129,8 @@
 	if(!canUnEquip(I))
 		return
 	SEND_SIGNAL(src, COMSIG_CLOTH_DROPPED, I)
```
TheShown911, line 131: What about this signal, it also uses `I`, is it s problem too?

MLGTASTICa: passing arguments as null is not a problem , but sending our signal to  null is.

TheShown911 (approved):

PR #6209, opened 2021-05-15
Kovacs tweaks

Jackofashepard:
## About The Pull Request

Lowered the amount of kovacs available and made it more expensive.

## Why It's Good For The Game

Too common for how good it is, too cheap too. 

## Changelog
:cl:
tweak: kovacs price from 2000 to 3000 and amount available from 5 to 2.
/:cl:


Sparrowbird611:
Can we stop removing every food source. This isn't fixing the bar at all, it just makes gameplay more tedious. Your insistence on making the only food sources roaches and the garden is painful.

Wouju:
Removing the ERP is a dumb idea, as sparrow said it just makes shit more tedious.

And who the FUCK would pay 800 credits for the FOOD??? No one in their right mind would spend 800 creds on some meh food when they can easily go to bar or break in there.

Better yet, the true reason why you would pay for the ERP is the chems rather than the food itself.

Tl;dr

Removing the ERP is a dumb idea, don't do it, removing snack vendors didn't fix any issues regarding club and neither will this

TheShown911 (approved):

PR #6213, opened 2021-05-16
removes health_deficiency

Sparrowbird611:
## About The Pull Request

Removes slowdown from damage. It is now based only on limb efficiency and pain.

## Why It's Good For The Game

This is a relatively small change, but it could potentially have a big impact on the overall game. First shot advantaged will be leve important, enabling more interesting fire fighters and more options to retreat, rather than just trying to outdps each other with automatics and praying for the best. This also feels more consistent, since you would expect pain and limb efficiency to dictate speed, not some random, arbitrary damage number to stack with these things and slow you down massively.

## Changelog
:cl:
balance: Removes slowdown from damage. It is now based only on limb efficiency and pain.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -27,10 +27,8 @@
 			var/obj/item/weapon/cruciform_upgrade/speed_of_the_chosen/sotc = upgrade
 			tally -= sotc.speed_increase
 
-	var/health_deficiency = (maxHealth - health)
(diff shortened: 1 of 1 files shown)


Sparrowbird611:
I tested this locally. There is a decent increase in movement speed here, which should make both melee tactics and retreats more viable, and overall allow more control over the character. Pain slowdown is still around, though it takes a bit more damage to become as slow as you could be previously.

mattroks101 (approved):

PR #6220, opened 2021-05-16
Organic buff: Makes limb HP separate from death.

Sparrowbird611:
## About The Pull Request

Makes it so limbs and groin don't contribute to death. They only deal pain and the associated advantages of disabling limbs now.

## Why It's Good For The Game

Further augments erismed 3 by giving a safer location to shoot someone, even if you're using lethal ammunition. The end result may make humans a tad tankier in terms of surviving death, but paincrit should be about the same. Shooting people in the head and chest is still lethal of course.

## Changelog
:cl:
balance: Humans don't have shared HP between all limbs and groin, making them safer to shoot for non-lethal takedowns. Aim for the chest or head to kill.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_damage.dm
+++ b/code/modules/mob/living/carbon/human/human_damage.dm
@@ -8,6 +8,7 @@
 	var/total_brute = 0
 	for(var/obj/item/organ/external/O in organs)	//hardcoded to streamline things a bit
 		if(BP_IS_ROBOTIC(O) && !O.vital)
+		if(BP_IS_ORGANIC(O) && !O.vital)
 			continue //*non-vital* robot limbs don't count towards shock and crit
 		total_brute += O.brute_dam
 		total_burn  += O.burn_dam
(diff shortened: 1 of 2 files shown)


FortuneMaster:
Not realistic, but maybe good for the game

Jackofashepard:
Its good for the game, besides people will die from blood loss if they lose limbs so it evens out.

Sparrowbird611:
> Not realistic, but maybe good for the game

Realism is overrated anyway.

Wouju:
if i can't be a Chivarly character with 3 missing limbs in my atmos sim

why live

Sparrowbird611:
You can technically do that RN, since a missing limb also doesn't count towards your death HP.

Wouju:
> You can technically do that RN, since a missing limb also doesn't count towards your death HP.

i know but with this it would be easier i believe

MadmanMartian:
I did this before on another server. Expect people to welderbomb themselves with impunity.

Sparrowbird611:
> I did this before on another server. Expect people to welderbomb themselves with impunity.

I'd be surprised if we saw more suicide welderbombers, but you are probably right that this would make explosives a bit less lethal.

Jackofashepard:
Nah, our explosives are a lot more lethal than most servers, we changed it to be so a loooooooooong time ago

TheShown911 (approved):
What could go wrong?

PR #6228, opened 2021-05-18
Blacklist ballistic NVGs from spawning

Firefox13:
## About The Pull Request

They are treated as abstract item and shouldn't be spawned anywhere else other than when using the tactical ballistic helmet

## Why It's Good For The Game

less bugs

## Changelog
:cl:
fix: Fixed ballistic night vision goggles appearing outside of tactical ballistic helmet
/:cl:

### DIFF
--- a/code/modules/clothing/glasses/night.dm
+++ b/code/modules/clothing/glasses/night.dm
@@ -22,7 +22,7 @@
 	darkness_view = 7
 	see_invisible = SEE_INVISIBLE_NOLIGHTING
 	flags = ABSTRACT
-	spawn_blacklisted = FALSE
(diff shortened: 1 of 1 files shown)


code/modules/clothing/glasses/night.dm
```diff
@@ -22,6 +22,7 @@
 	darkness_view = 7
 	see_invisible = SEE_INVISIBLE_NOLIGHTING
 	flags = ABSTRACT
+	spawn_blacklisted = FALSE
```
Firefox13, line 25:
```suggestion
	spawn_blacklisted = TRUE
```
I'm blind

TheShown911 (approved):

PR #6248, opened 2021-05-22
Deepmaint hiveroom aka hive nest

Wouju:
## About The Pull Request

Another kneejerk PR
adds a normal sized deepmaint room with some loot and one of each hivemind mob (except bomber), along with a abnormal PDA and 2 hivemind beakers (so a total of 60 units of each nanite)
image
image

## Why It's Good For The Game

It's not

I believe this could spice up deepmaint a bit more, by having a new room with "new" enemies in it, while not having broken easy to gain loot within nor being bloat

## Changelog
:cl:
add: When the hive is destroyed all the machinery is destroyed aswell, what about the little guys? A group of hive mobs have been left behind deep within the ship.
/:cl:


Firefox13 (changes requested):
The .dmm doesn't looks mapmerged.
Also, we have spawners for a reason, don't just straight put oddity there. Put an oddity spawner there or don't if we don't even have those.

Wouju:
> The .dmm doesn't looks mapmerged.
im confused a bi tthen
is there anything more to do when making a new map using strongdmm?
> Also, we have spawners for a reason, don't just straight put oddity there. Put an oddity spawner there or don't if we don't even have those.
aight, gimme a second

Wouju:
> Don't hive mobs just slowly die? It's entirely possible that by the time this is found the mobs are just dead.

they do but this is because of their malfunction chance, i set the mobs in the room to have a malfunction chance to be 0

mattroks101:
> The .dmm doesn't looks mapmerged.

The DMM looks fine.

Firefox13 (approved):

PR #6256, opened 2021-05-24
Adds random disk crate to Zarya

zombiemimic:
## About The Pull Request
Your gun fetish can now be sated with lootboxes at a price that can actually make a dent in your space bitcoin

## Why It's Good For The Game
makes guild less maint-dependant than it is

## Changelog
:cl:
add: Autolathe disk box now is sold from the trade beacon
/:cl:


zombiemimic:
travis being smoothbrained

mattroks101 (approved):

code/game/objects/items/weapons/storage/deferred.dm
```diff
@@ -90,6 +90,12 @@
 	icon_state = "box_serbian"
 	initial_contents = list(/obj/item/device/synthesized_instrument/trumpet = 1) //TODO: Add an accordian to this, sprites already made.
 
+/obj/item/weapon/storage/deferred/disks //for the beacon, cause we want it priced right and actually functional
+	name = "autolathe disk box"
+	desc = "A small collection of autolathe disks"
+	initial_contents = list(/obj/spawner/lathe_disk = 7)
+	price_tag = 2800
```
Wouju, line 97: I suggest making it cost way way more than 2800

zombiemimic: On the beacon, it’s about 3.3 thousand. I think that’s enough

Rumplenutskin: I'm trying to wrack my brain for the most expensive disk I've ever bought from someone. I think the max I usually pay is 1-2k. So 3.3 sounds fair.

GrayRachnid: This is 7 disks with some of them having the possibility of being an extremely rare one that is blacklisted from spawning in loot. The fact alone that this "lootbox" costs so little when it has the ability to print out an AMR, a Dallas or some other godforsaken gun we haven't thought about is the issue, not how much you pay for disks. You can print 5-10 guns out from a disk, that's worth more than just the disk on its own when the guild has access to minerals.

Iglpigl: Maint loot also has possibility of Dallas and other similar disks, moot point, unless the gotten disks are infinite(excelsior ones) the 5-10 number is absolutely false and  is 4 for most disks, regarding access to minerals Moebius has a barge that can go to the asteroid too, not a problem, the current price is absolutely alright especially considering the 200 credit shipping fee.

code/modules/trade/datums/trade_stations_presets/zarya.dm
```diff
@@ -39,6 +39,7 @@
 			/obj/item/weapon/tool/omnitool,
 			/obj/structure/reagent_dispensers/fueltank,
 			/obj/machinery/floodlight,
+			/obj/item/weapon/storage/deferred/disks, //it returns
```
AlexMorgan3817, line 42:
```suggestion
			/obj/item/weapon/storage/deferred/disks = custom_good_amount_range(list(-2, 3)),
```
No comments and less (or zero) count of loot boxes for half of hour

code/game/objects/items/weapons/storage/deferred.dm
MLGTASTICa: > This is 7 disks with some of them having the possibility of being an extremely rare one that is blacklisted from spawning in loot. The fact alone that this "lootbox" costs so little when it has the ability to print out an AMR, a Dallas or some other godforsaken gun we haven't thought about is the issue, not how much you pay for disks. You can print 5-10 guns out from a disk, that's worth more than just the disk on its own when the guild has access to minerals.

Only /obj/spawner/lathe_disk/advanced can spawn  those , basic ones cannot.

Rumplenutskin: Also let's not forget the lathe nerfs. A new disk can print 4 guns, but those gun will come out funky. And if it's a gun with a unique mag you are going to have to waste a ton of license points since mags no longer print fully loaded.

code/modules/trade/datums/trade_stations_presets/zarya.dm
```diff
@@ -39,6 +39,7 @@
 			/obj/item/weapon/tool/omnitool,
 			/obj/structure/reagent_dispensers/fueltank,
 			/obj/machinery/floodlight,
+			/obj/item/weapon/storage/deferred/disks,
```
Firefox13, line 42:
```suggestion
			/obj/item/weapon/storage/deferred/disks
```

AlexMorgan3817: Why

kappa-sama: for reference, the last item in a list does not use a comma because the presence of a comma indicates that the list should continue

AlexMorgan3817: > for reference, the last item in a list does not use a comma because the presence of a comma indicates that the list should continue

If it was c++ or other it would indicate, but it's DM, and it's allow this. And this won't create extra null element if it is last comma in list and this comma haven't anything after it.

Better to leave commas on last elements because we have down streams and this commas removing many problems with merge conflicts if someone on downstream will decide to add something to list

TheShown911 (approved):
This PR has twice the number of comments then the number of lines of code changed, I don't think it's going to be too strong so I approve. Maybe would put it on a station that's not available roundstart so you would need to do a bit of work before you get the disks, if disks weren't so easy to get in the first place.

code/game/objects/items/weapons/storage/deferred.dm
TheShown911: >  especially considering the 200 credit shipping fee.

The shipping fee is the price of the crate, it's just there so you don't order and export unlimited crates.

Firefox13:
The problem with the disks in beacon was because before there was the random spawner put into it, which cost 30 or so credits.
With the price adjusted it it'll be less of a problem until Guild finds yet another exploit to gain millions of credits.

PR #6263, opened 2021-05-26
Gives moebius maint access

Sparrowbird611:
## About The Pull Request

Gives all moebius roles maintenance access.

## Why It's Good For The Game

Everyone else on the ship has maintenance access, and recent PRs have added it to most of the ones that didn't. I feel its about time moebius received this treatment as well. Given they can already get it quite easily, I don't feel this is a problem and see no reason to withhold it from them.

## Changelog
:cl:
add: Added maintenance access to all moebius roles.
/:cl:

### DIFF
--- a/code/game/jobs/job/medical.dm
+++ b/code/game/jobs/job/medical.dm
@@ -75,7 +75,7 @@ Your second loyalty is to your career with Moebius corp, and to your coworkers i
 	outfit_type = /decl/hierarchy/outfit/job/medical/doctor
 
 	access = list(
-		access_moebius, access_medical_equip, access_morgue, access_surgery, access_chemistry, access_virology,
+		access_moebius, access_medical_equip, access_maint_tunnels, access_morgue, access_surgery, access_chemistry, access_virology,
 		access_genetics
 	)
 
@@ -136,7 +136,7 @@ Your second loyalty is to your career with Moebius corp, and to your coworkers i
 	outfit_type = /decl/hierarchy/outfit/job/medical/chemist
 
 	access = list(
-		access_moebius, access_medical_equip, access_morgue, access_surgery, access_chemistry, access_virology
+		access_moebius, access_medical_equip, access_maint_tunnels, access_morgue, access_surgery, access_chemistry, access_virology
 	)
 
 	stat_modifiers = list(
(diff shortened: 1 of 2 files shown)


Omni44:
The reason club worker got that treatment was Manager having it, while they hadn't, which was weird, inconsistent and detriment further keeping people away from playing this omewhat underdeveloped role.
I don't see any good reason for giving moebius this. It's basically only downside of playing moebius is lack of maint access, since they are coddled in every other way, and I can see neither lore reason for them to have it nor benefit in game from enabling them to joonk too. Also makes paramedic even more of a joke than it is. Just because about anyone can get spare quite easily, doesn't mean we should give everyone All Access.

Sparrowbird611:
> The reason club worker got that treatment was Manager having it, while they hadn't, which was weird, inconsistent and detriment further keeping people away from playing this omewhat underdeveloped role.
> I don't see any good reason for giving moebius this. It's basically only downside of playing moebius is lack of maint access, since they are coddled in every other way, and I can see neither lore reason for them to have it nor benefit in game from enabling them to joonk too. Also makes paramedic even more of a joke than it is. Just because about anyone can get spare quite easily, doesn't mean we should give everyone All Access.

MEO has maintenance access, while his followers lack it, which is weird, inconsistent, and detrimental, and MEO is the only head in the game to lack maintenance access, which is definitely weird, inconsistent, and detrimental. I don't see a reason for this to be the case.

If we're using this as a balancing mechanic it's a crappy one because you can get maintenance access in the first five minutes if you just get a permanent maint guest card or take an ID later in the game. I see it as a good thing to give moebius. If you got a problem with how moebius is "coddled," I recommend you suggest or make changes to fix it, rather than rely on this "downside" that is literally just an inconvenience.

I'm not gonna lie most of the moebius medical roles are kind of just extras outside of doctor, used to help people specialize and change their stats. Paramedic is just there so you can get a boost to VIG and TGH, and to designate yourself as the one who will definitely be leaving medical to rescue patients, instead of everyone rushing out to do it.

MLGTASTICa:
Moebius power creep

mattroks101:
All departments should get maint access 😎

Sparrowbird611:
> All departments should get maint access 😎

Chad

Firefox13 (approved):

PR #6272, opened 2021-05-28
Erismed 3 Knockout time nerf and self surgery buff

TheShown911:
## About The Pull Request

Lowered the knockout time from 2-3 minutes to a flat 1.5 minutes. Makes self surgery possible with either tramadol + paracetamol or oxycodone + inaprov.

## Why It's Good For The Game

Should have just made knockout play admin "man up, deal with it, move on", but this is an alternative solution.

## Changelog
:cl:
balance: pain knockout duration lowered to 1.5 minutes
balance: self surgery now possible by stacking 2 painkillers
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/shock.dm
+++ b/code/modules/mob/living/carbon/shock.dm
@@ -129,7 +129,7 @@
 		enter_hard_crit()
 
 /mob/living/carbon/human/proc/enter_hard_crit()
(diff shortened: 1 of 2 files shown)


Chaoyote:
1.5 minutes is still a massive amount of time. I'd suggest reducing it to under a minute at minimum and then having it go lower, to a maximum, based on toughness (NOT robustness)

mattroks101 (approved):
Step in the right direction.

PR #6276, opened 2021-05-29
Lexorin Spider Tweak

Chaoyote:
I'm putting this up again for the following reason:

- The first PR was clustered with misinformation and wrongful assumptions.
- If this doesn't do anything, then it doesn't cost anyone anything. It will purely be for my peace of mind and the peace of mind of everyone who agrees. You could still stack two spiders for the same effect if you really want to abuse this, it just becomes a tiny bit less ezkill.
- If the spider is eventually fixed from however you believe it to be broken, then you will want this change eventually.
- Yes, it is a bunch of one-line number changes. What do you expect, a complete rework of carrion accompanying this? My other PR was responded to with "Split this into 2 changes" so this clearly isn't a real concern you have.

Thank you.

## Changelog
:cl:
balance: Rebalances Lexorin spiders, 100u is too much.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/implant/implants/carrion/toxic_spider.dm
+++ b/code/game/objects/items/weapons/implant/implants/carrion/toxic_spider.dm
@@ -22,10 +22,10 @@
 
 /obj/item/weapon/implant/carrion_spider/toxicbomb/proc/prime()
 	var/location = get_turf(src)
-	gas_storage = new /datum/reagents(100, src)
-	gas_storage.add_reagent("lexorin", 100)
+	gas_storage = new /datum/reagents(50, src)
+	gas_storage.add_reagent("lexorin", 50)
 	var/datum/effect/effect/system/smoke_spread/chem/S = new
 	S.attach(location)
-	S.set_up(gas_storage, 10, 100, location)
+	S.set_up(gas_storage, 10, 50, location)
 	S.start()
 	die()


GrayRachnid:
The lexorin spiders have 100units because the gas spreads it out. This leads to 5units per tile. Your change will lead the gas to have 2.5units per tile, not enough to even bother anyone.

TheShown911:
The section of the PR titled "Why It's Good For The Game" is strangely missing, you might have made a mistake.

Chaoyote:
I'll grab some screenshots.

Chaoyote:
image
Yeah no, they definitely work. I did notice that they wouldn't work in places where smoke would be sucked down by different elevations.

One spider **will kill** a target **and those around them from full health** with some certainty after a time. 5 units is way more than enough for an AOE ability that can be activated from 4 floors down on the opposite end of the station whenever you choose.

I don't really get why this uses Lexorin in particular. To be quirky and unique?

There is definitely a way to make someone take the full dose. I've seen it before in game months ago (led to dialysing about a hundred lexorin out of someone) and I saw it pretty recently in a round too. I'll do more testing.

GrayRachnid:
> There is definitely a way to make someone take the full dose. I've seen it before in game months ago (led to dialysing about a hundred lexorin out of someone) and I saw it pretty recently in a round too. I'll do more testing.

The one from months ago has already been patched.

Chaoyote:
As I said, I saw it the other round. I'm not entirely sure how it's done, if it's RNG, if it's an exploit of some kind. Still, the spider doesn't need 100u.

GrayRachnid:
> The lexorin spiders have 100units because the gas spreads it out. This leads to 5units per tile. Your change will lead the gas to have 2.5units per tile, not enough to even bother anyone.

Chaoyote:
> > The lexorin spiders have 100units because the gas spreads it out. This leads to 5units per tile. Your change will lead the gas to have 2.5units per tile, not enough to even bother anyone.

And I'm saying 2.5 units will bother someone without bringing them from full health to dead.

GrayRachnid:
I've gone ahead and done some of my own tests, the issue isn't that there's 100/5 lexorin, it's that ingested lexorin has a metabolism rate of 0.01, Even with just 1 unit, you will pass out and almost die.

Chaoyote:
There's some interaction with Lexorin that makes you instantly pass out as well, but I can't replicate it.

mattroks101:
This is not getting merged. Don't open it again.

PR #6284, opened 2021-05-31
a button that lets you set sheets_per_tick on the ore processor

Hatterhat:
## About The Pull Request
adds a button to the ore processing console that just lets you set the sheets_per_tick, maximum 60. this also sets how many ores the unloading machine can pull out of an ore box at once
## Why It's Good For The Game
i hate waiting for ore to process and unload
## Changelog
:cl:
add: The ore processing console now has a button that lets you set how many sheets it can process per processing cycle. Maximum 60 sheets per tick.
/:cl:

### DIFF
--- a/code/modules/mining/machine_processing.dm
+++ b/code/modules/mining/machine_processing.dm
@@ -35,7 +35,7 @@
 	user.set_machine(src)
 
 	var/dat = "<h1>Ore processor console</h1>"
-
+	dat += "High-speed processing is <A href='?src=\ref[src];toggle_speed=1'>[(machine.speed_process ? "<font color='green'>active</font>" : "<font color='red'>inactive</font>")].</a>"
 	dat += "<hr><table>"
 
 	for(var/ore in machine.ores_processing)
@@ -70,6 +70,8 @@
 		return 1
 	usr.set_machine(src)
 
+	if(href_list["toggle_speed"])
+		machine.toggle_speed()
 	if(href_list["toggle_smelting"])
 		var/choice = input("What setting do you wish to use for processing [href_list["toggle_smelting"]]?") as null|anything in list("Smelting","Compressing","Alloying","Nothing")
 		if(!choice) return
@@ -108,6 +110,8 @@
 	var/list/ores_stored
 	var/static/list/alloy_data
 	var/active = 0
+	var/speed_process = 0 // are we on SSfastprocess?
+	var/tick = 0 // if we're on SSfastprocess, how many processes have we done? once we hit 10 processes then we update the linked console UI
 	var/input_dir = 0
 	var/output_dir = 0
 
@@ -140,10 +144,32 @@
 		if(marker)
 			output_dir = get_dir(src, marker)
 
+/obj/machinery/mineral/processing_unit/proc/toggle_speed(var/forced)
+	var/area/refinery_area = get_area(src)
+	if(forced)
+		speed_process = forced
+	else
+		speed_process = !speed_process // switching gears
+	if(speed_process) // high gear
+		STOP_PROCESSING(SSmachines, src)
+		START_PROCESSING(SSfastprocess, src)
+	else // low gear
+		STOP_PROCESSING(SSfastprocess, src)
+		START_PROCESSING(SSmachines, src)
+	for(var/obj/machinery/mineral/unloading_machine/unloader in refinery_area.contents)
+		unloader.toggle_speed()
+	for(var/obj/machinery/conveyor_switch/cswitch in refinery_area.contents)
+		cswitch.toggle_speed()
+	for(var/obj/machinery/mineral/stacking_machine/stacker in refinery_area.contents)
+		stacker.toggle_speed()
+
 /obj/machinery/mineral/processing_unit/Process()
 
 	if(!output_dir || !input_dir)
 		return
+
+	if(speed_process)
+		tick++
 
 	var/list/tick_alloys = list()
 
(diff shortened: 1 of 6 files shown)


Jackofashepard:
Is there any downside to this or is it just cheatmode speedup button?

code/modules/mining/machine_processing.dm
```diff
@@ -108,6 +110,8 @@
 	var/list/ores_stored
 	var/static/list/alloy_data
 	var/active = 0
+	var/speed_process = 0 // are we on SSfastprocess?
```
vode-code, line 113:
```suggestion
	var/speed_process = FALSE // are we on SSfastprocess?
```
you should probably use FALSE instead of just 0, like you did for the rest.

Hatterhat:
> 
> 
> Is there any downside to this or is it just cheatmode speedup button?

considering that the machinery runs 5x faster than usual, it draws 5x more power
that and the speed at which you process stuff may lead to an overproduction of things or running out of that one material you were alloying with

otherwise, uh - not really, no

Rumplenutskin:
One thing I would like to remind people is that the mining drills were buffed relatively recently to produce almost double material but the processor stayed the same. Mining already takes forever and hardly anyone does it. I'm not sure if this needs a downside.

L74814:
Not only drills were buffed, but also the minerals around the asteroids aswell. I would suggest just increasing the actual speed, since this smelter/ore processing machine its practically impossible to reach to upgrade without some real fuckery. And give the "fast-mode button" some downside, maybe consuming double materials? idk

L74814:
This also doesn't affect other conveyors and machinery that can be builded/exists already outside the Hulk? Or these are special conveyors? I don't remember, tbh. Wouldn't be weird if somehow all conveyors in guild started going x2 speed randomly because Miners are using this?

Hatterhat:
> This also doesn't affect other conveyors and machinery that can be builded/exists already outside the Hulk? Or these are special conveyors? I don't remember, tbh. Wouldn't be weird if somehow all conveyors in guild started going x2 speed randomly because Miners are using this?

it only applies to the hulk machines because there's an area check - that makes sure that only the machines in the same area (the hulk) get the processing bonus

Firefox13:
Wouldn't this just bog down the server if tou leave it on?

MLGTASTICa:
Why let conveyor and the stacking machine use fast process ? those don't need it , and fast processing looping conveyors can actually kill performance of the server.

Hatterhat:
> 
> 
> Wouldn't this just bog down the server if you leave it on?


i...really hope people don't just leave the processor on and/or are responsible enough to turn it off when they're not using it, but if it's that much a concern i can go ahead and throw in a timer that automatically turns off the fast processing mode in a minute or something

> 
> 
> Why let conveyor and the stacking machine use fast process ? those don't need it , and fast processing looping conveyors can actually kill performance of the server.

the stacker probably doesn't need fast processing, but the conveyors need fast processing so they can actually keep up with the unloaded ores/processed sheets coming out of the machinery - hopefully it shouldn't be too much a performance hit because it's literally only the handful of conveyors on the Hulk itself

drexample:
@Hatterhat Have you considered just
Increasing the amount of sheets the machine produces in a stack instead of this?

Hatterhat:
> @Hatterhat Have you considered just
> Increasing the amount of sheets the machine produces in a stack instead of this?
> Instead of making a separate redundant subsystem that ticks faster exclusively for this machine and then hotswap processing between it and normal processing subsystem?
> There's literally a var in processing machine called "sheets_per_tick". It does exactly that.

the thing is that ten sheets per processing tick, i think, is a nice enough round number that i don't particularly want to mess with
that, and increasing the amount of sheets getting ejected onto conveyors might make the conveyors worse when it comes to calling step_glide

if the maintainers just want me to do that instead, though, i mean shit i guess i could. i wouldn't be happy with it, but i could.

drexample:
@Hatterhat What? What are you talking about? Sheets in a stack are not separate items. Increasing the value increases the amount of sheets in the stack, not the amount of items created.

Hatterhat:
fair enough

Firefox13:
Couldn't you just simply increase the amount of sheets per stack it makes at this point without ant buttons at this point?

Hatterhat:
> Couldn't you just simply increase the amount of sheets per stack it makes at this point without any buttons at this point?

honestly the ore processor as it is now is just. fucking awful:

- the processor shits out PILES OF INDIVIDUAL SHEETS instead of stacks with amounts adjusted
- alloying doesn't respect sheets per tick and instead only makes enough alloyed material for one (1) full unit of it,
- you have to manage how much material you have processing
- else you compress all your plasma when you meant to make some plasteel, or you smelted all your carbon into plastic when you suddenly need some steel or some shit

sure the fast processor was a terrible idea in hindsight for a codebase as glued together as this is, but god damn this whole ore processing cycle is just fucking imprecise and awful as it is now

Hatterhat:
a port or an equivalent to the ORM from /tg/code could not fucking come soon enough

Iglpigl:
> a port or an equivalent to the ORM from /tg/code could not fucking come soon enough

Port it then.

GrayRachnid:
> > a port or an equivalent to the ORM from /tg/code could not fucking come soon enough
> 
> Port it then.

Frankly I'd rather keep a more functioning version of the interactive smelter than a boring machine that takes all the ores in and spits them out.

The issue with sheets_per_tick is that is completely ignored alloying. I'd that's fixed or the machine is refacyored from the ground up then it wouldn't have issues.

bigpep:
I'd use clamp instead, not sure if there'd be any noteworthy difference.

Hatterhat:
i kinda forgot clamp existed, actually, good point

PR #6285, opened 2021-05-31
Reclaims the Civilian Account for the Club

zombiemimic:
## About The Pull Request
Gives the Civilian Account to the club and adds a new account for Offship vendors

## Why It's Good For The Game
Gives a fund so the club is not caught with its pants down when starting

## Changelog
:cl:
add: Added Offship account
tweak: Changed around some vendor accounts
/:cl:


code/game/jobs/department.dm
```diff
@@ -126,3 +125,7 @@
 	account_initial_balance = 7500
 	funding_type = FUNDING_NONE
 
+/datum/department/offship //So we can pay the Club without giving them independant money
+	name = "Offship entities"
+	id = DEPARTMENT_OFFSHIP
+	account_budget = 0
```
Iglpigl, line 131: No newline.

Iglpigl (changes requested):

code/game/jobs/department.dm
zombiemimic: Will take care of in the morning

Firefox13:
Where the snack vendors are at? You understand that we've removed them all entirely? Putting them back even with club accounts linked will only aggravate the issue of it not being visited.

Sparrowbird611:
You should put a single one in the club IMO, that way people know the club always has food.

Firefox13 (approved):

PR #6286, opened 2021-05-31
Nerfs carrion maw's Consume Flesh and cleans it up a bit

vode-code:
## About The Pull Request
This PR changes the Carrion's Maw's Consume Flesh ability in these ways:

- it now is unable to be used on victims that are in a grab weaker than an aggressive grab.
- it now is unable to be initiated while being used.
- it now is unable to tear groin off.
- it now will tear an organ out of chest or groin.
- it now will backlash with a Stun(5) and item drop when targeting head without all limbs being disabled and without the victim being unconscious.
- it now has better messages.

## Why It's Good For The Game
The Maw is supposed to be a tool for advancement and disabling downed opponents, not something that you can use to just instakill whoever you get into a grab for five seconds. This PR makes Consume Flesh an utility ability, instead of an OP weapon.

## Changelog
:cl:Chickenish
add: The carrion's Maw now damages and removes a random organ when the target is the chest.
balance: The carrion's Maw now also does the same when the groin is targeted instead of ripping it off, can't initiated while in progress, does not work with a weak grab, and will stun you if you try to tear off your victim's head without KOing them or disabling all their limbs first.
spellcheck: There was extra "the"s in the maw limb tear off ability, no more.

/:cl:

### DIFF
--- a/code/modules/organs/internal/carrion.dm
+++ b/code/modules/organs/internal/carrion.dm
@@ -292,29 +292,40 @@
 			return
 		if(istype(H))
 			var/obj/item/organ/external/E = H.get_organ(owner.targeted_organ)
-			if (!isnull(target_to_tear) && target_to_tear != E)
-				to_chat(owner, SPAN_WARNING("Your maw is already focused on tearing something else off."))
+			if (!isnull(target_to_tear)) // one at a time, thank you.
+				to_chat(owner, SPAN_WARNING("Your maw is already focused on something."))
 				return
-			target_to_tear = E
 
 			if(E.is_stump())
 				to_chat(owner, SPAN_WARNING("You can't tear off a limb stump"))
 				return
+			target_to_tear = E // here is where it chooses what the maw is "focused on".
 
-			visible_message(SPAN_DANGER("\The [owner] bites into \the [H]'s \the [E] and starts tearing it off!"))
+			visible_message(SPAN_DANGER("[owner] bites into [H]'s [E.name] and starts tearing it [E.functions ? "off": "apart"]!"))
 			if(do_after(owner, 5 SECONDS, H))
 				target_to_tear = null
-				if(E.organ_tag == BP_GROIN) // would be able to tear off both legs at once in one consumption otherwise.
+				if (E.organ_tag == BP_HEAD)
+					if(!(H.incapacitated(INCAPACITATION_UNCONSCIOUS)))
+						var/datum/gender/G = gender_datums[H.gender]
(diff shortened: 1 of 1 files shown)


GrayRachnid:
> it now will tear an organ out of chest or groin.

considering all our limbs have organs, why limit to chest or goin?

code/modules/organs/internal/carrion.dm
```diff
@@ -286,18 +287,48 @@
...
+									visible_message(SPAN_DANGER("[H] smacks away [owner]'s maw with [G.his] [smacker.name]."))
+									return
+			
+				if (E.organ_tag == BP_GROIN || E.organ_tag == BP_CHEST) // torso bits should not be torn off.
+					E.take_damage(30)
+					var/obj/item/organ/internal/organ_to_remove = pick(E.internal_organs)
+					organ_to_remove?.loc = get_turf(src)
+					E.internal_organs -= organ_to_remove
+					visible_message(SPAN_DANGER("[owner] tears \a [organ_to_remove] out of [H]'s [E.name]!"))
+					playsound(loc, 'sound/voice/shriek1.ogg', 50)
```
GrayRachnid, line 324: This already exists 3 lines below unless it doesn't proc

vode-code: if you had included one more line, you would have seen the return.

GrayRachnid: me dum

vode-code:
> 
> 
> > it now will tear an organ out of chest or groin.
> 
> considering all our limbs have organs, why limit to chest or goin?

everything else has a joint to rip off, and the ability to remove arms so you can RP with your victim before whatever else happens is one I've found fun as the victim.

GrayRachnid:
> 
> 
> > > it now will tear an organ out of chest or groin.
> > 
> > 
> > considering all our limbs have organs, why limit to chest or goin?
> 
> everything else has a joint to rip off, and the ability to remove arms so you can RP with your victim before whatever else happens is one I've found fun as the victim.

Ah so legs and arms still get ripped off but the groin and chest do not. I misunderstood and assumed you took ribbing limbs off out completely. Also, why not do the same thing for the head? I'm pretty sure we have plenty of other organs in there not counting the brain and skull.

Also, make sure to black list the bones from torso and groin because removing bones destroys the bodypart.

Iglpigl:
> it now gets a third of the gene points from decayed organs.
Considering bones rot/decay too this sounds a like way too big of a change.

Firefox13 (approved):

PR #6295, opened 2021-06-04
Adds energy gun cell mount, but not as overpowered

zombiemimic:
## About The Pull Request
as before, but with much higher power cost and fire delay

## Why It's Good For The Game
gives a new tool for energy gun users.

## Changelog
:cl:
add: Added energy gun cell mount
code: changed var location for cell used
/:cl:


Firefox13:
You understand that 30+ shots for a laser cannon with this mod will still be abyssmally high?

zombiemimic:
That's why I also added the fire delay

Omni44:
Wih 15k cell that's 30 shots without shunt or excruciator. Given that it's rather large cell you won't always get your hand on, most important eguns got nerfed and (iirc) all eguns had additional pain effect removed that's barely enough to make them worth using again, wit that mod that is.

Firefox13:
Unfortunately there aren't really much ways to properly balance this, the terrible amount of charge you get with L-cells is gonna outweigh any downside short of making gun unusable. 
*dabs with 15000L cell with a laser cannon in distance*

PR #6315, opened 2021-06-10
Lesser Forms can't be cruciformed anymore.

L74814:
## About The Pull Request

something something, makes lesser form not being able to be cruciformed and if they somehow manage to get it and activate it on, it should gibs then.

I have to say that I initially added mutated people to the gibbing one aswell, but noticed that if you get a breakdown it could fuck you aswell plus heterocromia is a thing. So no.
 
So the check is on nonhuman species & some funny carbons.

This count as my fix n19
## Why It's Good For The Game

It was funny for players, it wasn't funny for jannie.

And this count as a bugfix.
## Changelog
:cl: HerBaon
balance: Only glorious humans can be part of NT. Everyone else gets gibbed or isn't worthy anymore.
/:cl:

### DIFF
--- a/code/__DEFINES/is_helpers.dm
+++ b/code/__DEFINES/is_helpers.dm
@@ -33,8 +33,6 @@
 #define isbrain(A) istype(A, /mob/living/carbon/brain)
 
 #define ishuman(A) istype(A, /mob/living/carbon/human)
-
-#define ismonkey(A) istype(A, /mob/living/carbon/human/monkey)
 //---------------------------------------------------
 
 #define isanimal(A) istype(A, /mob/living/simple_animal)
--- a/code/__DEFINES/is_helpers.dm
+++ b/code/__DEFINES/is_helpers.dm
@@ -33,6 +33,8 @@
 #define isbrain(A) istype(A, /mob/living/carbon/brain)
 
 #define ishuman(A) istype(A, /mob/living/carbon/human)
+
+#define ismonkey(A) istype(A, /mob/living/carbon/human/monkey)
 //---------------------------------------------------
 
 #define isanimal(A) istype(A, /mob/living/simple_animal)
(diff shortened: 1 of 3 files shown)


granodd:
🙏 amen

tested for any odd situations to make sure that all humans can be cruciformed without issue, I assume?

L74814:
I tested it and only monkeys, golems and human slimes gibbed when activating the cruciform (epiphany) and the first one required to be introduced through surgery, the other ones work as carrion does, at activation it gibs the holder of the activated cruciform.

Normal human were able to be cruciformed normally and it checks if your species are Human. Though I don't know if there's any odd situation where you could have a nonhuman species and still being cruciformed.

While the other nonhuman carbons as Roaches, normal slimes, mouses, etc can't have a installed cruciform, and since you can't do surgery on these, you shouldn't be able to surgically insert the cruciform to skip this check.

Firefox13:
> and the first one required to be introduced through surgery, the other ones work as carrion does, at activation it gibs the holder of the activated cruciform.

Might want to do something about cruciforming&gibbing monkeys as it gives plenty of energy to EOTP.

L74814:
> > and the first one required to be introduced through surgery, the other ones work as carrion does, at activation it gibs the holder of the activated cruciform.
> 
> Might want to do something about cruciforming&gibbing monkeys as it gives plenty of energy to EOTP.

Right. I forgot that people clearly wouldn't take advantage of this

code/__DEFINES/is_helpers.dm
```diff
@@ -33,6 +33,8 @@
 #define isbrain(A) istype(A, /mob/living/carbon/brain)
 
 #define ishuman(A) istype(A, /mob/living/carbon/human)
+
+#define ismonkey(A) istype(A, /mob/living/carbon/human/monkey)
```
TheShown911, line 37: Checking for monkeys should be done via a species, not a type check.

TheShown911 (approved):

PR #6324, opened 2021-06-12
Frag Grenade tweak

Sparrowbird611:
## About The Pull Request

Tweaks frag grenades. They now only put out 50 shrapnel instead of 150, do 15 damage instead of 5 damage, and now have 15 AP. I also tweaked the weak shrapnel that the shellshock fires, it deals 15 damage and 10 AP instead. I've also removed their ability to embed.

## Why It's Good For The Game

Frag grenades are kind of in a weird place right now. To anyone wearing decent armor, they deal practically no damage, but fill people with so much shrapnel a doctor will take one look at you and commit suicide. I'm hoping this will make grenades a bit more useful in terms of direct damage, while making their after affects less annoying to deal with.

## Changelog
:cl:
tweak: Frag grenades put out less shrapnel, but deal more damage and now have AP. This applies to pommies, china lake grenades, and shellshock grenades.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/grenades/frag.dm
+++ b/code/game/objects/items/weapons/grenades/frag.dm
@@ -6,8 +6,9 @@
 	loadable = TRUE
 
 	var/fragment_type = /obj/item/projectile/bullet/pellet/fragment/strong
-	var/num_fragments = 150  //total number of fragments produced by the grenade
-	var/fragment_damage = 5
+	var/num_fragments = 50  //total number of fragments produced by the grenade
+	var/fragment_damage = 15
+	var/fragment_penetration = 15
 	var/damage_step = 2      //projectiles lose a fragment each time they travel this distance. Can be a non-integer.
 
 	//The radius of the circle used to launch projectiles. Lower values mean less projectiles are used but if set too low gaps may appear in the spread pattern
@@ -42,5 +43,3 @@
 	icon_state = "frag_nt"
 	item_state = "frggrenade_nt"
 	matter = list(MATERIAL_BIOMATTER = 75)
-	fragment_damage = 7
-	damage_step = 3
--- a/code/modules/projectiles/projectile/projectilegrenades.dm
+++ b/code/modules/projectiles/projectile/projectilegrenades.dm
@@ -42,8 +42,9 @@
 	name = "frag shell"
 	var/range = 7
 	var/f_type = /obj/item/projectile/bullet/pellet/fragment/strong
-	var/f_amount = 100
-	var/f_damage = 4
+	var/f_amount = 50
+	var/f_damage = 15
+	var/f_penetration = 15
 	var/f_step = 2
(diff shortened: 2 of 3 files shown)


FortuneMaster:
reducing the amount of projectiles will also reduce the lag

thedarkcrusader:
> reducing the amount of projectiles will also reduce the lag

Nothing reduces lag Fortune son

Sparrowbird611:
> reducing the amount of projectiles will also reduce the lag

Primary reason I even started this was because someone told me the 150 projectiles were causing lag spikes.

TheShown911 (commented):
Have you tested the damage on this on armored/unarmored targets? I think the added damage can make a big difference due to GDR.

Sparrowbird611:
> Have you tested the damage on this on armored/unarmored targets? I think the added damage can make a big difference due to GDR.

I have. I wasn't sure how effective they should be compared to guns, but they can do some decent damage if you're near the grenade. My only concern is that anyone who stands directly on a grenade gets massively fucked up, which might be what we want, but I'd be hard pressed to change it.

Firefox13 (approved):

code/game/objects/items/weapons/grenades/frag.dm
```diff
@@ -6,8 +6,9 @@
 	loadable = TRUE
 
 	var/fragment_type = /obj/item/projectile/bullet/pellet/fragment/strong
-	var/num_fragments = 150  //total number of fragments produced by the grenade
-	var/fragment_damage = 5
+	var/num_fragments = 50  //total number of fragments produced by the grenade
+	var/fragment_damage = 15
+	var/fragment_penetration = 15
```
Firefox13, line 11: Where this variable is used?

code/modules/projectiles/projectile/projectilegrenades.dm
```diff
@@ -42,17 +42,19 @@
 	name = "frag shell"
 	var/range = 7
 	var/f_type = /obj/item/projectile/bullet/pellet/fragment/strong
-	var/f_amount = 100
-	var/f_damage = 4
+	var/f_amount = 50
+	var/f_damage = 15
+	var/f_penetration = 15
```
Firefox13, line 47: Where this variable is used?

code/game/objects/items/weapons/grenades/frag.dm
Sparrowbird611: I don't understand which variable you mean, most of these were already here except for fragment_penetration.

Firefox13: I'm indeed talking about it. It isn't used anywhere. Did you test your PR? It will cause an error or atleast a warning when compiling.

drexample (changes requested):
I'm not sure this even compiles

PR #6326, opened 2021-06-12
Cleans up body organ scan code

vode-code:
## About The Pull Request
This PR makes the external and internal organ parts of body scans use an other_wounds list instead of a lot of variables, set a variable for unknown body instead of add to it since it's only used in a boolean format, and do nothing on the infection switch if there is no significant infection instead of going through all of it.

## Why It's Good For The Game
It looked bad previously, and this should decrease lag a tiny bit, although I haven't tested.

## Changelog
:cl: Chickenish
refactor: refactored some body scanner code, this should also make it look cleaner.
/:cl:


Firefox13:
> although I haven't tested.

Did you just admit that you didn't test your own code? изображение

drexample:
> although I haven't tested.

Then go and test it.

GrayRachnid:
I think he means the lag part

vode-code:
I did in fact mean the lag part, I have gameplay tested all my code, I did not bother to use the profiler however.

Firefox13 (approved):

PR #6333, opened 2021-06-15
Medical Specialist appreciation day

Omni44:
## About The Pull Request

Medbrig now gets its own sleeper (I tried to restrict access to it so that only medspec can apply chems, it didn't really work though for whatever reason), proper surgery&oxyloss kit, organ freezer and a minor refurbishment. 

Removed area of space in bar's walls too, minor oversight would cause shield to pop in there.

31Z

## Why It's Good For The Game

There is, in my opinion, no point in keeping medspec neutered as he is now (since he has to spend considerable time and resources to get nearly close to actual medbay's capabilities). While Medbrig shouldn't get everything medbay does, many people feel like sleeper is bare minimum to actually do any medicine there semi-reliablly. Either that, or give up on medspec entirely.

## Changelog
:cl:
tweak: tweaked birgmed's equipment
fix: fixed a shield tiel appearing in bar
/:cl:

### DIFF
--- a/maps/CEVEris/_CEV_Eris.dmm
+++ b/maps/CEVEris/_CEV_Eris.dmm
@@ -4810,12 +4810,7 @@
 /area/eris/security/exerooms)
 "alp" = (
 /obj/structure/table/glass,
-/obj/item/weapon/tool/cautery{
-	pixel_y = 4
-	},
-/obj/item/weapon/tool/hemostat{
-	pixel_y = 4
-	},
+/obj/item/weapon/storage/freezer/medical,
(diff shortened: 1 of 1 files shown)


Sparrowbird611:
Sleeper is unnecessary IMO but since you already went and did it, why not. I don't think we should be putting strait jackets or blindfolds in there though. Everything else is swag.

Omni44:
Straight jackets and blindfolds have always been there.

Sparrowbird611:
> Straight jackets and blindfolds have always been there.

Wait really? Damn I'm retarded.

Firefox13 (approved):

PR #6336, opened 2021-06-16
Nerf the plastic bag

Firefox13:
## About The Pull Request

Plastic bag is irrederamably bad and shouldn't as much as space as duffel bag even if able to store even small items only for a price of 1 (one) plastic sheet. It can only store as much as five small items. As a trade-off, it can hold a nuclear disk, hurray!

## Why It's Good For The Game

balance

## Changelog
:cl:
balance: Plastic bag doesn't holds as much as duffle bag, but it can store a nuke disk now.
/:cl:


TheShown911 (approved):
Why remove the nuke disk blacklist though?

Firefox13:
We're long past gone that point of nukies stealing the disk, There's kinda no point to it already.

PR #6340, opened 2021-06-17
Mega optimization PR

drexample:
## About The Pull Request

This PR modifies GC in a few places and adds debug procs to trace down references on objects. 

This PR disables hard deletions. This is not a preferable measure and I hope we can find a better way in the process before this PR is merged. Everyone is welcome to research and contribute.

This PR also modifies mobs and the way they handle the environment. Roaches should now sleep when there's no one near at all.

This PR ALSO ports some more modern controllers for machinery and power so we don't die in lags.
Also divides machinery list into a lot of smaller lists because iterating through them is expensive and causes lag.
Also modifies NT obelisk to do the same job but eat less ticks.

SO FAR THE MAIN GC CLOG CULPRIT FROM ITEMS IS OLDIFICATOR

This PR also removes hydrobase from ruin rotation because bots need their own PR to fix their half-dead AI

## Why It's Good For The Game

Garbage collector is what causes like 40% of the lag and MC crashes, if not more. Un-fucking it is necessary to support any population higher than 20.

Machinery system lagged before I arrived to work on this codebase so see its load percentage stay at least in triple digits feels nice.

Optimize your code, goddamit.

## Changelog
:cl:
tweak: Drastically reworks several subsystems for the sake of optimization.
/:cl:


MLGTASTICa:
I think the oldificator component is causing hard dels because it holds onto a reference on the item , but doesn't null it themselves.

MLGTASTICa:
> I think the oldificator component is causing hard dels because it holds onto a reference on the item , but doesn't null it themselves when said items get deleted

code/controllers/subsystems/garbage.dm
```diff
@@ -269,12 +308,9 @@ SUBSYSTEM_DEF(garbage)
 
 
 	if(isnull(D.gc_destroyed))
-		if (SEND_SIGNAL(D, COMSIG_PARENT_PREQDELETED, force)) // Give the components a chance to prevent their parent from being deleted
-			return
 		D.gc_destroyed = GC_CURRENTLY_BEING_QDELETED
 		var/start_time = world.time
 		var/start_tick = world.tick_usage
-		SEND_SIGNAL(D, COMSIG_PARENT_QDELETING, force) // Let the (remaining) components know about the result of Destroy
```
LetterN, line 277: do not remove this wtf, this is used for dereferencing components or objects

LetterN (changes requested):

LetterN:
also disabling harddel is bad because you're causing more stuff to NOT get deleted, which means more wasted memory.

LetterN:
also here are fucks who keep clogging the gc https://cdn.discordapp.com/attachments/299262993617780737/855273365509636116/qdel.log

code/controllers/subsystems/garbage.dm
drexample: That comsig doesn't do anything.

drexample: Whoever ported it did half a job and had a full smile.

TheShown911:
Were the find_refs verbs locked behind permissions?

TheShown911:
With burrows and some oldificator bad dels fixed, and with many other optimizations, wouldn't it be better to enable hard dels and fix things that cause them?

AlexMorgan3817:
Fuck, time to optimize kibarspaise

AlexMorgan3817:
Fuck, time to optimize kibarspaise

drexample:
> Were the find_refs verbs locked behind permissions?

Yes!

> With burrows and some oldificator bad dels fixed, and with many other optimizations, wouldn't it be better to enable hard dels and fix things that cause them?

While I caught quite a bit of failed qdel cases, there's still a lot of unresolved ones, so skipping harddels so far wouldn't cause much trouble.

code/modules/scrap/oldificator.dm
```diff
@@ -375,4 +375,4 @@
 			var/obj/item/trash_mod = new trash_mod_path
 			if(SEND_SIGNAL(trash_mod, COMSIG_IATTACK, src, null))
 				break
-			qdel(trash_mod)
\ No newline at end of file
+			QDEL_NULL(trash_mod)
```
TheShown911, line 378: Newline.

TheShown911 (approved):
Skimmed over it, didn't notice anything wrong.

ACCount12:
This is a fucking mess, but some of the measures at least look sane. Down the hatch it goes, then.

PR #6351, opened 2021-06-25
Adds Gunsmith oddity perk and rebalances Gun Assemblies

vode-code:
## About The Pull Request
This PR adds the gunsmith oddity perk, which doubles the amount of ammo gained from ammo kits and adds three options to gun assemblies, reduces the number of options per Mec from 2.5 per ten to one per ten, reduces the number of parts needed from 15 to 10, and makes the maximum price scale based on the option number it is.

## Why It's Good For The Game
Gun Assemblies are currently too potent for the relative ease of obtainment, Gun Assemblies are currently too hard to get meaning that they only show up through luck, dedication, or destruction of an armory, and Gunsmith was planned and fit in this PR.

## Changelog
:cl: Chickenish
add: Adds Gunsmith oddity perk, which increases the amount of ammo from ammo kits and adds options to gun assemblies.
balance: Gun assemblies now have less options available per 10 Mec, cost two thirds as much, and have a maximum price per option based on option number.

/:cl:


code/modules/projectiles/guns/launcher/grenade_launcher.dm
```diff
@@ -174,6 +174,7 @@
 	force = WEAPON_FORCE_PAINFUL
 	matter = list(MATERIAL_STEEL = 20, MATERIAL_WOOD = 10)
 	force = 5
+	price_tag = 500
```
Evankhell561, line 177: why?

Evankhell561 (approved):
except for this change in the price of grenade launchers, everything is perfect.

code/modules/projectiles/guns/launcher/grenade_launcher.dm
vode-code: because the default price is 0, there was no override, and the minimum price for the first option, let alone the rest, is 100.
since the rest of the makeshift guns have a price tag #of 500, and otherwise the makeshift grenade launcher would not be eligible for the gun assembly, I made the price 500.

> 
> 
> why?

because the minimum price for the first option is 100, scaling per option, the default price is 0, and all the other makeshift guns have a price tag of 500; otherwise, the makeshift grenade launcher would not be a possibility for gun assemblies.

TheShown911 (approved):
I would lower the amount of gun parts needed to 10, if the assembly has been sufficiently nerfed.

code/datums/perks/oddity.dm
```diff
@@ -277,4 +283,4 @@
 	if((holder.r_hand && istype(holder.r_hand, /obj/item/weapon/cell)))
 		C = holder.r_hand
 		if(!C.fully_charged())
-			C.give(50)
+			C.give(50)
```
zombiemimic, line 286: newline

PR #6363, opened 2021-06-30
Roach Revamp 1

vode-code:
## About The Pull Request
Roach Gameplay stuff:
Roachlings now remember the friends list on growing up.
Nano-Roach Swarms from Kraftwerk roaches now inherit the friends list when produced.
Roaches eating corpses or laying eggs is now based on realtime checks instead of spawns, this also fixes roaches attempting to lay eggs every cycle and as such doing it effectively all at once.
Roaches now only take 30 seconds to eat a superior_animal such as a spider or roach.
Roach eggs now fit in containers, and will not hatch unless on the ground or in a closet.
Roaches now eat spider eggs instead of failing to lay an egg on a turf with a spider egg, resulting in the roach laying as many eggs as there would have been giant spiders from the spider eggs.
Roach eggs are now able to be laid on a tile with a roach egg.
Instead of a superior_animal being worth rand(1, 2), it is now worth rand(1, meat_amount), which results in roachling and kampfer roaches no longer being profitable to feed to roaches, and otherwise making a bigger superior_animal be worth more eggs.
Roachlings that have eaten a fuhrer roach or kaiser roach now have a chance to become a kaiser roach based on how fed they are.
I removed roaches having their presence spawn roach cubes and I don't know how to add it back, so I made Kaiser roaches lay roach cubes instead of eggs.
Roaches eating other roaches will add the number of eggs the other roach would have laid to its own count.
When grabbed by a friend, a superior_animal will stop doing things.
roaches will no longer immediately bite you after displaying the "hesitates for a moment...and then charges" message.

Gestrahlte roaches have been reworked!
- Gestrahlte roaches now shoot toxic and radioactive projectiles from a distance.
- Gestrahlte Bile now is able to be countered with Bio armor, the melee special is twice as radioactive as the projectile, and the Kaiser's equivalent is five times as radioactive as the melee special, to go with the increased Tox damage.
- Gestrahlte roaches now will not move closer after getting three tiles away, and have two more tiles of acceptable target range than other roaches.
- The Bile projectile cannot hit roaches.

Roach Refactoring stuff:
the taming window var is now in roach.dm instead of riding.dm .
the roach idle decision making process now uses a switch statement.
obj/item/roach was redundant and has now been removed.
instead of using stat, roach AI now uses check_AI_act.

Not Roach stuff:
You can now edit the variables of spawned stuff using a spawner.
lighting_atom.dm no longer redefines atom's new, but has the init_light proc, which is called by atoms.dm's New.
SECOND, MINUTE, HOUR, and DAY defines now exist, they are equivalent to their plural form and exist purely for readability.
All superior_animal that have ranged attacks will now use them if they are not close enough to melee.
It is now possible to reinforce railings, which will prevent the PASSTABLE flag from being enough to move over them, and add some extra health.
It is now possible to shoot railings, although you will need to point-blank them if they aren't reinforced.
railing examine now shows the damaged damage amount instead of only severely damaged and scraped.
Limb examination used to say "unusaly pale" for oxyloss, now it says "unusually pale".

## Changelog
:cl: Chickenish
add: You can now reinforce railings, which means that they block projectiles and roaches, along with some extra health.
add: a roach's friend can now wrangle them for uneventful transport.
add: Gestrahlte roaches are now ranged attackers, and their bile is radioactive.
add: grown up roachlings and nanoroach swarms now inherit their friends lists.
del: Roaches no longer spawn roach cubes with their presence, sorry.
tweak: roaches no longer bite you immediately after you get the message that they hesitate.
balance: roaches now lay more eggs, except for when they ate a roachling or kampfer corpse, then just one.
balance: roaches now take much less time to eat roaches or spiders.
fix: roaches no longer try to lay all the eggs at once.
fix: roach eggs now fit in containers.
imageadd: added reinforced railing icons.
spellcheck: Limb examination used to say "unusaly pale" for oxyloss, now it says "unusually pale".
refactor: refactored a bit of roach code, along with changing an atom/New() override to init_light().
/:cl:


code/modules/mob/living/carbon/superior_animal/roach/types/toxic.dm
```diff
@@ -14,12 +14,40 @@
...
 			playsound(src, 'sound/voice/insect_battle_screeching.ogg', 30, 1, -3)
-			L.visible_message(SPAN_DANGER("\the [src] globs up some toxic bile all over \the [L]!"))
+			L.visible_message(SPAN_DANGER("\the [src] globs up some glowing bile all over \the [L]!"))
+
+/obj/item/projectile/roach_spit
+	name = "Glowing bile"
+	icon = 'icons/obj/hivemind.dmi'
+	icon_state = "goo_proj"
+	damage_types = list()
+	irradiate = 20
```
TheShown911, line 39: Rad damage on basic mobs is a bad idea, it needs very special chemicals to cure.

code/modules/mob/living/carbon/superior_animal/roach/types/toxic.dm
```diff
@@ -14,12 +14,40 @@
...
 	. = ..()
-	if(isliving(A))
-		var/mob/living/L = A
-		if(istype(L) && prob(25))
+	if(prob(25))
+		if(isliving(A))
+			var/mob/living/L = A
 			var/damage = rand(melee_damage_lower, melee_damage_upper)
-			L.damage_through_armor(damage, TOX)
+			L.apply_effect(40, IRRADIATE)
```
TheShown911, line 29: Ditto.

code/game/objects/effects/roaches.dm
```diff
@@ -67,8 +62,11 @@
 			src.loc = O.owner ? O.owner.loc : O.loc
 
 		var/spawn_type = /mob/living/carbon/superior_animal/roach/roachling
-		new spawn_type(src.loc, src)
-		qdel(src)
+		if (isturf(src.loc) || istype(src.loc, /obj/structure/closet)) // suppresses hatching when not in a suitable loc
```
TheShown911, line 65: I'd make this also stop the eggs from growing so the eggs wouldn't instantly hatch when being dropped from the inventory,

code/modules/mob/living/carbon/superior_animal/roach/types/toxic.dm
```diff
@@ -14,12 +14,40 @@
...
+	. = ..()
+	if(isliving(target))
+		var/mob/living/L = target
+		var/damage = rand(3, 7)
+		L.damage_through_armor(damage, TOX, attack_flag = ARMOR_BIO)
+
+/obj/item/projectile/roach_spit/attack_mob(mob/living/target_mob, distance, miss_modifier=0)
+	if (isroach(target_mob))
+		return FALSE // so these pass through roaches
+	..()
```
TheShown911, line 53: Newline.

code/modules/mob/living/carbon/superior_animal/roach/types/toxic.dm
vode-code: I survived it in testing, this is not just any roach but specifically a "radioactiveroach", hyronalin isn't all that special with a recipe of dylovene and radium, and once it's done processing all you need is dylovene to cure the Tox.

Iglpigl (changes requested):
>add: Gestrahlte roaches are now ranged attackers, and their bile is radioactive.

Firstly there is absolutely no reason for this to be made, secondly it will screw up maintenance gameplay loop and force chemists to work constantly or chemistry will be broken into when there are no chemists available, thirdly, ranged radioactive roaches(icon_state and roach description aren't justified reasons for the change, it was probably thought up on a whim) and kaisers having a radioactive melee attack too are direct violations to contributing guidelines, especially fun above realism section, you should try reading them.
https://wiki.cev-eris.com/Content_GuidelinesEn

GrayRachnid:
Ignore the person above and let him seethe. That's a good change.

Alliostra:
Firstly the only reason for Eris to exist is this, secondly it will fix the maintenance gameplay loop and force chemists to actually do their job, chemistry will be broken into even when there are chemists available as per usual anyways, thirdly, ranged radioactive roaches(icon_state and roach description are justified by lore for this change, it was probably thought up on a whim by reere but who cares) and kaisers having a radioactive melee attack too are in perfect accordance to contributing guidelines, other coders should try reading them.
https://wiki.cev-eris.com/Content_GuidelinesEn

Alliostra:
seriously though, if the rad damage is capped at 100, it really isn't very dangerous or debiliating. It will hardly deal any toxin damage, which can be easily counteracted with dylovene, and genetic damage occurs only rarely and in limited amounts.

granodd:
Ironhammer fears the indoor roach farmer.

TheShown911 (commented):
Alliostra did some tests and it looks like the rad damage *hopefully*  won't be a problem, only other thing I don't like from this PR is MINUTE/MINUTES but if it's for qol of some coders I can approve when roach hatching is fixed.

drexample:
Looks pretty based

infraredbaron (approved):
For some reason there is change in railings, but otherwise the PR is good

vode-code:
> 
> 
> For some reason there is change in railings, I don't see clear reason to, but otherwise the PR is pretty good

@TheShown911 asked for it and it helps with roach ranching, so I added it.

TheShown911 (approved):

PR #6395, opened 2021-07-21
DNA-authorization gunmod

SirRichardFrancis:
## About The Pull Request

Added trigger that, upon being installed, prevents everyone, but person who installed it, from using the gun.

Planned to be removable, but after some tests this modification became permanent, for two reasons:
First - It's worthless when removable. Currently it's possible to remove mods (but also destroy) within 1-2 seconds, regardless of "removal_time". What's the point of having protection, that can be bypassed in literally one second?
Second - There was a bug that allowed to re-write "original" DNA sample without even removing the lock. With mod being permanent - this does not occur.
Spawns exactly like any other trigger.
And yes, once owner is dead - gun is unusable, unless carrion/scientist use previous owner's genes.
image

Comes from [this document](https://docs.google.com/document/d/1DECLfZjtbj92sTkSON6T3fstegJqYVbjtLu3-6WvTiA/edit).

## Why It's Good For The Game

One more gunmod
Biocoded weapon sounds pretty cyberpunk-ish

## Changelog
:cl:
add: Biocoded gunlock
imageadd: sprite for dna lock trigger, same as danger zone, but blue
/:cl:


SaltyOldBull:
How can this gun mod be acquired if it's not a bother to ask?

Firefox13:
Did you made sure to test the DNA lock? Is it actually spoofable by enzyme editing? Are you sure that 15 rarity is enough for this kind of gunmod? You can check the amount of it by using SDQL2 query "select <object path>" when on server

SirRichardFrancis:
> Did you made sure to test the DNA lock? Is it actually spoofable by enzyme editing?

Tested it quite a bit, DNA-stealing worked, but as of yet, only carrion does that properly, DNA mod machine just wont change UE, and generally works odd in our build, for some reason.

> Are you sure that 15 rarity is enough for this kind of gunmod?

What kind of gunmod? It's cool concept, but not very useful in practice (if at all). And I have no idea whatsoever of how rare you want those to be. I would like to put it vendors (for 3k or so) and even IHS personnel inventories at spawn.
But then again, I don't want PR to be turned down, because "that buffs someone too much" or something along those lines, hence the rarity value from similar mods.

> You can check the amount of it by using SDQL2 query "select " when on server

image

drexample:
Add it to uplink for a relatively cheap price. Let the bad boys lock out their cool toys from the general populace. Also, there NEEDS to be a way to remove it manually. Maybe costly or too-long taking, but there needs to be one.

SirRichardFrancis:
Further testing revealed yet another bug. Will make new PR when thing is actually ready.

PR #6404, opened 2021-07-30
Da crab increase

Wouju:
## About The Pull Request

image

makes the stat increase of the von krabin thing be 15 instead of 5

## Why It's Good For The Game

~~its not~~

von krabin felt underwhelming ever since it was introduced, hopefully this makes it see a bit more use in science instead of collecting dust

## Changelog
:cl:
tweak: tweaked Von-Krabin Stimulator stat increase to 15 (from 5)
/:cl:

### DIFF
--- a/code/game/objects/items/devices/von-krabin.dm
+++ b/code/game/objects/items/devices/von-krabin.dm
@@ -18,7 +18,7 @@
 	var/active = FALSE
 	var/area_radius = 7
 
-	var/buff_power = 5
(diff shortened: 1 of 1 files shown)


TheShown911 (approved):
Would like a complete rework of this item, but beggars can't be choosers.

Wouju:
> Would like a complete rework of this item, but beggars can't be choosers.

agreed, but got no clue what it should do instead

PR #6434, opened 2021-08-17
fixes wire splicings spawning off of cable

vode-code:
now will check if trap is present before splicing behaviour
now will return false if there is no cable in range of splicing

## About The Pull Request
This PR fixes wire splicings spawning off of cables, and reduces the likelyhood of two wire splicings spawning on the same tile.

## Why It's Good For The Game
Wire splicings were supposed to never spawn off of cables, but did. The bug this PR patches enables it.

## Changelog
:cl: Chickenish
fix: Wire Splicings no longer spawn off of cables.
/:cl:

### DIFF
--- a/code/game/objects/random/traps/traps.dm
+++ b/code/game/objects/random/traps/traps.dm
@@ -36,11 +36,13 @@
 //Checks if a trap can spawn in this location
 /proc/can_spawn_trap(turf/T, trap)
 	.=TRUE
-	if(ispath(trap, /obj/structure/wire_splicing))
-		if(locate(/obj/structure/cable) in dview(3, T))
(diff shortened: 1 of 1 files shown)


TheShown911:
Was this tested?

SirRichardFrancis:
> Wire splicings were supposed to never spawn off of cables, but did.

We literally have splicing spawners on cables all around the ship, is that a bug too? 

image
And where wire splicings coming from, in your opinion, if not from wires?

vode-code:
> 
> 
> Was this tested?

Yes, I found a wire splicing on a cable, found bear traps, and didn't find any wire splicings off of cables on a walk-around.

PR #6445, opened 2021-08-20
Pizza implants dupe and knife eating fix

SirRichardFrancis:
## About The Pull Request

image

Sliceable snacks no longer take any of the arm-embedded augments and any tool with cutting quality.
Once lost two cleavers in a pizza, that was pretty stupid.

## Why It's Good For The Game

Fixes cool.

## Changelog
:cl:

fix: pizza won't accept tools with cutting quality

/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/food/snacks.dm
+++ b/code/modules/reagents/reagent_containers/food/snacks.dm
@@ -267,7 +267,7 @@
 		var/hide_item = !has_edge(W) || !can_slice_here
 
 		if (hide_item)
-			if (W.w_class >= src.w_class || is_robot_module(W))
+			if (W.w_class >= src.w_class || is_robot_module(W) || (QUALITY_CUTTING in W.tool_qualities))
 				return
 
 			to_chat(user, SPAN_WARNING("You slip \the [W] inside \the [src]."))


vode-code:
ah, so you'll be able to smuggle a hammer in a pizza, but not a knife.

vode-code:
> Sliceable snacks no longer take any of the arm-embedded augments

It looks like it'd still be able to take any arm embedded augments without cutting quality though?

SirRichardFrancis:
> ah, so you'll be able to smuggle a hammer in a pizza, but not a knife.

No, there is a size limit.

> It looks like it'd still be able to take any arm embedded augments without cutting quality though?

No, I tested every augment and those either have cutting or too big.

vode-code:
> No, there is a size limit.

yes, the size limit is that it has to be smaller than ITEM_SIZE_NORMAL, and hammers(the item called hammer, not the tool class) are ITEM_SIZE_SMALL.

> No, I tested every augment and those either have cutting or too big.

The "does not accept items as big or bigger than itself" function of sliceables is future-proofed, and would work even with a huge size sliceable. The "does not accept robot modules" function of sliceables is also future-proofed, and would work even if a new robot module was added with weird variables. The "does not accept arm-embedded augments" function this PR would add to sliceables is not future-proofed, and appears to just be a side effect of the future-proofed "does not accept cutting tools" function.

SirRichardFrancis:
image

TheShown911 (commented):
Better way to do this is just to disable embedded augments and not disable everything with cutting quality.

SirRichardFrancis:
> Better way to do this is just to disable embedded augments and not disable everything with cutting quality.

I don't agree. Trying to cut pizza and loosing a knife inside is ridiculous. For jailbreak you don't need cutting tools anyway.
Also embedded augments just tools, would have to list each.

vode-code:
> Also embedded augments just tools, would have to list each.

It is possible to check if the user has an `obj/item/organ_module/active/simple` in the active arm, and if so check if the held item is the `obj/item/organ_module/active/simple`'s `holding`.

vode-code:
> would have to list each

whenever possible, the best solution is to find how things really work, and then use that information to design a simple way to do what you need; the "list each" type of solution ends up being clunky and breaking with a single change.

SirRichardFrancis:
> It is possible to check if the user has an `obj/item/organ_module/active/simple` in the active arm, and if so check if the held item is the `obj/item/organ_module/active/simple`'s `holding`.

You see, `organ_module/` is an implant that's in the body, it contains and retracts `obj/tool/implantname`, which is used for interactions.


> whenever possible, the best solution is to find how things really work, and then use that information to design a simple way to do what you need; the "list each" type of solution ends up being clunky and breaking with a single change.

I'm completely satisfied with my solution, it's one line long and does the trick.

vode-code:
> does the trick

For now, it does.

I found the embedded multitools, you would need to also check for `obj/item/organ_module/active/multitool`, and check if the held item is in the `items` list.

SirRichardFrancis:
> For now, it does.
> 
> I found the embedded multitools, you would need to also check for `obj/item/organ_module/active/multitool`, and check if the held item is in the `items` list.

Why don't you go and realize all of that wonderful ideas in your own PR?

vode-code:
> 
> 
> > For now, it does.
> > I found the embedded multitools, you would need to also check for `obj/item/organ_module/active/multitool`, and check if the held item is in the `items` list.
> 
> Why don't you go and realize all of that wonderful ideas in your own PR?

currently am writing a different one, can do.

vode-code:
> > 
> > Why don't you go and realize all of that wonderful ideas in your own PR?
> 
> currently am writing a different one, can do.

Done in #6454

Firefox13:
I suppose this is redundant now since #6454 was merged?

SirRichardFrancis:
On the most part, yes. You still able to lose knifes inside the pizza tho, which is super retarded. But whatever.

PR #6446, opened 2021-08-20
Meat grinding for synths fix

SirRichardFrancis:
## About The Pull Request

image

Added /meat to chemistry gripper list and allowed synthetics to interact with the grinder.

## Why It's Good For The Game

Fixes cool.

## Changelog
:cl:
fix: chemistry gripper not accepting meat
/:cl:

### DIFF
--- a/code/modules/reagents/machinery/grinder.dm
+++ b/code/modules/reagents/machinery/grinder.dm
@@ -76,9 +76,6 @@
 	SSnano.update_uis(src)
 	return 0
 
-/obj/machinery/reagentgrinder/attack_ai(mob/user as mob)
-	return 0
-
 /obj/machinery/reagentgrinder/attack_hand(mob/user)
 	. = ..()
 	if(.)
--- a/code/modules/mob/living/silicon/robot/gripper.dm
+++ b/code/modules/mob/living/silicon/robot/gripper.dm
@@ -267,7 +267,8 @@
 		/obj/item/reagent_containers/spray,
 		/obj/item/storage/pill_bottle,
 		/obj/item/hand_labeler,
-		/obj/item/stack/material/plasma
+		/obj/item/stack/material/plasma,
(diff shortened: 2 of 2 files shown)


code/modules/reagents/machinery/grinder.dm
```diff
@@ -76,9 +76,6 @@
 	SSnano.update_uis(src)
 	return 0
 
-/obj/machinery/reagentgrinder/attack_ai(mob/user as mob)
```
vode-code, line 79: does the usability tweak in attackby not work? attack_ai, as the proc name implies, is for AIs, not direct touching.
`//Useability tweak for borgs`
`if (istype(I,/obj/item/gripper))`
`	ui_interact(user)`
`	return`

SirRichardFrancis: Gripper transfer items just fine, but grinder's interface can't be opened by silicon.

vode-code: and clicking with an empty gripper doesn't work?

SirRichardFrancis: It doesn't.

SirRichardFrancis: Regardless, I see no reason why robot can remotely interface with almost any machine, but not an industrial grinder.
Some modules have to interact with it as part of their "job".

vode-code: I thought maybe it'd let them teleport beakers, but from what I saw the beakers and inserted items eject at the location of the grinder. Not sure if there's a button on the UI that lets you teleport stuff in though.

SirRichardFrancis: If such problem even exist, there is a better way of fixing it than simply locking machine for robots.

TheShown911 (approved):

PR #6451, opened 2021-08-20
increases door health

zombiemimic:
## About The Pull Request
makes shooting doors less viable

## Why It's Good For The Game
puts more emphasis on entry without shooting

## Changelog
:cl:
tweak: tweaked door health
/:cl:


Firefox13:
> puts more emphasis on entry without shooting

I do really doubt that this will matter other than putting few more bullets into an airlock. Is there really any other reason?

zombiemimic:
I hope that this will make breaking down doors with knives less of a concern and promote other ways of entry

PR #6468, opened 2021-08-24
Fix camera alarm on camera destruction

### LINKED ISSUE #6345: Cameras are somewhat bugged
When a camera gets busted, it shows a camera alarm, as it should. When it gets repaired, however, the alarm still persists, even if the alarm light on the camera itself is off. This has happened multiple times and annoys the shit out of me.

Funce:
## About The Pull Request
Fixes #6345 
New behaviour upon total camera destruction, the camera alarm is removed from the network!

And on the side , because its camera related,
Small oversight with one of the proc names, so now Cyborgs will properly know when their camera is and isn't being looked at.

## Why It's Good For The Game
Bugfix

## Changelog
:cl:
fix: Cameras no longer stay infinitely alarming when they're repaired.
fix: Malf borgs properly know when their cameras are being accessed.
/:cl:

### DIFF
--- a/code/game/machinery/camera/camera.dm
+++ b/code/game/machinery/camera/camera.dm
@@ -70,6 +70,9 @@
 		assembly = null
 	qdel(wires)
 	wires = null
+	if(alarm_on)
+		alarm_on = 0
+		camera_alarm.clearAlarm(loc, src)
 	return ..()
 
 /obj/machinery/camera/Process()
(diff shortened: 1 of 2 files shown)


MLGTASTICa:
You're removing the alarm upon total destruction , not fixing the issue mentioned. Either use signals to check when the camera is rebuilt and remove the alarm itself , or tag it as a balance change.

PR #6478, opened 2021-08-29
Fixes hitting turrets with tools on help intent

SugarBone:
## About The Pull Request

This PR returns 1 after the user uses a suitable tool on help intent, preventing him from attacking the turret. Also introduces a minor adjustment to the intents, making it closer to the common standards. 

## Why It's Good For The Game

It prevents the user from accidentally damaging/breaking the turret while using the appropriate tools and adjusts the intent code to suit the common standards.

## Changelog
:cl:
fix: the user no longer hits the turret while using appropriate tools on help intent
code: the user will now hit the turret on every intent except HELP
/:cl:


SugarBone:
Messed up horribly, made an unmaintained branch, which was chosen for the pull request.

PR #6485, opened 2021-08-30
(Possible) Fix to intense gunfight lag + temperature gun bugfix

MLGTASTICa:
## About The Pull Request
Makes autofire handler use callback instead of sleep (thus no longer pausing the firer, and perhaps preventing the "lag")
Fixes temperature gun not applying the set temperature to its projectile
Fixes temperature gun giving gun parts

Might cause the timer system to shit itself , or make some guns fire slower than they should.  tested as much as i could locally , but the real test is when more than 4 guys are firing.
## Why It's Good For The Game
Less lag , as for the temperature gun , idk .
## Changelog
:cl:
fix: Fixed lag during intense gunfights
fix: Fixed temperature gun temperature setting not being applied onto projectiles
fix: Fixed temperature gun giving gun parts.
/:cl:


mattroks101:
What changed in interface?

MLGTASTICa:
> What changed in interface?

No idea , its just a automatic change , can't git ignore it or anything , if you want pls do a separate pull for this without that file thing.

MLGTASTICa:
REdid this without skin.dmf

PR #6492, opened 2021-09-01
Allows the carrion to re-evolve organs in the case they were removed

SugarBone:
## About The Pull Request

Adds renewability variable to the carrion powers, allowing carrion to re-evolve powers involving organs in the case those organs were removed. The price for the power, however, remains the same. Also adds a unique tag to maw, spinneret and the chemvessel, thus preventing the carrion from inserting a second organ of the type (due to the issues concerning the balance)

## Why It's Good For The Game

Having those organs removed could practically render the carrion useless. Moreover, it doesn't make sense that an antag capable of recovering from serious wounds would not be able to regrow its organs.

## Changelog
:cl:
balance: makes spinneret, maw and chemvessel unique organs - one per mob
fix: allows carrion to re-evolve removed organs
code: introduces a renewability variable to carrion powers, enables a check for removed carrion organs
/:cl:

### DIFF
--- a/code/game/gamemodes/changeling/modularchangling.dm
+++ b/code/game/gamemodes/changeling/modularchangling.dm
@@ -10,6 +10,7 @@ var/list/datum/power/carrion/powerinstances = list()
 	var/genomecost = 500000 // Cost for the carrion to evolve this power.
 	var/organpath //Path to the organ that is getting evolved.
 	var/spiderpath //The path of the spider we spawn.
+	var/renewable = FALSE //Whether the power can be renewed (Primarily used for organs)
 
 /datum/power/carrion/flashbang_spider
 	name = "Flashbang spider"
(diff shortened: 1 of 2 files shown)


SugarBone:
> I think a carrion could potentially remove their organs, regenerate them and then re-implant them. How does the code interact with multiple organs? Especially the chemical vessel.

Tested this. Doing this increases the efficiency of the organ. For the chem vessel, if the owner has two chem vessels in him, his max chemical value would be doubled as would the speed of generating the chemicals. Might be bad, but considering the effort you would have to put in (plus the risk), I think it is fair game. Having examined the code, I have confirmed the testing results - you could examine the "/obj/item/organ/internal/proc/get_process_eficiency" proc to confirm this yourself.

`/mob/living/carbon/human/proc/get_organ_efficiency(process_define)
	var/list/process_list = internal_organs_by_efficiency[process_define]
	var/effective_efficiency = 0
	if(process_list && process_list.len)
		for(var/organ in process_list)
			var/obj/item/organ/internal/I = organ
			effective_efficiency += I.get_process_eficiency(process_define)
		
	return effective_efficiency ? effective_efficiency : 1`

`/mob/living/carbon/human/proc/carrion_process()
	var/vessel_efficiency = get_organ_efficiency(OP_CHEMICALS)
	var/maw_efficiency = get_organ_efficiency(OP_MAW)
	if(vessel_efficiency)
		carrion_stored_chemicals = min(carrion_stored_chemicals + (0.01 * vessel_efficiency), 0.5 * vessel_efficiency)`

TheShown911 (changes requested):
This is not how organs are supposed to work, the main idea of the erismed organ rework is that all organs should stack. This PR also gives carrions nigh infinite free food for their gene points and no way for people to neuter carrion bodies.

PR #6493, opened 2021-09-01
Makes Oborin colorblindness no longer bypassable through the usage of prosthetic eyes

SugarBone:
## About The Pull Request

Removes the check for inorganic eyes for changing the colour matrix

## Why It's Good For The Game

The Oborin syndrome provides a ton of buffs in exchange for colorblindness, thus this exploit allows the player to bypass this drawback as easily as choosing a different option in character selection, which is bad. Moreover, lore-wise, Oborin is a "psionic" condition, so it doesn't make much sense that changing out the eyes would just remedy it.

## Changelog
:cl:
balance: prevents inorganic eyes from removing the colorblindness effect of Oborin
/:cl:

### DIFF
--- a/code/modules/organs/internal/eyes.dm
+++ b/code/modules/organs/internal/eyes.dm
@@ -43,8 +43,8 @@
 	if(is_broken() && !oldbroken && owner && !owner.stat)
 		to_chat(owner, SPAN_DANGER("You go blind!"))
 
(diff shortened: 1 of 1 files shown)


FortuneMaster:
I actually liked this feature because grey filter(other filters too, but at least they are temporary) are a bit painful to look and tire my eyes.

Firefox13 (approved):

PR #6514, opened 2021-09-08
[Needs live testing] Removes sleep from process() in bullet code

MLGTASTICa:
## About The Pull Request
Removes sleeping in process code for bullets , unlikely to fix the issues people are having but its always worth a shot.

## Why It's Good For The Game
If this fixes the lag it'd be preety nice.

## Changelog
:cl:
fix: Changed sleep() to spawn() in bullet process code .
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -392,8 +392,8 @@
 
 /obj/item/projectile/Process()
 	var/first_step = TRUE
-
-	spawn while(src && src.loc)
+	LOOP_START
+	spawn if(src && src.loc)
 		if(kill_count-- < 1)
 			on_impact(src.loc) //for any final impact behaviours
 			qdel(src)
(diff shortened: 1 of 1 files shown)


mattroks101:
I don't think that's going to fix anything.

code/modules/projectiles/projectile.dm
```diff
@@ -429,7 +429,9 @@
 			tracer_effect(effect_transform)
 
 		if(!hitscan)
-			sleep(step_delay)	//add delay between movement iterations if it's not a hitscan weapon
+			spawn(step_delay) goto LOOP_START 	//add delay between movement iterations if it's not a hitscan weapon
+		else
+			goto LOOP_START
```
drexample, line 434: Goto moment

drexample (changes requested):

MLGTASTICa:
> I don't think that's going to fix anything.

Not sure myself either , but it might help , since sleep makes the bullet instance sleep . while spawn and goto keep it running , might be less resource intensive than pausing , unpausing , pausing , unpausing every process

drexample:
Absolutely not

PR #6523, opened 2021-09-11
Avoid search in world for del_event

hyperioo:
## About The Pull Request

del_event function was searching `in world` instead of using the global list that contains all existing ships.

Minor tweak of how Eris is stored in that list because that `ships[eris]` with eris a global class variable made little sense to me.

## Why It's Good For The Game

Optimization

## Changelog
:cl: Hyperio
refactor: Refactor del_event function and global list for ships
/:cl:

### DIFF
--- a/code/global.dm
+++ b/code/global.dm
@@ -4,9 +4,6 @@ var/global/datum/DB_search/db_search = new()
 var/global/list/all_areas                = list()
 var/global/list/ship_areas               = list()
 
-
-GLOBAL_LIST_EMPTY(ships) // List of ships in the game.
-
 //var/global/list/machines                 = list()		//Removed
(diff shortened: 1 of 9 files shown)


LetterN:
image
GLOB.ships doesnt exist?

code/global.dm
```diff
@@ -5,7 +5,7 @@ var/global/list/all_areas                = list()
 var/global/list/ship_areas               = list()
 
 
-var/global/list/ships 	= list() // List of ships in the game.
+GLOBAL_LIST_EMPTY(ships) // List of ships in the game.
```
TheShown911, line 8: It's a problem with these lines here, not sure exactly why it doesn't work though.

PR #6526, opened 2021-09-12
update parkour

HentaiStorm:
## About The Pull Request

Perk 'Parkour' gives the ability to walk around low walls and railings

## Why It's Good For The Game

1) Climb up previously inaccessible ledges
2) Escape from roaches

## Changelog
:cl:
add: Added perk 'parkour' ability to walk around low walls and railings
/:cl:

### DIFF
--- a/code/datums/perks/oddity.dm
+++ b/code/datums/perks/oddity.dm
@@ -63,7 +63,7 @@
 
 /datum/perk/oddity/parkour
 	name = "Parkour"
-	desc = "You can climb tables and ladders faster."
+	desc = "You can climb tables and ladders faster. Also clings to railings and low walls."
 	icon_state = "parkour" //https://game-icons.net/1x1/delapouite/jump-across.html
 
 /datum/perk/oddity/parkour/assign(mob/living/carbon/human/H)
--- a/code/modules/multiz/movement.dm
+++ b/code/modules/multiz/movement.dm
@@ -222,6 +222,12 @@
 /mob/living/carbon/human/can_fall(turf/below, turf/simulated/open/dest = src.loc)
 	// Special condition for jetpack mounted folk!
 	if (!restrained())
+		if(stats.getPerk(PERK_PARKOUR))
+			for(var/obj/structure/low_wall/LW in view(src, 1))
+				return FALSE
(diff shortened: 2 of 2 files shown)


code/modules/multiz/movement.dm
```diff
@@ -222,6 +222,12 @@
 /mob/living/carbon/human/can_fall(turf/below, turf/simulated/open/dest = src.loc)
 	// Special condition for jetpack mounted folk!
 	if (!restrained())
+		if(stats.getPerk(PERK_PARKOUR))
+			for(var/obj/structure/low_wall/LW in view(src, 1))
```
MLGTASTICa, line 226: why for and not if ? seems like unnecesarry looping.

Alliostra:
this is very cool, maybe the same should be possible with magboots? And maybe you should be slowed down while climbing so you can't just run off a ledge at full speed and run to the other side.

Omni44:
Will that also make possible to climb onto low walls and railings from below while standing on the table? Right now railings prevent you from climbing completely if I recall correctly, that'd be a nice niche.

code/datums/perks/oddity.dm
```diff
@@ -63,7 +63,7 @@
 
 /datum/perk/oddity/parkour
 	name = "Parkour"
-	desc = "You can climb tables and ladders faster."
+	desc = "You can climb tables and ladders faster. Also clings to railings and low walls."
```
TheShown911, line 66:
```suggestion
	desc = "You can climb tables and ladders faster and also cling to railings and low walls."
```

code/modules/multiz/movement.dm
```diff
@@ -220,10 +220,20 @@
 
 
 /mob/living/carbon/human/can_fall(turf/below, turf/simulated/open/dest = src.loc)
-	// Special condition for jetpack mounted folk!
-	if (!restrained())
-		if (CanAvoidGravity())
```
TheShown911, line 225: You would need to put back this `CanAvoidGravity()` check somewhere in here (if it's even used).

TheShown911 (approved):
Very cool, slowdown may be needed to balance it out but that can be done later.

PR #6529, opened 2021-09-14
Logistics disk

pexel5:
## About The Pull Request

Removed the limit from the logistics disk.

## Why It's Good For The Game

This disk is not used so often, and those who use it or try to, rest against the limit of 10 items, i think it is too little. 
The disk itself contains " Conveyor belt assembly", "Conveyor switch  assembly", "Smelter circuit", "Sorter circuit".

## Changelog
:cl:
tweak: Removed the license limit from the logistics disk
/:cl:


Firefox13:
You probably could remove the prices from items in the disk to avoid confusion

Firefox13:
I meant removing values altogether but that works too.

TheShown911 (approved):

PR #6552, opened 2021-09-23
Uplink implant is now hidden

Humonitarian:
## About The Pull Request

Implants now have a variable for being scannable, uplink implants by default are hidden from medical scanners.

## Why It's Good For The Game

Design document

## Changelog
:cl:
balance: uplink implants are hidden from advanced medical scanners
/:cl:

### DIFF
--- a/code/game/machinery/adv_med.dm
+++ b/code/game/machinery/adv_med.dm
@@ -379,11 +379,13 @@
 		if (e.implants.len)
 			var/unknown_body = FALSE
 			for(var/I in e.implants)
-				if(is_type_in_list(I,known_implants))
+				if(is_type_in_list(I,known_implants)) //ATTENTION ALL ANTAG GAMERS
 					var/obj/item/implant/device = I
 					other_wounds += "[device.get_scanner_name()] implanted"
 				else
-					unknown_body = TRUE
+					var/obj/item/implant/device = I
+					if(!device.scanner_hidden)
+						unknown_body = TRUE
 			if(unknown_body)
 				other_wounds += "Unknown body present"
 		if (e.is_stump() || e.burn_dam || e.brute_dam || other_wounds.len)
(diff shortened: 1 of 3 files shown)


FortuneMaster:
well, haven't thought about making all untag implants be invisible, but should actually work. Maybe the only exception to this rule should be the explosive one.

Humonitarian:
> well, haven't thought about making all untag implants be invisible, but should actually work. Maybe the only exception to this rule should be the explosive one.

Only the Uplink Implant is invisible, Implants ordered *through* Uplink are still detectable

FortuneMaster:
oh, its just not clear from that line. Can be easily misread as "implants bought through uplink are undetectable", so you may want to change description a bit(at least change implants to implant)

Firefox13 (changes requested):
Please remove the unnecessary comment and update the box description to mention the cooldown.

Humonitarian:
What do you refer to as "box description"? The item's description?

FortuneMaster:
Damn, I just noticed that the description is outdated. Maybe change it to something like this: "Use it if you ever get tied up. Works once every 10 minutes"

zombiemimic (approved):

code/game/objects/items/weapons/implant/implants/freedom.dm
```diff
@@ -18,6 +18,8 @@
 		return
 	if (emote == src.activation_emote)
 		src.uses--
+		spawn(6000)
```
TheShown911, line 21: You can put `10 MINUTES` here instead.

Humonitarian:
Like this works?

TheShown911:
Should work (test it (waste 10 minutes of your life (also you don't need the `src.uses--` `uses--` should be fine)))

Humonitarian:
Last change untested and I'm in no condition to test rn, DNM until next week

Firefox13:
> 
> 
> Last change untested and I'm in no condition to test rn, DNM until next week


@Humonitarian It is about time.

PR #6556, opened 2021-09-23
Buffs most pouches

MLGTASTICa:
## About The Pull Request
Buffs pouches

## Why It's Good For The Game
Generic or small tend to always be better than specialized , it should not be the case.
## Changelog
:cl:
balance: Medical pouch can hold now 4 items
balance:Engineering supply pouch can now hold 12 items , from 2 , but it can no longer hold materials , tools or stacks
balance:Ammo pouch can now hold 9 mags instead of 3.
balance: Vial pouch can no longer hold grenades and rpgs (tubular untouched) but now it can hold 10 vials and small cells and other misc small items
balance : Tubular pouch can now hold any type of beaker , small cells and medium sized cells.

/:cl:


Sparrowbird611:
Engineering pouch is useless in this form except as an ammo pouch for laser gunners and should probably be reverted or changed in a different way. Otherwise this is good.

Firefox13 (commented):
Vial pouch could and should be removed because it was a simple rename before. Its update should be moved to tubular pouch

code/game/objects/items/weapons/storage/pouches.dm
```diff
@@ -136,29 +136,27 @@
 
 /obj/item/storage/pouch/engineering_supply
 	name = "engineering supply pouch"
-	desc = "Can hold engineering equipment. But only about two pieces of it."
+	desc = "Can hold engineering equipment. 12 pieces of hardware , cells , rods or cables"
```
Firefox13, line 139:
```suggestion
	desc = "Can hold engineering equipment. 12 pieces of hardware, cells, rods or cables."
```

TheShown911 (approved):

PR #6561, opened 2021-09-23
Bolt rifle tweak

SirRichardFrancis:
## About The Pull Request

Makes bolt rifles eject ammo each time bolt is worked, no matter spent or not. AMR too.
Because that's how bolt rifles work. (Or so I've been told)
Changed bolt rifle ammo capacity to 5, (also increases time it takes to reload from a box) so ammo clips isn't useless anymore.

## Why It's Good For The Game

Ammo clips is good and should be used.

## Changelog
:cl:
tweak: bolt rifles eject ammo when bolt is worked, spent or not
balance: bolt rifle ammo capacity halved
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/projectile/battle_rifle/boltgun.dm
+++ b/code/modules/projectiles/guns/projectile/battle_rifle/boltgun.dm
@@ -17,7 +17,7 @@
 	recoil_buildup = 1.6 // reduced from the AK's/Takeshi's buildup of 1.7/1.8 because >lol boltgun
 	handle_casings = HOLD_CASINGS
 	load_method = SINGLE_CASING|SPEEDLOADER
-	max_shells = 10
+	max_shells = 5
 	magazine_type = /obj/item/ammo_magazine/lrifle
 	fire_sound = 'sound/weapons/guns/fire/sniper_fire.ogg'
 	reload_sound = 'sound/weapons/guns/interact/rifle_load.ogg'
@@ -64,11 +64,17 @@
 	playsound(src.loc, 'sound/weapons/guns/interact/rifle_boltback.ogg', 75, 1)
 	bolt_open = !bolt_open
 	if(bolt_open)
-		if(chambered)
-			to_chat(user, SPAN_NOTICE("You work the bolt open, ejecting [chambered]!"))
-			chambered.forceMove(get_turf(src))
-			loaded -= chambered
-			chambered = null
+		if(contents.len)
+			if(chambered)
+				to_chat(user, SPAN_NOTICE("You work the bolt open, ejecting [chambered]!"))
+				chambered.forceMove(get_turf(src))
+				loaded -= chambered
+				chambered = null
+			else
+				var/obj/item/ammo_casing/B = loaded[loaded.len]
+				to_chat(user, SPAN_NOTICE("You work the bolt open, ejecting [B]!"))
+				B.forceMove(get_turf(src))
+				loaded -= B
 		else
 			to_chat(user, SPAN_NOTICE("You work the bolt open."))
 	else
--- a/code/modules/projectiles/guns/projectile/sniper.dm
+++ b/code/modules/projectiles/guns/projectile/sniper.dm
@@ -62,11 +62,17 @@
 	playsound(src.loc, 'sound/weapons/guns/interact/rifle_boltback.ogg', 75, 1)
 	bolt_open = !bolt_open
 	if(bolt_open)
-		if(chambered)
-			to_chat(user, SPAN_NOTICE("You work the bolt open, ejecting [chambered]!"))
-			chambered.loc = get_turf(src)
-			loaded -= chambered
-			chambered = null
+		if(contents.len)
+			if(chambered)
+				to_chat(user, SPAN_NOTICE("You work the bolt open, ejecting [chambered]!"))
+				chambered.forceMove(get_turf(src))
+				loaded -= chambered
+				chambered = null
+			else
+				var/obj/item/ammo_casing/B = loaded[loaded.len]
+				to_chat(user, SPAN_NOTICE("You work the bolt open, ejecting [B]!"))
+				B.forceMove(get_turf(src))
+				loaded -= B
 		else
 			to_chat(user, SPAN_NOTICE("You work the bolt open."))
 	else


Sparrowbird611:
Have you not considered just increasing the ammo clip to 10 instead of nerfing the bolt for what seems to be no reason.

SirRichardFrancis:
> Have you not considered just increasing the ammo clip to 10 instead of nerfing the bolt for what seems to be no reason.

As a matter of fact - I did. But then I would have to make new sprite for ammo clip.

Sparrowbird611:
> > Have you not considered just increasing the ammo clip to 10 instead of nerfing the bolt for what seems to be no reason.
> 
> As a matter of fact - I did. But then I would have to make new sprite for ammo clip.

I'm terrified that you think nerfing a weapon this hard just so you can skip out on spriting a niche item that would frankly be fine to remove is ok.

SirRichardFrancis:
> I'm terrified that you think nerfing a weapon this hard just so you can skip out on spriting a niche item that would frankly be fine to remove is ok.

No need to be dramatic, it's not a hard nerf. Use ammo clips.

Sparrowbird611:
> > I'm terrified that you think nerfing a weapon this hard just so you can skip out on spriting a niche item that would frankly be fine to remove is ok.
> 
> No need to be dramatic, it's not a hard nerf. Use ammo clips.

Better idea, explain why it's a good change in the first place. Novak hasn't brought up any severe balance concerns as of late, especially none that call for nerfing its ammo capacity by half, making it one of the lowest ammo weapons in the game, further cementing FA rifle meta. Change for the sake of change just so we have to use these ammo clips is mind boggling.

heymando:
alternatively i'd perhaps maybe recommend making ammo clips load instantly instead of using an action bar, as i have to assume they currently use one if they're being labelled the "worse ammo box"

Sparrowbird611:
> alternatively i'd perhaps maybe recommend making ammo clips load instantly instead of using an action bar, as i have to assume they currently use one if they're being labelled the "worse ammo box"

They don't. Hand loading is just superior, and still will be because printing a box is cheaper than printing a speedloader last I checked.

SirRichardFrancis:
> Better idea, explain why it's a good change in the first place. Novak hasn't brought up any severe balance concerns as of late, especially none that call for nerfing its ammo capacity by half, making it one of the lowest ammo weapons in the game, further cementing FA rifle meta. Change for the sake of change just so we have to use these ammo clips is mind boggling.

Because I believe that using ammo clips would benefit gameplay.

> alternatively i'd perhaps maybe recommend making ammo clips load instantly instead of using an action bar, as i have to assume they currently use one if they're being labelled the "worse ammo box"

It's instant load, of course, unlike ammo box. "Worse ammo box" is true tho, and that's what I'm correcting.

heymando:
ok then, but how are they comparable when they don't serve the same function?
the ammo box is for storing large amounts of ammo and slow-loading your gun while safe.
the ammo clip is for instantly loading your gun during a fight.

furthermore, how does this PR change the functions of the ammo box and ammo clip when it doesn't actually have anything to do with the function or effectiveness of either? the description of the pr's intended effect is unrelated to its actual code and one of those factors should be amended if this is a mistake.

SirRichardFrancis:
> ok then, but how are they comparable when they don't serve the same function?
> the ammo box is for storing large amounts of ammo and slow-loading your gun while safe.
> the ammo clip is for instantly loading your gun during a fight.

That's correct, but slow-loading isn't anywhere near being slow, and ammo clip hold only half of the gun ammo capacity, while taking same space the box does.

> furthermore, how does this PR change the functions of the ammo box and ammo clip when it doesn't actually have anything to do with the function or effectiveness of either? the description of the pr's intended effect is unrelated to its actual code and one of those factors should be amended if this is a mistake.

I meant that reloading with ammo box would take more time simply because it should be done more often, and reloading time is the same.
Removed that part from description so people won't get confused.

Sparrowbird611:
> > ok then, but how are they comparable when they don't serve the same function?
> > the ammo box is for storing large amounts of ammo and slow-loading your gun while safe.
> > the ammo clip is for instantly loading your gun during a fight.
> 
> That's correct, but slow-loading isn't anywhere near being slow, and ammo clip hold only half of the gun ammo capacity, while taking same space the box does.
> 
> > furthermore, how does this PR change the functions of the ammo box and ammo clip when it doesn't actually have anything to do with the function or effectiveness of either? the description of the pr's intended effect is unrelated to its actual code and one of those factors should be amended if this is a mistake.
> 
> I meant that reloading with ammo box would take more time simply because it should be done more often, and reloading time is the same.
> Removed that part from description so people won't get confused.

Do your shit properly then and increase the mag size to 10. This isn't an issue of whether or not speedloader is good or bad, it's an issue of you choosing to "fix" this by nerfing something that doesn't need a nerf. It's half assed and lazy.

SirRichardFrancis:
> Do your shit properly then and increase the mag size to 10. This isn't an issue of whether or not speedloader is good or bad, it's an issue of you choosing to "fix" this by nerfing something that doesn't need a nerf. It's half assed and lazy.

What if I don't accept your demands? Would you fire me or what?
DPS drop could be negated by usage of clips, as I mentioned already. Also this weapon's prototype have that ammo capacity.

heymando:
> > furthermore, how does this PR change the functions of the ammo box and ammo clip when it doesn't actually have anything to do with the function or effectiveness of either? the description of the pr's intended effect is unrelated to its actual code and one of those factors should be amended if this is a mistake.
> 
> I meant that reloading with ammo box would take more time simply because it should be done more often, and reloading time is the same.
> Removed that part from description so people won't get confused.

 as of 2 minutes ago the incorrect information in pr description is still present.

 specifically, the pr description says that the justification for this pr is that clips are underused; however, there are no changes in the pr as of current that promote the use of clips, and in fact the pr only does the opposite by making the likelihood of someone using the actual gun lower, and in conjunction the likelihood that someone uses clips.

vode-code:
> That's correct, but slow-loading isn't anywhere near being slow, and ammo clip hold only half of the gun ammo capacity, while taking same space the box does.

In that case, I'd suggest decreasing the size of a clip, which would improve the quick reload factor by allowing you to fit more in a bag or whatever, without changing the direct attributes of any gun.

Omni44:
High octane autism itt

SirRichardFrancis:
Clip size changing was in the PR I've made earlier, have perfect sense to move it here.

zombiemimic:
stripper clips are used for loading fast, not fully. Changing the magazine will only make this used less. The clip is already a small size. This does not help the game, nor does it incentivize stripper clips.

SirRichardFrancis:
Okay okay, I give. No need to lunch the coder. Let it be 10 ammo, fine.

TheShown911 (approved):
Eat the coders.

PR #6562, opened 2021-09-24
Prefix for medspec

SirRichardFrancis:
## About The Pull Request

image

What it says on the tin.

## Why It's Good For The Game

Medspec is the only IH with no job prefix, for no particular reason.

## Changelog
:cl:
add: "specialist" prefix for medspec
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -247,6 +247,7 @@
 var/list/rank_prefix = list(\
 	"Ironhammer Operative" = "Operative",\
 	"Ironhammer Inspector" = "Inspector",\
+	"Ironhammer Medical Specialist" = "Specialist",\
 	"Ironhammer Gunnery Sergeant" = "Sergeant",\
 	"Ironhammer Commander" = "Lieutenant",\
 	"Captain" = "Captain",\


Humonitarian:
Medic would probably be more appropriate

Firefox13 (commented):
Specialist can stand for a lot of other types and it doesn't cut for just the medical specialist.

SirRichardFrancis:
> Specialist can stand for a lot of other types and it doesn't cut for just the medical specialist.

"Medspec" prefix looks silly, "medical specialist" is long... "Medic" will do? Sounds worse than "Specialist" imo.

Firefox13:
Medic doesn't sounds professional either. I suggest we should keep it current way without a prefix

Omni44:
Make it M.S. 
M.S. Doom

SirRichardFrancis:
> Make it M.S.
> M.S. Doom

That would look stupid.

Omni44:
Well only alternative  then is Corpsman, but that's sooo milrp sounding

SirRichardFrancis:
> Well only alternative then is Corpsman, but that's sooo milrp sounding

Yeah. I still think that medspec need a prefix and that "Specialist" is best fit for that role, due to the lack of better alternative.

Humonitarian:
None sound professional. In the military they use non-professional terms as well, referring to combat life savers as Medics.

Sparrowbird611:
I feel like medspec is the best pick. Specialist isn't bad per say, it just isn't completely clear that it's a medical specialist.

ValoTheValo:
> None sound professional. In the military they use non-professional terms as well, referring to combat life savers as Medics.

Here is the thing, Specialist is a real rank used in several armies so a Medspec holding the rank and title of specialist is perfectly acceptable and professional.

ValoTheValo:
> Specialist can stand for a lot of other types and it doesn't cut for just the medical specialist.

As i said above its a rank. And there isnt any other role on eris referred to as specialist. Not like there would be any confusion

hyperioo (approved):
Ok codewise.

PR #6572, opened 2021-09-25
Hopefully  fixes a mech related runtime + mech buffs

MLGTASTICa:
## About The Pull Request
Hopefully fixes a runtime related to mechs missing their chassis
Buffs mechs in general cause they are trash
Doesn't fix pilot overlays because that thing needs managed overlays
## Why It's Good For The Game
Fun

## Changelog
:cl:
fix: Mech related runtime
balance: Leg turn rate drastically lowered on all mech legs
balance: Leg move delay lowered on all mech legs
balance: Mech carbine buffed to use less power , fire in 3 shot bursts and be slightly innacurate
balance: Ion carbine modified to use less power
balance: Laser carbine modified to use less power , fires in 2 shot bursts
/:cl:

### DIFF
--- a/code/modules/mechs/premade/combat.dm
+++ b/code/modules/mechs/premade/combat.dm
@@ -5,12 +5,6 @@
 
 	rarity_value = 60
 
-	arms = /obj/item/mech_component/manipulators/combat
-	legs = /obj/item/mech_component/propulsion/combat
-	head = /obj/item/mech_component/sensors/combat
-	body = /obj/item/mech_component/chassis/combat
-
-
 	material = MATERIAL_PLASTEEL
 	exosuit_color = COLOR_DARK_GUNMETAL
 	installed_armor = /obj/item/robot_parts/robot_component/armour/exosuit/combat
@@ -24,6 +18,17 @@
 		HARDPOINT_HEAD = /obj/item/mech_equipment/light,
 	)
 
+/mob/living/exosuit/premade/combat/Initialize()
+	if(!arms)
+		arms = new /obj/item/mech_component/manipulators/combat(src)
+	if(!legs)
+		legs = new /obj/item/mech_component/propulsion/combat(src)
+	if(!head)
+		head = new /obj/item/mech_component/sensors/combat(src)
+	if(!body)
+		body = new /obj/item/mech_component/chassis/combat(src)
+
+	. = ..()
 
 /mob/living/exosuit/premade/combat/slayer
 	name = "S.C.U. 'Slayer'" //Space Combat Unit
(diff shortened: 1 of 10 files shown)


githubuser4141:
Thank you, mechs have needed a buff for a while. Would it be possible to slightly buff the HP of mech sensors while you're here? IIRC the thermal sensors have 50 HP, the light (Mesons) have 30 HP and so both will pretty quickly if your opponent is aiming for the head and if you're using a heavy/non-transparent chassis this will render you completely blind. 

Also this might be out of scope for this PR, but there's currently a bug/exploit where damaging the mech then repairing it can let you repair beyond its health limit for super amounts of health.

code/modules/mechs/premade/heavy.dm
```diff
@@ -56,7 +48,7 @@
 	exosuit_desc_string = "super-heavy reinforced manipulators"
 	icon_state = "heavy_arms"
 	desc = "Designed for durability and dishing out beatings, this heavy set of manipulators can both take and dish out beatings."
-	melee_damage = 25
+	melee_damage = 60 // You know , these things walk like a snail . Why would you even get close to this.
```
TheShown911, line 51: I'm gonna put heavy arms on a fast mech and punt heads off of people.

code/modules/mechs/premade/combat.dm
```diff
@@ -85,21 +80,21 @@
...
 	name = "combat legs"
 	exosuit_desc_string = "sleek hydraulic legs"
 	desc = "These combat legs are both fast and durable, thanks to a generous plasteel reinforcement and aerodynamic design."
 	icon_state = "combat_legs"
-	move_delay = 3
+	move_delay = 2
 	power_use = 20
 	matter = list(MATERIAL_STEEL = 15)
-	turn_delay = 3 //almost identical to light, but slightly worse turning and less power efficiency for extra durability
+	turn_delay = 1 // Better than light , turns fast and consts a lot
```
TheShown911, line 97:
```suggestion
	turn_delay = 1 // Better than light , turns fast and costs a lot
```
(?)

TheShown911 (approved):

code/modules/mechs/premade/heavy.dm
githubuser4141: Sugg: Maybe disable strafing on treads, because strafing basically negates the balance of turn_delay?

MLGTASTICa: You need atleast 80 damage to one-tap a head, if 60 could actually one-tap valkrie would wreck.

PR #6573, opened 2021-09-25
Lets robots look down stairs and also PB through them

MLGTASTICa:
## About The Pull Request

Fixes a bug not allowing robots to look upstairs due to lack of  the variable
## Why It's Good For The Game
It just works

## Changelog
:cl:
fix: Robots runtiming whenever trying to look up or down a ladder.
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/silicon.dm
+++ b/code/modules/mob/living/silicon/silicon.dm
@@ -29,6 +29,8 @@
 	var/list/access_rights
 	var/obj/item/card/id/idcard
 	var/idcard_type = /obj/item/card/id/synthetic
+	var/is_watching = FALSE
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/silicon/silicon.dm
```diff
@@ -29,6 +29,8 @@
 	var/list/access_rights
 	var/obj/item/card/id/idcard
 	var/idcard_type = /obj/item/card/id/synthetic
+	var/is_watching = FALSE
+	var/can_multiz_pb = FALSE
```
TheShown911, line 33: These vars should be probably moved to mob/living so there aren't double definitions of the same variables.

PR #6576, opened 2021-09-25
Cognition now influences chance to trip,Cognition/Biological is required to see reagents inside a container, Chemist buff

MLGTASTICa:
## About The Pull Request
Trip chance is now at base 50%
Cognition reduces chance to trip
Tripping is no longer prevented by slip-preventing items such as magboots
You now need scanning googles , 35 cognition or 50 biological to recognise reagents in drinks
Adds functionality for abilities that are shared between more perks with bitflags

## Why It's Good For The Game
Forces everyone on plating to play on even ground , makes cognition count more 
Lets people actually poison drinks , which can deal to interesting plots

## Changelog
:cl:
balance: Seeing reagents inside beakers now requires scanning goggles , 35 cognition or 50 biology
balance: Tripping can no longer be prevented by anything slip-preventing
balance: Trip chance is now 50% , and gets lowered by cognition stat
code: Added bitflag handling for perks that share a common ability
balance: Buffed chemist perk and made the chemist perk also give the ability to see reagents
balance: Tripping no longer hardstuns , instead if weakens.
/:cl:

### DIFF
--- a/code/datums/perks/job.dm
+++ b/code/datums/perks/job.dm
@@ -62,7 +62,7 @@
 /datum/perk/selfmedicated/chemist
 	name = "Chemical-junkie"
 	desc = "You know what the atoms around you react to and in what way they do. You are used to making organic substites and pumping them into yourself in the name of science! \
-			You get 10 more NSA points and a quarter more NSA ontop than a normal person. Your chance of getting addicted is also reduced to half."
+			You get 10 more NSA points and a quarter more NSA ontop than a normal person. Your chance of getting addicted is also reduced to half and you can also see all reagents in beakers."
 	perk_shared_ability = PERK_SHARED_SEE_REAGENTS
 
 /datum/perk/selfmedicated/chemist/assign(mob/living/carbon/human/H)
@@ -245,7 +245,7 @@
 
 /datum/perk/job/club
 	name = "Raising the bar"
-	desc = "You know how to mix drinks and change lives. People near you recover sanity."
+	desc = "You know how to mix drinks and change lives. People near you recover sanity and you always know what is in the glasses."
 	icon_state = "inspiration"
 	perk_shared_ability = PERK_SHARED_SEE_REAGENTS
 
--- a/code/datums/perks/job.dm
+++ b/code/datums/perks/job.dm
@@ -245,9 +245,8 @@
 
 /datum/perk/job/club
 	name = "Raising the bar"
-	desc = "You know how to mix drinks and change lives. People near you recover sanity and you always know what is in the glasses."
+	desc = "You know how to mix drinks and change lives. People near you recover sanity."
 	icon_state = "inspiration"
-	perk_shared_ability = PERK_SHARED_SEE_REAGENTS
 
 /datum/perk/job/club/assign(mob/living/carbon/human/H)
 	..()
--- a/code/datums/perks/job.dm
+++ b/code/datums/perks/job.dm
@@ -61,7 +61,7 @@
 
 /datum/perk/selfmedicated/chemist
 	name = "Chemical-junkie"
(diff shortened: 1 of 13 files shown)


Omni44:
I gotta be honest, I don't like the magboots change, because active magboots make you so considerably slower when active that when coupled with pain effects you are probably more of an static object. It's enough of an disadvantage that not even now anyone really uses magboots. Besides it tripping with active magboots makes little sense to me, in universe. Otherwise everything seems fine.

Humonitarian:
Revert magboot, everything else is great (maybe raise cog required? Basic no oddity cog gain is pretty fast!)

MLGTASTICa:
> Revert magboot, everything else is great (maybe raise cog required? Basic no oddity cog gain is pretty fast!)

Ain't reverting , tripping is tripping. Magboots are meant for atmospheric changes , not avoiding metal beams. Use them properly instead of running on plating.

FortuneMaster:
Add to the bar perk ability to see chemecals

code/modules/clothing/glasses/science.dm
```diff
@@ -1,6 +1,6 @@
 /obj/item/clothing/glasses/powered/science
 	name = "Science Goggles"
-	desc = "The goggles do nothing!"
+	desc = "These goggles scan the reagents within beakers , displaying them to you!"
```
Firefox13, line 3:
```suggestion
	desc = "These goggles scan the reagents within beakers, displaying them to you!"
```

code/modules/clothing/glasses/science.dm
```diff
@@ -1,6 +1,6 @@
 /obj/item/clothing/glasses/powered/science
 	name = "Science Goggles"
```
Firefox13, line 2:
```suggestion
	name = "science goggles"
```

Firefox13 (commented):
imo COG requirement should be much higher for checking reagents in reagents.
Also, great time to add perks for chemists/club guys to be able to see exact reagents no matter what

SaltyOldBull:
a 30 or 35 cog requirement would be better

code/datums/perks/job.dm
```diff
@@ -59,6 +59,24 @@
 		holder.metabolism_effects.nsa_threshold_base -= 10
 	..()
 
+/datum/perk/selfmedicated/chemist
+	name = "Chemical-junkie"
+	desc = "You know what the atoms around you react to and in what way they do. You are used to making organic substites and pumping them into yourself in the name of science! \
```
SaltyOldBull, line 64: " You are used to making organic *substites*" typo. Use substitutes instead.

code/game/turfs/flooring/flooring.dm
```diff
@@ -255,12 +255,11 @@ var/list/flooring_types
...
 	if(!ishuman(M)|| M.incorporeal_move || !has_gravity(get_turf(M)))
 		return
+	var/mob/living/carbon/human/our_trippah = M
 	if(MOVING_QUICKLY(M))
-		if(prob(5))
-			M.adjustBruteLoss(5)
-			M.slip(null, 6)
-			playsound(M, 'sound/effects/bang.ogg', 50, 1)
-			to_chat(M, SPAN_WARNING("You tripped over!"))
+		if(prob(50 - our_trippah.stats.getStat(STAT_COG))) // The art of calculating the vectors required to avoid tripping on the metal beams requires big quantities of brain power
```
TheShown911, line 260: Maybe put COG * 20, as having 45 COG to have the same base tipping chance as before is high.

hyperioo (approved):
Could take the best of Cognition or Vigilance for the trip chance since someone really watchful could be able to notice a metal beam on the floor.

PR #6581, opened 2021-09-27
Fix job ban panel

hyperioo:
## About The Pull Request

Fix the issue with job ban panel where all jobs appeared under all categories.

## Changelog
:cl: Hyperio
fix: Fixed job ban panel
/:cl:


code/modules/admin/topic.dm
```diff
@@ -9,7 +9,7 @@
 
 /datum/admins/proc/formatJobGroup(var/mob/mob, var/title, var/color, var/bantype, var/list/joblist)
 	. += "<tr bgcolor='[color]'><th><a href='?src=\ref[src];jobban3=[bantype];jobban4=\ref[mob]'>[title]</a></th></tr><tr><td class='jobs'>"
-	for(var/jobPos in GLOB.joblist)
+	for(var/jobPos in joblist)
 		. += formatJob(mob, jobPos, GLOB.joblist[jobPos])
```
TheShown911, line 13: You sure this will work, it still takes `GLOB.joblist[jobPos]`?

SirRichardFrancis: It will.
image

PR #6589, opened 2021-09-29
Biosyphon update

Alliostra:
## About The Pull Request

The biosyphon now actually serves a purpose, as it produces stat boosting donuts when fed with enough maint fauna meat, with more rare types of meat (fuhrer and kaiser) producing donuts faster. The donuts only raise one stat by 10 points for 20 minutes and do not stack, so this should not be a balance concern (nobody will use it anyways other than antags who steal it for memes)

## Why It's Good For The Game

This update will thoroughly break the maintenance economy by making all of these items rarer, it also has literally no reason to exist and ironhammer mains will demand katanas to be mapped into the armory for "butchering purposes"

## Changelog
:cl:
tweak: the biosyphon can now be fed maint fauna meat to produce masterpiece donuts.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/biosyphon.dm
+++ b/code/game/objects/items/devices/biosyphon.dm
@@ -1,6 +1,6 @@
 /obj/item/biosyphon
 	name = "Bluespace Biosyphon"
-	desc = "Hunts on flora and fauna that sometimes populates bluespace, and use them to produce donuts endlessly."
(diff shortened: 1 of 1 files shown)


Alliostra:
That thing claims I've changed changelog shit. what do

FortuneMaster:
wanted to buff it that way a long time ago

Alliostra:
Noooo I don't want to fulfil the design document

hyperioo (changes requested):
Try to modify only biosyphon.dm since there is no reason to change the others

PR #6593, opened 2021-09-30
Sanctifier fix

garrulousTeamotei:
## About The Pull Request

Sanctifier tool mod now burns carrions, as intended.
Sanctifier tool mod added to Sword of Truth.

## Why It's Good For The Game

Intended effect failed to hit the mark, best help it reach it's target.
If even the lowliest NT weapon comes with a sanctifier pre-installed, why not the Sword of Truth?

## Changelog
:cl:

tweak: Sanctifier added to Sword of Truth
fix: Sanctifier works against Carrions

/:cl:

### DIFF
--- a/code/modules/mob/living/living_defense.dm
+++ b/code/modules/mob/living/living_defense.dm
@@ -45,7 +45,7 @@
 	if(ishuman(src) && isitem(used_weapon))
 		var/mob/living/carbon/human/H = src
 		var/obj/item/I = used_weapon
-		if((H.faction == "spiders" || mutations.len) && (SANCTIFIED in I.aspects))
+		if((is_carrion(H) || mutations.len) && (SANCTIFIED in I.aspects))
 			sanctified_attack = TRUE
 	//Feedback
 	//In order to show both target and everyone around that armor is actually working, we are going to send message for both of them
--- a/code/modules/mob/living/living_defense.dm
+++ b/code/modules/mob/living/living_defense.dm
@@ -45,7 +45,7 @@
 	if(ishuman(src) && isitem(used_weapon))
 		var/mob/living/carbon/human/H = src
 		var/obj/item/I = used_weapon
-		if((H.has_organ(BP_SPCORE) || mutations.len) && (SANCTIFIED in I.aspects))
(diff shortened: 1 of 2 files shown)


Humonitarian:
Can't this be used for validhunting? Feels like a big concern.

code/modules/mob/living/living_defense.dm
```diff
@@ -45,7 +45,7 @@
 	if(ishuman(src) && isitem(used_weapon))
 		var/mob/living/carbon/human/H = src
 		var/obj/item/I = used_weapon
-		if((H.has_organ(BP_SPCORE) || mutations.len) && (SANCTIFIED in I.aspects))
+		if((H.faction == "spiders" || mutations.len) && (SANCTIFIED in I.aspects))
```
vode-code, line 48: does is_carrion(H) not work?
my main concern is if someone made some spider equivalent of fuhrerole.

Humonitarian: A spiderole making you vulnerable to sanctifier sounds like a feature

garrulousTeamotei:
> Can't this be used for validhunting? Feels like a big concern.

For the carrion bit, this isn't a new feature, it's a bug fix. The original intention of the sanctifier is that it can both burn carrions, and mutants made by moebius.
In code/__DEFINES/damage_organs BP_SPCORE is defined as "spider core". The original intent of (H.has_organ(BP_SPCORE), and the sanctfier's description mentioning its effect on carrions seems clear cut.

code/modules/mob/living/living_defense.dm
garrulousTeamotei: Yea, kinda new to this coding bit. A combo of not knowing that is_carrion(H) existed, and realizing that a "spiderole" probably wouldn't exist due to it being a carrion false flag, and I went to the one I figured would work. I'll try it out.

garrulousTeamotei: Works like a charm. Thanks for the info on that.

Humonitarian:
Oh right, can't carrions remove themselves from the spider faction temporarily? That would also make them immune to this to prevent validhunting

PR #6599, opened 2021-10-02
Overmap Extension Part 2

hyperioo:
## About The Pull Request

Implementation of "Locators and scanners" category of the KOT map:
* Extended view on overmap to 21x21 instead of 7x7
* Passive scan range increased to 6 tiles
* New active scan ("Pulse" button on helm console) that consumes a lot of energy from the long range scanner and increases the scan range to 10 tiles during 30 seconds
* Energy consumption and duration of active scan are improved depending on the components installed in the long range scanner (possibility to swap them will come later)
* Increase size of long range scanner room because it was too tiny before to be able to place 3 spying sensors in it for Agent contract
* Points of interests are now spawned on the overmap at round start. Eris has to find them and stand still on their tile to be able to scan them with the "Scan Point of Interest" button on the helm console.
* One "Space Wrecks" point of interest adds a new "SpaceWrecks" affinity junk field to the pool of available fields when it is scanned. It contains spaceship-oriented 25 x 25 chunks. You will see in the list of the junk tractor beam console.
* One "Trading station" point of interest which adds a new trading station in cargo console when it is scanned. Roundstart number of trading station has been lowered from 8 to 7, so now you have to search the 8th one by scanning the point of interest.
* Three "Blacksite" points of interest (one small, one medium, one large) which add shuttle navigation points when they are scanned. They lead to new zlevels containing abandonned OneStar facilities. You can go there using Aster or Moebius shuttles. The small blacksite and the medium blacksite have their own maps. The large blacksite uses a copy of the medium blacksite for now as a placeholder because it will need more work to really be end-game content. Those facilities are only loaded when a shuttle travels to them so that they do not impact performance till someone decides to visit them (using a new trigger_landmark function that is called when a navigation point is being travelled to)
* Adding airless variants of most turfs so that blacksites turfs exposed to space are directly spawned without atmosphere to reduce performance impact.
* All Junk Tractor Beam chunks are now using those airless tiles which should reduce issues with items flying everywhere as the initial atmosphere is vented

## Why It's Good For The Game

Content and kot map.

## Changelog
:cl: Hyperio
add: Overmap extension part 2
/:cl:


Firefox13:
I've grabbed one of the .dmi myself and sent it to Rosa

PR #6605, opened 2021-10-04
Walking on plating without any form of lightning and without any form of seeing turfs will now make you trip regardless

MLGTASTICa:
## About The Pull Request
Walking on plating without any form of lightning / mesons / NVGS / etc makes you trip regardless of stats

## Why It's Good For The Game
People don't use mesons because they find no use for them , another reason is because they simply don't need lightning on turfs , this should change. Lightning should matter
THE PDA DEFAULT LIGHT IS ENOUGH TO COUNTERACT THIS.

## Changelog
:cl:
balance: You will now trip regardless of stats on plating if you don't have any form of lightning or ability to see
/:cl:

### DIFF
--- a/code/game/turfs/flooring/flooring.dm
+++ b/code/game/turfs/flooring/flooring.dm
@@ -256,11 +256,15 @@ var/list/flooring_types
 	if(!ishuman(M)|| M.incorporeal_move || !has_gravity(get_turf(M)))
 		return
 	var/mob/living/carbon/human/our_trippah = M
+	var/turf/our_plating = get_turf(our_trippah)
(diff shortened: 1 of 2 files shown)


FortuneMaster:
Trip chance is already high, what's the point of buffing it even more?

ValoTheValo:
this is just change for the sake of change. tripping chance is already high enough

TheShown911:
This just forces players to turn on their PDA light, it has an atomcell and never runs out? Just why? Only thing this could change is sneaking around in the dark on plaiting and I don't think that's a big problem.

Firefox13:
Well, what already people said. Tripping on underplating is already (reasonably tho) annoying enough, but making people trip on normal plating is pure agony. This would kill any kind of stealth and cuck the perk that speeds you up in darkness.

PR #6608, opened 2021-10-05
Lets detectives get DNA from blood inside beakers or syringes

MLGTASTICa:
## About The Pull Request
Lets detectives sample the blood reagent for its DNA
fixes some code style 

## Why It's Good For The Game
Litteraly can't check syringes of blood :((((( how do i valid

## Changelog
:cl:
balance: Detectives can now swab syringes or beakers for the DNA of the blood inside
/:cl:


code/game/objects/effects/decals/Cleanable/humans.dm
```diff
@@ -60,6 +60,7 @@ var/global/list/image/splatter_cache=list()
 
 /obj/effect/decal/cleanable/blood/Process()
 	if(world.time > drytime)
+		STOP_PROCESSING(SSobj, src) // Ok no more processing after this
```
SirRichardFrancis, line 63: Why do we have blood processing in the first place...

image

code/modules/detectivework/tools/swabs.dm
```diff
@@ -55,17 +55,17 @@
...
 	else
-		return
+		return FALSE
 
 	if(sample_type)
 		set_used(sample_type, H)
-		return
-	return 1
+		return TRUE
+	return TRUE
```
SirRichardFrancis, line 68: Probably no need in `return TRUE` after `set_used(sample_type, H)`, since there is only another `return TRUE` after that.

code/game/objects/effects/decals/Cleanable/humans.dm
MLGTASTICa: nanako code

PR #6604, opened 2021-10-04
moon-walking backwards is now 50% slower, facedir is no longer binded to alt, Binds changing move intent to Ctrl and adds a verb for it.

MLGTASTICa:
## About The Pull Request
Moon-walking behind you now makes you 50% slower, facedir is no longer binded by default 
Moon-walking to the sides makes you 10% slower
Adds a verb for changing move intent , binded to Ctrl by default , players are free to rebind to any key
## Why It's Good For The Game
General balance change and QoL
## Changelog
:cl:
add: Added a verb to change moving intent , and it is binded to Ctrl by default
balance: Walking backwards is now 50% slower
balance: Walking to the sides is now 10% slower.
del: Removed the  binds for facing dir . Macro it if you want to keep using it.
/:cl:

PR #6613, opened 2021-10-05
Penetration

Humonitarian:
## About The Pull Request

Piercing walls now uses armor_penetration instead of damage, and successfully piercing walls decreases the damage, penetration and velocity of the bullets. This will both make penetrating walls with good guns easier, but also reduce their effectiveness.
Probably a lot more interesting way to nerf thermal vision than just constantly increasing the price.

The PR also completely reworks the way shields work against bullets. The 50% block chance now no longer applies to projectiles. Instead, bullets will attempt to penetrate shields as if they were machinery, requiring "penetrating", and a good amount of armor penetration, and even if they succeed, their overall damage and armor penetration is reduced.

In effect, high-end weapons will see their effective damage output reduced by 10-20% as opposed to 50%, whereas smaller caliber weapons and rubber bullets are completely, or mostly ineffective. (Especially weapons with low multipliers, such as SMGs).
As shields now also have a list of bodyparts they protect (head+body+groin+holding hand), it is possible to use low-caliber weapons to target a shield user's legs or offhand. To combat this, when wielding a shield, it will cover the entire body.

It is hard to tell if values are anywhere balanced at the moment, playtesting might be required.

Beams are worthy of mentioning too, they will always penetrate shields, but suffer a decrease in raw damage when doing so.

## Why It's Good For The Game

Makes thermal-sniping through walls less lethal, more immersive shields no longer use RNG magic to block projectiles.

## Changelog
:cl:
balance: Piercing walls now uses armor penetration, inconsistent penetration results in loss of momentum
balance: Shields have been reworked, becoming a lot more consistent, stronger against smaller arms fire, but much less effective against high-end firearms.
tweak: Shields now only protect particular bodyparts, depending on how they are held
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/nt_melee.dm
+++ b/code/game/objects/items/weapons/nt_melee.dm
@@ -210,6 +210,7 @@
 	price_tag = 1000
 	base_block_chance = 40
 	item_flags = DRAG_AND_DROP_UNEQUIP
+	shield_integrity = 120
 	var/obj/item/storage/internal/container
 	var/storage_slots = 3
 	var/max_w_class = ITEM_SIZE_HUGE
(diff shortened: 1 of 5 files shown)


Humonitarian:
#6604 I believe the change to arc might also be important in that PR

PR #6641, opened 2021-10-10
Fixes default eris style hud for 514 and adds a move intent verb for macro'ing

MLGTASTICa:
## About The Pull Request
FIXES THE DEFAULT ERIS STYLE HUD
and adds a verb that can be macro'ed for toggling run or walk

## Why It's Good For The Game
I assure you this will break everything else.

## Changelog
:cl:
fix: Fixed huds for versions past 514.16
add: Added a verb to toggle walk or run . Not binded , has to be macroed , verb is change-move-intent.
/:cl:


SirRichardFrancis:
> Yes they are , the warning is meaningless. It doesn't need a static type , it will always be a hud.

Ok, Nanako.

PR #6664, opened 2021-10-19
Blacklists of most of the hastily added armor disks try 2

Iglpigl:
## About The Pull Request

Blacklists all expect generic armor disks from the game.

## Why It's Good For The Game

They were added on a whim of ex-delegate who is now gone, they are an idea thats plain bad especially with how common they were, and with everyone running with full bulletproofs other cloths are neglected, the armors from those disks can still be acquired using other ways.

## Changelog
:cl:
tweak: Blacklists all expect generic armor disks, the armors in them can still be acquired with other means like junking, guild, etc.
/:cl:


Humonitarian:
Scavenger is still an option as well. The only balance concern this PR has is making the average Eris inhabitant less armored, which honestly, is a good change.

Albeit changing the rarity of the disks seems like a better solution, personally.

PR #6667, opened 2021-10-20
Vapour

TGCMbetcher:
## About The Pull Request
Changes the charge use from .5 to .2. Turns the basic battery use from 10 units to 25 units per charge. 200 unit batteries will use the whole amount of reagents inside now instead of requiring 500 units of charge to use the whole mask.

## Why It's Good For The Game
Masks cant be used for smoking desires due to code issues I cant fixed. Time to make them a slow chem delivery systems, theyre good with space drugs and you need to know how to modify them but theyre good for pain chem delivery that lets you cut it off hen you want to instead of measuring pill size or liquid amounts. Balanced by the fact you cant use internals with it.

## Changelog
:cl:
tweak: changed the amount of power to use the vapour mask fully from 500 units to 200 units.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/cigs_lighters.dm
+++ b/code/game/objects/items/weapons/cigs_lighters.dm
@@ -566,7 +566,7 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	var/voltage = 0
 	var/quality_multiplier = 1
 
-	var/charge_per_use = 0.5
+	var/charge_per_use = 0.2
 	var/obj/item/cell/cell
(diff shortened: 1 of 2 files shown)


drexample:
Linter moment

Kegdo:
Ima look into this
Keeping you informed

Firefox13 (commented):
I suppose you could recreate the PR or something to avoid that Install.bat deletion because we really it was *probably* put there for a reason.
Given this PR also is sitting here for a month we'll close it on next friday.
LInter shitting itself also was fixed.

PR #6695, opened 2021-10-30
Nerfs ciggaretes

MLGTASTICa:
## About The Pull Request
Ciggarete cases now cost 1000 credits ( 400 credits cheaper than buying them individually)
All cigars now overall last less
All cigars now give more sanity
Loadout cigars cases now start with a random number of cigars.
Cigar cases are now small sized.

## Why It's Good For The Game
Cigars are too good at gaming the sanity system.

## Changelog
:cl:
balance: Cigar cases now cost 1000 credits , cigar packets now cost 150 and special cigar packets cost 200
balance: Ciggaretes contain less nicotine overall
balance: Ciggaretes now give more sanity overall
balance: Loadout cigar case starts with a random number of cigars beween 0 and 8
balance: Cigar case is now small , instead of tiny
/:cl:


code/game/machinery/vending.dm
```diff
@@ -1107,14 +1107,14 @@
...
 	prices = list(/obj/item/clothing/mask/smokable/cigarette/cigar = 200,
-				  /obj/item/storage/fancy/cigarettes = 100,
+				  /obj/item/storage/fancy/cigarettes = 150,
 				  /obj/item/storage/fancy/cigcartons = 800,
 				  /obj/item/storage/box/matches = 10,
 				  /obj/item/flame/lighter/random = 5,
-				  /obj/item/storage/fancy/cigar = 450,
-				  /obj/item/storage/fancy/cigarettes/killthroat = 100,
+				  /obj/item/storage/fancy/cigar = 1000,
+				  /obj/item/storage/fancy/cigarettes/killthroat = 200,
```
zombiemimic, line 1115: I'm surprised the cigarettes with literal poison in them are more expensive than the normal ones

Kegdo:
Aight, I like your idea of nerfing cigars/tobacco to make other sanity restoring methods """""""""viable"""""""""
However, the method you are using to do so is horrendous. It's balance for the sake of balance, and nobody'd like it. Instead of changing funny numbers and adding even more RNG to the game, make, in this case, tobacco have a noticeable drawback. If Erismed 4'd be implemented, this drawback would be lung cancer, which could be a pain in the ass to cure, therefore not everyone'd use tobacco. Unfortunately, everyone is a lazy shmuck and it's still not coded in. Therefore, you should either MAKE ERISMED 4(gigachad move), or think of an another, similarly unique, disadvantage for smokers.
That, or just make vaping good. Damn, I'd love to vape myself, but the vapour masks are not even in the loadout menu.

Kegdo:
Also, don't nerf and then buff; either buff first, or buff and nerf in the same pr. Nerfing something without making something else better makes your pr look like a salt pr.

glaciusinfinity:
Ok, you make cigarettes' cost more, but then you also make them worse. Dude that's two different nerfs in one. Do one or the other, not both at the same time.

Alanii:
This PR makes no sense.

It isn't a proper balance change at all. You nerf one side, and buff the other. Not increase the price, while also lessening the effects.

It just looks like you're upset people can afford fancier smokes than you, and you want to make it harder for those people to get them, and make it worse off to use them because you are too lazy to use them.

ValoTheValo:
If you want to make other sanity sources more viable. Buff them. This is a lazy nerf and doesnt fix the root problem

mattroks101:
This is the wrong way to go about this. This is not going to get merged.

PR #6712, opened 2021-11-03
Removes water-pot explosion reaction , replaces it with fire

MLGTASTICa:
## About The Pull Request
Water and pot no longer explode , but instead create a area denying fire of weak intensity with a low chance to actually ignite anyone
## Why It's Good For The Game
We nerfed water pot into the ground , might as well make it have a more sparky role with area-denial role instead of an useless explosion

## Changelog
:cl:
balance: water-pot no longer creates a explosion , it now creates a fire with duration  and range scaling off amount.
/:cl:

### DIFF
--- a/code/game/objects/effects/effect_system.dm
+++ b/code/game/objects/effects/effect_system.dm
@@ -491,7 +491,7 @@ steam.start() -- spawns the effect
 
 /datum/effect/effect/system/pottasium_sparkle_explosion/start()
 	started = world.time + duration
-	looping_fire(fire_stacks)
+	looping_fire()
 
 /datum/effect/effect/system/pottasium_sparkle_explosion/proc/looping_fire()
 	var/particles_to_throw = amount_rating / 50 // 10 big particles in the worst case that break into 30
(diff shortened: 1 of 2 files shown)


Omni44:
This would be okay and all,b ut fire in itself is such a joke, that it's not much of an effect anyways, sadly.

Humonitarian:
Make fire deadly too while you're at it

Alliostra:
SC PR

TheShown911:
I don't think "water/pot is useless so we should nerf it even more" is a good idea

MLGTASTICa:
> I don't think "water/pot is useless so we should nerf it even more" is a good idea

imo its kinda useless in its state , and in real life it is an exothermic reaction , soooo. its a really good grenade if you want to set off a plasmafire off very fast.

Firefox13:
Removing pot-water explosion would be a crime against humanity.

code/game/objects/effects/effect_system.dm
```diff
@@ -474,3 +474,65 @@ steam.start() -- spawns the effect
 				round(min(light, BOMBCAP_LIGHT_RADIUS)),
 				round(min(flash, BOMBCAP_FLASH_RADIUS))
 				)
+
+/datum/effect/effect/system/pottasium_sparkle_explosion
```
Firefox13, line 478: also it's potassium not pottasium

PR #6714, opened 2021-11-03
IH medical HUD + EMT belt for medspec

hyperioo:
## About The Pull Request

Replace Medical specialist standard medical belt by an emergency response team belt so that they can store more rescue stuff on them (such as crowbar).

Create a medical HUD with flash protection. Replace the sec HUD in Medspec lockers by one of those.

## Why It's Good For The Game

Makes Medical Specialists' life easier.

## Changelog
:cl: Hyperio
add: Added new IH medical HUD
balance: Medspec now spawns with an ERT medical belt
/:cl:

### DIFF
--- a/code/datums/outfits/jobs/security.dm
+++ b/code/datums/outfits/jobs/security.dm
@@ -59,7 +59,7 @@
 	uniform = /obj/item/clothing/under/rank/medspec
 	suit = /obj/item/clothing/suit/storage/toggle/labcoat/medspec
 	pda_type = /obj/item/modular_computer/pda/forensics
-	belt = /obj/item/storage/belt/medical
+	belt = /obj/item/storage/belt/medical/emt
 	l_hand = /obj/item/storage/briefcase/crimekit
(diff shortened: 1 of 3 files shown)


Jan-Blasiak:
Oh, yes, you wouldn't believe how many times I've flashed the IH medspec with stationary flashes.

MLGTASTICa:
For balance concerns , could we not have this ? med hud with flash resistance + thermal lenses is going to be super powerfull in general.

mattroks101 (approved):

ValoTheValo:
> For balance concerns , could we not have this ? med hud with flash resistance + thermal lenses is going to be super powerfull in general.

for one you have to steal it off of them and thermal lenses are currently broken anyway so its fine

githubuser4141:
> For balance concerns , could we not have this ? med hud with flash resistance + thermal lenses is going to be super powerfull in general.

Would it be possible to add 'medHUD lenses' and make it so glasses can only hold one type of lenses? Maybe later down the line it'd open up eyewear customization

Humonitarian:
Personally, none of the HUDglasses should accept lenses - they already have a built-in computer messing with the image.

PR #6715, opened 2021-11-03
Makes havelock a little bit better

Iglpigl:
## About The Pull Request

Makes havelock do 10% damage and have 10% penetration.

## Why It's Good For The Game

It's a gun that was almost never used and the general consensus is that it's quite weak.

## Changelog
:cl:
tweak: tweaked havlocks damage and penetration, it now does 10% more of both
:cl:

### DIFF
--- a/code/modules/projectiles/guns/projectile/revolver/havelock.dm
+++ b/code/modules/projectiles/guns/projectile/revolver/havelock.dm
@@ -13,8 +13,8 @@
 	magazine_type = /obj/item/ammo_magazine/slpistol
 	matter = list(MATERIAL_PLASTEEL = 12, MATERIAL_WOOD = 6)
 	price_tag = 600
-	damage_multiplier = 1.4 //because pistol round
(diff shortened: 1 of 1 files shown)


MLGTASTICa:
It doesn't need a buff , 40% damage and 40% penetration at .35 is already a very good deal.

Humonitarian:
No buff needed, it can already compete with a Novak, especially when wielded.

A buff is fine if you raise the price, though, to something like 1200-1400.

TheShown911:
A small, cheap and mass produced gun is supposed to be bad, the problem isn't that this gun is useless, but that you can acquire top tier guns easily so this is never used.

Firefox13:
> 
> 
> A small, cheap and mass produced gun is supposed to be bad, the problem isn't that this gun is useless, but that you can acquire top tier guns easily so this is never used.

This. The current problem is amount of top tier guns, not the low-tier guns being awful.

PR #6726, opened 2021-11-05
General timer fix

MLGTASTICa:
## About The Pull Request

Prevents stacking timers in general , reason being that so far it only has caused issues and each item made their own shitty fix for it

## Why It's Good For The Game
No more stacking timers
## Changelog
:cl:
fix: Timers being stackable
/:cl:


TheShown911 (approved):
I foresee unforeseen consequences.

code/__HELPERS/mobs.dm
```diff
@@ -210,11 +210,18 @@ Proc for attack log creation, because really why not
...
+		
+/mob
+	var/currently_timing = FALSE
 
 /proc/do_after(mob/user, delay, atom/target, needhand = 1, progress = 1, var/incapacitation_flags = INCAPACITATION_DEFAULT, immobile = 1)
 	if(!user)
 		return 0
 
+	if(user.currently_timing)
+		to_chat(user, SPAN_NOTICE("You cannot do more than 1 action at a time!"))
```
drexample, line 222: Replace the numeral with the world "one"

drexample (changes requested):

Kegdo:
oh no 
ohn n noononononononono 
PLEASE NO
nononononononononononononononono

Humonitarian:
Maybe an argument should be given to the proc for whether it can stack?

For actions that **should** stack, for more flexibility.

TheShown911:
> Maybe an argument should be given to the proc for whether it can stack?
> 
> For actions that **should** stack, for more flexibility.

This is a valid concern, it's true that stackable timers cause bugs all of the time, but they also make a lot of other actions faster for the player.

MLGTASTICa:
> > Maybe an argument should be given to the proc for whether it can stack?
> > For actions that **should** stack, for more flexibility.
> 
> This is a valid concern, it's true that stackable timers cause bugs all of the time, but they also make a lot of other actions faster for the player.

To my knowledge , none of the current features make use of stacking intensively , and it always causes bugs everywhere. If there was any feature that made use of action stacking i wouldve added a override , but theres none as of speaking , atleast to my knowledge.

Humonitarian:
Shoveling dirt? Can we still stack that over multiple piles?

Using syringes on someone, can we still inject 15 units at once by stacking those timers?

I could think of more, probably

MLGTASTICa:
> Shoveling dirt? Can we still stack that over multiple piles?
> 
> Using syringes on someone, can we still inject 15 units at once by stacking those timers?
> 
> I could think of more, probably

dirt shoveling is fast and provides a lot per tile... and syringe stacking itself is also not meant to happen

ValoTheValo:
Syringe stacking is meant to be injecting entire syringe in one shot no? Also opening firedoors is another thing that needs stacking to not be absolute pain

Humonitarian:
> Syringe stacking is meant to be injecting entire syringe in one shot no? Also opening firedoors is another thing that needs stacking to not be absolute pain

This.

Kegdo:
> syringe stacking itself is also not meant to happen

maybe it wasn't, but it became a feature now. Really, some timers need to be locked, but actions like plowing dirt, injecting, **!!CONSTRUCTING!!**(seriously I will personally strangle you if I'll have to manually screw each and every window instead of just doing them all at once), these MUST be stackable.

MLGTASTICa:
> > syringe stacking itself is also not meant to happen
> 
> maybe it wasn't, but it became a feature now. Really, some timers need to be locked, but actions like plowing dirt, injecting, **!!CONSTRUCTING!!**(seriously I will personally strangle you if I'll have to manually screw each and every window instead of just doing them all at once), these MUST be stackable.

K , made them stack. You got a valid point.

Humonitarian:
Materials? Pretty cbt to place mats on girders one by one, although I'm not a techno main

Kegdo:
> Materials? Pretty cbt to place mats on girders one by one, although I'm not a techno main

That too.. Maybe add a define/type of sorts, so there would be hard(unstackable) timers and soft(stackable) timers?

PR #6727, opened 2021-11-05
Revert "Fixes not being able to harm borgs with certain tools"

drexample:
This change is good and all but it somehow screwed up the linters.


Kegdo:
>Linter is angry because if(ABORT_CHECK) is always true since ABORT_CHECK is defined as -1 but let's not break everything since it seems to be a legacy thing.

It broke everything...........

PR #6743, opened 2021-11-08
Buff excel turret

SirRichardFrancis:
## About The Pull Request

Makes excelsior turret shoot in bursts of 8 bullets, wildly inaccurate at that.
Still an easy kill, if not rushing in melee, but that's something at least.

Demonstration:
https://www.youtube.com/watch?v=cv0WPZXzpgE
UPD: Turret direction no longer lags behind and updates before each shoot instead of before each burst.

<details>
<summary>
Suggestions
</summary>

image

image

</details>

## Why It's Good For The Game

Considering how expensive that turret is to make (requires Maxim LMG), performance of it is laughable.

## Changelog
:cl:
balance: excelsior turret shoot in bursts of 8
/:cl:

### DIFF
--- a/code/game/machinery/excelsior/ex_turret.dm
+++ b/code/game/machinery/excelsior/ex_turret.dm
@@ -145,11 +145,11 @@
 
 	if(target)
 		last_target = target
-		set_dir(get_dir(src, target))
 		for(var/i; i < burst_lenght; i++)
 			if(!ammo)
 				break
 			sleep(2)
+			set_dir(get_dir(src, target))
 			shootAt(target)
 
 		return 1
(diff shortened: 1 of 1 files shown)


Voldirs:
Some suggs on that matter:
Turrets have a problem, they can be easily borked using scope. So if turret gets shot and there are no enemies in view, but it was shot from within line of view, check if it isn't a stupid commie that shot it by mistake. If turret doesn't see where did it get shot from, have turret blast several bursts offscreen to location where shot came from initially.
Same logic could be used on some boss mobs, but that's out of this PR scope

TheShown911 (approved):
As accurate as an average excel player.

PR #6747, opened 2021-11-08
[RTM] Cover combat update 1

Humonitarian:
## About The Pull Request

Makes bullets use targeted zones when flying over tables and low walls, now hitting them only when the legs or groin were targeted.
Refactors the targeting check that updates the def_zone of the bullet based on distance travelled, now can be called multiple times.

Allows holding items while resting, as well as manipulating them. Tactical rest can be used to dodge bullets while reloading, or managing your inventory. Players are prevented from firing guns, hittings things, using tools, or otherwise manipulating the world in any way with their newfound items.

While on paper ready for review, it should not be merged until my say so due to potential exploits still present within the code.

## Why It's Good For The Game

XCOM Eris > shoot until horizontal

## Changelog
:cl:
balance: tables and low walls protect the leg instead of having a random chance to eat bullets
refactor: refactored the targeting check of bullets that updates the targeted limb when hitting a mob
balance: resting now doesn't force you to drop items
balance: you can manage your inventory while resting, but not use them to interact with the world
/:cl:


Jan-Blasiak:
I think there still should be SOME chance of hitting a table or low wall when shooting at the chest.

Also: Make it so when people LIE behind a low wall or overturned table they are impossible to target outside of 1/2 tiles range.

code/modules/projectiles/projectile/bullets.dm
```diff
@@ -59,6 +59,10 @@
 		if(D.glass) chance *= 2
 	else if(istype(A, /obj/structure/girder))
 		chance = 100
+	else if(istype(A, /obj/structure/low_wall))
+		chance = round(penetrating * armor_penetration * 2 / 150 * 180) // hardcoded, value is same as steel wall, will have to be changed once low walls have integrity
```
MLGTASTICa, line 63: don't hardcode , we have material integrity for a reason.

MLGTASTICa (commented):
I suggest that you do all the checks when a bullet is fired and change its target if the cover is meant to block it than handle this way.

Humonitarian:
> I think there still should be SOME chance of hitting a table or low wall when shooting at the chest.
> 
> Also: Make it so when people LIE behind a low wall or overturned table they are impossible to target outside of 1/2 tiles range.

There is, whenever a bullet flies over a table it will roll on the target location randomization table depending on distance since the last check, as well as accuracy.

So the longer the distance, the more chance a bullet hits the table.

code/modules/projectiles/projectile/bullets.dm
Humonitarian: low walls don't

Humonitarian:
> I suggest that you do all the checks when a bullet is fired and change its target if the cover is meant to block it than handle this way.

Wouldn't take distance into account, especially against a moving target.

MLGTASTICa:
> > I think there still should be SOME chance of hitting a table or low wall when shooting at the chest.
> > Also: Make it so when people LIE behind a low wall or overturned table they are impossible to target outside of 1/2 tiles range.
> 
> There is, whenever a bullet flies over a table it will roll on the target location randomization table depending on distance since the last check, as well as accuracy.
> 
> So the longer the distance, the more chance a bullet hits the table.

This should be lowered by vigilance.

Humonitarian:
> > > I think there still should be SOME chance of hitting a table or low wall when shooting at the chest.
> > > Also: Make it so when people LIE behind a low wall or overturned table they are impossible to target outside of 1/2 tiles range.
> > 
> > 
> > There is, whenever a bullet flies over a table it will roll on the target location randomization table depending on distance since the last check, as well as accuracy.
> > So the longer the distance, the more chance a bullet hits the table.
> 
> This should be lowered by vigilance.

Vigilance is already taken into account in the miss-chance.

This is bullet spread, player accuracy is not taken into account here.

MLGTASTICa:
> > > > I think there still should be SOME chance of hitting a table or low wall when shooting at the chest.
> > > > Also: Make it so when people LIE behind a low wall or overturned table they are impossible to target outside of 1/2 tiles range.
> > > 
> > > 
> > > There is, whenever a bullet flies over a table it will roll on the target location randomization table depending on distance since the last check, as well as accuracy.
> > > So the longer the distance, the more chance a bullet hits the table.
> > 
> > 
> > This should be lowered by vigilance.
> 
> Vigilance is already taken into account in the miss-chance.
> 
> This is bullet spread, player accuracy is not taken into account here.

should be taken into account , again. I wouldn't want tables fully blocking my shots because im using a scoped wintermule with 90 vig. You are taking bullet  distance into account that means people will be forced to use CQC when anyone ever does tables , which becomes a 1-sided fight the very moment someone would set them up in a tight hallway.

Humonitarian:
> > > > > I think there still should be SOME chance of hitting a table or low wall when shooting at the chest.
> > > > > Also: Make it so when people LIE behind a low wall or overturned table they are impossible to target outside of 1/2 tiles range.
> > > > 
> > > > 
> > > > There is, whenever a bullet flies over a table it will roll on the target location randomization table depending on distance since the last check, as well as accuracy.
> > > > So the longer the distance, the more chance a bullet hits the table.
> > > 
> > > 
> > > This should be lowered by vigilance.
> > 
> > 
> > Vigilance is already taken into account in the miss-chance.
> > This is bullet spread, player accuracy is not taken into account here.
> 
> should be taken into account , again. I wouldn't want tables fully blocking my shots because im using a scoped wintermule with 90 vig. You are taking bullet distance into account that means people will be forced to use CQC when anyone ever does tables , which becomes a 1-sided fight the very moment someone would set them up in a tight hallway.

It doesn't matter how good of a shoot you are. If you shoot toast from a toaster, chances are it will have a pretty wild trajectory.

I could refactor the whole bit, on paper, but it wouldn't be in the scope of a quick cover rework.

Humonitarian:
The scope of this PR has changed.
To the moon!

code/_onclick/click.dm
```diff
@@ -125,7 +125,7 @@
 			return 1
 		throw_mode_off()
 
-	var/obj/item/W = get_active_hand()
+	var/obj/item/W = get_active_hand()
```
TheShown911, line 128:
```suggestion
	var/obj/item/W = get_active_hand()
```

code/modules/projectiles/projectile/plasma.dm
```diff
@@ -26,4 +26,17 @@
...
+		for(var/i in damage_types)
+			damage_types[i] *= loss
 
+		A.visible_message(SPAN_WARNING("\The [src] is weakened by the \the [A]!"))
+		playsound(A.loc, 'sound/weapons/shield/shielddissipate.ogg', 50, 1)
+		return 1
+	else if(istype(A, /obj/structure/barricade) || istype(A, /obj/structure/table) || istype(A, /obj/structure/low_wall))
+		return 0
+
+	return 1
```
TheShown911, line 42: Newline.

TheShown911 (approved):
Please make returns return TRUE or FALSE, not 1 or 0. Also make the results of if statements indented under them, not continued in the same line.

PR #6751, opened 2021-11-10
Armor stacking with fall-off and fractional multiplication

Humonitarian:
## About The Pull Request

Pretty radical change, makes armor stack multiplicatively (eg. 60% bulletproof with 10% rockerboy jumpsuit becomes 64% bullet protection). There are other safety features in mind, such as any armor against melee, bullet and energy is averaged with 75 once above 75. (eg. 80 armor becomes (75+80)/2)

_**Do not merge**_
Until properly tested.

## Why It's Good For The Game

Armor will make a lot more sense now, limbs will be able to take armor from organs and from being robotic, generally just a lot more pleasant to make features for.

## Changelog
:cl:
balance: armor stacks
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -166,7 +166,7 @@ meteor_act
 	var/protection = 0
 	var/list/protective_gear = list(head, wear_mask, wear_suit, w_uniform, gloves, shoes)
 	if(def_zone.armor)
-		protection = 100 - (100 - def_zone.armor.getRating(type)) * (100 - protection) * 0.0001 // Converts armor into multiplication form, stacks them, then converts them back
+		protection = 100 - (100 - def_zone.armor.getRating(type)) * (100 - protection) * 0.01 // Converts armor into multiplication form, stacks them, then converts them back
 
 	for(var/gear in protective_gear)
 		if(gear && istype(gear ,/obj/item/clothing))
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -137,11 +137,11 @@ meteor_act
 				armorval += getarmor_organ(organ, type) * weight
 				total += weight
 	
-	armorval =/ max(total, 1)
+	armorval = armorval/max(total, 1)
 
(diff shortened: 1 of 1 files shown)


Humonitarian:
This took me about 5 minutes to make.

SpadesSlick:
I was thinking of doing exactly that back when I originally made armor rework. I decided against that on account that it would lead to minmaxing of all possible gear instead of just suits. I see you addressing this, but it might not be enough.

All in all, at the moment I lack involvement to make a proper call on that, but there is something I'm sure about about.

Original armor rework was done to make armor actually work and be viable, instead of tying it all to RNG. It achieved that, but several other problems were created. If there is really a need to actually change armor, it should be done by creating a new system that takes in account those things:

* Armor and it's balance should make customization and general player expression viable. Style system was done to do just that, but I'm unaware if it's viable at the moment.
* Departments and factions should have options of armor to remain distinct from each.
* Armor must not only be viable, but also should support more dynamic and fun combat.

Trying to adjust numbers, formulas and doing another massive armor rebalance (as you mentioned in this PR) is essentially defeat in detail, a cargo cult of development that only strains maintainers. Moving around numbers won't make game more fun, won't progress development, and won't make players exited for new updates. New approaches and ideas must be applied and tested.

New coders, in general, should focus their attention on creating new things. Even if they are bad, degrade game and are generally bad idea, like Blitzshell, it would be a lesson learnt, or maybe idea that can be salvaged. A rejected PR with a feature is something you still can learn from. But endless stream of rebalances would not make you improve yourself in any way.

maikilangiolo:
I miss you reere come back we can make out in the club bathrooms.

Also I think this is an elegant PR to fix armor issue. The only thing I'd suggest is preventing armor stacking of heavy armours and suit plates. Like you can't layer IH ballistic with idk riot plates for giant melee and bullet resist. It would have to be done manually I think but there are not many heavy armours in game

Alliostra:
how

sl0w3st:
> I miss you reere come back we can make out in the club bathrooms.

The hell happened here, where is reere?

drexample:
holy shit I thought reere died

Humonitarian:
> **_Do not merge_**
> Until properly tested.

https://c.tenor.com/_bTaLmoLSc4AAAAM/troll-pilled.gif

PR #6752, opened 2021-11-10
update animal ai

HentaiStorm:
## About The Pull Request

upgrade superior animal AI

## Why It's Good For The Game

Mobs will look for new nearby targets when receiving damage

## Changelog
:cl:
tweak: superior animal ai
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/superior_animal/defense.dm
+++ b/code/modules/mob/living/carbon/superior_animal/defense.dm
@@ -186,7 +186,7 @@
 	if (overkill_dust && (amount >= overkill_dust) && (getFireLoss() >= maxHealth*2))
 		dust()
 
-/mob/living/carbon/superior_animal/proc/reagr_new_targets(var/reagr_radius = 1))
+/mob/living/carbon/superior_animal/proc/reagr_new_targets(reagr_radius = 1)
 	for(var/mob/living/carbon/superior_animal/SA in view(reagr_radius))
 		if(SA.stance == HOSTILE_STANCE_ATTACK)
 			SA.target_mob = SA.findTarget()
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/superior_animal/defense.dm
```diff
@@ -185,6 +186,11 @@
 	if (overkill_dust && (amount >= overkill_dust) && (getFireLoss() >= maxHealth*2))
 		dust()
 
+/mob/living/carbon/superior_animal/proc/reagr_new_targets(var/reagr_radius = 1))
```
SirRichardFrancis, line 189:
```suggestion
/mob/living/carbon/superior_animal/proc/reagr_new_targets(reagr_radius = 1)
```

MLGTASTICa:
Why is this needed ? the AI works fine most of the time. At most this change might be laggy when you have 10 roaches on a tile fighting a spider and them constantly cycling eachother by looping through view and through the mobs nearby needlessly .

HentaiStorm:
> 
> 
> Why is this needed ? the AI works fine most of the time. At most this change might be laggy when you have 10 roaches on a tile fighting a spider and them constantly cycling eachother by looping through view and through the mobs nearby needlessly .

There is a point when the mob pursues the target and does not pay attention to the attacks in the back. This is fix mob.

MLGTASTICa:
> HOSTILE_STANCE_ATTACK - This is moving to target, not attacking. If the mob is already attacking someone, the reagr will not execute.
> 
> https://github.com/discordia-space/CEV-Eris/blob/68d2bbbbab3bb4231f8501e73130f29b0b14be66/code/modules/mob/living/carbon/superior_animal/superior_animal.dm#L278

it still goes through everything in view(1) and calls find_target for how many superior animals are there. target_mob is also.. the target , not just moving. For what the performance impact could be i don't see why this is needed and if the behaviour it provides is of any use at all.

TheShown911 (approved):
This should work then the current system, might be problems with mobs re-aggro-ing on different targets in the same fight though.

PR #6786, opened 2021-11-25
Breakable splints

Kegdo:
## About The Pull Request

spcr made a salt pr and then closed it without fleshing it out, I decided that the idea is good so I stole his code and tweaked it.

splints fall off after getting more than 20 damage. For scale, wielded untoolmodded crowbar deals 19.5, unwielded untoolmodded butcher cleaver deals 20.
## Why It's Good For The Game

"No longer you can combat pre-splint to avoid a major part of the medical system and balance"

## Changelog
:cl:
balance: splints break and fall off after being hit, 20+ damage to break them.
/:cl:

### DIFF
--- a/code/modules/mob/living/living_defense.dm
+++ b/code/modules/mob/living/living_defense.dm
@@ -68,7 +68,7 @@
 		if(ishuman(src) && def_zone)
 			var/mob/living/carbon/human/H = src
 			var/obj/item/organ/external/o = H.get_organ(def_zone)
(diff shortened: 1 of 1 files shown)


TheShown911 (approved):
You sure this works?

Kegdo:
YES I TEST MY CODE

hyperioo:
You could make a small proc and call it. That way you won't have the exact that code at two places.

Kegdo:
on armored it's actual damage, and on unarmored it's, uh, total damage or something. So I think it's fine.

PR #6791, opened 2021-11-26
Fix smoke not being inhaled

Kegdo:
## About The Pull Request

Legacy code made it so glasses can block smoke(???), so it was removed
For smoke proc not to work it needs not only the airtight tag(e.g. voidsuits have airtight), but the block smoke tag, which only the gasmasks have. Shouldn't inhale smoke if internals are on.

## Why It's Good For The Game

Everyone wants the smoke to be fixed, so here it is. I am not 100% this will fix the issue, because the smoke is handled horrendously spaghetti, and nanako also "fixed" it years ago. Still, it at least SHOULD work.

## Changelog
:cl:
fix: smoke should be inhaled properly now. Maybe.
/:cl:


TheShown911:
I don't think this will do anything, also is there any clothing with both the airtight and block gas smoke flags, looks like overcomplication to me.

Kegdo:
> I don't think this will do anything, also is there any clothing with both the airtight and block gas smoke flags, looks like overcomplication to me.

There is, and it's called GAS MASKS.
Right now, some glasses can block the smoke just because they feel like it

PR #6833, opened 2021-12-04
Engineering supply pouch, but instead of reverting the change there's another pouch. Two pouches.

Wouju:
## About The Pull Request

Remember the old engineering supply pouch? where it had 2 slots and could also hold sheets of materials? Similar situation here. New pouch has  slots but can only hold sheets, rods and cable coil.

I am still salty about the nerf

image
image

My main worry is that the sprite may be comically large.

## Why It's Good For The Game

I didnt want to revert the supply pouch change, so instead i made a new one that fulfills the pre-change supply pouch's role

## Changelog
:cl:
add: Added new engineering pouch, can only hold sheets, rods and cable coil.
/:cl:


code/game/objects/items/weapons/storage/pouches.dm
```diff
@@ -162,6 +162,25 @@
 		/obj/item/gun/projectile/flare_gun
 		)
 
+/obj/item/storage/pouch/engineering_material
+	name = "engineering material pouch"
+	desc = "Can hold sheets, rods and cable coil. Has only 2 slots."
```
Kurgis, line 167: You should bring this in line with the other descriptions, "has only 2 slots" isn't fit.

Wouju: done, removed the "has only 2 slots". Couldnt think of anything else so its just "Can hold sheets, rods and cable coil"

TheShown911 (approved):

PR #6842, opened 2021-12-05
Reduces face shield vision obstruction for paramedic helmet

ACCount12:
## About The Pull Request

Knocks paramedic helmet face shield vision tint down from `TINT_MODERATE` (default value, same as Altyn) to `TINT_LOW` (same as IH "heavy operator" helmet).

It has less armor than the IH "heavy operator" helmet, even with its faceplate down - and is harder to obtain than Altyn.

## Why It's Good For The Game

Dunno. I was asked to make this PR, and so I did.

## Changelog
:cl:
balance: Reduces face shield vision obstruction for paramedic helmet
/:cl:

### DIFF
--- a/code/modules/clothing/head/armor.dm
+++ b/code/modules/clothing/head/armor.dm
@@ -575,6 +575,7 @@
 	up = TRUE
 	spawn_blacklisted = TRUE
 	style = STYLE_HIGH
+	tint_down = TINT_LOW
 	var/speaker_enabled = TRUE
 	var/scan_scheduled = FALSE
 	var/scan_interval = 15 SECONDS


Wouju:
good pr

SirRichardFrancis:
Buff armor values while you're at it.

Humonitarian:
> Buff armor values while you're at it.

No need for that considering

> Buff armor values while you're at it.

No need for that, as it is already stronger than the default operator armor, while offering other advantages as well.

Style should be lowered to very low for the whole set, considering this is armor, the suit even gives a speed buff which is pretty strong.

Don't forget that these are paramedics we are speaking of. They are less often targets of killing due to their position, thus head armor does not need to be better than the suit.

We don't want even more armor bloat.

At the very moment they are literally the best choice for a style suit, even beating the triad overcoat and odd helmet in protection, while offering the same style

TheShown911 (approved):

PR #6845, opened 2021-12-05
Fixing punching with one line of code.

hex37:
## Why It's Good For The Game
Allows Punching to work properly for people with Special(tm) organs implanted, and removes a previously unknown, super niche exploit for everyone else.

Up until now, the organ efficiency in the active arm of THE PERSON BEING PUNCHED determined how much damage they took from punching. This meant many stupid things could happen, such as: If someone removed the blood vessels, nerves, and muscles in their right arm, they would be almost immune to being punched by other players, so long as they had their right hand selected.

This was intensely stupid, so I fixed it to use the limb efficiency of the person doing the punching instead.

Textbook example of why using bad variable names results in bad readability, which results in bad code. H.

## Changelog
:cl:
fix: Fix punching that was computed with the target's hand instead of the attacker's hand
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_attackhand.dm
+++ b/code/modules/mob/living/carbon/human/human_attackhand.dm
@@ -144,7 +144,7 @@
 			var/accurate = 0
 			var/hit_zone = H.targeted_organ
 			var/obj/item/organ/external/affecting = get_organ(hit_zone)
-			var/obj/item/organ/external/current_hand = organs_by_name[hand ? BP_L_ARM : BP_R_ARM]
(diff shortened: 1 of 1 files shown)


mattroks101 (approved):

Jan-Blasiak:
Filleting your arms to become a God of CQC is quite based.

TheShown911:
Clean up the CL, way too long.

Kegdo:
comically large one-liner fix changelog

hyperioo:
Updated the changelog myself so that it could be merged today. I put the previous text in "Why It's Good For The Game"

PR #6847, opened 2021-12-06
REMOVE melee and throw dodging, fix missing buckled mobs, fix division by zero in throwing knives

Kegdo:
## About The Pull Request
removes the old ass bay12 leftover get_zone_with_missing_chance proc, that overwrote some given vars and did useless checking + was calculating missing buckled mobs incorrectly. It was also removed to implement stat-based dodging.
dodging melee formula: __((base_miss_chance[hit_zone] * (5 /(1 + 100 / dodge) + 10))__ //soft cap
dodging  the thrown items formula: __(base_miss_chance[zone] * (10 / (1 + 100 / dodge ) + 10)) /10__ //soft cap
## Why It's Good For The Game
more ROB scaling, better dodge, fixes cool, bay leftover bad

## Changelog
:cl:
balance: dodging melee attacks and thrown items scales off ROB
del: bay leftover get_zone proc
fix: missing buckled mobs with thrown items
fix: division by zero in throwing knives code
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -276,7 +276,7 @@
 	if(istype(item, /obj/item/stack/throwing_knife))
 		var/obj/item/stack/throwing_knife/V = item
 		var/ROB_throwing_damage = stats.getStat(STAT_ROB)
-		V.throwforce = 35 / (1 + 100 / ROB_throwing_damage) + 10 //soft cap; This would result in knives doing 10 damage at 0 rob, 20 at 50 ROB, 25 at 100 etc.
+		V.throwforce = 35 / (1 + 100 / ROB_throwing_damage + 1) + 10 //soft cap; This would result in knives doing 10 damage at 0 rob, 20 at 50 ROB, 25 at 100 etc. + 1 is here to avoid division by zero
 		if(V.amount == 1)
 			drop_from_inventory(V)
 			V.throw_at(target, item.throw_range, item.throw_speed, src)
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -275,8 +275,8 @@
 
 	if(istype(item, /obj/item/stack/throwing_knife))
 		var/obj/item/stack/throwing_knife/V = item
-		var/ROB_throwing_damage = stats.getStat(STAT_ROB)
-		V.throwforce = 35 / (1 + 100 / ROB_throwing_damage + 1) + 10 //soft cap; This would result in knives doing 10 damage at 0 rob, 20 at 50 ROB, 25 at 100 etc. + 1 is here to avoid division by zero
+		var/ROB_throwing_damage = max(stats.getStat(STAT_ROB), 1)
+		V.throwforce = 35 / (1 + 100 / ROB_throwing_damage + 10) //soft cap; This would result in knives doing 10 damage at 0 rob, 20 at 50 ROB, 25 at 100 etc.
 		if(V.amount == 1)
 			drop_from_inventory(V)
 			V.throw_at(target, item.throw_range, item.throw_speed, src)
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -213,9 +213,19 @@ meteor_act
 
 	if(user == src) // Attacking yourself can't miss
 		return target_zone
-
-	var/hit_zone = get_zone_with_miss_chance(target_zone, src)
-
+	var/hit_zone = check_zone(target_zone)
+	//check if we hit
+	var/dodge = src.stats.getStat(STAT_ROB)
+	var/miss_chance = ((base_miss_chance[hit_zone] / 4) + (20 / (1 + 100 / (dodge + 1)) + 10)) //soft cap, minumum 10 at rob 0, 16.5 at 50, 20 at 100, 22 at 150,
+	if(prob(miss_chance))																//max - ~24, ignoring the base_miss_chance that adds a bit more
+		hit_zone = null																//+ 1 is there to avoid division by zero
+	// you cannot miss if your target is prone or restrained
+	if(src.buckled || src.lying)
+		return hit_zone
+	// if your target is being grabbed aggressively by someone you cannot miss either
+	for(var/obj/item/grab/G in src.grabbed_by)
+		if(G.state >= GRAB_AGGRESSIVE)
+			return hit_zone
 	if(!hit_zone)
 		visible_message(SPAN_DANGER("\The [user] misses [src] with \the [I]!"))
 		return null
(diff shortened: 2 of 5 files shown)


Kegdo:
my calculations with the examples for the formulas may be incorrect

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -276,7 +276,7 @@
 	if(istype(item, /obj/item/stack/throwing_knife))
 		var/obj/item/stack/throwing_knife/V = item
 		var/ROB_throwing_damage = stats.getStat(STAT_ROB)
-		V.throwforce = 35 / (1 + 100 / ROB_throwing_damage) + 10 //soft cap; This would result in knives doing 10 damage at 0 rob, 20 at 50 ROB, 25 at 100 etc.
+		V.throwforce = 35 / (1 + 100 / ROB_throwing_damage + 1) + 10 //soft cap; This would result in knives doing 10 damage at 0 rob, 20 at 50 ROB, 25 at 100 etc. + 1 is here to avoid division by zero
```
Humonitarian, line 279: You might want to put the ROB_throwing_damage +1 in parentheses, or better yet, as max(rob, 1)

CeUvi:
why does everything scale with ROB.
we have FIVE other stats that barely do anything, and TWO of them are combat stats.

SugarBone:
Why is the base miss chance divided by 4? It decreases chances with higher miss probabilities by a greater amount, thus resulting in a substantial increase in the probability of hitting those attacks, while for easy to hit zones, it barely offsets it. Thus, at the minimum ROB of 0, you would have increased miss probabilities for easy-to-hit parts and lowered probabilities for the head. For instance, at ROB 0: head = ((45/4 + 10) = 21.25) vs. chest = ((10/4 + 10) = 12.5)

Kegdo:
> why does everything has to scale with ROB. 

cuz melee. ROB is everything about melee, not only strength. And yes, throwing is melee, too
Vig is about hitting/dodging(?) bullets, so it won't work here.
Tgh is your toughness, not agility.

>we have FIVE other stats that barely do anything, and TWO of them are combat stats.

>barely do anything

funny, laughed

>a substantial increase in the probability of hitting those attacks, while for easy to hit zones, it barely offsets it.

indeed, what's your solution then?

CeUvi:
> Vig is about hitting/dodging(?) bullets, so it won't work here.

ah yes, VIG is """gun only""" stat.

ROB is about hitting/blocking, so it won't work here.
(ROB is basically "strength" stat.)

Kegdo:
> ROB is about hitting/blocking, so it won't work here. (ROB is basically "strength" stat.)

hmm, yes, allow me to quote the ROB stat desc from our wiki:
In terms of combat, it __reduces the chances of your weapon becoming embedded in a target__, as __well as increasing the chance of actually hitting a target.__ 
Not only ROB currently doesn't actually increase the chance of hitting the target, but it also reduces the embed chance. This is melee mastery, a knife won't stop embedding in flesh if you just push it harder.

Kegdo:
>Why do you need all this then? It sounds fun, but, as you are pointing out, there is little incentive to use it

style. And rage cage fighting. Also there is, again, currently no special grab moves with chest and groin.

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -213,9 +213,19 @@ meteor_act
 
 	if(user == src) // Attacking yourself can't miss
 		return target_zone
-
-	var/hit_zone = get_zone_with_miss_chance(target_zone, src)
-
+	var/hit_zone = check_zone(target_zone)
+	//check if we hit
+	var/dodge = src.stats.getStat(STAT_ROB)
+	var/miss_chance = ((base_miss_chance[hit_zone] / 4) + (20 / (1 + 100 / (dodge + 1)) + 10)) //soft cap, minumum 10 at rob 0, 16.5 at 50, 20 at 100, 22 at 150,
```
Humonitarian: I think altogether swapping the addition with a multiplication would be better (base chance * (1+softcapped dodge), with the value tweaked accordingly for a similar final result, but works well eitherways in terms of balance.

Humonitarian: This would allowing removing the /4, satisfying Sugarbone's criteria.
It can be further balanced down by decreasing the "1" into 0.9, decreasing the default while allowing it to keep its strong scale with the Head.

Humonitarian:
> > Vig is about hitting/dodging(?) bullets, so it won't work here.
> 
> ah yes, VIG is """gun only""" stat.
> 
> ROB is about hitting/blocking, so it won't work here. (ROB is basically "strength" stat.)

Wrong there, Robustness is general brawling and bar-fu ability, as well as the ability to handle melee and thrown weapons.

Humonitarian:
> why does everything has to scale with ROB. we have FIVE other stats that barely do anything, and TWO of them are combat stats.

According to Montague, Cog is the best stat at the moment. I'm not joking.

code/modules/mob/living/carbon/human/human_defense.dm
Kegdo: just gimme the formula, my head hurts when think

TheShown911: Don't, make him think

SaltyOldBull:
Wait for style dodge, do not merge.

GrayRachnid:
It would be much better if this scaled off Vigilance instead of Robust

Humonitarian:
Very good points are being raised, RNG for melee and especially throwing is pretty bad for the game. The very act of using these combat methods over a burst-fire gun is already a massive gamble. No need to punish NeoTheology mains with even more RNG.
Throwing in general is pretty easy to miss, and opponents can pretty easily throw things back. A very hit or miss kind of deal with a lot of risk involved.

Making Rob allow players to catch thrown items, or prevent others from catching theirs would be a good feature, however. I believe that would involve scaling the speed of thrown items by Rob, as well as the catching speed limit, which I believe to be 5 at the moment.
I don't actually see a lot of issues with removing melee missing altogether given thrown items aren't hitscan, save for the body-part based missing. At the moment for guns Vigilance scales to increase the chance of hitting the head and limbs, you could use the same calculation but scaling with Rob for thrown items and punching.

Kegdo:
I want to remind you again that there already is a RNG chance to miss, it's just a shitty bay leftover right now

Kegdo:
> Very good points are being raised, RNG for melee and especially throwing is pretty bad for the game.

It's already there????¿¿??¿¿¿??¿

>no need to punish NeoTheology mains with even more RNG.

Don't they have a 75dmg throwing spear now?
Also, NT is not the only ones who use melee and throwing weaponry. I use it too.

> Throwing in general is pretty easy to miss

Clearly a skill issue
Really
It's not hard

> Making Rob allow players to catch thrown items, or prevent others from catching theirs would be a good feature, however.

You can already catch it? And autocatching is a bad idea?

Humonitarian:
> > Very good points are being raised, RNG for melee and especially throwing is pretty bad for the game.
> 
> It's already there????¿¿??¿¿¿??¿

Yes, remove it.

> > no need to punish NeoTheology mains with even more RNG.
> 
> Don't they have a 75dmg throwing spear now? Also, NT is not the only ones who use melee and throwing weaponry. I use it too.

Only if they get RNG. Imagine, after 2 hours of gaming you finally get the RNG to print javelins. You throw the javelin. RNG fucks you over and you miss, and become shredded cheese from a Takeshi.

> > Making Rob allow players to catch thrown items, or prevent others from catching theirs would be a good feature, however.
> 
> You can already catch it? And autocatching is a bad idea?

Please read the second part of the sentence slowly, deliberately.

Kegdo:
>remove it

not a solution, because, apparently, SugarBone will use the missing mechanic in his melee rework. He already asked me to give him the miss sound files.


>they miss and die

If you print only one pilum, that's a you issue. Their shield can hold three of them.
And if you can't run away/stay in groups you are already fucked.


>read properly

no.

Voldirs:
Just remove random dodging altogether

Firefox13:
RNG in PvP combat is a sin.

drexample (commented):
RNG in PvP combat is a sin.

Kegdo:
then the rng goes into the trash bin
guhhhhhhhhhhhhhhhhhhhh.

drexample (approved):
Controversial, but let's see how players feel.

Kegdo:
spoiler alert, they WILL be salty, one way or another. I just hope I didn't just make throwing knives and javelins OP as balls

PR #6848, opened 2021-12-06
[TESTING REQUIRED]Remove #2811 click catchers 

Kegdo:
## About The Pull Request

the said click catchers spawn on every tile on your screen(like hud does), only to catch your clicks. Now, what does click catching do?
well, if you click on it with LMB/RMB, it calls the Click proc.
And if you click on it with MMB, it swaps your hand. That's it.
We have the MiddleClickOn proc for that.
image
Ergo it doesn't do anything but just... exist here. The fact that it is called "unnamed" terrifies me when I test stuff on local.

(also I am currently sleep deprived so it may break some shit but I am 59% sure it won't)
## Why It's Good For The Game
probably huge optimisation. And some QoL, you will never wonder 'what's that "unnamed" thing' any more.
## Changelog
:cl:
del: click catchers that catched your clicks only to swap your hand
/:cl:

### DIFF
--- a/code/__DEFINES/_planes+layers.dm
+++ b/code/__DEFINES/_planes+layers.dm
@@ -51,8 +51,6 @@ What is the naming convention for planes or layers?
 //Defines for atom layers and planes
 //KEEP THESE IN A NICE ACSCENDING ORDER, PLEASE
 
-#define CLICKCATCHER_PLANE -99
-
 #define PLANE_SPACE -95
 #define PLANE_SPACE_PARALLAX -80
 
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -366,46 +366,6 @@
 	set_dir(ndir)
 	return 1
 
-
-
-GLOBAL_LIST_INIT(click_catchers, create_click_catcher())
-
-/obj/screen/click_catcher
-	icon = 'icons/mob/screen_gen.dmi'
-	icon_state = "click_catcher"
-	plane = CLICKCATCHER_PLANE
-	mouse_opacity = 2
-	screen_loc = "CENTER-7,CENTER-7"
-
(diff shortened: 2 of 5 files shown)


zombiemimic:
you better have tested that if you want your lungs to continue functioning

Kegdo:
well yes, I tested it
the game didn't explode and turfs didn't perish to black death
but uhhhhhhhh
should there be a TESTING REQUIRED tag?

SirRichardFrancis:
That does break MMB hotkey, doesn't it?

Kegdo:
funnily enough, no. It's a thing that does another thing upon being clicked on with MMB, the MMB clicking itself is untouched,

MLGTASTICa:
Did you test if full auto still works after this ? same for charge mode and gun firing in general ? they all use CH, also these very little performance gained by removing them.

TheShown911:
Click handlers are different to click catchers, though i feel this will breaks stuff.

mattroks101:
Click catchers were a shitty port from bay when having click catchers was trendy. We never made use of them the way Bay did. There's no point in them being here. Good to test it, but I'm pretty sure nothing will break.

Humonitarian:
Do RIG modules still work? They use middleclicking, and I have a feeling click catchers are used to determine the turf.

Kegdo:
no, they use OnMiddleClick proc, not a click catcher

SirRichardFrancis:
Breaks shooting if target area is not visible.

image

Kegdo:
oh my fucking god OnMiddleClick doesn't work if you click on a dark empty hollow out of vision tile

Kegdo:
> Breaks shooting if target area is not visible.

Despite this CLEARLY being a skill issue, I will remake this pr, but to do it I need to start from the scratch

PR #6852, opened 2021-12-07
Buffs gun assembly so it does not require great luck to get good guns

vode-code:
## About The Pull Request
This PR rebalances the gun assembly so each option increases the minimum price by 300 instead of 100.

## Why It's Good For The Game
The gun assembly is bad, and a streamlined and better version is coming, so this PR is made so that it can be good for those that don't have great luck.

## Changelog
:cl: Chickenish
balance: Gun assemblies now should have good guns without relying on RNG.
/:cl:


TheShown911 (approved):

hyperioo:
The fact that they named the variable `canidates` instead of `candidates` pains me ;-;

PR #6856, opened 2021-12-10
Tweaks Oxycodone

Anonmare:
## About The Pull Request
Oxycodyne once again has the same metabolic rate as the rest of the painkillers and has 10 more pain relief, putting it on the same level as ParaTram (but still takes 20 more NSA). It was either this or make every other painkiller metabolise at the same speed as oxycodone and I took the path of least resistance.

## Why It's Good For The Game
Inconsistency isn't fun design and the painkiller that takes more effort to make shouldn't be objectively worse than its precursors. At least now it's on par with its contemporary, Tramadol, in terms of how much pain relief is provided per NSA.

## Changelog
:cl:
balance: Oxycodyne once again has the same metabolic rate as the rest of the painkillers and has 10 more pain relief, putting it on the same level as ParaTram (but still takes 20 more NSA).
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/medicine.dm
+++ b/code/modules/reagents/reagents/medicine.dm
@@ -237,11 +237,11 @@
 	reagent_state = LIQUID
 	color = "#800080"
 	overdose = REAGENTS_OVERDOSE * 0.66
-	metabolism = 0.1
+	metabolism = 0.02
 	nerve_system_accumulations = 60
 
 /datum/reagent/medicine/oxycodone/affect_blood(mob/living/carbon/M, alien, effect_multiplier)
-	M.add_chemical_effect(CE_PAINKILLER, 65)
+	M.add_chemical_effect(CE_PAINKILLER, 75)
 	M.druggy = max(M.druggy, 10)
 
 /datum/reagent/medicine/oxycodone/overdose(mob/living/carbon/M, alien)


Kegdo:
Sounds meh, maybe also increade oxy pain tolerance? 20+ nsa is a lot, paratram should be weaker than oxy.

Humonitarian:
The fast metabolization is what made Oxy unique. I'd rather see the recipe changed to result in 2 units, so it isn't that slow to make.

Don't forget Oxy is balanced around being an Antag item rather than being a chem made in chemistry.

Anonmare:
@Kegdo I'm already pushing my luck by even suggesting undoing the nerf by this much
@Humonitarian Any antag worth his salt would get Tramadol for it's incredibly long processing time, ease of making and much higher OD threshold. Oxy processing 10x as fast is _crippling_ and makes it an objectively inferior choice

Humonitarian:
> @Kegdo I'm already pushing my luck by even suggesting undoing the nerf by this much @Humonitarian Any antag worth his salt would get Tramadol for it's incredibly long processing time, ease of making and much higher OD threshold. Oxy processing 10x as fast is _crippling_ and makes it an objectively inferior choice

Good. When it comes to things like medical, uplink should be inferior to what factions can produce.

Anonmare:
If it's antag balancing, then antags should be getting the weaker painkillers, or, Hell, give them their own antag-specific chems instead of having to balance something both for and against crew. It makes no sense that the painkillers that take more effort to make are the worst

TheShown911 (approved):
There was no nerf on painkillers, as before they didn't work and now they do. Items should not be buffed for consistency, but because they are underpowered. Comparing it to other painkillers it is so I approve.

Humonitarian:
Great, another identical painkiller that remains in the system for minutes outshining all other chems, no variety.

TheShown911:
Like every other painkiller?

Humonitarian:
> Like every other painkiller?

Exactly my point!

MLGTASTICa:
Should just remove the druggy effect and make tramadol slightly weaker

PR #6858, opened 2021-12-10
Super minor QoL tweaks

koghs:
## About The Pull Request

Added material cost to items that obviously miss it (including newly added throwing knives), added ability to put appropriate oddities into wallet, increased melee damage for Cog (it's basically a laser equivalent to Novakovic, except it has nothing interesting or funky about it). I think it is fair considering Cog is an old and bulky police weapon, and those guys sure love bashing people skulls. 

## Why It's Good For The Game

Makes joonking and getting rid of items slightly less annoying and disappointing, adds some interesting property to old good Cog, makes so you can now put money into your wallet, wow. 

## Changelog
:cl:
tweak: Added material cost to toolboxes, throwing knives, suit cooling unit, increased OS LG "Cog" melee damage, added ability to put appropriate oddities into wallet.
/:cl:

### DIFF
--- a/code/game/objects/items/devices/suit_cooling.dm
+++ b/code/game/objects/items/devices/suit_cooling.dm
@@ -5,6 +5,7 @@
 	icon = 'icons/obj/device.dmi'
 	icon_state = "suitcooler0"
 	slot_flags = SLOT_BACK	//you can carry it on your back if you want, but it won't do anything unless attached to suit storage
+	matter = list(MATERIAL_STEEL = 8, MATERIAL_GOLD = 4)
 
 	//copied from tank.dm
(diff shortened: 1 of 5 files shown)


koghs:
All tested I swear, works as intended.
Minus fact that throwing knives can be accidentally recycled in autolathe because they are technically stacks, but they are for throwing anyway right?

Humonitarian:
Make sure oddities work while in wallets on ID slot, that's the only problem I can imagine

koghs:
Ya I did 
Works fine, they are the same thing as boxes after all.

hyperioo (approved):

PR #6863, opened 2021-12-12
Paramedic armor slowdown rework

SirRichardFrancis:
## About The Pull Request

Replaced constant 0.3 speed boost of paramedic's armor with 0.5, that can be active for ~15~ 30 seconds and have a cooldown of 5 minutes.
Added materials of medium atomcell to armor cost.
If paramedic helmet is equipped, there will be notifications on activation, deactivation and cooldown end.

## Why It's Good For The Game

People are salty over constant speed boost, probably that will work out better.

## Changelog
:cl:
tweak: paramedic armor's constant 0.3 speed boost replaced with 0.5, that can be active for 30 seconds and have a cooldown of 5 minutes
/:cl:

### DIFF
--- a/code/modules/clothing/suits/armor.dm
+++ b/code/modules/clothing/suits/armor.dm
@@ -587,7 +587,7 @@
 	var/speed_boost_ready = TRUE
 	var/speed_boost_active = FALSE
 	var/speed_boost_power = -0.5
-	var/speed_boost_length = 15 SECONDS
+	var/speed_boost_length = 30 SECONDS
 	var/speed_boost_cooldown = 5 MINUTES
 	var/matching_helmet = /obj/item/clothing/head/armor/faceshield/paramedic
 
--- a/code/modules/clothing/suits/armor.dm
+++ b/code/modules/clothing/suits/armor.dm
@@ -568,7 +568,9 @@
 		MATERIAL_PLASTEEL = 10,
 		MATERIAL_STEEL = 5,
 		MATERIAL_PLASTIC = 5,
-		MATERIAL_PLATINUM = 3
+		MATERIAL_PLATINUM = 3,
+		MATERIAL_URANIUM = 4,
+		MATERIAL_SILVER = 2
 		)
 	armor = list(
 		melee = 30,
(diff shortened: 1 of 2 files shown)


Humonitarian:
Imo remove platinum from price and make base slowdown 0.05, because with the cooldown added its price won't justify the advantages.
Also, consider giving them a helmet light that only works while the mask is open, would be better to give the open helmet an advantage than the closed helmet a tint disadvantage.

GrayRachnid:
If the cooldown is 5 minutes make it last 30 seconds. 15 seconds won't get you far.

And keep the platinum cost in IMO, the suit is still strong.

code/modules/clothing/suits/armor.dm
```diff
@@ -581,4 +583,52 @@
 	body_parts_covered = UPPER_TORSO|LOWER_TORSO|LEGS|ARMS
 	spawn_blacklisted = TRUE
 	style = STYLE_HIGH
-	slowdown = -0.3
+	action_button_name = "Toggle Acceleration"
+	var/speed_boost_ready = TRUE
+	var/speed_boost_active = FALSE
+	var/speed_boost_power = -0.5
+	var/speed_boost_length = 15 SECONDS
```
SirRichardFrancis, line 590:
```suggestion
	var/speed_boost_length = 30 SECONDS
```

Sparrowbird611:
> If the cooldown is 5 minutes make it last 30 seconds. 15 seconds won't get you far.
> 
> And keep the platinum cost in IMO, the suit is still strong.

15 seconds is a lot of time for a speed boost. 30 is too much IMO.

Humonitarian:
I'm also in favour of making it a 0 slowdown suit.

"They will just swap it off if they find something better" is probably a bad argument, progression is part of the arms race that is Eris.

TheShown911:
> "They will just swap it off if they find something better" is probably a bad argument, progression is part of the arms race that is Eris.

No. Eris is not a race to get the best gear as soon as possible, nor an RPG where you rush the most powerful items and follow strategy guides. It's a cyberpunk roleplaying game, where the goal is to have interesting roleplaying experiences, not the most powerful gear. Unlike other factions medbay has no progression, witch is not a problem for doctors as they don't have to get into combat situations often, unlike paramedics. Giving them vagabond progression is detrimental to their job as it drags them away from doing it, and forcing them to buy items if they don't want to dig trash to not die instantly in a firefight is pointless as medbay has no form of money generation. There is another faction with no progression that gets into combat often and that is IH, and yet I only see them getting buffed, even though their role is only detrimental to round progression by removing people from it, instead of saving them. So no, the argument is not bad, your vision of the game is.

garrulousTeamotei:
No. Eris is not a race to get the best gear as soon as possible, nor an RPG where you rush the most powerful items and follow strategy guides. It's a cyberpunk roleplaying game, where the goal is to have interesting roleplaying experiences, not the most powerful gear. 

What game have you been playing?

Sparrowbird611:
Just to address a few points...

> Giving them vagabond progression is detrimental to their job as it drags them away from doing it, and forcing them to buy items if they don't want to dig trash to not die instantly in a firefight is pointless as medbay has no form of money generation. 

The reason everyone gravitates towards vagabond tier progression, AKA joonking, is because it's the only universally available one and because it costs only time instead of money (the FS vendor or guild). Most other forms of progression either encourage joonking (moebius research deconstruction) or have joonking as one of the possible ways to complete it (even if there was no decent mats in maintenance, the guild will buy either guild junk or byproducts of it from NT so they can build obelisk for the EYE). If you don't want people gravitating towards that, then we need new, effective methods of progression related to jobs. 

Paramedics also have a lot of free time. Eris is dangerous, yes, but it's not beestation where people are being murdered in the hallways constantly. Paramedics are going to get gear that will help them save others and defend themselves, both of which are forms of progression. As mentioned previously, there's no reason not to go joonk if you have nothing better to do, and with nothing better to do, they'll strengthen themselves in maint. Not only is it not detrimental to their work, it's optimal. That's why they still do it now, even with good roundstart gear.

> There is another faction with no progression that gets into combat often and that is IH, and yet I only see them getting buffed, even though their role is only detrimental to round progression by removing people from it, instead of saving them.

Everyone should have progression of some kind, otherwise the game stagnates and becomes boring. The fact that IH is being constantly given stupid shit is a travesty as it discourages them from doing anything other than killing and waiting around for people to kill, and does not mean we should be repeating this dumbfuckery for other factions, we should be fixing that problem. It's also not even entirely true, eris is by nature a game of progression, otherwise there would be no reason to use drugs, cigarettes, the bar, etc (stats), and IH in particular benefits from obtaining any boost possible to be effective combatants since its their entire job (buying painkillers and combat drugs increased my own success as IH significantly).

> Unlike other factions medbay has no progression, witch is not a problem for doctors as they don't have to get into combat situations often, unlike paramedics.

Having no progression is bad as I mentioned, and everyone gets into combat is common is eris, not some rare event. Doctors buy weapons like everyone else, and there was no reason to give paramed roundstart gear to do it.

TheShown911:
I agree, having no progression is a bad thing, but the fact they do have no progression is true. By removing things they have and giving them nothing in return it just makes the problem worse. Paramedics job isn't to fight people, but to save them. Ideally they get in, drag away the dying person and get out. The speed boost is useful here, unlike digging trash in maintenance. And from my experience paramedics have the access to the best loot in the game. It's not gathered by digging trash, or getting arbitrary points but by being in the aftermath of a gunfight. Already modded items people actually use, gained organically by the virtue of the fight happening at all. I was able to aquire decent gear, while keeping my starting gun as a sidearm and the starting armor. A person digging trash in the maints brings nothing to the round while someone running around the ship actively looking for conflict to either tip the scales of the fight or bring a player who participated in the conflict back does. I am aware junking can not be removed as even IH after many, many buffs still does it, but it should at least not be encouraged. Imagine scattering parts of the engine around maints and calling it progression.

Humonitarian:
> I agree, having no progression is a bad thing, but the fact they do have no progression is true. By removing things they have and giving them nothing in return it just makes the problem worse.

No need to strawman me if you can get to the point ^

In my experience playing Paramedic, I have found MoeSci was ready to provide me with anything I needed, in exchange for the occassional item I found in maint while I wasn't getting injury alarms from anyone.

It has been valid progression, although the lack of useful items that they could print was a little sad. Atomcells, weapon mods, ammo and a spare anne is all they could supply me with. While useful, only the atomcells could provide real help in my job.

The armor was so useful in fact, when I found a tan heavy armor in a pile of junk I passed it to a scientist.

This experience defines my opinion, and as always, don't forget: the most permanent change is a temporary one.

TheShown911:
Overall neutral on this one, up to other coders to decide.

Firefox13 (approved):

PR #6869, opened 2021-12-13
Migration balance

Kurgis:
## About The Pull Request

Reduces burrow migrate interval so additional migrations will be conducted, a minute counts. Base migrate time for burrows reduced from eighty to twenty, this will cause short sudden attacks while in maintenance. Migrate time should only primarily effect maintenance-to-maintenance migrations, burrow story events have their own time set usually. Removed the storyteller switch, this worked as intended for maintenance-to-crew area interactions, but it actually caused infestation events to spawn anywhere instead of solely within crew areas. Sealed burrows are more likely to be breached than before. Warrior is way more aggressive than previously, with a focus on gaining moderate event points for various combat events, and added new tag weight multipliers. The Warrior will now be more likely to spawn external antags such as the Blitzshell and Serbians, though it still doesn't get that many roleset points.

Adds a period to Warrior description, and moves the placement of a file in the .dme because of Windows Explorer pathing.

## Why It's Good For The Game

Stuff should start working properly at least regarding migrations. The Warrior changes are "in violation" of Reere's guidelines which he implemented in order to get people to stop changing values radically, but I didn't notice that much in damage or effect after 1 hour on an uninhabited ship. Though originally I just wanted to spike the value to 100% to get things rolling, Warriors are about fighting after all.

## Changelog
:cl:
tweak: Removed Jester and Warrior causing natural maintenance to crew area migrations due to problems with implementation.
balance: Burrow migrations will trigger one minute more often. Migrations now only take twenty seconds now as well, beware maintenance burrows from now on as there may be little warning.
balance: The Warrior will now be more aggressive in spawning more moderately severe events than before, such as infestation events. Blitzshell and Serbians have also been weighted to be somewhat more common than other antagonist choices with this storyteller.
/:cl:


code/game/gamemodes/roleset/faction/mercenary.dm
```diff
@@ -5,6 +5,7 @@
 	weight = 0.4
 	ocurrences_max = 1
 	req_crew = 5
+	tags = tags = list(TAG_DESTRUCTIVE, TAG_NEGATIVE)
```
SirRichardFrancis, line 8: Bruh

SirRichardFrancis (approved):
Don't see any problem with the code and change is good.

TheShown911 (approved):
Radical inactivity of the storyteller requires radical changes.

PR #6870, opened 2021-12-13
Material Wall Icons

Kurgis:
## About The Pull Request

Materials will now use the solid wall base as default, this doesn't affect materials which have their own icon_base set. This can lead to weird behavior but we're beyond that now.

Iron walls adjacent to some steel ones:
iron

Old iron wall adjacent to the newer ones:
iron_old

## Why It's Good For The Game

Less 2D, more Faux-3D.

## Changelog
:cl:
tweak: Some materials will form regular-looking walls now.
/:cl:

### DIFF
--- a/code/modules/materials/materials.dm
+++ b/code/modules/materials/materials.dm
@@ -98,9 +98,9 @@ var/list/name_to_material
 
 	// Icons
 	var/icon_colour                                      // Colour applied to products of this material.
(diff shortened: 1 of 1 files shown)


Kegdo:
can you make a comparison?
~~just because~~

TheShown911 (approved):
It's been so long.

PR #6881, opened 2021-12-14
Condenses Deepmaint

Kurgis:
## About The Pull Request
Increases the amount of stuff you have to deal with while down inside of deep maintenance. This needs testing, and likely some mapping work to keep things non-repetitive. The map now also uses iron walls as a base which it's getting cut out of.

## Why It's Good For The Game
May as well fill the map with a dungeon instead of just walls if we're going to generate it at all. I mean, more stuff equals more good, right? Right?

## Changelog
:cl:
tweak: Deep Maintenance is now more tightly packed with stuff.
/:cl:


Kurgis:
I'd rather see if this can make rounds on the test server before taking any.

drexample (changes requested):
Large "core" rooms were designed to be unique and one-per deepmaint level.

Also during testing the generator reached its iteration cap several times, so this configuration needs tweaking.
image

drexample:
image
Also it looks kinda the same.

PR #6886, opened 2021-12-17
Preacher rank prefix

Ajax05:
## About The Pull Request

I think the preacher could use a title for people to consistently address him by. Currently, people will refer to him as preacher, father, or just outright use his name. This doesn't seem to fit the head of the church, so I've come up with a universal title which can be used by all members of the crew respectfully:

Reverend

This title has been used to address members of the clergy since the Middle Ages and was used commonly in colonial America, though it has fallen out of favor in modern times. I believe it fits well with Neotheology's distant American origins, and that it is one that might be used in-universe by non-NT to address clergy.

## Why It's Good For The Game

This PR provides a more consistent way to address the preacher in character and helps in identifying the preacher over radio.

## Changelog
:cl:
add: Job prefix for NeoTheology Preacher
/:cl:


hyperioo (dismissed):
Don't delete ` config/example/admin_ranks.txt` and `config/example/admins.txt` as part of this pull request.

Firefox13 (approved):

PR #6887, opened 2021-12-17
crafting stations should drop their materials on deconstruction

Hatterhat:
## About The Pull Request
replaces a code snippet referring to the autolathe to make it refer to the crafting station. also adds the proc for crafting stations to eject materials. doesn't make it accessible other than deconstruction though
## Why It's Good For The Game
i feel like deconstructing a crafting station for the mats inside it is a good idea
## Changelog
:cl:
fix: Crafting stations should now, theoretically, drop their loaded materials when deconstructed.
/:cl:


hyperioo (approved):

hyperioo:
When you say "Crafting stations should now, theoretically, drop their loaded materials when deconstructed." it sounds like you did not test it...

hyperioo:
It works, I checked it.

PR #6888, opened 2021-12-17
Multi-Tile Turf Icons + Maint Turf Retexture

greenteaguzzler:
## About The Pull Request

- Turfs now have the option to have textures that repeat over multiple tiles. This allows for hexagonal turf textures, which is impossible without distortion in a single 32x32 tile.
image
- Catwalks now only show caution strips when bordering plating/underplating (implied elevation difference)
- Catwalks, underplating, plating get new textures. They're less blue now.
### Before:
image
### After:
image
image

## Why It's Good For The Game

Multitile patterns break up a rigid 32x32 grid.

## Changelog
:cl:
add: Multitile patterns for floors
tweak: Floor smoothing, textures
/:cl:


Kegdo:
honestly I think that the former sprites are better, these look dark and... not eris-y
and the hexagonal tiles don't really look nice, too

MLGTASTICa:
The perfectionist within me is screaming , but those new maintenance platings do look nice.

CeUvi:
bring back the blue tint

SirRichardFrancis:
Design is interesting, yet old ones simply fit much better.
Sad to say, but that just doesn't really work with other Eris sprites and makes overall look worse.
😕

drexample:
Very inconsistent with the base palette.

PR #6892, opened 2021-12-18
Solidification

TGCMbetcher:
## About The Pull Request

Adds the same solidification of plasma to gold and uranium

## Why It's Good For The Game

Allows science, after like 20-30 minutes of work, to get the materials they need

## Changelog
:cl:
add: Added recipes that allow the solidification of gold an uranium
/:cl:

### DIFF
--- a/code/modules/reagents/recipes.dm
+++ b/code/modules/reagents/recipes.dm
@@ -583,6 +583,20 @@
 	new /obj/item/stack/material/plasma(get_turf(holder.my_atom), created_volume)
 	return
 	
+/datum/chemical_reaction/goldsolidification
+	result = null
+	required_reagents = list("iron" = 5, "frostoil" = 5, "gold" = 20)
+	result_amount = 1
+
+/datum/chemical_reaction/goldsolidification/on_reaction(var/datum/reagents/holder, var/created_volume)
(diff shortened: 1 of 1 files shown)


Voldirs:
Why not adding the same for silver?

TGCMbetcher:
> Why not adding the same for silver?

Because there is no way to get liquid silver in the wild. Its not in circuit boards or anything an autolathes give it to you naturally. This is mostly for botany to make the plant gold/uranium you make have a way to turn into solid matter. If you have liquid silver chances are you were making drinks an used enough of it that you cant resolidify.

glaciusinfinity:
> > Why not adding the same for silver?
> 
> Because there is no way to get liquid silver in the wild. Its not in circuit boards or anything an autolathes give it to you naturally. This is mostly for botany to make the plant gold/uranium you make have a way to turn into solid matter. If you have liquid silver chances are you were making drinks an used enough of it that you cant resolidify.

If nothing else you should add it for consistency's sake, maybe at some later point, a way to get liquid silver will be added and being able to solidify it would be convenient to have

TGCMbetcher:
alright Ill add silver in as well as a carbon iron way of making steel. Frankly frost oil was way more valuable then steel so if thats how they want to do it whatever. Id rather spend paltry amounts at the guild for 480 sheets then spen forever slaving on a chem dispensor

PR #6910, opened 2021-12-26
Removes file that causes Weird Things to happen with Git

Shadow-Quill:
## About The Pull Request

It's an exact duplicate, but windows/git Does Not Like It, as windows is insensitive to file name capitalization.


MacBlaze1:
I have experienced issues with this too

Shadow-Quill:
https://github.com/discordia-space/CEV-Eris/pull/6909CI failure caused by runtime fixed in

TheShown911 (approved):

PR #6933, opened 2022-01-05
High-velocity bullets don't have any damage boost

Kegdo:
## About The Pull Request

removes the 5 dmg point boost from the HV ammo, making it less bonkers, armor penetration and everything else untouched.

## Why It's Good For The Game

less lethal combat? The council of coderbus wanted this thing to be added.

## Changelog
:cl:
tweak: HV ammo doesn't have the 5 dmg point boost now, everything else untouched.
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/bullettypes.dm
+++ b/code/modules/projectiles/projectile/bullettypes.dm
@@ -18,7 +18,6 @@ There are important things regarding this file:
 	penetrating = 1
 
 /obj/item/projectile/bullet/pistol/hv
-	damage_types = list(BRUTE = 32)
 	armor_penetration = 20
 	step_delay = 0.75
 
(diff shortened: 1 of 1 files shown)


Sparrowbird611:
I do want to point out that more AP realistically means more damage, so less lethal combat isn't really true except in cases where people are totally unarmored (in which case they'll probably die anyway). I'm not saying this is bad, but I do want to ask if this was made because HV is overperforming or just because someone got in their head that damage = more lethality while AP somehow doesn't. As far as I am aware HV got a decent price hike in one of those uplink rebalances, and I haven't seen very often since then.

ValoTheValo:
Salt PR moment

hyperioo (commented):
No problem codewise.

Firefox13:
HV ammo is pretty much "pay more for more bonk" and is for antags; they should have more damage

PR #6935, opened 2022-01-06
[TGUI] remap nanoui

LetterN:
## About The Pull Request
Title. Litteraly just renames nanoui functions to have `nanoui_` prefix


hex37:
Probably necessary if we're bringing in TGUI, which uses similar function names.

hyperioo (approved):

PR #6938, opened 2022-01-07
VSC Debugging

Shadow-Quill:
## About The Pull Request

Adds the functionality to [debug code with Visual Studio Code](https://code.visualstudio.com/docs/editor/debugging).

## Why It's Good For The Game

Can help track down issues.

## Changelog
:cl:
code: Can now debug using VSC.
/:cl:

### DIFF
--- a/tools/hooks/install.bat
+++ b/tools/hooks/install.bat
@@ -1,2 +1,0 @@
-@call "%~dp0\..\bootstrap\python" -m hooks.install %*
-@pause
--- a/SpacemanDMM.toml
+++ b/SpacemanDMM.toml
@@ -1,2 +1,5 @@
 [langserver]
 dreamchecker = true
+
+[debugger]
+engine = "auxtools"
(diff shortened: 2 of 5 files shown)


tools/hooks/install.bat
```diff
@@ -1,2 +0,0 @@
-@call "%~dp0\..\bootstrap\python" -m hooks.install %*
```
LetterN, line 1: why removing this?

Shadow-Quill: https://github.com/discordia-space/CEV-Eris/pull/6910

hyperioo: @LetterN Do you think removing this file could break things? It seems to be a duplicate of Install.bat and to not be called in the code but maybe we're missing something, since you made the PR that set that up.

Shadow-Quill: Removed via another PR.

hyperioo (approved):
Seems ok. Might be tested on test server first, just in case.

Shadow-Quill:
> Seems ok. Might be tested on test server first, just in case.

I've used this a few times. It doesn't affect normal servers in any way, only if running it with VSC's debugger.

drexample:
Imagine using anything but god-given Dream Maker IDE.

PR #6941, opened 2022-01-09
Some price updates

TGCMbetcher:
## About The Pull Request

Changes the prices to match beacons prices an fixes an infinite money loop

## Why It's Good For The Game

Since youre buying surplus in bulk it make no sense for some of them to be x price. They export for less and you have to buy multiple unwanted things so I downed gun prices. Fixed cigs infinite money loop, sure its only 50 per 120 used but free money bad according to the top notes.  Added vacuum to the janitorial crate. Left most of the other stuff if someone wants to RP
## Changelog
:cl:
add: One item to a list
tweak: prices
fix: infinite money loop
/:cl:


Humonitarian:
Vacuum is an NT item, you sure putting that accessible to guild is a good idea?

TGCMbetcher:
> Vacuum is an NT item, you sure putting that accessible to guild is a good idea?

It's a single vacuum, they can already but NT weapons unless that's commented out and I haven't read it, it's mostly so people can clean their departments without begging spacecleaner from moe and if it's a deal breaker I'll just lock the order behind NT access since it is a janitorial crate. We already have code for that in the commented out miners explosives crate.

Humonitarian:
You could probably include a sprayer of space cleaner, actually

Lorewise it makes more sense, and the ghetto chem opportunities it gives probably aren't game-breaking

Humonitarian:
> Probably but the vacuum is easy, it helps NT out and it's cheap as well so there is zero reason NT doesnt mass sell them like their pouches and other things which you can also get in cargo and the holsters need a price reduction because the beacon is cheaper by far but I forgot to add that. Since guild already sells other NT things no reason a far overpriced one cant be sold here.

Doesn't the vacuum produce biomatter?

Also selling pouches and holsters via cargo is no improvement at all, if anything they should be encouraged to be sold to crew.

What makes pouches better than biomatter is that you can bypass guild through selling to the crew.

TGCMbetcher:
Guild has random holsters and pouch crate. They're priced higher then the individual ones purchased from the beacons. I'm going to have to turn alot of things into surplus so they cant be resold. Also do you know what modifier is used to make the things printed in lathes with cheaper costs? There are a few easy money exploits if you have access to a bio printed and I want to specifically put them on the medical crates.

Humonitarian:
> As for the vacuum it's more that it is alot faster with no slips compared to the mop, it produces filth which can be turned into biomatter like meat can, and I tried all weekend to put orders in for the chemist at their desk and they were never there. Probably should have left a paper. Since the crate is 1000 credits one of three things are going to happen. I take the vacuum out and drop the price to 500 for the old crate. I lock it to NT so the order has to go through them, or I throw in 2 bottles of space cleaner, more then that and you can abuse for infinite money. Maybe a separate crate of just cleaner.

Yep, that's what I'm saying. A single space cleaner spray instead of the vacuum. This is a non-NT station, right? So it would make sense if they had a non-NT cleaning tool.

TGCMbetcher:
Except you're importing things from other places. Also it's a nom NT ship but its cleaned by NT and there are NT nades INSIDE the pack already. I dont get what the big deal with this is. There are tons of stuff from other factions already in packs and you're not even accepting locking it to NT only. The issue with space cleaner is you need to charge 1 credit per unit you sell or its infinite money. Its excessively expensive for the little use you'll get from it

Humonitarian:
> Except you're importing things from other places. Also it's a nom NT ship but its cleaned by NT and there are NT nades INSIDE the pack already. I dont get what the big deal with this is. There are tons of stuff from other factions already in packs and you're not even accepting locking it to NT only. The issue with space cleaner is you need to charge 1 credit per unit you sell or its infinite money. Its excessively expensive for the little use you'll get from it

IMO job-locking would only make sense with department-specific items, for example Ironhammer having a premium FS subscription, or NT having access to the more occult items on an NT ship.

Sparrowbird611:
If the concern is biomatter production, make a subtype that doesn't create biomatter. Otherwise I hear no problem with having an NT vacuum, lots of factional equipment is sold at the guild. Neotheo is a supposedly high tech cult, renowned for cloning. They are not renowned for inventing the vacuum cleaner.

TGCMbetcher:
Alright how about this. I make a new chem that reacts with water to create space cleaner at 5 units and a water tank to make whatever water into space cleaner just like how enzyme works. Throw a bottle into the thing and presto you have your cleaning supplies, no issues with infinite money, no NT vacuum problems just easy cleaning solution so the halls dont look like shit all game.

TGCMbetcher:
> If the concern is biomatter production, make a subtype that doesn't create biomatter.

You literally have to clean the whole ship to get 20 units of biomatter, it's more of an annoyance to empty it occasionally them a solid biomatter production method. You get more from a backpack.

hyperioo (approved):
Seems ok. Four holsters at 8000 credits is intriging when you can have guns and rig modules for around 2000 credits, but it's for consistency with beacon prices so fixing that would imply changing a lot of things.

TGCMbetcher:
> Seems ok. Four holsters at 8000 credits is intriging when you can have guns and rig modules for around 2000 credits, but it's for consistency with beacon prices so fixing that would imply changing a lot of things.

My bad typo on the 8000, but yea alot of prices are because crates used to be worth 200 and the stamps from merchant an extra 100 so they required all crates to be 400 minimum to prevent infinite money. Alot of things are just bring it it down for that. The RNG stuff should be cheaper since its RNG but the rest is either fixes for infinite or more in line with what you can buy at the beacon.

PR #6957, opened 2022-01-18
Fix mouse goonchat

SirRichardFrancis:
## About The Pull Request

Turns whatever mouse says into "Squeak!" in a floating message.

Before:
image

After:
image

## Why It's Good For The Game

Fixes cool.

## Changelog
:cl:
fix: horrible case of language barrier breaching by mice
/:cl:

### DIFF
--- a/code/modules/mob/living/say.dm
+++ b/code/modules/mob/living/say.dm
@@ -308,8 +308,9 @@ var/list/channel_to_radio_key = new
 			speech_bubble_recipients += M.client
 		M.hear_say(message, verb, speaking, alt_name, italics, src, speech_sound, sound_vol, 1)
 
-	INVOKE_ASYNC(GLOBAL_PROC, .proc/animate_speechbubble, speech_bubble, speech_bubble_recipients, 30)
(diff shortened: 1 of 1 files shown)


zombiemimic:
No more tempting souls into violence. sad

Fatherperhaps:
Rat is the most enriching role in this game.

Alliostra:
https://www.youtube.com/watch?v=3KcJ5Ji4WfM

Rumplenutskin:
this would be a sad change. I've had so many fun mouse interactions. don't become tg on me

Omni44:
Remove the part where you remove the whole chat thing from mice, and add part where player spawned mice do not disintegrate after few seconds and you'll have good PR and a half there.

mattroks101:
This is a really big bandaid fix for this. Can we not do a deep dive onto why this isn't respecting mice language in the first place?

TheShown911:
It should also be tested how goonchat works with other simplemobs.

hyperioo (commented):
I agree with matt, it hides the issue rather than fix it. The same kind of issue would remain with current/future talking mobs.

drexample:
Yeah that doesn't fix the issue of parts of language system being ignored by runechat.

PR #6951, opened 2022-01-14
Recoil overhaul [RTM]

Humonitarian:
## About The Pull Request

Recoil was so far very much unfinished, so I have decided to use some spare time and have some fun with it.
Copious alcohol, neglected sleep, and the joy of painting with Bob Ross later, I realised that I let this rework get a little out of hand. Feels pretty good ingame, though.

Removes delay from recoil, refactors large swaths of recoil code, removes multiple aiming cursor bugs (and potentially adds new ones), reworks armor slowdown into a combination of slowdown and "stiffness", a new stat that causes armor to give you recoil as you walk, a new stat that causes helmets to provides extra base offset (finally we can reduce our dependence on the publicly hated TINT value).

A couple minor armor value tweaks where I realised I left plotholes in previous PRs (riot mask), and a new fix to armor plates that prevents them from being placed on already armoured clothing (overcoats, armor vests, etc.), as well as gives more hefty debuffs to them.

Hopefully I didn't forget anything.

## Why It's Good For The Game

Less bug, more fleshed out recoil, less annoying slowdown and tint.

Thanks to a few cheaty lines of code, the recoil of guns has been adjusted to be in effect identical to before the change. This means we get all the benefits of a better feel for recoil, without the drawback of balance issues.

Whether the style PR goes in or not is not a concern for this PR's balance changes.

This PR makes the armor stacking PR ready to test, as it addresses most balance issues brought by it.
The armor stacking PR is not ready to be merged live by merging this alone, due to potential risks with One Star limbs, and other potentially overpowered items.

## Changelog
:cl:
balance: All armor and clothing now provide less slowdown, but run and gunning in heavy armor reduces accuracy
balance: Recoil now decreases immediately after firing, the recoil of every weapon adjusted accordingly
balance: Thanks to reducing the dependence on TINT, much less guns will obscure your vision
/:cl:

PR #6959, opened 2022-01-22
Revert "Recoil overhaul [RTM]"

ValoTheValo:
Reverts discordia-space/CEV-Eris#6951

original PR not tested and was half baked, and nobody so far likes it. should be reverted until made playable and balanced. and was not thought through aswell


drexample:
cope

PR #6960, opened 2022-01-22
Unrecognizable voice when wearing gasmask or bandana

SirRichardFrancis:
## About The Pull Request

Basically this suggestion:
image

Works with bandanas and most gasmasks (excluding clown masks), but not scarves.
Voice recognized anyway if hearer have "Ear of quicksilver" perk or standing near the speaker.

image

## Why It's Good For The Game

Features cool.

## Changelog
:cl:
add: gasmasks and bandanas replacing speaker name with "Unknown"
/:cl:


code/modules/clothing/masks/miscellaneous.dm
```diff
@@ -225,7 +225,8 @@
 	w_class = ITEM_SIZE_SMALL
 	siemens_coefficient = 0.9
 	body_parts_covered = HEAD|FACE|EYES
-	style_coverage = COVERS_WHOLE_HEAD
+  style_coverage = COVERS_WHOLE_HEAD
```
SirRichardFrancis, line 228:
```suggestion
	style_coverage = COVERS_WHOLE_HEAD
```

code/modules/clothing/masks/miscellaneous.dm
```diff
@@ -297,8 +299,9 @@
 	item_flags = FLEXIBLEMATERIAL
 	w_class = ITEM_SIZE_SMALL
 	price_tag = 20
-	style = STYLE_LOW
-
+ 	style = STYLE_LOW
```
SirRichardFrancis, line 302:
```suggestion
	style = STYLE_LOW
```

SirRichardFrancis (commented):
I love smell of merge conflict in the morning

Jan-Blasiak:
As per my SUGG just below the one shown in the image, I would also allow for proximity based recognition: from 1-2 tiles away, at least for bandanas.

SirRichardFrancis:
> As per my SUGG just below the one shown in the image, I would also allow for proximity based recognition: from 1-2 tiles away, at least for bandanas.

Done.

Sparrowbird611:
I think proximity recognition is detrimental to the feature. Otherwise, nice addition.

Jan-Blasiak:
Bandana or a Gas Mask is now a very cheap version of Traitor's voice modulator. It NEEDS to have some downside or nobody will waste TCs for it.

Sparrowbird611:
> Bandana or a Gas Mask is now a very cheap version of Traitor's voice modulator. It NEEDS to have some downside or nobody will waste TCs for it.

Nobody wastes TC on it in the first place. If you don't care about voice mimicry (which most people don't), then you wear a mask and just flat out don't talk. This is really not such a strong feature that it needs a downside.

Jan-Blasiak:
And I also think it makes more sense this way, while the voice could be lost in the echoes of the Eris corridors, when you stand directly next to someone with muffled voice you can understand them much easier.

I used to work in a cheap gas mask IRL, it muffles and distorts your voice, but you can easily communicate and recognize people's voices up close.

Sparrowbird611:
Realism is not really relevant here. I'm just stating that making it based on range is detrimental to the feature. The only need for this mechanic is for those rare situations where either a group of people cares about hiding their identity or a single person needs to communicate while remaining hidden, both of which are relatively rare scenarios. When the situations for the usefulness of this are rare, there's little need to add a feature that removes the whole use of the item, and no place I've seen that has added this mechanic has done so.

Jan-Blasiak:
Well, whispering works similarly, while you whisper you can be heard only few tiles away, and fully recognized only up close.

Here, with gas masks on you can be heard normally, but recognized through the distortion only up close.

SirRichardFrancis:
> Realism is not really relevant here.

la classique.

Humonitarian:
Could be interesting if whispering removed the range recognition

Jan-Blasiak:
That it could! You have to borderline yell in a gas mask to be understandable!

Sparrowbird611:
You all can do what you like, but it seems like an unnecessary amount of complication for no gain when the feature could just be "your voice is concealed, bam, done." I don't really care either way, I just roleplay a mute like every normal crazed gunman ;)

ValoTheValo:
Does the voice supression work on radio or if your have your ID on? Cause Operative Unknown could be an issue

SirRichardFrancis:
> Does the voice supression work on radio or if your have your ID on? Cause Operative Unknown could be an issue

It doesn't work on radio, clearly seen on the attached screenshot. ID doesn't matter.

code/modules/mob/hear_say.dm
```diff
@@ -14,7 +14,11 @@
 	var/speaker_name = speaker.name
 	if(ishuman(speaker))
 		var/mob/living/carbon/human/H = speaker
-		speaker_name = H.rank_prefix_name(H.GetVoice())
+		// GetVoice(TRUE) checks if mask hiding the voice
+		speaker_name = H.rank_prefix_name(H.GetVoice(TRUE))
+		// If we have the right perk or standing close - GetVoice() again, but skip mask check
+		if(((get_dist(src, H) < 2)) || (src.stats.getPerk(PERK_EAR_OF_QUICKSILVER)))
```
TheShown911, line 20:
```suggestion
		if((get_dist(src, H) < 2) || stats.getPerk(PERK_EAR_OF_QUICKSILVER))
```
may work

TheShown911: Why do the check for the perk here, and than again in GetVoice?

SirRichardFrancis: This PR have been trough a couple of iterations and I simply forgot to cleanup here. ¯\\_(ツ)_/¯

TheShown911 (approved):

PR #6966, opened 2022-01-25
Themis and Nemesis fix

Humonitarian:
## About The Pull Request

Gun code was not meant to handle the agony damage of crossbow bolts, that is fixed. Apparently these two guns are overpowered, downing an antagman (full mercenary armor, that gives 50 energy protection against the 0 penetration of the crossbows) in 5 shots, so their firepower was nerfed. They have no recoil, afterall.

## Why It's Good For The Game

Themis and Nemesis fix good.

## Changelog
:cl:
fix: Themis and Nemesis now do agony damage as intended
balance: Themis and Nemesis nerfed
/:cl:

### DIFF
--- a/code/modules/mob/living/living_defense.dm
+++ b/code/modules/mob/living/living_defense.dm
@@ -128,6 +128,10 @@
 		to_chat(src, SPAN_WARNING("You have been hit by [P]!"))
 		qdel(P)
 		return TRUE
+	
+	if(P.agony > 0)
+		hit_impact(P.agony, hit_dir)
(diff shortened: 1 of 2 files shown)


TheShown911 (approved):

code/modules/mob/living/living_defense.dm
```diff
@@ -128,6 +128,10 @@
 		to_chat(src, SPAN_WARNING("You have been hit by [P]!"))
 		qdel(P)
 		return TRUE
+	
+	if(P.agony > 0)
```
TheShown911, line 132: Wait a minute, how does this work with stun lasers?

PR #6973, opened 2022-01-29
Reducing armor hallos coefficient

Iglpigl:
## About The Pull Request

Reduces armor hallos coefficient from 0.4 to 0.1, makes armor more effective.

## Why It's Good For The Game

Requested by TheShown.

## Changelog
:cl:
tweak: armor hallos coefficient changed from 0.4 to 0.1, armor more effective now
/:cl:

### DIFF
--- a/code/modules/mob/living/living_defense.dm
+++ b/code/modules/mob/living/living_defense.dm
@@ -1,4 +1,4 @@
-#define ARMOR_HALLOS_COEFFICIENT 0.4
+#define ARMOR_HALLOS_COEFFICIENT 0.1
 #define ARMOR_GDR_COEFFICIENT 0.1
 
 //This calculation replaces old run_armor_check in favor of more complex and better system


TheShown911 (approved):
Good PR speedmerge.

Sparrowbird611:
What's armor and GDR co-eff doing? I assume it reduces pain damage?

TheShown911:
It's % of armor converted to GDR, for example 50% armor gets converted to 5 GDR so if you take 25 damage it's reduced by 5 and also by 50% from armor.

Sparrowbird611:
> 

So what's armor halloss in this case?

TheShown911:
% of damage armor absorbs converted to hallos

Humonitarian:
> > 
> 
> So what's armor halloss in this case?

Say, armor is 60

Say, damage is 106

GDR blocks 6, resulting in 100 effective.

60 damage is turned into halloss, which is then multiplied by the halloss coefficient (after this PR, this means 60 damage is turned into 6 halloss (see: 0))

The 40 damage unblocked is dealt as brute to the targeted limb.

Sparrowbird611:
> 

So this fixes our issue of armor not really doing much to protect against pain?

Humonitarian:
> > 
> 
> So this fixes our issue of armor not really doing much to protect against pain?

In a way, yes, but also makes it much more likely to die in armour instead of being knocked out, especially with high toughness

Sparrowbird611:
> > > 
> > 
> > 
> > So this fixes our issue of armor not really doing much to protect against pain?
> 
> In a way, yes, but also makes it much more likely to die in armour instead of being knocked out, especially with high toughness

What decreased pain does to a mfer. Still better than being paincrit and executed.

PR #6978, opened 2022-01-31
Prevent wave dashing cripples

hex37:
## Why It's Good For The Game
Makes the weakened condition work as intended in limiting player movement.

## Changelog
:cl:
fix: prevents players with the weakened status from being able to speed up movement by diving repeatedly instead of crawling.
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -630,7 +630,7 @@ default behaviour is:
 		var/speed = movement_delay()
 		resting = TRUE
 		var/dir = C.true_dir
-		if(ishuman(src) && (dir))// If true_dir = 0(src isn't moving), doesn't proc
+		if(ishuman(src) && !src.weakened && (dir))// If true_dir = 0(src isn't moving), doesn't proc
 			var/mob/living/carbon/human/H = src
 			livmomentum = 5 // Set momentum value as soon as possible for stopSliding to work better
 			to_chat(H, SPAN_NOTICE("You dive onwards!"))


code/modules/mob/living/living.dm
```diff
@@ -630,7 +630,7 @@ default behaviour is:
 		var/speed = movement_delay()
 		resting = TRUE
 		var/dir = C.true_dir
-		if(ishuman(src) && (dir))// If true_dir = 0(src isn't moving), doesn't proc
+		if(ishuman(src) && !src.weakened && (dir))// If true_dir = 0(src isn't moving), doesn't proc
```
SirRichardFrancis, line 633: Needless `src.`, same in the code above with `src.client` and `src.stats`.
```suggestion
		if(ishuman(src) && weakened && (_dir))// If true_dir = 0(src isn't moving), doesn't proc
```

code/modules/mob/living/living.dm
```diff
@@ -630,7 +630,7 @@ default behaviour is:
 		var/speed = movement_delay()
 		resting = TRUE
 		var/dir = C.true_dir
```
SirRichardFrancis, line 632: Var with this name already exist.

```suggestion
		var/_dir = C.true_dir
```

code/modules/mob/living/living.dm
```diff
@@ -630,7 +630,7 @@ default behaviour is:
...
-		if(ishuman(src) && (dir))// If true_dir = 0(src isn't moving), doesn't proc
+		if(ishuman(src) && !src.weakened && (dir))// If true_dir = 0(src isn't moving), doesn't proc
 			var/mob/living/carbon/human/H = src
```
SirRichardFrancis, line 634: What if mob isn't human? More runtimes to runtime god.

code/modules/mob/living/living.dm
hex37: If you're in a situation where you actually need to use src to differentiate vars you're slready writing bad code. This just improves readability and has no effect on speed at compile time.

code/modules/mob/living/living.dm
hex37: Unrelated to my change.

code/modules/mob/living/living.dm
hex37: ishuman ensures the mob is human

Kegdo:
so, is this just a one-line change? i cant really see because of the reviews

hex37:
> so, is this just a one-line change? i cant really see because of the reviews

yes.

Firefox13 (approved):

PR #6994, opened 2022-02-08
Shield of holding no more

Alliostra:
## About The Pull Request

You can no longer store the NT spear in the NT buckler.

## Why It's Good For The Game

The NT spear does not fit into backpacks, but the NT buckler does. Thus, you shouldn't be able to put the NT spear into the buckler and store the buckler inside of a backpack. Also change for the sake of change (I was asked to do this)

## Changelog
:cl:
tweak: you cannot store the NT spear inside the NT buckler anymore (not balance, I swear!)
/:cl:


Firefox13 (approved):

Humonitarian:
1984

PR #6999, opened 2022-02-10
Molotov fix

Alliostra:
## About The Pull Request

Bottles properly break when thrown with harm intent again.

https://github.com/discordia-space/CEV-Eris/pull/5211 
broke bottles breaking when thrown (I now realize how stupid this sounds, but I'm not going to change it), because bottles check if the thrower is in harm intent when the bottle smashes, but that PR made the thrower var on the item get set to null before the throw_impact. This is a workaround.

I also removed the random chance from bottles breaking when thrown, they now always break regardless of range traveled when thrown by someone with harm intent. Fire is weak enough already, no need to add a random chance for a molotov not to break on impact.

## Why It's Good For The Game

No good revolution without molotovs.

## Changelog
:cl:
balance: bottles now always break when thrown with harm intent active
fix: bottles can now break again when thrown with harm intent active
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/food/drinks/bottle.dm
+++ b/code/modules/reagents/reagent_containers/food/drinks/bottle.dm
@@ -16,6 +16,7 @@
 	var/rag_underlay = "rag"
 	var/icon_state_full
 	var/icon_state_empty
+	var/bottle_thrower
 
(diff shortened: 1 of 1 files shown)


TheShown911 (approved):

pexel5:
noice

PR #7025, opened 2022-02-18
Stinger grenades, HESH rocket, fixed fragmentation

Humonitarian:
## About The Pull Request

Frag grenades up until now used shrapnel/strong instead of shrapnel, while redefining their damage value back to 5, which seemed like an oversight, especially because any GDR above 5 means they do no damage nor pain.

The fix now increases their damage back to 15, but as this would be too extreme, the amount of fragmentation was nerfed heavily.
Also, fragmentation will now ricochet. Fun.

The end result is a grenade that will deal ~25 damage to heavily armoured or stylish personnel in close proximity, draining style extremely quick. Those without style or armour can easily take a hundred or upwards damage.

Stinger grenades are a new type of grenade working similarly to frag grenades, but dealing mainly agony rather than raw damage. They are more effective against armor and less effective against style, but as a less-than-lethal option they underperform overall.

IH has two boxes of them roundstart.
A guild variant exists in the code, but is unused. Leaving it to guild coders to implement via trade stations, or shuttle.

Rubber shot for grenade launchers was replaced both in code and on the map with weak stinger variants, making the use of a grenade launcher feel more like a grenade launcher than a big beanbag shotgun.

A new type of rocket was added exclusive to serb RPGs, the HESH rocket. 4 of the 12 rockets are HESH. HESH rockets offer more AP on a direct hit, as well as directed shrapnel released through the target upon impact (alllowing you to frag through walls), but have a significantly smaller explosion.

Stinger sprites drawn by Shazbot.

## Why It's Good For The Game

Frags fixed, they actually cause damage more
More tacticool options to IH and serbs

## Changelog
:cl:
balance: frags are deadlier
add: HESH rocket
add: stinger grenades
balance: grenade launcher rubber shot replaced with stinger rounds in armoury
/:cl:


Humonitarian:
https://imgur.com/a/630A5C5

Humonitarian:
Colour of guild sting grenades swapped to be more similar to FS one, but with a green stripe instead of blue

Firefox13:
🅱️ESH

PR #7040, opened 2022-02-22
Piercing, scoping and AMR buff

Alliostra:
## About The Pull Request
These changes have been approved by Humon (dev delegate):

a bullet piercing an object no longer makes the bullet lose AP
when using a scope, the initial offset of the AMR and valkyrie is reduced by 8 and that of all other guns by 3. This is to allow for precise long-range shots even with offset added through helmet obscuration
doubles the amount of objects all bullets can pierce at maximum without increasing the chances of piercing through an object.
triples the penetration power of the AMR before, it could only pierce a single wall total.)
(do not be confused the the pierce_multiplier changes- it's not actually a multiplier, but rather added to the penetrating var of the bullet. This is why they had to be increased as well)

These changes have not been approved by Humon (dev delegate):

massively lowered the price of all Ammunition types for the AMR (consider that the AMR itself still costs 20 TC)
increased the direct damage on the AMR breacher bullets, as the shrapnel produced by it can no longer hit the same person the projectile itself hit.

Also makes talking and control spiders no longer get activated by the carrion activate all button, since if you hit activate all, you surely won't want to transfer your mind into a roach.

## Why It's Good For The Game

AMR buffs good (Montagu complained about it being too weak)

## Changelog
:cl:
balance: doubled the maximum amount of walls bullets are able to pierce and piercing a wall no longer makes the bullet lose AP.
balance: made AMR ammo from the uplink cheaper and increased the damage of the Breacher shells
balance: tripled the penetration power of the AMR
balance: using a scope makes shots more accurate now
tweak: the talking spider and the control spiders will no longer be triggered when activating all carrion spiders.
/:cl:


Sparrowbird611:
I'd let the AMR hold multiple shells and give it a fire delay similar to the valk. It's a tad ridiculously that it can't compared to its cousin the valkyrie. If you're concerned about power level, recall that the valkyrie literally uses batteries for ammo, and that someone autistic enough to try and speed reload an AMR will reach similar firing speed to the valkyrie.

You should also add mention of the slight carrion tweaks to the description of this PR.

hyperioo (approved):
Seems ok. May need a balance later if it proves to be too strong

PR #7041, opened 2022-02-22
carrion spider balance

Alliostra:
## About The Pull Request

Don't worry, this is the last balance PR in a while, content PR coming soon.

Lowers the gene point cost of the talking, identity and holographic spiders from 3 to 1, and also lowered the chem point cost of the talking and identity spider- All of these spiders aren't too powerful, but actually using them can make for fun rounds, so they should be cheaper to acquire and produce.

Reduces the gene point cost of the blight spider from 7 to 5, doubles the amount of amatoxin it injects from 5 to 10 and makes it inject 5 units of pararein as well. Previously, the toxins from a single blight spider would just get outhealed by passive toxin regeneration, and even if multiple were stacked, the toxins buildup was still extremely slow and could be easily healed.

Now, a single spider is at least able to deal a bit of toxin damage to someone and debuff their toughness, while stacking many still doesn't have a huge amount of toxins buildup, but at least enough so it can't just be ignored in the long run.

## Why It's Good For The Game

hopefully more use of interesting funny spiders

## Changelog
:cl:
balance: buffed the blight spider
balance: lowered the prices of some carrion spiders
/:cl:

### DIFF
--- a/code/game/gamemodes/changeling/modularchangling.dm
+++ b/code/game/gamemodes/changeling/modularchangling.dm
@@ -39,7 +39,7 @@ var/list/datum/power/carrion/powerinstances = list()
 /datum/power/carrion/blight_spider
 	name = "Blight spider"
 	desc = "Evolves a spider filled with a sickening venom."
-	genomecost = 7
+	genomecost = 5
 	spiderpath = /obj/item/implant/carrion_spider/blight
 
 /datum/power/carrion/breeding_spider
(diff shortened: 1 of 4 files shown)


TheShown911 (approved):
Doesn't parenin also lower sanity? It could be used with mindboil spiders for faster mental breakdown but that's cool.

Alliostra:
> Doesn't parenin also lower sanity? It could be used with mindboil spiders for faster mental breakdown but that's cool.
yeah that's true, but when I stacked a bunch of them, I didn't get a breakdown even as a regular vagabond, so it would be more situational  to use when someone is already low on sanity, I guess

PR #7048, opened 2022-02-25
Revert "Makes cover work at a distance"

ValoTheValo:
Reverts discordia-space/CEV-Eris#7045

it broke the SM, causing the emitter to destroy the low wall infront of crystal and delam every single round. will conduct futher testing.


ValoTheValo:
i have tested it and yes cover was what broke the SM

PR #7064, opened 2022-03-03
Style rebalance 2

vode-code:
## About The Pull Request
Dodging now increases recoil proportionally to the style damage, to prevent perfect accuracy while under fire.
Throwing items now deals 30 style damage instead of 20.
Banana Cream Pies now deal 60 style damage, even if they hit.
Recoil gain from spinning a held item is reduced by 10 to 50.
Spinning a levergun now works the "bolt"(lever).
Vomiting and slipping is now three times as embarrassing, with 30% style damage.
tripping now deals 10% style damage, for when you trip on bleeping plating again.
It is now possible to disable dodging with the Toggle Dodging verb, although riding shockwaves is an innate property of awesomeness and cannot be disabled through any means other than depletion of awesomeness.

## Why It's Good For The Game
Before this PR stylish spacemen were able to have perfect accuracy while under fire, which lead to one-sided fights between stylish and non-stylish spacemen; recoil on dodge should fix this.
Throwing items requires you have something to throw, which generally is more intensive than using guns to do it in Eris.
Cream Pies used to not have any special effect for their embarrassing appearance on throw, now they are expensive to dodge and embarrassing to get hit by.
Stiffness meant you would have the item fly out of your hand immediately if you spun your held item while moving, which is why I gave it 10 buffer... didn't I do this already though?
Leverguns being able to be levered by spinning them is a cool thing that can be done IRL and should be doable here.
Vomiting and slipping should be more embarrassing than dodging is, and now it is.
Tripping is somewhat embarrassing, although I decided not to go too far as only plating does it.
Sometimes people just don't wanna dodge, especially after recoil gets added to dodging.
## Changelog
:cl: Chickenish
add: Cream Pies are now massively effective against style.
add: It is now possible to lever a levergun with the spin-held-item verb.
add: You can now toggle dodging with the Toggle-Dodging verb.
tweak: Throwing items now deals 30 style damage instead of 20.
tweak: Recoil gain from spinning a held item is reduced below the flying out of your hand threshold so you can spin while moving.
tweak: Vomiting and slipping is now three times as embarrassing, with 30% style damage, and tripping now deals 10% style damage, for when you trip on bleeping plating again.
balance: Dodging now increases recoil proportionally to the style damage, to prevent perfect accuracy while under fire.
fix: fixed a few things
/:cl:


githubuser4141:
Maybe add a fail chance to lever arm spinning, a quick google search shows that IRL gun cocking isn't practical or safe and there's no legit reason to do it. So a chance to have the gun clatter to the floor or a small chance of shooting yourself would be pretty funny.

Wouju:
As much as I hate style, having someone’s gun fall out of their hands because of funny RNG is a terrible idea

Firefox13 (approved):

ValoTheValo:
"It is now possible to disable dodging with the Toggle Dodging verb"
Extremely based. I now do not have to deal with more recoil bullshit

PR #7069, opened 2022-03-07
NT pilum throwing damage reduction

Alliostra:
## About The Pull Request

reduced the damage the NT spear does when thrown from 75 to 45. the 75 value is a remnant of when all throwforces were multiplied by 0.6 for the purpose of applying damage. This changes it back to it's old damage.

also increases the style damage of the spear to 50 (requested by humon)

## Why It's Good For The Game

fix but also balance

## Changelog
:cl:
balance: The NT spear now deals the intended 45 damage when thrown instead of 75 damage, and deals 50 style damage instead of 20.
/:cl:


Humonitarian:
If its damage is being nerfed, its style damage should be raised. Without the 50+ damage armour is fairly effective against it, and style outright denies any attempts to use it as it is stuck with 20 style damage despite being such an unwieldy weapon.

TheShown911 (approved):

PR #7072, opened 2022-03-08
Chameleon clothing update

Alliostra:
## About The Pull Request

~~I'm deeply sorry~~

Wearing the chameleon jumpsuit will now allow you to quickly switch between one of 4 loadouts, which you can customize yourself (the default loadouts are a vagabond, an IH op, a scientist and a NT acolyte)

Also adds a chameleon satchel, so you are no longer bound to using a backpack if you want to go full chameleon clothes.
Chameleon clothes will now properly respect what the item they are mimicking is supposed to block from view. (Previously, having a chameleon mask with the appearance of a cigarette would not display glasses)
Chameleon clothes no longer contain biomatter, but plastic instead, so moebius can now actually print the chameleon kit.

Hex helped with the nanoUI stuff. Others who I forgot about probably helped, too.

this is the result of a copy-paste coder writing original code. has mining gone too far?

grafik
## Why It's Good For The Game

Chameleon clothes OP, plz nerf!
(makes chameleon clothes actually usable)

## Changelog
:cl:
add: added the ability to quickly switch between different loadouts for chameleon clothes if the chameleon jumpsuit is worn.
add: Added the chameleon satchel
fix: chameleon items will no longer block other clothing sprites from appearing when they weren't supposed to
tweak: chameleon clothes are now made of plastic instead of bioatter
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/storage/uplink_kits.dm
+++ b/code/game/objects/items/weapons/storage/uplink_kits.dm
@@ -129,7 +129,7 @@
 
 /obj/item/storage/box/syndie_kit/chameleon
 	name = "chameleon kit"
-	desc = "Comes with all the clothes you need to impersonate most people.  Acting lessons sold seperately."
+	desc = "Comes with all the clothes you need to impersonate most people. Acting lessons sold seperately. Wearing the uniform will allow fot quick switching between appearances."
 
 /obj/item/storage/box/syndie_kit/chameleon/populate_contents()
 	new /obj/item/clothing/under/chameleon(src)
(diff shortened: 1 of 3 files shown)


ValoTheValo:
can you make it so chameleon clothing don't use biomatter, as currently moesci cant actually print thier cham kits due to them costing biomatter due to not having resource values assigned

FortuneMaster:
Would you kindly attach screenshot or two of chameleon loadout screen

Alliostra:
grafik
click on a button to set a new item to the slot.

FortuneMaster:
> grafik click on a button to set a new item to the slot.

attach it to PR description itself(for convenience of rosa who'll be making devlog too)

Firefox13 (approved):

PR #7073, opened 2022-03-09
Gives the voice changer mask chameleon capabilities

Alliostra:
## About The Pull Request

The voice changer mask now works exactly like a chameleon mask, meaning that you can change it's appearance at will.
Also moves the "toggle voice" and "change voice" verbs to the Chameleon items tab.

## Why It's Good For The Game

harder to spot someone obviously mimicking someone's voice

## Changelog
:cl:
tweak: The voice changer mask can now change appearances like a chameleon mask. 
tweak: The "change voice" and "toggle voice" verbs of the voice changer have been moved to the chameleon item tab
/:cl:

### DIFF
--- a/code/datums/uplink/stealth and camouflage items.dm
+++ b/code/datums/uplink/stealth and camouflage items.dm
@@ -33,9 +33,9 @@
 	path = /obj/item/storage/box/syndie_kit/cleanup_kit
 
 /datum/uplink_item/item/stealth_items/voice
-	name = "Voice Changer"
+	name = "Chameleon Voice Changer"
 	item_cost = 5
-	path = /obj/item/clothing/mask/gas/voice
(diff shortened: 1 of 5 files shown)


FortuneMaster:
I'd prefer to change voice changer to attachment that you apply to items that fit in gas mask slot for them to get properties of voice changer(also that chameleon stuff should be left to chameleon kit), but your idea may be done as well

Alliostra:
> I'd prefer to change voice changer to attachment that you apply to items that fit in gas mask slot for them to get properties of voice changer(also that chameleon stuff should be left to chameleon kit), but your idea may be done as well

I'd like that to be the case as well, but for now this solution works and requires much less coding

TheShown911 (approved):

PR #7075, opened 2022-03-10
Mandella buff

Alliostra:
## About The Pull Request

Yoooooo balance changes???

The mandella got hit by the nerf to non-pistol bullets, as it uses caseless rifle ammunition. Even before the nerf it wasn't the best choice, since at it's damage of 27.6 and with an AP value of 25.5, The mandella is quite weak for a gun that costs only 1 TC less than the miller.
an example: vs 60 bullet resist, the mandella deals 14.5 damage, while the miller deals 24 damage, the paco deals 16.2 damage and even the humble giskard deals 14.3 damage.

This is why I'm increasing the mandella's damage from 27.6 to 36.8 and it's AP from 25.5 to 45. With these stats, it deals 26.2 damage vs 60 bullet resist (only 2.2 more damage than the miller versus the same bullet resist).

The bullet velocity of the mandella has also been increased (similar velocity to that of a regular bullet with the accelerator gunmod) to make it a bit more unique and worth using.

## Why It's Good For The Game

a difference of 1 tc between guns should imply similar performance levels.

## Changelog
:cl:
balance: greatly buffed the damage, AP and bullet velocity of the mandella
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/projectile/pistol/mandella.dm
+++ b/code/modules/projectiles/guns/projectile/pistol/mandella.dm
@@ -18,8 +18,9 @@
 	load_method = MAGAZINE
 	mag_well = MAG_WELL_PISTOL
 	magazine_type = /obj/item/ammo_magazine/cspistol
-	damage_multiplier = 1.2
-	penetration_multiplier = 1.7
+	proj_step_multiplier = 0.8
+	damage_multiplier = 1.6
+	penetration_multiplier = 3
 	recoil_buildup = 2
 
 	spawn_tags = SPAWN_TAG_FS_PROJECTILE


zombiemimic (approved):
Looks good, codewise. I like it

FortuneMaster:
looks alright

Humonitarian:
Given that the mandella receives no buffs against style (aka. it has the default 13 of CL rounds), this PR is solid. Makes the mandella outperform the miller against armoured targets, which is a solid move.

Might want to consider reducing the spawnrate of mandella disks however, if they aren't rare enough as is.

TheShown911 (approved):
Blah blah blah boring

PR #7076, opened 2022-03-10
Makes weaken() drop items carried in hands again.

Alliostra:
## About The Pull Request

Sources of getting weakened include falling down a zlevel, being electrocuted, tripping on underplating and many more. All of these now make you drop the items you are holding in your hands again, with the exceptions of being weakened by shuttle acceleration and by radiation effects. Resting and diving will NOT make you drop items in your hands.
Jumping on top of someone from above will now also weaken the person who is being jumped on regardless of whether or not they are wearing a helmet.

## Why It's Good For The Game

If I manage to push someone over a railing, I want him to drop whatever the fuck he is holding. This primal need of man is adressed by this PR.

## Changelog
:cl:
tweak: weaken() makes you drop items again
/:cl:

### DIFF
--- a/code/modules/multiz/turf.dm
+++ b/code/modules/multiz/turf.dm
@@ -166,7 +166,7 @@ see multiz/movement.dm for some info.
 			var/fall_damage = mover.get_fall_damage()
 			if(M == mover)
 				continue
-			if(M.getarmor(BP_HEAD, ARMOR_MELEE) < fall_damage || ismob(Mover))
+			if(M.getarmor(BP_HEAD, ARMOR_MELEE) < fall_damage || ismob(mover))
 				M.Weaken(10)
 			if(fall_damage >= FALL_GIB_DAMAGE)
 				M.gib()
--- a/code/modules/multiz/turf.dm
+++ b/code/modules/multiz/turf.dm
@@ -166,7 +166,7 @@ see multiz/movement.dm for some info.
 			var/fall_damage = mover.get_fall_damage()
 			if(M == mover)
 				continue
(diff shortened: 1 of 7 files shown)


Humonitarian:
Goes against intended design. I suggest as an alternative to give Weaken() a parameter for force-dropping items.

There are multiple cases where Weaken() should not force you to drop your items, and this PR breaks those cases.

Firefox13 (changes requested):
I agree with Humon comment on this one - not all Weaken() should make you drop items, but then not all should *not* make you drop them and this should be a parameter
Probably set to true (so you drop items), because we have like two or three cases when you shouldn't drop items

code/modules/multiz/turf.dm
```diff
@@ -166,7 +166,7 @@ see multiz/movement.dm for some info.
 			var/fall_damage = mover.get_fall_damage()
 			if(M == mover)
 				continue
-			if(M.getarmor(BP_HEAD, ARMOR_MELEE) < fall_damage)
+			if(M.getarmor(BP_HEAD, ARMOR_MELEE) < fall_damage || ismob(Mover))
```
hyperioo, line 169:
```suggestion
			if(M.getarmor(BP_HEAD, ARMOR_MELEE) < fall_damage || ismob(mover))
```
_*Not testing if PR compiles screaming*_

hyperioo (changes requested):

hyperioo (approved):
Might need quick fixes later if we realize we forgot to set dropitem to FALSE somewhere. There are so many Weaken in the code it's hard to check.

PR #7081, opened 2022-03-12
Unrandomizes wires on most types of machinery

Alliostra:
## About The Pull Request

This PR undoes the randomization of wires for every instance of a type of machinery, such as airlocks. This means that the wires in one airlock are similar to those in all other airlocks, instead of assigning every airlock a random set of wires.

The only things that keep their randomized wires are robots, RIG wires, the nuke, the smartfridge, cameras and secure airlocks, because they did so before the randomisation, too.

Currently, seeing a series of locked doors I want to pass makes me not want to hack them open, but instead brute-force them open with a weapon or murder someone for access, because that feels like less effort than having to check the wires on every single door while possibly bolting and shocking the door in the process.
I could understand randomized wires if hacking was actually quick to pull off and people would use it constantly to effortlessly break into places, but that's not the case.

Randomized wires only serve the purpose of cucking people who didn't powergame for high MEC by making wiring pure pain to interact with. Wires are such a cool system, and it's sad that actually using them is discouraged through randomized wiring.

## Why It's Good For The Game

~~I just want to be able to pull off funny spark spider shenanigans~~

I would like to be able to interact with the systems of the game in a way that makes me not want to bash my head into a brick wall without having to result to powergaming.

## Changelog
:cl:
tweak: most types of machinery will no longer have randomized wiring for every instance.
/:cl:

### DIFF
--- a/code/datums/wires/airlock.dm
+++ b/code/datums/wires/airlock.dm
@@ -1,6 +1,7 @@
 // Wires for airlocks
 
 /datum/wires/airlock/secure
+	random = 1
 	wire_count = 14
 
 /datum/wires/airlock
--- a/code/datums/wires/camera.dm
+++ b/code/datums/wires/camera.dm
@@ -1,6 +1,7 @@
 // Wires for cameras.
 
 /datum/wires/camera
+	random = 1
 	holder_type = /obj/machinery/camera
 	wire_count = 6
 	descriptions = list(
--- a/code/datums/wires/nuclearbomb.dm
+++ b/code/datums/wires/nuclearbomb.dm
@@ -1,4 +1,5 @@
 /datum/wires/nuclearbomb
+	random = 1
 	holder_type = /obj/machinery/nuclearbomb
 	wire_count = 7
 	descriptions = list(
--- a/code/datums/wires/robot.dm
+++ b/code/datums/wires/robot.dm
@@ -1,4 +1,5 @@
 /datum/wires/robot
+	random = 1
 	holder_type = /mob/living/silicon/robot
 	wire_count = 5
 	descriptions = list(
--- a/code/datums/wires/smartfridge.dm
+++ b/code/datums/wires/smartfridge.dm
@@ -8,6 +8,7 @@
 	)
 
 /datum/wires/smartfridge/secure
+	random = 1
 	wire_count = 4
 
 var/const/SMARTFRIDGE_WIRE_ELECTRIFY	= 1
--- a/code/datums/wires/wires.dm
+++ b/code/datums/wires/wires.dm
@@ -11,7 +11,7 @@ var/list/wireColours = list("red", "blue", "green", "darkred", "orange", "brown"
 
 /datum/wires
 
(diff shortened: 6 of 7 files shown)


garrulousTeamotei:
Well, hacking was quick to pull off and people did use it constantly to effortlessly break into places. If anything, increase door health, or increase the minimum damage a weapon needs to do to damage a door. Make it so if you want a door gone, you need a bit more than a crowbar or a fire extinguisher. And, powergaming any stat has it's own opportunity cost. Min max one stat, and the rest suffer, not to mention any other task one might do other than increasing your stats.

Alliostra:
> Well, hacking was quick to pull off and people did use it constantly to effortlessly break into places.

Not my experience, especially since hacking takes considerable amounts of time and you're likely to get noticed and caught by someone if you actually hack your way into a department through multiple doors. And with MEC showing the functions on the wires, hacking into places must be even easier and quicker now if one has powergamed high MEC, no? Compared to "vanilla" SS13, hacking is utilized much more rarely on eris, and hacking into departments carries a serious risk of getting shot if someone catches you. And if you want to, make a PR that buffs door health yourself if you're willing to get hated by 50% of the playerbase.

garrulousTeamotei:
>Not my experience, especially since hacking takes considerable amounts of time
Well, separate experiences are separate experiences. But, breaking it down, unscrewing a panel is just about instant, pulsing the power wire is just about instant (just need to find out which is which on a door out of the way first, and remember where the wire is, what color it is) and crowbaring the door open is somewhere between just about instant to somewhere around 3 seconds, as long as your mech score isn't around 10 or less.

>And with MEC showing the functions on the wires, hacking into places must be even easier and quicker now if one has powergamed high MEC, no? 
It's actually harder now. Before, you can remember where the wire is in comparison to others (second to top, middle, last wire, etc.). Now, you need to look over all the wires looking for the correct word.

>Compared to "vanilla" SS13, hacking is utilized much more rarely on eris and hacking into departments carries a serious risk of getting shot if someone catches you.
Don't really have the best point of reference, I tend to find a server I like and stick to it. But, it's still pretty common here. They have a gun, I have a gun, everyone has a gun

Alliostra:
I find that, if you're not using an omnitool, putting your screwdriver, multitool and corwbar away/ switching hands between steps takes up the bulk of the time you'll be spending hacking airlocks, and so does closing the panel and the closing the door after hacking them (so it's not immediately obvious someone broke in). I don't know if I'm the only one, but sometimes the  nanoUI windows will also take a significant amount of time to load.

>It's actually harder now. Before, you can remember where the wire is in comparison to others (second to top, middle, last wire, etc.). Now, you need to look over all the wires looking for the correct word.
fair enough

Overall, I can only say that when playing as a department role, I've only rarely seen someone hack into a place without them either being caught or just nothing about their break-in being noticed (such as stolen items or spying sensors etc).

hyperioo (commented):
No problem codewise.

hyperioo:
image

Alliostra:
most based headcoder.

PR #7086, opened 2022-03-16
Adds matter lists to a ton of thing that previously didn't have them

githubuser4141:
## About The Pull Request

Adds materials to various weapon mods, grenades and heavy ammunition (AMR/RPG rounds) so they can be broken down at an autolathe. Recycling = good!

## Why It's Good For The Game

More recycling, more loot. Actually, less loot but 99% of maintenance stuff isn't useful anyways, and most people would prefer a new power cell over some AMR shells or thermal visors found in maints. Reduces the amount of clutter created by scientists tossing unusable rubbish about on the floor.

## Changelog
:cl:

balance: Adds materials to grenades, weapon mods and a few various knives you'd find on uplinks or in maints. Now traitors can donate materials to science by purchasing crap from their uplinks!

/:cl:

### DIFF
--- a/code/game/objects/items/weapons/grenades/emgrenade.dm
+++ b/code/game/objects/items/weapons/grenades/emgrenade.dm
@@ -4,6 +4,7 @@
 	icon_state = "emp"
 	item_state = "empgrenade"
 	origin_tech = list(TECH_MATERIAL = 2, TECH_MAGNET = 3)
+	matter = list(MATERIAL_STEEL = 3, MATERIAL_IRON = 2, MATERIAL_URANIUM = 2)
 
 /obj/item/grenade/empgrenade/prime()
 	..()
--- a/code/game/objects/items/weapons/grenades/emgrenade.dm
+++ b/code/game/objects/items/weapons/grenades/emgrenade.dm
@@ -4,7 +4,7 @@
 	icon_state = "emp"
 	item_state = "empgrenade"
 	origin_tech = list(TECH_MATERIAL = 2, TECH_MAGNET = 3)
-	matter = list(MATERIAL_STEEL = 3, MATERIAL_IRON = 2, MATERIAL_URANIUM = 2)
+	matter = list(MATERIAL_STEEL = 3, MATERIAL_PLASMA = 1, MATERIAL_SILVER = 1)
 
 /obj/item/grenade/empgrenade/prime()
 	..()
--- a/code/game/objects/items/weapons/grenades/emgrenade.dm
+++ b/code/game/objects/items/weapons/grenades/emgrenade.dm
@@ -4,7 +4,8 @@
 	icon_state = "emp"
 	item_state = "empgrenade"
 	origin_tech = list(TECH_MATERIAL = 2, TECH_MAGNET = 3)
-	matter = list(MATERIAL_STEEL = 3, MATERIAL_PLASMA = 1, MATERIAL_SILVER = 1)
+	matter = list(MATERIAL_STEEL = 3, MATERIAL_PLASMA = 1, MATERIAL_SILVER = 1, MATERIAL_IRON = 1)
+	matter_reagents = list("uranium" = 5)
 
 /obj/item/grenade/empgrenade/prime()
 	..()
@@ -19,6 +20,7 @@
 	icon_state = "lyemp"
 	item_state = "empgrenade"
 	origin_tech = list(TECH_MATERIAL = 2, TECH_MAGNET = 3)
+	matter_reagents = list("uranium" = 3)
 
 /obj/item/grenade/empgrenade/low_yield/prime() // Inheritance is a fuck . this made low yields as effective as normal.
 	var/turf/T = get_turf(src)
--- a/code/modules/projectiles/guns/mods/mods.dm
+++ b/code/modules/projectiles/guns/mods/mods.dm
@@ -72,7 +72,7 @@
 /obj/item/gun_upgrade/barrel/blender
 	name = "OR \"Bullet Blender\" barrel"
 	desc = "A curious-looking barrel bearing the Oberth insignia. A small label reads \"No refunds for any collateral damage caused\"."
-	matter = list(MATERIAL_STEEL = 1, MATERIAL_PLASTEEL = 4, MATERIAL_DIAMOND = 1)
+	matter = list(MATERIAL_STEEL = 2, MATERIAL_PLASTEEL = 3, MATERIAL_DIAMOND = 1)
 	icon_state = "Penetrator"
 	rarity_value = 30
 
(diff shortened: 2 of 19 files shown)


Alliostra:
I'd advise against giving AMR shells plasteel and especially uranium, since they can be bought from the uplink rather cheaply.

ValoTheValo:
to comment specifically just on the weapon mods. you don't need to give everything components of rare materials, plastic, steel and glass are fine for the more mundane and basic stuff. ill see if i can figure out the review feature thing to go into specifics

code/modules/projectiles/guns/mods/mods.dm
```diff
@@ -128,6 +132,7 @@
 /obj/item/gun_upgrade/trigger/cop_block
 	name = "Frozen Star \"Cop Block\" Trigger"
 	desc = "A simpler way of making a weapon display-only"
+	matter = list(MATERIAL_PLASTEEL = 2)
```
ValoTheValo, line 135: a trigger to make a gun be a dud could just cost steel or plastic. should be cheap

code/modules/projectiles/guns/mods/mods.dm
```diff
@@ -282,6 +294,7 @@
 /obj/item/gun_upgrade/scope/watchman
 	name = "Frozen Star \"Watchman\" scope"
 	desc = "Scope that can be attachet to avarage gun."
+	matter = list(MATERIAL_PLASMAGLASS = 2, MATERIAL_PLASTEEL = 2)
```
ValoTheValo, line 297: a little fancy for a extremely basic scope, could just be plastic and glass. its a low zoom scope not a targeting system

code/modules/projectiles/guns/mods/mods.dm
```diff
@@ -111,6 +114,7 @@
 /obj/item/gun_upgrade/trigger/dangerzone
 	name = "Frozen Star \"Danger Zone\" Trigger"
 	desc = "When you need that extra edge."
+	matter = list(MATERIAL_STEEL = 2, MATERIAL_DIAMOND = 1)
```
ValoTheValo, line 117: Why does this cost diamond? could just be silver or maybe gold since technology

code/modules/projectiles/guns/mods/mods.dm
```diff
@@ -144,6 +149,7 @@
 /obj/item/gun_upgrade/trigger/dnalock
 	name = "Frozen Star \"DNA lock\" Trigger"
 	desc = "There are many guns, but that one will be yours. Prevents others from using weapon with this trigger."
+	matter = list(MATERIAL_STEEL = 2, MATERIAL_DIAMOND = 1)
```
ValoTheValo, line 152: same as danger zone, doesn't make sense for it to cost diamond

code/modules/projectiles/guns/mods/mods.dm
```diff
@@ -443,6 +463,7 @@
 /obj/item/gun_upgrade/trigger/better
 	name = "Refined trigger"
 	desc = "This trigger seems to be made of durable alloys and cut to the precision of milimeters."
+	matter = list(MATERIAL_STEEL = 2, MATERIAL_PLASTEEL = 3, MATERIAL_DIAMOND = 1)
```
ValoTheValo, line 466: Inbuild mods from techno lathes are especially a problem if they give mats as these now give valuable mats and can be done theoretically infinitely as inbuilt designs on the lathe don't have license points

code/modules/projectiles/guns/mods/mods.dm
githubuser4141: It's a fairly rare find in maint and requires two minutes of standing still to remove, so it'd be logical for it to be made out of something tough

code/modules/projectiles/guns/mods/mods.dm
githubuser4141: DNA locking tech in regular TG costs diamond, so I added diamond for familiarities'

githubuser4141:
> I'd advise against giving AMR shells plasteel and especially uranium, since they can be bought from the uplink rather cheaply.

The e-crossbow on the uplink already costs 6 TC and can be recycled for 6 uranium + other materials. In testing, the shells could be purchased for 4 TC and recycling each bundle gave 2 uranium, so.. yeah you're right. Lowering it to 1 uranium each should be enough?

githubuser4141:
Removed materials from nanoforged mods.

code/game/objects/items/weapons/grenades/emgrenade.dm
```diff
@@ -4,6 +4,7 @@
 	icon_state = "emp"
 	item_state = "empgrenade"
 	origin_tech = list(TECH_MATERIAL = 2, TECH_MAGNET = 3)
+	matter = list(MATERIAL_STEEL = 3, MATERIAL_IRON = 2, MATERIAL_URANIUM = 2)
```
TheShown911, line 7: I remember these could be ordered rather cheaply from the guild, so maybe no uranium.

code/modules/projectiles/guns/mods/mods.dm
githubuser4141: Done.

code/game/objects/items/weapons/grenades/emgrenade.dm
Humonitarian: Actually, it should contain the chemical price of EMP. There is an EMP effect in the chem recipes, I recall it costs iron and some other reagent.

githubuser4141: Probably wouldn't be worth recycling if it just contained some chemicals you could make in a dispenser for free though.

githubuser4141: I'll add uranium reagent and iron material to it then.

Alliostra:
> > I'd advise against giving AMR shells plasteel and especially uranium, since they can be bought from the uplink rather cheaply.
> 
> The e-crossbow on the uplink already costs 6 TC and can be recycled for 6 uranium + other materials. In testing, the shells could be purchased for 4 TC and recycling each bundle gave 2 uranium, so.. yeah you're right. Lowering it to 1 uranium each should be enough?

I don't know if that works, but couldn't you separate each stack of bullets into 5 single bullets which could each be recycled?

githubuser4141:
> 

Tried it, when you separate a round it invalidates the entire stack as a recyclable object so you can't recycle either, so.

Humonitarian:
> > 
> 
> Tried it, when you separate a round it invalidates the entire stack as a recyclable object so you can't recycle either, so.

Should be modifiable within the stack code

githubuser4141:
> > > 
> > 
> > 
> > Tried it, when you separate a round it invalidates the entire stack as a recyclable object so you can't recycle either, so.
> 
> Should be modifiable within the stack code

Probably, 

> > > 
> > 
> > 
> > Tried it, when you separate a round it invalidates the entire stack as a recyclable object so you can't recycle either, so.
> 
> Should be modifiable within the stack code

Probably? That'd be out of the scope of this PR which is to just add mats to things that previously didn't have them, sorry.

SirRichardFrancis (approved):

hyperioo (approved):

PR #7087, opened 2022-03-16
Chemsmoke fix (not really)

Alliostra:
## About The Pull Request

This PR changes the way chemsmoke is handled. Previously, reagents would only get transferred to a mob if they touched one of the visible gas clouds created by a chem reaction. And This didn't work like 90% of the time. Now, if you are inside the area of the smoke reaction when it occurs and you're not wearing a gas mask/or active internals, 5 units of whatever was in the smoke will be transferred to your bloodstream, and 5 units will be ingested (These amounts are the same as the ones you could take in from one visible smoke cloud). 
Theoretically this means that you could take in twice the amount of chems from a single smoke reaction since you can still breathe in the smoke clouds, but that only happens extremely rarely, so I hope that's fine.

Also fixes a bug where the actually visible smoke cloud would ignore someone with a gasmask having protection from the cloud.
## Why It's Good For The Game

Fixes good
I will not be held responsible for any warcrimes committed.

## Changelog
:cl:
fix: chemical reagent smoke should be much more reliable now.
/:cl:

### DIFF
--- a/code/game/objects/effects/chem/chemsmoke.dm
+++ b/code/game/objects/effects/chem/chemsmoke.dm
@@ -162,6 +162,17 @@
 					continue
 				else if (ismob(A))
 					chemholder.reagents.touch_mob(A)
+					if(istype(A, /mob/living/carbon))
+						var/mob/living/carbon/H = A
+						var/internals = H.get_breath_from_internal()
+						var/gasmask = FALSE
+						if(H.wear_mask)
+							gasmask = (H.wear_mask.item_flags & BLOCK_GAS_SMOKE_EFFECT)
+						to_chat(world, "gasmaks: [gasmask]")	
+						to_chat(world, "internals: [internals]")					
+						if(!internals && !gasmask)
+							chemholder.reagents.trans_to_mob(H, 5, CHEM_INGEST, copy = 1)
+							chemholder.reagents.trans_to_mob(H, 5, CHEM_BLOOD, copy = 1)
 				else if(isobj(A) && !A.simulated)
 					chemholder.reagents.touch_obj(A)
 
--- a/code/game/objects/effects/chem/chemsmoke.dm
+++ b/code/game/objects/effects/chem/chemsmoke.dm
@@ -167,12 +167,10 @@
 						var/internals = H.get_breath_from_internal()
 						var/gasmask = FALSE
 						if(H.wear_mask)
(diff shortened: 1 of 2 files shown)


code/game/objects/effects/chem/chemsmoke.dm
```diff
@@ -162,6 +162,17 @@
...
+						var/mob/living/carbon/H = A
+						var/internals = H.get_breath_from_internal()
+						var/gasmask = FALSE
+						if(H.wear_mask)
+							gasmask = (H.wear_mask.item_flags & BLOCK_GAS_SMOKE_EFFECT)
+						to_chat(world, "gasmaks: [gasmask]")	
+						to_chat(world, "internals: [internals]")					
+						if(!internals && !gasmask)
+							chemholder.reagents.trans_to_mob(H, 5, CHEM_INGEST, copy = 1)
+							chemholder.reagents.trans_to_mob(H, 5, CHEM_BLOOD, copy = 1)
```
MLGTASTICa, line 175: DGotta remove debug messages , no need for () in the & check , and why are you using copy = 1 , and doing a double stack on the chems , this is both unnecesarry and will let people dupe reagents. Most predominanty , since you put copy to true , the cloud will never run out of reagents , which combined with the double transfer , will let people dupe reagents extremly easily (can't wait to see moebius gamers doing 300u uranium 1 tile nades and then having 90000 uranium reagent in their body) Should just transfer to blood a fixed amount and not let it copy..

Alliostra: Forgot about the messages. I just copied the rest from the chemsmoke cloud where it worked the same.

TheShown911 (commented):
I am not sure normal players would use their bodies to dupe chemicals, yet i am 100% sure SPCR would set up a monkey chem dupe farm with sleepers for dialysis so my hands are tied.

ValoTheValo:
> I am not sure normal players would use their bodies to dupe chemicals, yet i am 100% sure SPCR would set up a monkey chem dupe farm with sleepers for dialysis so my hands are tied.

SPCR Moment

MLGTASTICa:
> I am not sure normal players would use their bodies to dupe chemicals, yet i am 100% sure SPCR would set up a monkey chem dupe farm with sleepers for dialysis so my hands are tied.

he removed the copy so now no duping can take place.

hyperioo (approved):

PR #7088, opened 2022-03-17
Makes rigs update instantly when pressing buttons

MLGTASTICa:
## About The Pull Request
Makes the RIG UI more responsive in general by calling an update everytime a button is pressed
At higher player loads , the MC progressively gets more loaded , and the MC update tick gets called less, resulting in very unresponsive rig UI's.
## Why It's Good For The Game
nanoui autoupdate is a curse upon this codebase and a pinnacle of  coder lazyness

## Changelog
:cl:
code: RIG UI's now update at clicks.
/:cl:

### DIFF
--- a/code/modules/clothing/spacesuits/rig/rig.dm
+++ b/code/modules/clothing/spacesuits/rig/rig.dm
@@ -100,6 +100,7 @@
 
 	var/emp_protection = 0
 
+	// 2022- Just so everyone knows , this doesn't get checked at all down the line. It only checks if its on the back , regardles of its value.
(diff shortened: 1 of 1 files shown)


code/modules/clothing/spacesuits/rig/rig.dm
```diff
@@ -591,6 +592,9 @@
 		if (locked != -1)
 			locked = !locked
 
+	// Makes it so the UI instantly updates , instead of using the MC tick, way faster at high stress.
+	ui_interact(usr)
```
TheShown911, line 596: You can use `SSnano.update_uis(src)`

MLGTASTICa: changed it to use it .

PR #7098, opened 2022-03-19
Adds upgrade tiers to chem dispensers

MLGTASTICa:
## About The Pull Request
Adds tiers to the chem dispenser , making upgrading it now worth it
tier 2 parts  - inaprov , dylo kelo
3 - tricord , spaceaccilin , dermaline
4 - blattedin , polystem
5 - carpotoxin , bicaridine
6  - meralyne , nanites
## Why It's Good For The Game
All machinery should in generally have upgrades + gives a good reason for moebius to do expeditions , for those sweet T5 one-star parts

## Changelog
:cl:
balance: Upgrading the chem dispenser now provides it with more reagent options.
/:cl:

### DIFF
--- a/code/modules/reagents/machinery/chem_dispenser.dm
+++ b/code/modules/reagents/machinery/chem_dispenser.dm
@@ -34,7 +34,7 @@
 		2 = list("inaprovaline","anti_toxin","kelotane"), // basic upgrade
 		3 = list("tricordrazine","spaceacillin","dermaline"), // max moebius tech
 		4 = list("blattedin", "polsytem"), // excel tech
-		5 = list("carpotoxin, bicaridine"),// one-star
+		5 = list("carpotoxin", "bicaridine"),// one-star
 		6 = list("meralyne", "nanites") // alien
 	)
 	var/maximum_reagent_tier = 0
--- a/code/modules/reagents/machinery/chem_dispenser.dm
+++ b/code/modules/reagents/machinery/chem_dispenser.dm
@@ -33,7 +33,7 @@
 		1 = list(),
 		2 = list("inaprovaline","anti_toxin","kelotane"), // basic upgrade
 		3 = list("tricordrazine","spaceacillin","dermaline"), // max moebius tech
-		4 = list("blattedin", "polsytem"), // excel tech
+		4 = list("blattedin", "polystem"), // excel tech
 		5 = list("carpotoxin", "bicaridine"),// one-star
 		6 = list("meralyne", "nanites") // alien
 	)
--- a/code/modules/reagents/machinery/chem_dispenser.dm
+++ b/code/modules/reagents/machinery/chem_dispenser.dm
@@ -29,11 +29,27 @@
 		"radium","water","ethanol",
 		"sugar","sacid","tungsten"
 	)
+	var/list/tiered_reagents = list(
+		1 = list(),
+		2 = list("inaprovaline","anti_toxin","kelotane"), // basic upgrade
+		3 = list("tricordrazine","spaceacillin","dermaline"), // max moebius tech
+		4 = list("blattedin", "polsytem"), // excel tech
+		5 = list("carpotoxin, bicaridine"),// one-star
+		6 = list("meralyne", "nanites") // alien
+	)
+	var/maximum_reagent_tier = 0
+	var/has_tiered_reagents = TRUE
 	var/list/hacked_reagents = list()
 	var/obj/item/reagent_containers/beaker
 
 /obj/machinery/chemical_dispenser/RefreshParts()
 	cell = locate() in component_parts
+	var/sum = 0
+	for(var/obj/item/stock_parts/item in component_parts)
+		sum += item.rating
+	sum = round(sum / 4)
+	if(sum)
+		maximum_reagent_tier = sum
 
 /obj/machinery/chemical_dispenser/proc/recharge()
 	if(stat & (BROKEN|NOPOWER)) return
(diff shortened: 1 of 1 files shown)


ValoTheValo:
being able to instantly make infinite tricord and dermaline seems abit broken. and considering how common onestar parts are if you try infinite bicardine is just totally OP

MLGTASTICa:
> being able to instantly make infinite tricord and dermaline seems abit broken. and considering how common onestar parts are if you try infinite bicardine is just totally OP

moebius chemist starts with enough beakers of dylovene and inaprovaline in his 4 lockers to make 960 units of tricord , not really that big of a deal since they are both chems that are easy to make but just tedious. I believe chemists should spend more time on more complex stuff instead of on the repetitive tasks.

SaltyOldBull:
can you make it so you need a minimum amount of cog or bio to use the chem machine

MLGTASTICa:
> can you make it so you need a minimum amount of cog or bio to use the chem machine

Fuck no we ain't sojourn.

ValoTheValo:
still dont think dermaline is a good idea, nor some of the high tier stuff but im fine with tricord and paracet

Cyaneer:
If people REE too hard about instachems, could be a thing like in TG style chemdisps where you can save chem presets (aka save a recipe of chems the machine autodispenses into the beaker), with upgrades determining the max complexity

garrulousTeamotei:
> And considering how common onestar parts are if you try infinite bicardine is just totally OP.

It's easy enough to make infinite bicardine by other means, roach farming and botany. It's all time investments. Onestar components may be easy to obtain, but it's a time problem. The junk tractor beam is likely the quickest way to get the parts, but it doesn't garuentee finding a onestar field, and the cooldown on it is painful.

You need to assume the dispenser will always be upgraded to T3 if it gives good enough benefits. And, it's not like T2 and 3 chems are made of anything that isn't in the standard T1 list.

MemedHams:
This might make sense if:
Add a "cost" var to reagents that can be used to make some reagents require more power per synthesis unit.
Give the chems here increasing cost based on their tier
Nanite synthesis is a bit silly unless it can only be done in small amounts.

garrulousTeamotei:
>Nanite synthesis is a bit silly unless it can only be done in small amounts.

Isn't T6 parts admin parts?

MemedHams:
> > Nanite synthesis is a bit silly unless it can only be done in small amounts.
> 
> Isn't T6 parts admin parts?

actually fair point- but why do we have these options as admin part level in the first place?
An admin could just spawn in these reagents for themself directly

MLGTASTICa:
> > > Nanite synthesis is a bit silly unless it can only be done in small amounts.
> > 
> > 
> > Isn't T6 parts admin parts?
> 
> actually fair point- but why do we have these options as admin part level in the first place? An admin could just spawn in these reagents for themself directly

T6 isn't admin parts , just Super rare alien parts. Admin parts have 100 rating and are bluespace.

githubuser4141:
Maybe increase each tier by a level, T3 is basically a non-issue and doesn't even have special material needs, T2 isn't even used, ever, T5-T6 are super rare, though, so probably isn't that unbalanced

Autisem:
Having a consistant way to make blatadine would be fun.

TheShown911:
Won't even touch this one.

Firefox13 (changes requested):
> T6 isn't admin parts , just Super rare alien parts

drink your pills
изображение

Memed gives a good idea - tiered reagents should have a heightened power requirement - like 4 (which is a fair share on T2 ones but is better on T3+)

code/modules/reagents/machinery/chem_dispenser.dm
```diff
@@ -29,11 +29,27 @@
 		"radium","water","ethanol",
 		"sugar","sacid","tungsten"
 	)
+	var/list/tiered_reagents = list(
+		1 = list(),
+		2 = list("inaprovaline","anti_toxin","kelotane"), // basic upgrade
+		3 = list("tricordrazine","spaceacillin","dermaline"), // max moebius tech
+		4 = list("blattedin", "polsytem"), // excel tech
```
Firefox13, line 36:
```suggestion
		4 = list("blattedin", "polystem"), // excel tech
```

MLGTASTICa:
> > T6 isn't admin parts , just Super rare alien parts
> 
> drink your pills изображение
> 
> Memed gives a good idea - tiered reagents should have a heightened power requirement - like 4 (which is a fair share on T2 ones but is better on T3+)

Un Bruh Momento, ill heigten it since it should be easy enough. I think the T6 parts should be unblacklisted . just a thought.

Firefox13:
> I think the T6 parts should be unblacklisted . just a thought.

No

code/modules/reagents/machinery/chem_dispenser.dm
```diff
@@ -29,11 +29,41 @@
...
+		4 = list("blattedin", "polystem"), // excel tech
+		5 = list("carpotoxin", "bicaridine"),// one-star
+		6 = list("meralyne", "nanites") // alien
+	)
+	var/list/tiered_reagents_cost = list(
+		"inaprovaline" = 8,
+		"anti_toxin" = 8,
+		"kelotane" = 8,
+		"tricordrazine" = 8,
+		"spaceaccilin" = 8,
```
Firefox13, line 45:
```suggestion
		"spaceacillin" = 8,
```

TGCMbetcher:
> If people REE too hard about instachems, could be a thing like in TG style chemdisps where you can save chem presets (aka save a recipe of chems the machine autodispenses into the beaker), with upgrades determining the max complexity.
> 
> Or for chems to be dispensed in smaller+larger increments? T1 one only lets you dispense 5u+10u, T2 4u+15u and so on.

Please this, the free blatt is the only real bottle neck in chem I would rather have saved mixes so I dont have to measure out 5 units then 10 for 15 units in certain recipies

Firefox13 (approved):

PR #7119, opened 2022-03-28
Carrion bug fix

Pink-Chink:
## About The Pull Request

Fixes some unintended behavior in the living Life() proc where player-controlled nonhuman mobs would sleep, preventing environmental effects among others from being applied. So, carrion spider cores will no longer be able to survive indefinitely in space. Also, the spider core would die in roughly 60 seconds in a vacuum if it was at 100%., which is about the time it takes to create a new body. The timer was changed to 66 seconds.

## Why It's Good For The Game

Removes abusable behavior.

## Changelog
:cl:
tweak: added mind check to living Life() sleep check logic
balance: slightly extends the generate_body() timer to 66 seconds from 60
/:cl:

### DIFF
--- a/code/modules/mob/living/life.dm
+++ b/code/modules/mob/living/life.dm
@@ -5,7 +5,7 @@
 	. = FALSE
 	..()
 	if(config.enable_mob_sleep)
-		if(stat != DEAD)
+		if(stat != DEAD && !mind)	// Check for mind so player-driven, nonhuman mobs don't sleep
 			if(life_cycles_before_scan > 0)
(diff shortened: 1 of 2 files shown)


SirRichardFrancis (commented):
Imo all of the math is unnecessary, adjusting that `if(stat != DEAD && !mind)` and time-to-make-body will do the thing.

Pink-Chink:
I put it there in case the health was changed for whatever reason.

Firefox13 (approved):

PR #7129, opened 2022-04-01
Unleash Dusty

hyperioo:
## About The Pull Request

The veil has thinned between reality and bluespace.

Also make dirty fixes for two PRs that keeps spamming, just to avoid having Dusties constantly because of them.

## Changelog
:cl: Hyperio
tweak: Unleash Dusty
fix: Dirty fix of some PRs
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/friendly/cat.dm
+++ b/code/modules/mob/living/simple_animal/friendly/cat.dm
@@ -268,8 +268,8 @@
 
 // Runtime cat
 
-var/cat_cooldown = 20 SECONDS
-var/cat_max_number = 10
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/simple_animal/friendly/cat.dm
```diff
@@ -268,8 +268,8 @@
 
 // Runtime cat
 
-var/cat_cooldown = 20 SECONDS
-var/cat_max_number = 10
+var/cat_cooldown = 0 SECONDS  // UNLIMITED .... PAW-ER
```
Kegdo, line 271: God has abandoned us

PR #7139, opened 2022-04-05
Removes bad return in borer_powers.dm

### LINKED ISSUE #7115: Bad return in borer's infest verb
https://github.com/discordia-space/CEV-Eris/blob/8b59cb303b102335ded989e2977932299acb8206/code/modules/mob/living/simple_animal/borer/borer_powers.dm#L110-L113
This line returns when it shouldn't - after disabling the vision, it should go on to continue infestation instead of exiting.

greggbot:
## About The Pull Request

Presently, attempting to infest a host with biograde vision active will cause it to deactivate, and then return, blocking infestation when it shouldn't - it even happens after the message for wiggling into the host's ear, which can cause a lot of confusion for both borers who don't know why they're not in the host, and humans who think they're infested when they're not.
Closes #7115 

## Why It's Good For The Game

The proc shouldn't stop midway through after removing biograde vision's effects.

## Changelog
:cl:
fix: having biograde vision active as a borer no longer prevents infesting a host.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/borer/borer_powers.dm
+++ b/code/modules/mob/living/simple_animal/borer/borer_powers.dm
@@ -110,7 +110,6 @@
 	if(sight & SEE_MOBS)
 		sight &= ~SEE_MOBS
 		to_chat(src, SPAN_NOTICE("You cannot see living being through walls for now."))
-		return
 
 	host = M
 	host.status_flags |= PASSEMOTES


SirRichardFrancis (approved):
Won't save borer from being shitty antagonist tho.

githubuser4141:
> Won't save borer from being shitty antagonist tho.

that's more of a player quality issue, borers CAN BE fun and interesting if they realise there's more to life than trying to grief/kill people

githubuser4141:
> Won't save borer from being shitty antagonist tho.

I think that's more of a player quality issue, borers CAN BE fun and interesting if they realise there's more to life than trying to grief/kill people like 99% of the non-borer players do anyway.

TheShown911 (approved):

TheShown911:
Borer balance issues should be solved in balance PRs.

hyperioo (approved):

PR #7144, opened 2022-04-08
Adds a storage bag to every RIG, minor RIG buffs, minor exosuit buffs

githubuser4141:
## About The Pull Request

- Adds storage bags to each type of RIG in the game by default
- Increases medical/EVA/SCI RIG melee/ballistic armour by 10 because 10% proj resist on a light RIG seemed a bit low (light, unarmored 'punk' themed jackets are 20% resist.
- Buffs the mech flashlight, adds printable mech drill pieces to the exofab and makes the orderable cargo Ripley more viable by adding required parts.
- (QoL) adds a diagnosis unit to the mechfab exosuit tab

## Why It's Good For The Game

RIGs don't seem worth it currently if they don't come with a storage module, nobody wants to haul a bag in one hand everywhere they go or spend 30 minutes begging cargo/IH for a storage bag. Paramedic/Expedition/Light or even techno RIGs are very rarely seen anyway. Buffs a few under armored RIGs (light-EVA, medical and Science) to be more useful.

Mech flashlights seemed a bit dim compared to a heavy hand-held flashlight so I increased the brightness/range a bit to make it more comfortable to use in a dark environment. The orderable exosuit crate from classic cargo was missing almost every part necessary for it to function which should be fixed.

## Changelog
:cl:
balance: Buffs RIG armour values, adds storage modules to (hopefully) every RIG, buffs the mech flashlight. Makes the fugly MEO suit a bit faster.
tweak: Mech drillbits can be printed from the mechfab now and the orderable exosuit crate now contains all the necessary components.
/:cl:

### DIFF
--- a/code/modules/cargo/packs.dm
+++ b/code/modules/cargo/packs.dm
@@ -734,6 +734,7 @@ var/list/all_supply_groups = list("Operations","Security","Hospitality","Enginee
 /datum/supply_pack/mech_ripley
 	name = "exosuit assembly crate"
 	contains = list(
+		/obj/structure/heavy_vehicle_frame,
 		/obj/item/mech_equipment/drill,
 		/obj/item/mech_equipment/clamp,
 		/obj/item/mech_equipment/light,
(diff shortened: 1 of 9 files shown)


TGCMbetcher:
Mining eva suit has no slowdown so I would leave the armor as is. Dont add slowdown to compensate it's supposed to be that way it's the only one. Giving medical rig more armor will make it better right now its eva rig with slowdown so worse than useless

UltraNumeron:
Really agree on the bag thing rigs are pretty much worthless without it so making it a default thing is very good

drexample (approved):

PR #7153, opened 2022-04-12
Fix runtime in radio.dm

SirRichardFrancis:
## About The Pull Request

What it says on the tin.

## Why It's Good For The Game

Fixes cool.

## Changelog
:cl:
fix: runtime in radio.dm
/:cl:

### DIFF
--- a/code/game/objects/items/devices/radio/radio.dm
+++ b/code/game/objects/items/devices/radio/radio.dm
@@ -502,33 +502,28 @@ var/global/list/default_medbay_channels = list(
 	// what the range is in which mobs will hear the radio
 	// returns: -1 if can't receive, range otherwise
 
-	if (wires.IsIndexCut(WIRE_RECEIVE))
+	if(!listening || !on || !freq || wires.IsIndexCut(WIRE_RECEIVE))
 		return -1
-	if(!listening)
-		return -1
+
 	if(!(0 in level))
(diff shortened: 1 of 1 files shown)


code/game/objects/items/devices/radio/radio.dm
```diff
@@ -502,33 +502,28 @@ var/global/list/default_medbay_channels = list(
...
+	var/can_recieve = (freq == frequency)
+
+	if(!can_recieve)
+		for(var/i in channels)
+			var/datum/radio_frequency/RF = secure_radio_connections[i]
+			if(RF && RF.frequency == freq && (channels[i] & FREQ_LISTENING))
+				can_recieve = TRUE
+				break
+
+		if(!can_recieve)
```
TheShown911, line 526: There's got to be a better way to do this.

hyperioo: Since there is a given number of predefined channels, we could associate them with constants that are multiples of 2 and add that value to the headset when the radio frequency is installed and listened to? That way checking for a can recieve will just be a bit check like we do for flags.
It requires a more thorough rework though.

SirRichardFrancis:
Yes, current telecoms system is silly and does require a trough rework, that I have no time on.
No is no I guess, maybe someone will rework it in the future.

PR #7155, opened 2022-04-12
Add "Follow player" verb

SirRichardFrancis:
## About The Pull Request

New "Follow player" verb that is similar to "Follow mob", but only shows mobs that have a `ckey` assigned to them, so user don't have to scroll trough a ton of roaches, monkeys and other mobs to find a fellow human.

Removed "Follow" verb from ghost menu, now it's only visible on right click, as it should.
When called from ghost menu, it showed a list of every mob and object in view, which is pretty useless and only used by mistake.

## Why It's Good For The Game

QoL is good.

## Changelog
:cl:
add: "follow player" verb to ghost panel
tweak: "follow" verb no longer visible in ghost panel
/:cl:


Alliostra:
speedmerge

TheShown911 (approved):

PR #7160, opened 2022-04-13
Vending machine tweaks

Humonitarian:
## About The Pull Request

Converts some of the on-ship vending machines into custom ones 

- medbay's supply vendor (products remain free and requiring access)
- the MiniMed in medbay lobby
- the NeoTheology Theo-Mat

Also replaces the drink showcase minifridge with a vending machine of identical appearance.

The medbay vendors can only be supplied with medical equipment and chem containers, while the drink showcase continues to only hold items it used to be able to, meaning glasses, drinks and condiments.

The NT vending machine is not locked and can hold any item. It is now only able to serve disciplines however.

## Why It's Good For The Game

The medbay vending machines will experience a large QoL improvement, allowing the inner one to retain use through restocks as the round goes on, and the lobby one to distribute chems to the crew, while making profit for the department.

The NT vending machine sees little use beyond being a cahors dispenser, by allowing it to be stocked with products, NT will be able to supply more items to their followers.
The lock to cruciform prevents NT from completely bypassing Guild, but allows them to underprice items to encourage joining the Church, as well as allows ~disloyal~ entrepreneur disciplines to resell the cheap items at a profit.

The more than usually unused club showcase is pretty much worthless beyond storing large amounts of alcohol, which can be effortlessly accessed by anyone and stolen from. By turning it into a vending machine, the Club can prevent them from being taken, while still keeping them on display, or use the vending machine as a way to sell booze while they are busy somewhere else.
Since it cannot sell food or other similar goods, it should not immediately replace the need for a vending machine, which they can already construct using the vending machine board in their storageroom.

## Changelog
:cl:
tweak: Turned the Club showcase into a custom vending machine
tweak: Turned the medical vending machines into custom ones, allowing stocking them with chems and supplies
tweak: Turned the Theo-Mat vending machine into a custom one, allowing stocking with the products of NeoTheology
spellcheck: Fixed a few typos for vending machines
code: Changed some vending machine code
balance: TheoMat now only serves the cruciformed
/:cl:

### DIFF
--- a/code/game/machinery/vending.dm
+++ b/code/game/machinery/vending.dm
@@ -144,7 +144,7 @@
 	var/seconds_electrified = 0 //Shock customers like an airlock.
 	var/shoot_inventory = 0 //Fire items at customers! We're broken!
 
-	var/custom_vendor = TRUE //If it's custom, it can be loaded with stuff as long as it's unlocked.
+	var/custom_vendor = FALSE //If it's custom, it can be loaded with stuff as long as it's unlocked.
 	var/locked = TRUE
 	var/datum/money_account/machine_vendor_account //Owner of this vendomat. Used for access.
 	var/datum/money_account/earnings_account //Money flows in and out of this account.
@@ -182,10 +182,6 @@
 
 	if(product_ads)
 		ads_list += splittext(src.product_ads, ";")
-
-	if(!machine_vendor_account && vendor_department)
-		earnings_account = department_accounts[vendor_department]
-		machine_vendor_account = department_accounts[vendor_department]
 
 	build_inventory()
 	power_change()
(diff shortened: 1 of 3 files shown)


code/game/machinery/vending.dm
```diff
@@ -1150,6 +1165,8 @@
 	contraband = list(/obj/item/reagent_containers/pill/tox = 3,/obj/item/reagent_containers/pill/stox = 4,/obj/item/reagent_containers/pill/antitox = 6)
 	idle_power_usage = 211 //refrigerator - believe it or not, this is actually the average power consumption of a refrigerated vending machine according to NRCan.
 	auto_price = FALSE
+	custom_vendor = TRUE // Chemists can load it for MDs
```
SirRichardFrancis, line 1168: You sure money from sales won't go to the Guild? Same question for `/wallmed`.

Humonitarian: Good catch 👍

code/game/machinery/vending.dm
```diff
@@ -1452,6 +1472,32 @@
...
+		var/obj/item/clothing/C = H.w_uniform
+		if(istype(C))
+			for(var/obj/item/I in C.accessories)
+				if(istype(I, /obj/item/clothing/accessory/cross))
+					bingo = TRUE
+					break
+	if(!bingo)
+		to_chat(user, SPAN_WARNING("[src] flashes a message: Theo-Mat detected no cruciform."))
+		return
+	..()
```
SirRichardFrancis, line 1500:
```suggestion
/obj/machinery/vending/theomat/try_to_buy(obj/item/W, var/datum/data/vending_product/R, var/mob/user)
	var/bingo = FALSE
	if(ishuman(user))
		var/mob/living/carbon/human/H = user
		if(is_neotheology_disciple(H))
			bingo = TRUE

		else if(istype(H.get_active_hand(), /obj/item/clothing/accessory/cross))
			bingo = TRUE

		else if(istype(H.wear_mask, /obj/item/clothing/accessory/cross))
			bingo = TRUE

		else if(H.w_uniform && istype(H.w_uniform, obj/item/clothing))
			var/obj/item/clothing/C = H.w_uniform
			for(var/obj/item/I in C.accessories)
				if(istype(I, /obj/item/clothing/accessory/cross))
					bingo = TRUE
					break

	bingo ? ..() : to_chat(user, SPAN_WARNING("[src] flashes a message: Unauthorized Access."))
```

SirRichardFrancis (approved):
Looks okay. Let's hope that Guild won't be able to empty all vendors somehow.

hyperioo (approved):

PR #7169, opened 2022-04-15
deletes clickcatchers and possibly reduces Full Auto lag

vode-code:
## About The Pull Request
This PR cleanly deletes clickcatchers, and might reduce FA lag a bit from that and one removal of a proc.
My apologies for the commit spam, I stopped the new ones being made and thought a previous PR being merged would stop them showing up but I guess not?

## Why It's Good For The Game
Clickcatchers are bad and so is full auto lag.

## Changelog
:cl: Chickenish
del: Removed those unnamed things everywhere, proper name "click catcher"s.
del: middle mouse button no longer switches hands.
/:cl:


MLGTASTICa:
Can you explain  how this actually would HELP with the full auto lag and HOW guns now handle full auto?

TheShown911:
Does this just remove full auto?

vode-code:
> Does this just remove full auto?

No, I tested and it still works after.

vode-code:
> Can you explain how this actually would HELP with the full auto lag and HOW guns now handle full auto?

1. less abstraction means less computing the abstraction and thus less lag.
2. exactly the same as before, just with a proc used in only one place replaced with its contents and deleted.

vode-code:
I deleted click catchers because their only use is middle click hand swap hotkey, they have a bit of lag for /datum/click_handler/fullauto/MouseDrag , and seeing them on rightclicks everywhere is ugly.

MLGTASTICa:
> > Can you explain how this actually would HELP with the full auto lag and HOW guns now handle full auto?
> 
> 1. less abstraction means less computing the abstraction and thus less lag.
> 2. exactly the same as before, just with a proc used in only one place replaced with its contents and deleted.

right ,but the thing is that full auto firing itself never lagged enough in the first place to be a significant problem . The real issues that are lagging the codebase are machines , atmospherics and a few other problems. 
The motive people lag when they fire a weapon is mostly because of all the visual rendering and clipping done with the lightning system.

MLGTASTICa:
> > > Can you explain how this actually would HELP with the full auto lag and HOW guns now handle full auto?
> > 
> > 
> > 
> > 1. less abstraction means less computing the abstraction and thus less lag.
> > 2. exactly the same as before, just with a proc used in only one place replaced with its contents and deleted.
> 
> right ,but the thing is that full auto firing itself never lagged enough in the first place to be a significant problem . The real issues that are lagging the codebase are machines , atmospherics and a few other problems. The motive people lag when they fire a weapon is mostly because of all the visual rendering and clipping done with the lightning system
for the bullets.

SirRichardFrancis:
That will break MMB hotkey (commonly used by me, and most likely by many others), and ability to target tiles in the dark and behind smoke.
Here is picture from #6848 to demonstrate the issue.
image

Reducing lag is great, but please don't break stuff while you do so.
Especially if performance gain is... quite elusive. ¯\\\_(ツ)\_\/¯

drexample:
Don't remove click catchers.

PR #7188, opened 2022-04-21
Gun frames in trash/maintenance spawn with 1-2 parts installed

Pink-Chink:
## About The Pull Request

Gun frames that spawn in trash piles or maintenance will come with part(s) preinstalled. One part is guaranteed and the second has a 50% of spawning. Additionally, using a screwdriver on the frame will let you take out any parts currently installed.

## Why It's Good For The Game

Reduces randomness in gun part hunting.

## Changelog
:cl:
tweak: gun parts can be removed from frames with a screwdriver
balance: maint/trash gun frames spawn with 1-2 parts installed
/:cl:

### DIFF
--- a/code/datums/craft/gun_parts.dm
+++ b/code/datums/craft/gun_parts.dm
@@ -35,7 +35,7 @@
 	if(spawn_with_preinstalled_parts)
 		var/list/parts_list = list(pick(gripvars), mechanism, barrel)
 
-		parts_list -= pick(parts_list)
+		pick_n_take(parts_list)
 		if(prob(50))
 			parts_list -= pick(parts_list)
 
--- a/code/datums/craft/gun_parts.dm
+++ b/code/datums/craft/gun_parts.dm
@@ -91,9 +91,8 @@
 	if(istool(I))
 		if(I.get_tool_quality(QUALITY_SCREW_DRIVING))
 			var/list/possibles = contents.Copy()
-			possibles += "Cancel"
 			var/obj/item/part/gun/toremove = input("Which part would you like to remove?","Removing parts") in possibles
-			if(toremove == "Cancel")
+			if(!toremove)
 				return
 			if(I.use_tool(user, src, WORKTIME_INSTANT, QUALITY_SCREW_DRIVING, FAILCHANCE_ZERO, required_stat = STAT_MEC))
 				eject_item(toremove, user)
--- a/code/datums/craft/gun_parts.dm
+++ b/code/datums/craft/gun_parts.dm
@@ -37,7 +37,7 @@
 
 		pick_n_take(parts_list)
 		if(prob(50))
-			parts_list -= pick(parts_list)
+			pick_n_take(parts_list)
 
 		for(var/part in parts_list)
 			if(ispath(part, grip))
--- a/code/datums/craft/gun_parts.dm
+++ b/code/datums/craft/gun_parts.dm
@@ -21,9 +21,18 @@
 	var/barrel = /obj/item/part/gun/barrel
 	var/barrel_attached = FALSE
 
-/obj/item/part/gun/frame/New(loc, junk)
+/obj/item/part/gun/frame/New(loc)
 	..()
-	if(junk)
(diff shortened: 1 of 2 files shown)


SaltyOldBull:
alright yeah this is based.

code/datums/craft/gun_parts.dm
```diff
@@ -21,6 +21,40 @@
...
+/obj/item/part/gun/frame/New(loc)
+	..()
+	var/spawn_with_preinstalled_parts = FALSE
+	if(istype(loc, /obj/structure/scrap_spawner))
+		spawn_with_preinstalled_parts = TRUE
+	else if(in_maintenance())
+		var/turf/T = get_turf(src)
+		for(var/atom/A in T.contents)
+			if(istype(A, /obj/spawner))
+				spawn_with_preinstalled_parts = TRUE
```
TheShown911, line 33: This is very cursed but okay.

TheShown911 (approved):

code/datums/craft/gun_parts.dm
```diff
@@ -21,6 +21,40 @@
...
+
+	if(spawn_with_preinstalled_parts)
+		var/list/parts_list = list(pick(gripvars), mechanism, barrel)
+
+		parts_list -= pick(parts_list)
```
hyperioo, line 38:
```suggestion
		pick_n_take(parts_list)
```

code/datums/craft/gun_parts.dm
```diff
@@ -21,6 +21,40 @@
...
+		if(prob(50))
+			parts_list -= pick(parts_list)
```
hyperioo, line 40:
```suggestion
			pick_n_take(parts_list)
```

code/datums/craft/gun_parts.dm
```diff
@@ -53,6 +87,23 @@
...
 			to_chat(user, SPAN_NOTICE("You have attached the barrel to \the [src]."))
 			return
+
+	if(istool(I))
+		if(I.get_tool_quality(QUALITY_SCREW_DRIVING))
+			var/list/possibles = contents.Copy()
+			possibles += "Cancel"
+			var/obj/item/part/gun/toremove = input("Which part would you like to remove?","Removing parts") in possibles
+			if(toremove == "Cancel")
+				return
```
hyperioo, line 97:
```suggestion
			var/obj/item/part/gun/toremove = input("Which part would you like to remove?","Removing parts") in possibles
			if(!toremove)
				return
```
There is already a Cancel button to the input prompt, no need to add a "Cancel" option in the list.

code/datums/craft/gun_parts.dm
```diff
@@ -21,6 +21,40 @@
...
+
+		for(var/part in parts_list)
+			if(ispath(part, grip))
+				new part(src)
+				grip_attached = TRUE
+			else if(part in gripvars)
+				var/variantnum = gripvars.Find(part)
+				result = resultvars[variantnum]
+				new part(src)
+				grip_attached = TRUE
```
hyperioo, line 50: Could you explain why this is needed? 
part is picked from `list(pick(gripvars), mechanism, barrel)` so it's either a grip, mechanism or barrel. Or do you know a case where gripvars do not contain only items of type `/obj/item/part/gun/grip`?

hyperioo (changes requested):

code/datums/craft/gun_parts.dm
Pink-Chink: It's not needed with the current gun frames, but I wanted to have something in case a variant grip was added when a specific grip was defined. Say, the grip is `/obj/item/part/gun/grip/wood`, but `/obj/item/part/gun/grip/rubber` is a variant. It would fail that first check and not add anything.

code/datums/craft/gun_parts.dm
Pink-Chink: I couldn't figure out an elegant way to make only the maint and trash pile spawns have the extra parts without creating infinite part exploits.

hyperioo (approved):

PR #7195, opened 2022-04-23
readds the chameleon satchel

Alliostra:
## About The Pull Request

readds the chameleon satchel to the chameleon kit. the kit now comes in a box again because of this.

## Why It's Good For The Game

Being able to choose between using satchel or backpack functionality when wearing chameleon clothes good.

## Changelog
:cl:
tweak: The chameleon kit now comes in a box instead of in a backpack, and contains both a satchel and a backpack.
/:cl:

### DIFF
--- a/code/datums/uplink/stealth and camouflage items.dm
+++ b/code/datums/uplink/stealth and camouflage items.dm
@@ -24,7 +24,7 @@
 	name = "Chameleon Kit"
 	desc = "Comes with all the clothes you need to impersonate most people. Acting lessons sold separately. Wearing the uniform will allow for quick switching between appearances."
 	item_cost = 5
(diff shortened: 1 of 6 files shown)


SirRichardFrancis:
It won't be able to mimic anything other than satchels, consider instead adding a satchel-backpack switch (two variables difference there) to existing chameleon backpack.
Also return of magic box is bad imo, it's a cardboard box that fit a lot of items it shouldn't with only that as an explanation:

image

TheShown911 (commented):
Just make the backpack act as a satchel.

cdb-is-not-good:
Why not just make the rest of the kit come in the satchel instead of in a magic box?

PR #7201, opened 2022-04-28
Fix and tweak serial numbers 

SirRichardFrancis:
## About The Pull Request

In serial types:
Replaced "M" (Moebius) with "ML" (Moebius Laboratories).
Replaced "REV" with "Excelsior" on Excel guns.
Replaced "REV" with "SA" (Serbian Arms) on Serbian guns.
Replaced "PAR-TS" (TS for "top secret") with "PAR".
Removed "SM" (self-manufactured) from handmade guns.

Added radial menu for gun and tool interactions.
Added serial number to gun frames, it will be kept on gun assembly/disassembly.
Added requirement to stand nearby to read serial numbers.

image

image

## Why It's Good For The Game

Makes serial number mechanic implementation less half-assed.

## Changelog
:cl:
add: radial menu for tool-gun interaction
add: serial numbers attached to a gun frame
tweak: serial types changed to make more sense
/:cl:


code/datums/craft/gun_parts.dm
```diff
@@ -20,6 +20,15 @@
...
 	var/barrel = /obj/item/part/gun/barrel
 	var/barrel_attached = FALSE
+	var/serial_type = ""
+
+
+/obj/item/part/gun/frame/New(loc, ...)
+	. = ..()
+	var/obj/item/gun/G = new result(null)
+	if(G.serial_type)
+		serial_type = G.serial_type
```
TheShown911, line 30: You can use `initial()` here instead of instancing a new gun (or at least delete the new gun afterwards)

SirRichardFrancis: That gun instance are being deleted when proc is finished (hence null spawn location).
Not sure if initial could be called on a type, but I'll try.

MLGTASTICa: it can be called on any var

SirRichardFrancis: Also gun's proper serials only generated on new() call.

SirRichardFrancis: > But `var/obj/item/gun/G` is a variable only in the scope of this proc

It is, but I don't understand where you're going with this.

> it can be called on any var

Doesn't really work for me.

SirRichardFrancis: It won'd _qdel_ it exactly, but since there is no references, byond garbage collector won't go scan everything for references.

TheShown911:
Merge conflict.

SirRichardFrancis:
Solved.

PR #7204, opened 2022-04-28
Fix jumping over deployable barriers

SirRichardFrancis:
## About The Pull Request

Deployable barriers stopped being an obstacle with implementation of diving/sliding PR, which is clearly unintended change.

UPD: Players with "Parkour" perk still able to jump over barricades, because why not.

image

## Why It's Good For The Game

Fixes cool.

## Changelog
:cl:
fix: no longer can jump over security barrier
/:cl:

### DIFF
--- a/code/game/machinery/deployable.dm
+++ b/code/game/machinery/deployable.dm
@@ -290,16 +290,11 @@ for reference:
 		icon_state = "barrier[locked]"
 
 /obj/machinery/deployable/barrier/CanPass(atom/movable/mover, turf/target, height=0, air_group=0)//So bullets will fly over and stuff.
-
 	if(istype(mover,/obj/item/projectile))
 		return (check_cover(mover,target))
 
-	if(air_group || (height==0))
-		return 1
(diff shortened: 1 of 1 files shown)


code/game/machinery/deployable.dm
```diff
@@ -290,16 +290,16 @@ for reference:
...
-	if(istype(mover) && mover.checkpass(PASSTABLE))
-		return 1
-	else
-		return 0
+		return TRUE
+
+	if(ishuman(mover))
+		var/mob/living/carbon/human/H = mover
+		if(H.checkpass(PASSTABLE) && H.stats?.getPerk(PERK_PARKOUR))
+			return TRUE
```
TheShown911, line 302: Do you need a return FALSE after this?

SirRichardFrancis: Why? If no conditions for `return TRUE` is met, `null` would be returned, which is the same as `FALSE` for the purpose of `CanPass()` check.

TheShown911 (approved):

PR #7232, opened 2022-05-19
Colt 1911 now only takes normal magazines

ValoTheValo:
## About The Pull Request
When most pistols were given the ability to load high-cap the colt 1911 was included by mistake i think. this PR makes it only load single stack mags like the real colt 1911. Salt PR

## Why It's Good For The Game
Differentiates the colt from other pistols and makes it more """realistic"""

## Changelog
:cl:
balance: Colt 1911 can no longer load highcap magazines
/:cl:


SirRichardFrancis:
>Differentiates the colt from other pistols 

By making it shittier?

>and makes it more """realistic"""

image

ValoTheValo:
> 

i honestly just found it wierd that the colt could take double stack mags.

tristain01:
2011s exist, though they are chambered in 9mm most of the time, considering we have a catch all cartridge in .35-auto, I would leave it as is. Just my opinion though.

TheShown911:
Cringe!

FortuneMaster:
I'm not the one to judge eris combat balance, but this looks like a nerf for the sake of nerf. Making things realistic doesn't make game more fun

PR #7242, opened 2022-05-21
Export and Import multiplier tweak

TGCMbetcher:
## About The Pull Request

Changes export prices and sets the import to prevent abuse

## Why It's Good For The Game

Knowing you lose almost half the money on export is depressing. I would rather lower all the prices than know I lose 40% on export. This also is for the recycling vendor because it gives at cost so guild ill be at a loss until this is merged. Fixes the export scanner being wrong by proxy

## Changelog
:cl:
tweak: import and export prices
/:cl:

### DIFF
--- a/code/__DEFINES/trade.dm
+++ b/code/__DEFINES/trade.dm
@@ -4,7 +4,7 @@
 
 // Export price multipliers
 #define NONEXPORTABLE 0
-#define JUNK 0.1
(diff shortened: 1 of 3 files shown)


hyperioo:
We'll have to see if it does not allow again stuff like infinite crowbar production thing with that 100% export price.

hyperioo (approved):

PR #7245, opened 2022-05-24
Fix roach layers

SirRichardFrancis:
## About The Pull Request

Makes dead roaches always display under stunned roaches, with conscious roaches always being on top.

image

## Why It's Good For The Game

Pretty sure dead roaches not meant to be on top, that's just inconvenient.

## Changelog
:cl:
fix: live roaches sometimes displaying below dead roaches
/:cl:


hyperioo:
Ameno

PR #7251, opened 2022-05-27
Aiming rework

Humonitarian:
## About The Pull Request

Target bodypart randomization now only happens when a bullet is fired, chance depends on distance and the recoil of the user.

Base miss chance now takes size into account (monkeys are harder to hit, mechs are easier), Vigilance only reduces miss chance on the actual target, not bystanders. This also weakens unaimed sprayfire, but lessens friendly fire.

## Why It's Good For The Game

Fullauto and high recoil weapons are less lethal, but still just as good at downing. Justifies a lower base recoil for all firearms, improving user experience.

## Changelog
:cl:
balance: bullet miss chance takes size into account, vigilance hit chance bonus only triggers on the target clicked on
code: bodypart randomization is determined when a bullet is launched, rather than when the bullet reaches a target
/:cl:


code/__DEFINES/maths.dm
```diff
@@ -224,11 +224,18 @@ proc/n_ceil(var/num)
...
+			pos++
+	return pos
 
 
 #define T100C 373.15 //  100.0 degrees celsius
 
 
 
-#define CELSIUS + T0C
+#define CELSIUS + T0C
```
hyperioo, line 241: End of line
```suggestion
#define CELSIUS + T0C

```

Humonitarian: woah

hyperioo (approved):

PR #7293, opened 2022-07-01
Fix runtime in mech_movement.dm

SirRichardFrancis:
## About The Pull Request

image

## Why It's Good For The Game

Fixes cool.
## Changelog
:cl:
fix: runtime in mech_movement.dm
/:cl:

### DIFF
--- a/code/modules/mechs/mech_movement.dm
+++ b/code/modules/mechs/mech_movement.dm
@@ -120,7 +120,8 @@
 	return !pilots.len
 
 /mob/living/exosuit/get_fall_damage(var/turf/from, var/turf/dest)
-	return (from.z - dest.z > 1) ? (50 * from.z - dest.z) : 0 //Exosuits are big and heavy //but one z level can't damage them
+	//Exosuits are big and heavy, but one z level can't damage them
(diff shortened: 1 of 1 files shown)


code/modules/mechs/mech_movement.dm
```diff
@@ -120,7 +120,8 @@
 	return !pilots.len
 
 /mob/living/exosuit/get_fall_damage(var/turf/from, var/turf/dest)
-	return (from.z - dest.z > 1) ? (50 * from.z - dest.z) : 0 //Exosuits are big and heavy //but one z level can't damage them
+	//Exosuits are big and heavy, but one z level can't damage them
+	. = (from && dest) ? ((from.z - dest.z > 1) ? (50 * from.z - dest.z) : 0) : min(15, maxHealth * 0.4)
```
hyperioo, line 124: Why put `min(15, maxHealth * 0.4)`? Do you expect it to trigger in some cases? Shouldn't `from` never be null where we are calling `get_fall_damage` in the code?

SirRichardFrancis: >Why put min(15, maxHealth * 0.4)?

It's what parent proc does.

>Do you expect it to trigger in some cases?

Yes.

>Shouldn't from never be null where we are calling get_fall_damage in the code?

It probably should, but isn't, this proc often called with no arguments.

TheShown911 (approved):

PR #7313, opened 2022-07-23
Add mechs picking up ore from the floor

SirRichardFrancis:
## About The Pull Request

Mekas could pick up ores from the floor only via the drill module, meaning that mechas can't pick up ores if it's mined via a plasma cutter or dropped by golems.
This PR adds interaction of mech clamp with the turf.

## Why It's Good For The Game

I've been asked to make this, so probably someone actually uses mechs for mining.

## Changelog
:cl:
add: mechs picking up ore from the floor with mounted clamp
/:cl:

### DIFF
--- a/code/modules/mechs/equipment/utility.dm
+++ b/code/modules/mechs/equipment/utility.dm
@@ -83,9 +83,6 @@
 					step_away(M, owner)
 					to_chat(user, "You push [target] out of the way.")
 					owner.visible_message("[owner] pushes [target] out of the way.")
-//			else if(isturf(target))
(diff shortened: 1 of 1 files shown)


githubuser4141:
Awesome, the update we've all (citation needed) been waiting for, for two years. Uh. While you're here, mind looking at this Occulus PR? Apparently there's a oversight/shitcode with the meka drills that causes lag when drilling.

https://github.com/Occulus-Server/Occulus-Eris/pull/800

TheShown911 (approved):

PR #7324, opened 2022-07-30
Fixes cover functionality when overshooting [RTM]

Humonitarian:
## About The Pull Request

When a bullet reaches someone behind cover, the bullet will always miss unless they are the intended target (were directly clicked on, or an NPC is targeting them).

Tested on test server.

## Why It's Good For The Game

Removes an exploit, makes cover more effective, especially against sprayfire. Does not prevent long-range snipers from shooting over cover, and if they aimed correctly, will still enable them to hit someone directly behind cover.

## Changelog
:cl:
add: Cover properly triggers when overshooting
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -313,6 +313,20 @@
 
 	var/result = PROJECTILE_FORCE_MISS
 
+	if(target_mob != original) // If mob was not clicked on / is not an NPC's target, checks if the mob is concealed by cover
+		var/turf/cover_loc = get_step(get_turf(target_mob), get_dir(get_turf(target_mob), starting))
+		for(var/obj/O in cover_loc)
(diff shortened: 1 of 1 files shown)


MLGTASTICa:
All good except that theres no skill check for it hitting ? how is original target lost based on skill ? for all i know vig only makes it sure the  bullet hits the targeted organ , but not the mob itself.

Humonitarian:
> All good except that theres no skill check for it hitting ? how is original target lost based on skill ? for all i know vig only makes it sure the bullet hits the targeted organ , but not the mob itself.

"Original target" is the target you click on

Vig is unrelated in this calculation. If someone's behind cover and you're not aiming at them, you're not going to hit them. If they are out of cover, or you're aiming specifically to hit them, everything works as it used to.

PR #7325, opened 2022-07-30
Vendors fixes

Humonitarian:
## About The Pull Request

Drinks Showcase showcases drinks now
Theomat no longer eats cash
Custom vendors no longer automatically guild-affiliated
MiniPharma finally shut up

## Why It's Good For The Game

Bugfix good

## Changelog
:cl:
fix: various vending machine issues
/:cl:

### DIFF
--- a/code/game/machinery/vending.dm
+++ b/code/game/machinery/vending.dm
@@ -1512,7 +1512,7 @@
 	return FALSE
 
 /obj/machinery/vending/theomat/attackby(obj/item/I, mob/user)
-	if(!I.tool_qualities && check_NT(user))
+	if(I.tool_qualities || check_NT(user))
 		..()
 
 /obj/machinery/vending/powermat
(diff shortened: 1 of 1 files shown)


code/game/machinery/vending.dm
```diff
@@ -1504,12 +1511,8 @@
...
 	return FALSE
 
-/obj/machinery/vending/theomat/vend(datum/data/vending_product/R, mob/user)
-	if(check_NT(user))
-		..()
-
-/obj/machinery/vending/theomat/try_to_buy(obj/item/W, var/datum/data/vending_product/R, var/mob/user)
-	if(check_NT(user))
+/obj/machinery/vending/theomat/attackby(obj/item/I, mob/user)
+	if(!I.tool_qualities && check_NT(user))
```
SirRichardFrancis, line 1515: Can only use non-tool and only if NT disciple? Why?

Humonitarian: Ah good catch, those two should be separate.

code/game/machinery/vending.dm
```diff
@@ -1749,6 +1752,7 @@
 	name = "Custom Vendomat"
 	desc = "A custom vending machine."
 	custom_vendor = TRUE
+	vendor_department = null
```
TheShown911, line 1755: Sure this won't runtime without any department?

Humonitarian: This is what they were before I set the default to guild

PR #7333, opened 2022-08-03
Turns agony into halloss [test this]

Humonitarian:
## About The Pull Request

Was surprisingly simple.

Important note: taser_effect does everything nodamage does. If taser_effect is set to true, damage other than agony damage cannot be done (apparently). Not that many projectiles have to worry about this, and ebows were fixed in #6966.

## Why It's Good For The Game

Cleaner code, 32 bytes freed up from /projectile.

## Changelog
:cl:
refactor: agony is halloss now
/:cl:


TheShown911 (commented):
This is very much a major balance change as i suppose hallos would be affected by damage modifiers?

Humonitarian:
> This is very much a major balance change as i suppose hallos would be affected by damage modifiers?

I think so, actually, yes

Humonitarian:
#7353 implements this

PR #7341, opened 2022-08-05
flash-protected sunglasses for head stuff and fix desc.

pexel5:
## About The Pull Request

Gave the heads the opportunity to take in loadout glasses, with flash protection. Guild Merchant now has normal protective glasses in preloadout. I changed the name of ordinary glasses so that there would be no misunderstandings.

## Why It's Good For The Game

Fixes good. Protection from flashes for command stuff is also not bad.

## Changelog
:cl:

add: The ability to take glasses that protect from flash for heads of departments/
fix: Sunglasses desc.

/:cl:

### DIFF
--- a/code/modules/clothing/glasses/misc.dm
+++ b/code/modules/clothing/glasses/misc.dm
@@ -66,7 +66,7 @@
 	body_parts_covered = 0
 
 /obj/item/clothing/glasses/sunglasses
-	desc = "Strangely ancient technology used to help provide rudimentary eye cover. Enhanced shielding reduces many flashes."
+	desc = "Strangely ancient technology used to help provide rudimentary eye cover. Doesn't really protects from flashes, but stylish."
 	name = "sunglasses"
 	icon_state = "sun"
 	item_state = "sunglasses"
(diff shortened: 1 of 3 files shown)


Firefox13 (approved):

code/modules/clothing/glasses/misc.dm
```diff
@@ -66,7 +66,7 @@
 	body_parts_covered = 0
 
 /obj/item/clothing/glasses/sunglasses
-	desc = "Strangely ancient technology used to help provide rudimentary eye cover. Enhanced shielding reduces many flashes."
+	desc = "Strangely ancient technology used to help provide rudimentary eye cover. Doesn't really protects from flashes, but stylish."
```
Humonitarian, line 69: This is wrong. Regular sunglasses do provide flash protection. What they don't provide is flash immunity. Pairing sunglasses with a helmet makes you flash immune.

SecHUDs also only provide limited protection, the same level regular sunglasses do. Consider accounting for this.

hyperioo: Yeah, there might be a better way to express that but the "Doesn't really protect" sort of mean "flash protection but not flash immunity".

PR #7346, opened 2022-08-05
[Not tested]The latest nerf from the thing that killed me (Makes acid take into account clothing properly)

MLGTASTICa:
## About The Pull Request
Acid no longer teleports instantly to the players skins and melts them from the inside . it now must go layer by layer until it has its effect applied

## Why It's Good For The Game
Polyacid 120u beakers should be no longer a easy insta-kill + its funny to strip people of their armor with acid.
## Changelog
:cl:
bal: Acid now melts clothes before dealing damage
/:cl:


pexel5:
BRUH THIS WORKED

PR #7353, opened 2022-08-08
[Ready] Armor and penetration system rework

Humonitarian:
## About The Pull Request

Completely reworks our armor system to use SUBSTRACTIVE armor instead of MULTIPLICATIVE.
Removes GDR, which was a substractive armor system.

Replaces Armor Penetration with Armor Division, a divisor value used to reduce a target's armor.

All armor values and all melee weapons/projectiles have been adjusted.

Implements ablative protection for RIGs, an extra layer of armor that decreases as it blocks damage.

Gun stats updated to the new values, ablative values show up in the RIG menu.

Contains the agony to halloss PR, and relevant adjustments.

## Why It's Good For The Game

Makes balancing significantly easier, solves major issues with how penetration functions (specifically for weapons like the milller), while offering an armor system just as robust as what Reere designed.

Also, ablative RIGs are better.

## Changelog
:cl:
balance: New armor system implemented
/:cl:


Humonitarian:
An explosion proc errored the integration test apparently?

maikilangiolo:
Humon wanted me to write the explanation of the PR in the comments instead of updating the OP post, so here it is.

First of all, the scope of the PR was not to rebalance all guns and change them, but rather to change the underlying armor and AP system not to be retarded both to work with and mechanically. Nearly all guns have not been touched, except in numbers to make them do the same damage they did before, but a few exceptions exist which I will list in every category. The change keeps in mind the current eris balace/and trend: pistols are high damage since small mags and only semi, revolvers are essentially better pistols with even less ammo. smgs are smgs: low recoil, fast firing, lots of ammo, lower damage vs the pistol counterparts. shotguns are high damage, low bullet speed and low ammo count and rifles are basically the generalists: average damage, average rof, good magazine size.
That said, lets get into a little bit more detail.

- first, the armor system: armor was changed to be a direct 1:1 reduction of damage, and the current armor values are the old ones divided by 4. The old % system was done so that low armor was still meaningfull and high armor was not too good. Now low armor doesnt protect as well as it did before, however the only low armor items were style clothing which had "meaningful" armor and also giving every benefit of style, so it is essentally a buff to armor and a slight nerf to style.
- second, we added a new gun "modifier": wounding modifier. Taking inspiration from gurps, guns now will have a modifier that changes damage past armor. these go from half damage, normal, 50% more damage and twice the damage. Why did we do this? Well, we hit a brick wall when we found out rifles were worse or the same as smgs, and not just that: the three rifles calibers were the same, and could not be differentiated without complex and chaotic ap/dmg modifers we wanted to clean out. It allows us to have high ap weapons by giving them high damage which gets halved past armor, and vice versa for low ap guns. At the same times, calibers have been cleaned so much: all calibers have 1 AP (now called divisor), with rifle calibers being 1.5. More easy to compare calibers between each other.

Now onto the guns:

- .35 pistols/revs: same as before. notable differences: the paco is not on par with the 1911 and .40 guns doing 40 damage a shot, its now a medium pistol like the olivaw, and baseline mk58 but gets the best recoil out of every pistol. Skydriver now does the same damage no matter the armor since highest divisor possible, and gets slightly faster bullets. Its an artefact gun, people. The makarov (as per humon request) was made into a lighter miller: good vs flesh, loses effectiveness vs armor, since the excel drozd and vintorez are ap weapons, and before it was as bad as a giskard.
- .40 pistols/revs: lamia and avarasala used to do more or the same damage as the .40 revolvers. they do slightly less damage now but have crazy good recoil (I did not change this). the .40 revolvers were brought in line to be less direct upgrades: consul is a baseline, no frills revolver. the deckard is the ap revolver (also with faster bullets) and the miller works properly by being a handcannon that rapes low armor but sucks vs armor (like it was intended to, so says the code and the old stats). The unique mateba gets deckard's levels of ap with the miller's damage and 7 round cylinder. very good.
- .35/.40 smgs: whew, this was hell. smgs being pistol calibers did as much or more damage than rifles did, and the c20r had like 1/2 points of damage less than the sts, with more ammo and less recoil, smaller size and technically a scope. Compared to before, smgs were changed in how they performed compared to other guns (less good vs armor), but internally they are the same. the antag c20r is the best smg and the moebius c20m is the second best, still pretty good. The zoric and drozd, which both stated having poor recoil control in the fluff, actually had the best recoil than any gun on eris (better than pistols); the zoric could be used with one hand with 0% deviation, and despite having less than half the Ap of the drozd it was better vs armor due to sheer high damage. no more: zoric is not the best anti armor smg, and both .40 smgs have worse recoil on paper traded for the higher damage and more controllable rof (and thus recoil). atreides got the second lowest recoil to be the one handed smg it was meant to be, straylight is the same as before and the molly gets a sol-like fast burst since it was either identical to the clarissa/anne or identical to the straylight.
- the rifles: you know how I said all guns were kept the same? half truth, but they were changed into what they were supposed to be, so technically they have the original balance. like i mentioned before, rifles were the same with minor differences (sts higher damage, krinkov a joke gun doing little damage). hell wintermute, sol and IH ak were the same gun. thanks to the magic of gurps, now eris can benefit from !!cool guns!! that arent just cosmetic preferences. .20 guns have the lowest flesh damage of all rifles but the highest damage against medium to heavy armor (and .20 was already a lower recoil caliber). .30 rifles are the opposite, and .25 is a middle ground caliber featuring higher bullet velocities to make it an easy to use caliber with no significant strenghts or weaknesses. the kovacs is now a proper dmr/battle rifle thanks to spcr rof fix, the IH ak no longer does more damage than both the antag AKs, but has better recoil in the category. the civilian ak does slightly less damage rather than being identical to IH ak except not foldable, the krinkov has less damage but is roughly on par with the other AKs vs high armor (it used to have less damage but higher AP, but didnt work as intended).
- the shotguns: lol. lmao. buffed a bit the kammerer sawn off since it did the same damage as the molly but with worse stats everywhere, nerfed the lever shotgun from doing 80/60 damage (same or better than AMR) to being simply very good (a mix between glad and regulator, analogous to the mateba). the gladstone is a proper ap shotgun (same damage as before, ap that works), while the regulator is a cannon vs low armor but is not better than the gladstone vs high armor now. bull is the same old gun, the boj doesnt suck as badly as before (it was a full auto gladstone, but both were way worse than the regulator and lever). buckshot got a bit of a buff, though it has 6 pellets from 8, so less shrap memes.
- the lmgs: same as before. IH lmg got better recoil (dont say I hate IH and nerf it) compared to the others, the maxim HMG no longer has lower damage than the other lighter machineguns.
- bolt actions: same as before, would you look at that. except tosshin doesnt have the meme super high AP it had before to be an ap bolt action, but it works like intended and the other .20 guns: lower "flesh" damage compared to the .30 bolt actions, but way better vs armor, and recoil too since .20. incel mosin bolt got the better recoil vs the vendor, non-antag novak.
- misc: motherfucker same as before, the lever isnt a better mateba (and now the sawn off has worse stats instead of being the same) but its still a very strong gun. the amr doesnt lose close to 20 damage vs armor but only 10, and instead of tanking vs low armor it does better. all makeshift guns are slightly worse than the baseline guns of each category, except the makeshift pistol (which nothing can make good except it not being single shot with meme jam chance) and the shotgun which humon explicitly wanted to be a "fuckoff gun".

Additional notes: internal balance was kept the same, so the powerful (antag) guns are still the most powerful guns but some got tweaks as mentioned, and may not be as over or underpowered as they were before (namely the makarov, c20r and lever guns). Arguably the areas that could use more tweaking are the SMGs. Pistols/revs are the same as before and I am very happy with the rifles and shotguns (of which only kammerer sawn off, regulator and lever action were changed from their previous stats).

Energy is coming next, but first we want to test these out.

hyperioo (approved):
I skimmed through it. Most of the line changes are numerical values changes. The bugs in armor code itself should have been ironed out with the several test server rounds.

PR #6936, opened 2022-01-06
Cyborgs can no longer be hard-stunned by flashes. Security cyborgs are immune to flashes. Adds framework for modular traits.

MLGTASTICa:
## About The Pull Request
Cyborgs no longer get hard-stunned by flashes
Sec cyborgs are now immune to flashes
Adds in framework for robot_Traits

## Why It's Good For The Game
Stuns are cringe , even so on extremly slow robots. You should bother with the effort for an ion gun atleast if you wish for a insta-win
## Changelog
:cl:
balance: Cyborgs no longer get hard-stunned by flashes , instead they get a speed debuff and the flash overlay.
balance: Security cyborgs are now flash immune
code: Added framework for module-based and component-based  robot traits , which provide functionality
code: "Floor Cleaning" trait added for the janitor cyborg module , "Flash resistant" added for security , "Deflective armor" trait added for admin fun.
/:cl:

PR #7360, opened 2022-08-09
Weakens no longer paralyse robots

MLGTASTICa:
## About The Pull Request
Weakening a robot no longer paralyses it , instead it gives them a high slowdown and forces them to drop equipment.

## Why It's Good For The Game
Having a flash is the cheapest instant-win against any robots , with a flash , a person can just keep using it on a robot and beat them to death with a crowbar, the same is not available for humans , which can reduce its effects by wearing sunglasses , and don't get paralyzed for 12 seconds everytime they get flashed.
This makes it so flashes are no longer a instant win and the death of all robots, but still a viable weapon for self-defense and escape if necesarry.
## Changelog
:cl:
balance: Stuns / Flashes no longer paralyse robots , instead it forces them to drop all items and gives them a major amount of slowdown.
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/life.dm
+++ b/code/modules/mob/living/silicon/robot/life.dm
@@ -93,10 +93,10 @@
 
 
 	if (src.stat != 2) //Alive.
-		if (src.paralysis || src.stunned || !src.has_power || src.weakened) //Stunned etc.
+		if (src.weakened > 0)
+			AdjustWeakened(-1)
+		if (src.paralysis || src.stunned || !src.has_power) //Stunned etc.
 			src.stat = 1
-			if (src.weakened > 0)
-				AdjustWeakened(-1)
 			if (src.stunned > 0)
 				AdjustStunned(-1)
 			if (src.paralysis > 0)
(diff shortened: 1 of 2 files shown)


Firefox13:
is this just #6936 but without the security borg part?

MLGTASTICa:
> is this just #6936 but without the security borg part?

yes

assortedbeads:
> > is this just #6936 but without the security borg part?
> 
> yes, and if you dare use the same argument on the old like "its impossible to emag robots without a flash" , contractors can buy emps for cheap, and they stun for 20 seconds on hit . Emagging a robot also gives the traitor full AA since robots can give acces on ID modification program.

A contractor can just buy a hacking tool, or an agent ID and pick off a straggler, or fly to the caps window and scan their id, or find a wares disk and give themselves AA.

There are easier ways to get access to places that dont involve hoping someone plays a silicon.

glaciusinfinity:
The one good SPCR Pr. Seriously you have no clue how infuriating it is to get flashed as a synthetic. All it takes for you to get killed is ONE CLICK. You can't run, you can't tell anyone what's going on, you can't even escape once it's started. There is no counterplay other than being the first one to make a move, which given how laws work already puts you on the back foot.  Even soporific, the anti-fun chem needs a way to get passed armor if injected and even then it's not instantaneous like flashes are. I don't care if it makes borgs harder to hack, since, in my experience, hacking tends to be a minority of cases instead of just crowbarring them to death. Flashes on synths should have been nerfed long ago, I am glad its getting addressed now.

pexel5:
SO F SALTY PR

Firefox13:
> yes, and if you dare use the same argument on the old like "its impossible to emag robots without a flash"

I'm afraid that argument is still valid as well other mentioned in feedback of other PR

PR #7364, opened 2022-08-10
Reduces the fire delay for most pistols and revolvers , and a set of energy guns.

MLGTASTICa:
## About The Pull Request
Brings guns firing speed on par with full autos.

## Why It's Good For The Game
Most guns are far below full autos ,this brings those more niche guns up to power.
Where the old fire delay is not mentioned , it used to be 6.
## Changelog
:cl:
balance: Energy crossbow bow now has 3 MS delay
balance: Decloner gun has 4 MS delay
balance:  Spider rose has 3 MS delay
balance: Counselor has 4 MS delay
balance: Svallin has 4 MS delay
balance : Shellshock fire delay is now 8 MS (From 12)
balance: Valkrie fire delay is now 12 MS (from 35)
balance: Kovacs fire delay is now  4 (from 6.5)
balance: Moebius plasma gun now has 5 MS delay
balance: Boltguns now have 5 MS fire delay (From 8)
balance: Avasarala now has 4 MS fire delay
balance: Lamia now has 4 MS fire delay
balance: Giskard now has 3 MS fire delay
balance:Mandella now has 4 MS fire delay
balance: Paco has 5 MS fire delay
balance: Consul has 4 MS fire delay
balance: Deckard has 3 MS fire delay
balance :Havelock has 3 MS fire delay
balance: Mateba now has 4 MS fire delay
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy/sniperrifle.dm
+++ b/code/modules/projectiles/guns/energy/sniperrifle.dm
@@ -10,7 +10,7 @@
 	projectile_type = /obj/item/projectile/beam/sniper
 	slot_flags = SLOT_BACK
 	charge_cost = 300
-	fire_delay = 35
+	fire_delay = 12
 	force = 10
 	w_class = ITEM_SIZE_BULKY
 	matter = list(MATERIAL_PLASTEEL = 20, MATERIAL_PLASTIC = 8, MATERIAL_SILVER = 9, MATERIAL_URANIUM = 6)
--- a/code/modules/projectiles/guns/energy/crossbow.dm
+++ b/code/modules/projectiles/guns/energy/crossbow.dm
@@ -13,6 +13,7 @@
 	fire_sound = 'sound/weapons/Genhit.ogg'
 	projectile_type = /obj/item/projectile/energy/bolt
 	self_recharge = 1
+	fire_delay = 3
 	charge_meter = 0
 	charge_cost = 200
 	price_tag = 2500
--- a/code/modules/projectiles/guns/energy/decloner.dm
+++ b/code/modules/projectiles/guns/energy/decloner.dm
@@ -8,6 +8,7 @@
 	origin_tech = list(TECH_COMBAT = 5, TECH_MATERIAL = 4, TECH_POWER = 3)
 	can_dual = TRUE
 	projectile_type = /obj/item/projectile/energy/declone
+	fire_delay = 4
 	charge_cost = 100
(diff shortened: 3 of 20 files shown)


code/modules/projectiles/guns/energy/sniperrifle.dm
```diff
@@ -10,7 +10,7 @@
 	projectile_type = /obj/item/projectile/beam/sniper
 	slot_flags = SLOT_BACK
 	charge_cost = 300
-	fire_delay = 35
+	fire_delay = 12
```
TheShown911, line 13: I think valk should keep it's very high fire delay.

MLGTASTICa: Valk sucks as it is.  Don't see why it shouldn't be more on par with the lightfall.

ValoTheValo:
man paco gets a fire delay reduction and MK58 dont. sad

code/modules/projectiles/guns/energy/sniperrifle.dm
assortedbeads: > Valk sucks as it is.  Don't see why it shouldn't be more on par with the lightfall.

Because it is a sniper weapon and has high damage

MLGTASTICa: Ah yes , the highly lethal energy rifle that has a maximum potential of 250 damage that is not enough to kill anyone should remain as it is , while an AK .30 with a lower rarity can easily kill a person  with a mag dump.

ValoTheValo: > Ah yes , the highly lethal energy rifle that has a maximum potential of 250 damage that is not enough to kill anyone should remain as it is , while an AK .30 with a lower rarity can easily kill a person with a mag dump.

valk sucks but it doesnt need to become a machinegun

Firefox13: I'll agree with TheShown and Valo, it has high damage and doesn't need fire_delay decrease

maikilangiolo: > man paco gets a fire delay reduction and MK58 dont. sad

if this pr gets closed because SPCR is too stubborn to change, i will just re-do it but including the mk58, and any other gun you think might deserve it

Humonitarian:
If the energy crossbow bow is the new one implemented recently, don't make it shoot faster. The delay is intended to act as the "windup time"

assortedbeads:
> If the energy crossbow bow is the new one implemented recently, don't make it shoot faster. The delay is intended to act as the "windup time"

It's the nemesis, the self charging one that does like toxin from what I can see

PR #7347, opened 2022-08-06
Map Oversight Fixes

maikilangiolo:
## About The Pull Request
A hotfix so hot it will headgib you. Fixes some mapping errors on my part with the new club layout, but also addresses ancient map reports years old. Yes, incinerator room has shield emitters now, the SM eject button works (beware of AI) and last shelter showcase button retrieved from shadowrealm.

## Why It's Good For The Game
Old bugs bad, new bugs good. Feel free to @ me if you want to requests changes

## Changelog
:cl:
add: Added beer keg back to club
add: added more scrubbers and vents to club
add: added more shield diffuers to junk beacon
add: added shield diffusers to incinerator room
fix: fixed conference buttons with wrong id
fix: fixed SM eject button not working
fix: fixed bad and incomplete piping in club
fix: fixed that one normal window in courtroom
fix: fixed double wall and missing hull plating
fix: probably more I forgot
/:cl:

PR #7366, opened 2022-08-12
Fixing old oversights

maikilangiolo:
## About The Pull Request
Does what #7347 does but I think I fixed the merge conflict???? Firefox said to try again, hopefully I cheated the system because I cannot download a fresh map somehow, without having to download the entire repo again, again. Again.

## Why It's Good For The Game
Half the changes you cannot see since they are ID changes, or chairs that are invisible in the mapping programs.

image
image
image
image
image

## Changelog
:cl:
add: Added beer keg back to club
add: added more scrubbers and vents to club
add: added more shield diffuers to junk beacon
add: added shield diffusers to incinerator room
fix: fixed conference buttons with wrong id
fix: fixed SM eject button not working
fix: fixed bad and incomplete piping in club
fix: fixed that one normal window in courtroom
fix: fixed double wall and missing hull plating
fix: probably more I forgot
/:cl:


Humonitarian:
Tbh leave beer keg in the bar, so in case club is absent they don't need to hack the vendor or the doors to access alcohol.

Club players will be able to just move it to the back, and beer isn't a very effective alcohol by volume anyways.

PR #7371, opened 2022-08-12
Fixes Name Error in extended mags, and change to .20 extended

maikilangiolo:
## About The Pull Request

Changes the old name of makseshift ak when it ws caleld kalash to "Sermak", which is the current name, in the descriptions of .20 and .30 drum/ext mags. 
Also makes all .20 rifles take the .20 ext, since its only 5 more rounds vs the more powerful .30 (also good for when gun rework is ready)

## Why It's Good For The Game

Fixing name mistake good, and .20s are very weak right now, and when they will get changed, they will be the lower damage guns so 35 round mags you can only get by crafting is ok.

## Changelog
:cl:
balance: z8 and kovacs can now take .20 extended
fix: fixed wrong makeshift AK name in extended and drum mags descs
/:cl:


Firefox13:
> Also makes all .20 rifles take the .20 ext, since its only 5 more rounds vs the more powerful .30 (also good for when gun rework is ready)

Are you sure it won't make all other rifles invisible?

maikilangiolo:
> Are you sure it won't make all other rifles invisible?

uuuuuuuuuuuuuuhhhh. Consider, the sermak is already invisible.

TheShown911 (approved):
I hope this is tested and works, can't check icon diffs.

maikilangiolo:
> I hope this is tested and works, can't check icon diffs.

even though it takes like 20 minutes to compile and test, I always check. wintermute and vintorez were fine already, only the makeshift ak, z8 and kovacs didnt have the sprites. I tested all five and no invisible sprites.

PR #7337, opened 2022-08-04
Yet another Major Gun Rebalance ( Fixes several issues with gun code,  full auto is now based off fire_delay)

MLGTASTICa:
## About The Pull Request
Burst fire can no longer be stacked
Full auto's fire delay is now based on the shown value , and not on the guns burst_delay
Guns  are no longer hard-locked to the 4 second delay between attacks.The fire_delay is now properly applied
Bull no longer has 0 fire_delay (since you can insta-fire all shots with the new fix)
SEMI_AUTO_NODELAY changed to SEMI_AUTO_300 since it now has the same firerate as a 300 shot auto would.

Related video : https://cdn.discordapp.com/attachments/299262993617780737/1004766001642086540/2022-08-04_18-00-08.mp4

## Why It's Good For The Game
Any gun that would get lower than 4 MS fire delay would have its effects not applied , this made most pistols underwhelming even with upgrades meant to make them better.
Makes Semi-Auto properly apply its delay if its firerate is under 4 MS.

## Changelog
:cl:
fix: Burst fire can no longer stack
fix: Full auto fire_delay is now based on the shown value , no longer on the gun's burst_delay
fix: Guns are no longer hard locked to a 4 MS fire_delay.
balance: Bull shotgun now has a fire_delay of 2 MS.
balance: SEMI_AUTO_NODELAY has been changed to SEMI_AUTO_300, with a firerate delay of 2 MS
/:cl:

PR #7374, opened 2022-08-13
Hotfix 1ms fullauto

Firefox13:
## About The Pull Request

Hotfixes tesla chip until (SPCR wakes up) we find a better solution

## Why It's Good For The Game

Removes this:

https://user-images.githubusercontent.com/57810301/184504466-16d85d34-9486-47c3-9a78-68f3ce6d4b54.mp4

## Changelog
:cl:
fix: removed death rays (fullauto from tesla chip is gone because it allows for 1ms killing machine)
/:cl:


Firefox13:
probably caused by #7337 
@MLGTASTICa

hyperioo (approved):
Now we just have to pray the whole server does not blow up.

Humonitarian:
The mod itself is buggy and we should limit it to only work on specific laser guns with the right tag.

PR #7380, opened 2022-08-15
Reduces Junk Tractor Cooldown

maikilangiolo:
## About The Pull Request

Simply changes the tractor cooldown after you release a junk field from 5 -> 3 minutes

## Why It's Good For The Game

The days of the old junk beacon creating a 5x5 grid of trash piles at your feet are gone. Now you have to navigate a new map with lots of swathes of zero G ground, no atmosphere and dangerous fauna. A reduction to 3 minutes isn't much and it wont buff junking, just help reset the field faster if someone looted it already.
It might be worth to add breathable atmosphere to the beacon so that roaches dont suffocate and die. Pressure is livable, and roaches and spiders can be added as extra danger outside carps and bears simple mobs. Extra traps such as the new mines and old bear traps should be considered as well.

## Changelog
:cl:
tweak: tractor cool down lowered from 5 to 3 minutes
/:cl:

### DIFF
--- a/code/modules/dungeons/junk_tractor_beam.dm
+++ b/code/modules/dungeons/junk_tractor_beam.dm
@@ -99,7 +99,7 @@
 	var/list/jf_pool = list()  // Pool of junk fields you can choose from
 
 	var/beam_cooldown_start = 0  // Starting time of the cooldown for the progress bar
-	var/beam_cooldown_time = 5 MINUTES
+	var/beam_cooldown_time = 3 MINUTES
 	var/beam_capture_time = 20 SECONDS
 
 	var/list/preloaded_25_25 = list()  // Need to preload maps in SOUTH direction


MLGTASTICa:
I believe this is a bad change because it makes it easier to just keep rerolling for  rare junkfields. Also why should it be faster ? it should be slower instead,  you have a 100x100 map full of junk and even randomly generated parts of eris with insuls / etc.
Id be surprised if someone manages to junk a entire field and still wants to spend 5 minutes waiting there instead of storing his loot.

TheShown911 (commented):
Controversial.

hyperioo (commented):
Yeah, I don't really see the point either, 2 minutes will not change much anyway.

And the 5 minutes cooldown was also there to avoid choking (too much) the garbage collector.

Firefox13:
It's both a balance (go do something else instead of junking) and performance (lol garbage controller) issue and the latter is a really serious concern
Junk field has enough stuff to junk for atleast five minutes, if you're that lazy

PR #7381, opened 2022-08-15
[Should be testmerged]Updates NanoUI JQuery and JQuery UI libraries

MLGTASTICa:
## About The Pull Request
Updates JQuery from v 1.1.0 to v 3.6.0
Updates JQuery UI from v 1.11 to 1.13.2 
Adds special handling for get_html to prevent pushing updates to non-existant HTML pages (might be cause of resulting white screens)

## Why It's Good For The Game
Likely fixes the issue with the white screen (which is porbalby caused by a undefined receive updata data) or which is caused by using methods not made for Microsoft edge

## Changelog
:cl:
code: Updates NanoUI related libraries to their newest counterparts
code: Added special handling to prevent pushing HTML updates to non-existant HTML pages.
/:cl:


MLGTASTICa:
Byond uses Mozilla Firefox with a maximum IE explorer version of 7. There is no way to update it without using external DLL's or server-side rendering / other frameworks that would require too much work.

PR #7383, opened 2022-08-16
Adds robotic sprites for brain , heart , liver and lungs. Makes robotic remove item use bolt turning quality. 

MLGTASTICa:
## About The Pull Request
Adds robotic and assisted  sprites for brain, heart , liver , kidneys , lungs.
Makes robotic organs name start with "Robotic"
Makes assisted organs start with "assisted"
Robotic remove item (for organs and shrapnel) now uses bolt turning.
Robotic people don't get an appendix.
## Why It's Good For The Game
Using clamping on robotic organs never made any sense.
Robo sprites help differentiate the more vital organs
Robotic pre-name just so people have a damn way to tell apart organic from robotic.

## Changelog
:cl:
add: Added new robotic  and assisted sprites for organs
balance: Robotic remove item now uses bolt turning instead of clamping
/:cl:


assortedbeads:
Since you're on organs already can you fix appendixes? Fbps spawn with flesh ones

code/modules/organs/body_modifications.dm
```diff
@@ -234,9 +238,15 @@ var/global/list/modifications_types = list(
 /datum/body_modification/organ/robotize_organ/create_organ(var/mob/living/carbon/holder, O, color)
 	var/obj/item/organ/I = ..(holder,O,color)
 	I.nature = MODIFICATION_SILICON
+	if(istype(I, /obj/item/organ/internal/appendix))
```
TheShown911, line 241: These appendix checks feel very wrong.

MLGTASTICa: its litteraly by the book.

GrayRachnid:
The posibrain animations aren't playing in the icon check, can you test and confirm they work?

GrayRachnid:
Also, update the base brain sprite to the new one please.

code/modules/organs/body_modifications.dm
hyperioo: It's just that there is no assisted appendix sprite right? Don't you want to just not set `I.icon_state = "[I.icon_state]_assisted"` but still applying the other lines so that it's named assisted appendix and have the right stats?

MLGTASTICa: k

hyperioo:
For posterity, when creating a robotic appendix, it returns null so the organ is not actually spawned inside the person. This happens when creating a robotic groin (which has no reason to have a robotic appendix inside since it has no purpose)

PR #7384, opened 2022-08-16
Fixes smoke related problems (blattedin farm fix)

Alliostra:
## About The Pull Request
Due to balance concerns, in https://github.com/discordia-space/CEV-Eris/pull/7087 I made chemsmoke lose some of its reagents when it transferred them to a mob, in order to prevent duping. This caused the visible smoke to lose its reagents and color due to the reagents being transferred to mobs beforehand. This is why blattedin farms no longer work properly.

By not removing reagents from the smoke when it is transferred to a mob, this problem is solved. And no, this is not a balance problem, SPCR . That's because it's handled the same way with reagents being transferred to containers or splashed. In fact, the blattedin farms are technically dupes as well.

## Why It's Good For The Game

Blattedin farms work again, and smoke now has the correct color again. People complained about blattedin farms not working anymore

## Changelog
:cl:
tweak: makes smoke not lose reagents anymore when some are transferred to a mob.
/:cl:

### DIFF
--- a/code/game/objects/effects/chem/chemsmoke.dm
+++ b/code/game/objects/effects/chem/chemsmoke.dm
@@ -169,8 +169,8 @@
 						if(H.wear_mask)
 							gasmask = H.wear_mask.item_flags & BLOCK_GAS_SMOKE_EFFECT			
 						if(!internals && !gasmask)
-							chemholder.reagents.trans_to_mob(H, 5, CHEM_INGEST, copy = FALSE)
-							chemholder.reagents.trans_to_mob(H, 5, CHEM_BLOOD, copy = FALSE)
+							chemholder.reagents.trans_to_mob(H, 5, CHEM_INGEST, copy = TRUE)
+							chemholder.reagents.trans_to_mob(H, 5, CHEM_BLOOD, copy = TRUE)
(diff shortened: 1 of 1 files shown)


pexel5:
noice

MLGTASTICa:
Still a balance and even a exploit concern. Anything that makes a smoke cloud will dupe if you set copy to true. The proper fix would be to have smoke apply its effects on each tile with a set amount.

maikilangiolo:
> Still a balance and even a exploit concern. Anything that makes a smoke cloud will dupe if you set copy to true. The proper fix would be to have smoke apply its effects on each tile with a set amount. Unlike container transfer and dumping , dumping always results in less amount and never places the reagent in multiple/any beakers on the target tile. Transfer is preety straight foward , amount in , amount out.

unknown

TheShown911 (approved):
I have never seen someone use smoke for duping reagents, this fix is fine in my opinion.
(inb4 someone uses it just after this PR to prove a point)

hyperioo:
> (inb4 someone uses it just after this PR to prove a point)
I would expect no less

MLGTASTICa:
Jesee , we have nitroglycern to dupe.

PR #7226, opened 2022-05-07
Refactor ammo kits and crafting station

SirRichardFrancis:
## About The Pull Request

What it says on the tin. Most bugs and RNGs have been removed in the process.
Goodbye, sheets of -120 plasteel, you won't be missed.

Ammo kits now able to craft most magazine types.
Magazines spawn empty, with the exception of speed loaders.

Ammo kit initially get 15 "material points", that player could use to pick what they wanna craft.
MEC stat increases "material points", up to plus 15 at skill level 80.
Gunsmith perk gives additional 15.

Crafting station uses ammo kit system to craft ammunition.
Crafting station's components contribute to "material points", up to 10 extra points.
Crafting station's initial storage capacity halved, but benefits more from better components.
Also crafting station's animations isn't derpy anymore.

<details>
<summary>
"Price list"
</summary>

.20 ammo pile (10 ammo, 3 points)
.20 standard magazine (empty, 25 ammo, 5 points)
.20 extended magazine (empty, 35 ammo, 10 points)
.20 drum magazine (empty, 60 ammo, 15 points)
.20 ammo box (50 ammo, 15 points)

.25 ammo pile (10 ammo, 3 points)
.25 standard magazine (empty, 30 ammo, 5 points)
.25 ammo box (60 ammo, 15 points)

.30 ammo pile (10 ammo, 3 points)
.30 standard magazine (empty, 30 ammo, 5 points)
.30 ammo box (60 ammo, 15 points)

.35 ammo pile (15 ammo, 5 points)
.35 speedloader (6 ammo, 3 points)
.35 standard magazine (empty, 10 ammo, 5 points)
.35 extended magazine (empty, 16 ammo, 10 points)
.35 SMG magazine (empty, 35 ammo, 12 points)
.35 ammo box (70 ammo, 15 points)

.40 ammo pile (6 ammo, 3 points)
.40 speedloader (6 ammo, 5 points)
.40 standard magazine (empty, 10 ammo, 5 points)
.40 SMG magazine (empty, 25 ammo, 12 points)
.40 ammo box (30 ammo, 15 points)

.50 slug pile (5 ammo, 3 points)
.50 pellet pile (5 ammo, 3 points)
.50 beanbag pile (5 ammo, 3 points)
.50 slug box (30 ammo, 15 points)
.50 pellet box (30 ammo, 15 points)
.50 beanbag box (30 ammo, 15 points)

.60 ammo pile (5 ammo, 5 points)
.60 ammo box (30 ammo, 30 points)

[...]

PR #7401, opened 2022-08-19
Tweaks to Crafting Station Ammo crafting (and .40 ammo storage)

maikilangiolo:
## About The Pull Request
Kirov's #7226 changed the crafting station and ammo gun crafting for the better. This PR builds upon it to improve QoL and do some tweaks for materials costs.
1. Added .20 and .30 speedloaders to the crafting station, for bolt actions
2. Changed magazine prices according to Humon's suggestion. Ammo packets and .20 drum prices unchanged
3. reduced ridiculous cost of armor parts. really, 20 steel, plastic, wood and cardboard? for a single part? do know that these prices were balanced when armor parts were considered rare and you could craft multiple armor parts per 20/stack. since you can buy those and they arent as rare, the price was lowered to just 20 steel and plastic
4. .40 ammo packets were changed from 30 -> 50 round capacity. not long ago .35 packets were increased to 70 so you could reload two .35 smg mags. rifle packets also hold enough to reload two mags, so .40 got changed to 50 since .40 smg mags hol 25 rounds (like .20)
5. .40 ammo piles hold 6 -> 10 rounds. since the packets are not 50 and not 30, they are no longer divisible by 6 and if the bigger rifle bullets can be in 10 round stacks, so can the smaller .40. adjusted the HV packet cost in the uplink to reflect the change
6. changed the recycle value of ammo packets. in case you didnt know, you could multiply steel by crafting rifle ammo packets which recycled for 20 steel (despite costing 10 to make), now all ammo packets recycle for 10 steel to avoid this issue
7. ammo crafting carboard cost for the crafting station reduced 5 -> 1. Packets recycled for 10/20 steel and 1 cardboard but costed 5 carboards to make. now that steel multiply was fixed, and you cant multiply packets like before the refactor, I felt like it was a decent cost.
8. fixed kirov putting the .25 caseless pistol magazine in the .20 section (and calling it .20 pistol mag)
9. lowered the .60 box cost from 30 -> 20, because it was near impossible to craft beforehand since the crafting station has around 20 points, and the other boxes/packets prices were more efficient than printing singular bullet piles. It is also scrap ammo for a very rare gun

## Why It's Good For The Game
Crafting station change was based, but still unpolished using old balance when you could make 2+ items (sometimes even 4+) per each crafting instance. arbitraty magazine costs were changed to be more consistent as Humon proposed.

## Changelog
:cl:
add: Added .20 and .30 scrap stripper clips to crafing station
tweak: tweaked .40 ammo pile and ammo packet size
balance: rebalanced armor part and ammo crafting cost
balance: rebalanced magazine crafting costs
tweak: changed ammo packet recycle cost to avoid steel multiplication
tweak: fixed wrong .25 caseless pistol mag category and name in crafting station
/:cl:


maikilangiolo:
yeah shaz updated modular speed loaders but didnt fix the wrong .20 scrap name (it had _f in the name instead of _s), causing the .20 scrap stripper clip to have invisible bullets

PR #7404, opened 2022-08-19
Brings .35 and .40 Speedloaders in line with the stripper clips

maikilangiolo:
## About The Pull Request
Revolver speedloaders are now tiny, instead of small

## Why It's Good For The Game
1. The rifle speedloaders (stripper clips) were tiny but the rev ones didnt get updated
2. a bullet pile of 10-15 rounds is tiny, while a speedloader of 6 rounds is small. 
3. it will make speedloaders more worth using, instead of having 30 rounds in two bullet piles for the same size a single speedloaders.

## Changelog
:cl:
tweak: speedloaders are now tiny sized from small
/:cl:


maikilangiolo:
#7408 does what this does but better

PR #7408, opened 2022-08-19
Smaller Dynamic speedloaders

shazbot194:
## About The Pull Request

This PR does two things, it makes speedloaders smaller (tiny) and makes their sprites dynamic with what they have loaded. 

It also appears to be an alternative to #7404

## Why It's Good For The Game

Currently, speedloaders are in a weird spot being just a tiny bit faster than a pile of loose bullets while also being twice the size for less than half the ammo. The size change should make them a more competitive choice. As well, it brings revolver speedloaders in line with rifle speedloaders (stripper clips)

image
In addition, this PR also updates the sprites to be both more in line with the current bullet sprites while also allowing people to see what exactly is loaded in the speedloaders, even if it's a rainbow of bullets.

## Changelog
:cl: Shazbot194
add: Dynamic sprites for speedloaders
tweak: Speedloaders are now Tiny
/:cl:


Firefox13 (approved):
begone broken speedloaders

PR #7413, opened 2022-08-20
stops disposal from accessing null

LetterN:
## About The Pull Request
Fixes trunks accessing bins that is never linked

fixes the issue at https://github.com/discordia-space/CEV-Eris/pull/7406


hyperioo:
Thanks for the fix

PR #7414, opened 2022-08-20
Fixes inactivity proper

MLGTASTICa:
## About The Pull Request
The asset update removed the duration check from the is_afk very , causing it to kick everyone since we had no config for kick_time
I made a shitty fix because i was super tired(and had no idea inactivity was used for something else)
This properly fixes the problem

## Changelog
:cl:
fix: Fixed inactivity properly
/:cl:


hyperioo:
Thanks for the fix

PR #7437, opened 2022-08-24
Grab speed now also depends on the target's robustness. Flip_in_hand no longer forces humans to lie down.

MLGTASTICa:
## About The Pull Request
Grab speed now takes into account the affected's robustness, increasing its speed.
Flip_in_hand no longer forcibly rests people.
## Why It's Good For The Game
Instant-grabs are bad for the game , and flip-in-hand was just obscure abuse.
## Changelog
:cl:
balance: Grab speed also takes the affected's robustnesss into account now
balance: flip-in-hand no longer forces people to rest
/:cl:

### DIFF
--- a/code/modules/mob/mob_grab.dm
+++ b/code/modules/mob/mob_grab.dm
@@ -274,9 +274,9 @@
 		// Negative ROB is a flat warmup increase
 		warmup_increase = assailant_stat
 	if(affecting_stat > 0)
-		warmup_increase = affecting_stat ** 0.8
+		warmup_increase += affecting_stat ** 0.8
 	else
-		warmup_increase = -(affecting_stat ** 0.6)
+		warmup_increase += -(affecting_stat ** 0.6)
 
 	var/total_warmup = max(0, UPGRADE_WARMUP + round(warmup_increase))
 
--- a/code/modules/mob/mob_grab.dm
+++ b/code/modules/mob/mob_grab.dm
@@ -276,7 +276,7 @@
 	if(affecting_stat > 0)
 		warmup_increase += affecting_stat ** 0.8
 	else
-		warmup_increase += -(affecting_stat ** 0.6)
+		warmup_increase += affecting_stat ** 0.6
 
 	var/total_warmup = max(0, UPGRADE_WARMUP + round(warmup_increase))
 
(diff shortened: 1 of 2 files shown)


code/modules/mob/mob_grab.dm
```diff
@@ -265,13 +265,18 @@
...
 	if(assailant_stat > 0)
 		// Positive ROB decreases warmup, but not linearly
 		warmup_increase = -(assailant_stat ** 0.8)
 	else
 		// Negative ROB is a flat warmup increase
 		warmup_increase = assailant_stat
+	if(affecting_stat > 0)
+		warmup_increase = affecting_stat ** 0.8
+	else
+		warmup_increase -(affecting_stat ** 0.6)
```
TheShown911, line 279: What?

MLGTASTICa: bruh!

hyperioo (commented):
Don't you rather when to multiply/divide or add/substract to `warmup_increase`?

Here you are directly erasing the value that was put by the lines just before, linked to the assailant's stats.
```
	if(assailant_stat > 0)
		// Positive ROB decreases warmup, but not linearly
		warmup_increase = -(assailant_stat ** 0.8)
	else
		// Negative ROB is a flat warmup increase
		warmup_increase = assailant_stat
```

hyperioo:
Also, I just noticed it, shouldn't it be?
```
// Negative ROB is a flat warmup increase
warmup_increase = -assailant_stat
```
Because if the ROB is negative, using `warmup_increase = assailant_stat` just decreases the warmup time

MLGTASTICa:
> Also, I just noticed it, shouldn't it be?
> 
> ```
> // Negative ROB is a flat warmup increase
> warmup_increase = -assailant_stat
> ```
> 
> Because if the ROB is negative, using `warmup_increase = assailant_stat` just decreases the warmup time

i didn't code this , in the first place , i don't really know

hyperioo:
So now without the minus, it increases warmup both if `affecting_stat` is negative or positive.

hyperioo:
I'll ask Vode code about it

MLGTASTICa:
> So now without the minus, it increases warmup both if `affecting_stat` is negative or positive.

only if negative.

TheShown911 (approved):

PR #7441, opened 2022-08-25
Robots now climb ladders 50% faster

MLGTASTICa:
## About The Pull Request
Reduces the climb delay multiplier from 6 to 3

## Why It's Good For The Game
No reason to have it 6 times higher. The sound que is more than enough anyway.

## Changelog
:cl:
balance: Robots now climb ladders 50% faster than before.
/:cl:


MLGTASTICa:
> Make sure drones won't become too fast, they could use a buff too, but they were already fairly fast at it.

drones were never affected b the robot climb increase in the first place.

TheShown911 (approved):

PR #7446, opened 2022-08-26
Poncho Accessory Creation

assortedbeads:
## About The Pull Request
This PR makes a brown poncho accessory.
It's slightly larger than the suit version sprite wise, it costs one point in the loadout and it can be attached to/detached from your suit/rig/voidsuit.
It has no stats, it affects no interactions, it's purely a visual change.

Currently while it can be attached to all rigs, the lack of a standardized rig shape or back form means it looks off on 
- IH Rig (Backpack section covers the chest as well)
- Rescue Rig (Same as above)
- Combat rig (Bulky shoulders)
- Techno Exhault Rig (Already has a cape)
- Merc Rig (Spiky shoulders)
If you don't like how it looks, it's an accessory and you can just detach it.

Here are some pictures of it on the technomancer rig, which the rig I edited the sprite against.
Do note that the back of the poncho will be covered by whatever the rig shape is, since it acts like a backpack and backpacks show above everything.

Full1
Full2
Full3
Full4
Headless1
Headless2
Headless3
Headless4

## Why It's Good For The Game
It's more fashion options, and eventually when the sprites are updated cloaks will be added to the currently created poncho/cloak/cape entry.

## Changelog
:cl:
add: Added oversized poncho sprite
add: Added beginning of code for future cloaks and ponchos
/:cl:


Humonitarian:
What if it could be attached on the RIG backpack? That way it would show up over it

assortedbeads:
> What if it could be attached on the RIG backpack? That way it would show up over it

Gonna be a mild issue since the rigs act as storage as well, unless  you mean making it a module which is its own can of worms.

tristain01:
Why not make it so it's the same thing for normal clothes, you know put it over your clothing and jacket? Like an actual poncho

assortedbeads:
> Why not make it so it's the same thing for normal clothes, you know put it over your clothing and jacket? Like an actual poncho

Yea that's how it works
Can be attached to the uniform/jacket/rig/voidsuit. I should have worded the description better

hyperioo:
Would be nice to add a recipe to craft it out of hide or something like that, so that players can make it even if they did not choose it as an accessory.

Humonitarian:
> Would be nice to add a recipe to craft it out of hide or something like that, so that players can make it even if they did not choose it as an accessory.

We do not have hide implemented. Roach organ update is a decent ways off.

assortedbeads:
> Would be nice to add a recipe to craft it out of hide or something like that, so that players can make it even if they did not choose it as an accessory.

Made a recipe that requires makeshift/normal wirecutters and a bedsheet

assortedbeads:
@hyperioo I've updated the sprites after getting feedback

shazbot194:
The biggest issue I see is that they very well might clash with the other voidsuits while also being very busy.

It also looks quite flat on the top of the front and especially the back, looking like it has a stiff bar in it.

assortedbeads:
> The biggest issue I see is that they very well might clash with the other voidsuits while also being very busy.
> 
> It also looks quite flat on the top of the front and especially the back, looking like it has a stiff bar in it. 

For clashing issue I dont think it's possible to account for all voidsuits/armors, and if it looks bad the player can just detach it.

hyperioo (approved):

PR #7473, opened 2022-09-04
Unlocks fullauto limits, Molly burstfire mode

Humonitarian:
## About The Pull Request

Gives molly a 3- and 5-shot rapid burstfire mode instead of fullauto, to differentiate it from the Straylight.

Unlocks full-auto firerate limit, allowing guns to actually vary in firing speed, but reduces the general firerate of all fullauto modes. In effect, this will slow down weapons with very low firerate (such as .40 SMGs), but turn high firerate guns into spray simulator.

Adjusts firerate for tick_lag.

Removes damage debuff from firemodes, significantly buffing all fully automatic and burst weapons. This last change is experimental, and will likely unbalance guns.

Fixes UI related bugs.

## Why It's Good For The Game

Firerate works better, Molly is more unique.

## Changelog
:cl:
balance: gun firerate now more accurate
balance: Molly now has a burst-mode instead of FA
fix: bugs related to firerate and gun UI
/:cl:


tristain01:
I look forward to this with great interest

code/datums/datum_click_handlers.dm
```diff
@@ -102,8 +103,10 @@
 	object = resolve_world_target(object)
 	if(object)
 		target = object
+		time_since_last_shot = world.time
+		to_chat(owner.mob, "[world.tick_lag], [world.time]")
```
hyperioo, line 107: ._.

code/modules/projectiles/gun.dm
```diff
@@ -896,8 +896,8 @@
 				"desc" = F.desc,
 				"burst" = F.settings["burst"],
 				"minimum_fire_delay" = GUN_MINIMUM_FIRETIME,
-				"fire_delay" = F.settings["fire_delay"],
-				"move_delay" = F.settings["move_delay"],
+				"fire_delay" = F.settings["fire_delay"] * 5,
+				"move_delay" = F.settings["move_delay"] * 5,
```
hyperioo, line 900: Why the `* 5` ?

hyperioo (changes requested):
Small conflict that you can fix with the web interface

code/modules/projectiles/gun.dm
Humonitarian: It calls ticks ms, but ticks are not real, so I used fortune telling to determine how many ms one tick will be during launch.

hyperioo (approved):

PR #7498, opened 2022-09-12
Make combat less lethal

ErisMeleeEnjoyer:
## About The Pull Request
increases DROPLIMB_THRESHOLD_EDGE from 0.75 to 1.15

increases max_health from 75 to 90

lowers the droplimb proc's chance to dismember with a blunt weapon from brute/2 to brute/3

## Why It's Good For The Game

less funny funky rng instakills, call it a salt pr, but if I could I'd delete that proc's random chance to dismember completely and nothing of value would be lost

## Changelog
:cl:
balance: made combat less lethal (salt pr)
/:cl:

### DIFF
--- a/code/modules/organs/external/_external.dm
+++ b/code/modules/organs/external/_external.dm
@@ -3,7 +3,7 @@
 ****************************************************/
 
 //These control the damage thresholds for the various ways of removing limbs
-#define DROPLIMB_THRESHOLD_EDGE 0.75
+#define DROPLIMB_THRESHOLD_EDGE 1.15
 #define DROPLIMB_THRESHOLD_TEAROFF 1.35
 #define DROPLIMB_THRESHOLD_DESTROY 1.5
 
--- a/code/modules/organs/organ_description.dm
+++ b/code/modules/organs/organ_description.dm
@@ -79,7 +79,7 @@
 	default_type = /obj/item/organ/external/head
 	default_bone_type = /obj/item/organ/internal/bone/head
 
-	max_damage = 75
+	max_damage = 90
 	min_broken_damage = 60
 	vital = TRUE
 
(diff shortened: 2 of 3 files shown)


assortedbeads:
Based

code/modules/organs/external/_external.dm
```diff
@@ -3,7 +3,7 @@
 ****************************************************/
 
 //These control the damage thresholds for the various ways of removing limbs
-#define DROPLIMB_THRESHOLD_EDGE 0.75
+#define DROPLIMB_THRESHOLD_EDGE 1.15
```
Humonitarian, line 6: As only melee weapons use this, I'd suggest keeping it at 1, so even with little damage they can still remove arms and the likes.

code/modules/organs/organ_description.dm
```diff
@@ -79,7 +79,7 @@
 	default_type = /obj/item/organ/external/head
 	default_bone_type = /obj/item/organ/internal/bone/head
 
-	max_damage = 75
+	max_damage = 90
```
Humonitarian, line 82: This will be very good for the economy

Humonitarian: For numbers, this will mean a bullet will require to reach
121 damage for a chance to tear off a head, and
135 damage for a chance to gib a head.

As opposed to the old numbers of
101 damage to tear off the head and
112 damage to gib the head.

Since the chance to gib is also reduced, this will make shooting the head a whole lot less lethal, and when you do shoot enough, it is more likely for the head to be torn off first before gib, making resusciation easier.

code/modules/organs/external/_external.dm
ErisMeleeEnjoyer: >little damage
>remove arms
guh?? nuh

tristain01:
Based

Alliostra:
salt PR

assortedbeads:
Now buff FBPs so they take less pain

hyperioo (approved):

PR #7499, opened 2022-09-12
Penetration improvements and bugfixes [ready]

Humonitarian:
## About The Pull Request

Shields and walls can now be penetrated, and will decrease the damage of projectiles.
Mobs now properly dodge when shot at based on size.
Also cleans up some code and fixes how agony is calculated.

## Why It's Good For The Game

Bugfix.

## Changelog
:cl:
fix: fixed penetrating objects
fix: small creatures can dodge
fix: various agony- and damage calculations
/:cl:


code/modules/mob/living/living_defense.dm
```diff
@@ -132,7 +146,7 @@
 		return
 	shake_animation(damage)
 
-
+ // return PROJECTILE_CONTINUE if bullet should continue flying
```
hyperioo, line 149: Where is that return PROJECTILE_CONTINUE  in `bullet_act`? I'm only seeing some return TRUE

SirRichardFrancis: image
image

Those are only used situationally.

hyperioo (approved):

PR #7505, opened 2022-09-13
More melee additions

ErisMeleeEnjoyer:
## About The Pull Request
Removed vore grab special.................................... sad..................
Added fireman throw grab special instead. Procs the same as that vile vore special, click on self with the grab. You throw your victim behind yourself, dealing holodamage, and even more holodamage if they land on a window/table/into a wall
swing grab special - grab intent neck level, swing the victim around yourself in a circle and throw them VERY far away; can't proc if there are structures/walls in range of 1 tile around you, mobs in that range get weakened. Needs the victim to be on the ground to work.
Throatslit grab special nerfed; it deals regular knife damage to victim, but with wounding multiplier of 2, so basically double the damage if unarmored, grab is not killed afterwards. Oxyloss damage is completely removed.
Vomit stun duration reduced from 5 to 2 seconds 
Neckgrab no longer sets victim's click cooldown to 1 second
also removed(i hope) a runtime that showed up when someone tabled someone else while trying to upgrade the grab

## Why It's Good For The Game
NO MORE VORE
MORE GRABS INTEAD
Throatslit was OP, now it still is, but far less
Vomit stun was too long even before gut punch was added, so here it is now
Neckgrab shouldn't give click cooldown by itself, victim already spams resist, which itself gives click cooldown

So, basically, salt pr. Especially the vore part.

## Changelog
:cl: Kegdo
add: swing grab special
add: fireman grab special
del: vore grab special
balance: throatslit damage refactored
balance: neckgrab doesn't give click cooldown
balance: vomit stun duration reduced from 5 to 2
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon_defense.dm
+++ b/code/modules/mob/living/carbon/carbon_defense.dm
@@ -71,12 +71,12 @@ true, and the mob is not yet deleted, so we need to check that as well*/
 		if(!(G && G.assailant == user && G.affecting == src)) //check that we still have a grab
 			return 0
 
-		affecting.damage_through_armor(W.force, W.damtype, BP_HEAD, wounding_multiplier = 2, sharp = W.sharp, edge = W.edge, used_weapon = W)
+		damage_through_armor(W.force, W.damtype, BP_HEAD, wounding_multiplier = 2, sharp = W.sharp, edge = W.edge, used_weapon = W)
 		
 		user.visible_message(SPAN_DANGER("\The [user] cut [src]'s neck with \the [W]!"))
 
 		if(W.hitsound)
-				playsound(loc, W.hitsound, 50, 1, -1)
+			playsound(loc, W.hitsound, 50, 1, -1)
 
 		G.last_action = world.time
 		flick(G.hud.icon_state, G.hud)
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/carbon/carbon_defense.dm
```diff
@@ -66,40 +66,25 @@ true, and the mob is not yet deleted, so we need to check that as well*/
...
-	adjustOxyLoss(min(oxyloss, 100 - getOxyLoss())) //don't put them over 100 oxyloss
-
-	if(total_damage)
-		if(oxyloss >= 40)
-			user.visible_message(SPAN_DANGER("\The [user] slit [src]'s throat open with \the [W]!"))
-		else
-			user.visible_message(SPAN_DANGER("\The [user] cut [src]'s neck with \the [W]!"))
+		damage_through_armor(W.force, W.damtype, BP_HEAD, wounding_multiplier = 2, sharp = W.sharp, edge = W.edge, used_weapon = W)
+		
+		user.visible_message(SPAN_DANGER("\The [user] cut [src]'s neck with \the [W]!"))
```
hyperioo, line 76:
```suggestion
		user.visible_message(SPAN_DANGER("\The [user] cuts [src]'s neck with \the [W]!"), SPAN_DANGER("You cut [src]'s neck with \the [W]!"))
```

hyperioo (changes requested):
Also need to test if the removal of devour grab does not block carrions from eating their prey. Because yeah, that's something useful.

ErisMeleeEnjoyer:
image
this grab special devours the WHOLE target. Carrions use a different proc.

ErisMeleeEnjoyer:
It works.

hyperioo (approved):

PR #7506, opened 2022-09-13
GYM

ErisMeleeEnjoyer:
## About The Pull Request
A draft for now!

>It's a club depo room that requires everyone but club staff to pay the fee(ranging from 0 to 500 credits, 0 on roundstart) to enter; inside there are two instances of three gym structures, each type of which works as a stationary oddity for a single combat stat(+10). Using these structures takes 15 seconds.
How to use them as an oddity? That's where the pain in the ass for coding shows up: Upon use, it makes a temporary wound that reduces the user's pain threshold for three minutes, thrice less if the user drinks a protein shake that can be created by combining milk and animal protein.
To use it as an oddity, I first have to make a bug a feature: insight stacking rework. Right now, there is no cap on how much insight points a player has. I will change this so that there can only be ONE active insight point, but you get no cap on how much rest points you have. A rest point is given when you finish resting(duh), or when you finish a personal objective. 
When a player uses a gym structure with 0 rest points, they gain a +15 buff to the structure stat for 5 minutes. They cannot use any gym structures while their temporary wounds are active.
If a player does have a point, however, they gain +10 to the stat permanently.
Now, to make regular oddities still viable: a rest point can be used anytime to improve your stats, using an oddity or not, whatever the player chooses. They can also pick an option to dump all their rest points into one oddity at once. There will be two buttons on the sanity UI page for that. It should also be possible to stop/resume insight gain with a third button on that page.

## Why It's Good For The Game
__**GYM**__
Might change our meta lol
Feature from kotmap

## Changelog
:cl: Kegdo
add: GYM!!!
del: holodeck
balance: insight stacking reworked

/:cl:

### DIFF
--- a/maps/CEVEris/_CEV_Eris.dmm
+++ b/maps/CEVEris/_CEV_Eris.dmm
@@ -9350,10 +9350,10 @@
 /turf/simulated/wall/r_wall,
 /area/eris/crew_quarters/fitness)
 "awt" = (
-/turf/simulated/floor/reinforced{
-	name = "Holodeck Projector Floor"
-	},
-/area/holodeck/alphadeck)
+/obj/structure/curtain/open/shower,
+/obj/machinery/shower,
+/turf/simulated/floor/tiled/white/brown_perforated,
+/area/eris/crew_quarters/fitness)
 "awu" = (
 /obj/machinery/atmospherics/unary/vent_scrubber/on,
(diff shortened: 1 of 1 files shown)


tristain01:
Truly, the boss of this gym.

SirRichardFrancis:
F

PR #7510, opened 2022-09-13
adds a new tip about being able to disable switching to the examine tab

Alliostra:
## About The Pull Request

This PR adds a tip that notifies people it's possible to turn off automatically switching to the examine tab every time you examine something.

## Why It's Good For The Game

It probably would have taken me a long time to find out that you can disable this IMO very annoying feature if I didn't check the PR that introduced the examine tab switching, and this probably applies to other people as well.

## Changelog
:cl:
add: Added a new tip about the preference of automatically switching to the examine tab
/:cl:


Firefox13 (approved):
Shouldn't it be just off by default?

MLGTASTICa:
> Shouldn't it be just off by default?

Considering how lackluster the wiki is , i believe having it enabled by default is far better for newbies , since ive added lots of documentation inside the examine tab for items.

TheShown911 (approved):
Newline missing.

code/modules/tips_and_tricks/gameplay.dm
```diff
@@ -46,3 +46,6 @@
 
 /tipsAndTricks/gameplay/cuffHardsuit
     tipText = "People in RIG/Hardsuits need to be cuffed using zipties or cable cuffs. Normal cuffs are not large enough to cover their gauntlets."
+
+/tipsAndTricks/gameplay/examinetab
+    tipText = "Automatically switching to the examine tab upon examining something can be disabled in your preferences."
```
hyperioo, line 51:
```suggestion
    tipText = "Automatically switching to the examine tab upon examining something can be disabled in your preferences."

```

PR #7514, opened 2022-09-14
Should fix .40 packets and .20 scrap stripper clips icons

maikilangiolo:
## About The Pull Request
Hopefully this fixes the problem with .40 ammo boxes having an empty sprite above 30 rounds, and .20 scrap bullets having the wrong name in the DMI, thus making the .20 clips invisible. The latter I managed to test and works (had to scrap the commit since merge conflicts), the former should work but my current code somehow throws random exceptions and bricks my computer when debugging.

## Why It's Good For The Game
invisible ammo bad.

## Changelog
:cl:
fix: fixed .20 scrap stripper clip icons
fix: fixed .40 ammo packet icons
/:cl:


TheShown911 (commented):
Is this tested properly?

maikilangiolo:
Clips are. The issue was the ammo having F in the same instead of S.

PR #7519, opened 2022-09-14
Technomancer Rig Speed Buff

assortedbeads:
## About The Pull Request
This gives the Technomancer Rig the same slowdown as the Technomancer Exhault Rig.

## Why It's Good For The Game
I feel like there's enough of a distinction between the TE and Techno rigs module and armor wise, that boosting the speed of the latter won't blur the lines between them.

Also it's weird that while all Clan are expected to be able to respond to trouble effectively, the majority of the clan is stuck lagging behind when it comes to simple movement.

## Changelog
:cl:
tweak: Techno Rig has the Light_Slowdown Flag now, same as the TE Rig
/:cl:

### DIFF
--- a/code/modules/clothing/spacesuits/rig/suits/station.dm
+++ b/code/modules/clothing/spacesuits/rig/suits/station.dm
@@ -211,6 +211,7 @@ Technomancer RIG
 	)
 
 	spawn_blacklisted = TRUE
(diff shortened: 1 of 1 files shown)


TheShown911 (approved):
Sure.

PR #7524, opened 2022-09-15
Conveyor and lever removal

Iglpigl:
## About The Pull Request

Removes 2 levers and 4 conveyors in guild considering shuttle is gone now.
Before:
before
After:
Screenshot_1

## Why It's Good For The Game

Map looks cleaner now that shuttle is gone.

## Changelog
:cl:
tweak: no more conveyors and levers in lower guild
/:cl:

### DIFF
--- a/maps/CEVEris/_CEV_Eris.dmm
+++ b/maps/CEVEris/_CEV_Eris.dmm
@@ -93995,12 +93995,6 @@
 	},
 /turf/simulated/floor/tiled/techmaint,
 /area/eris/maintenance/section3deck1central)
-"erN" = (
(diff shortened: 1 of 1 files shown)


Iglpigl:
> Why even leave the dock here if it's no longer used?

It looked ugly walled in, feels like this is probably better for the time being.

hyperioo (changes requested):
The top bulb is floating in the air. I know it was there before (for some reason...) but now it's even weirder with the lever gone.

Iglpigl:
Ghost bulb deleted.

hyperioo (approved):

PR #7530, opened 2022-09-16
Gun Crafting Expansion: Part I (of many) [GOOD TO MERGE]

maikilangiolo:
## About The Pull Request
Expanding on the system we created, half assed and abandoned, this PR aims to make gun smithing fun, worth the investment in time and resources, as well as finally allowing people to craft guns from frames since the parts are impossible to find (unless you ae guild and buy 20 part boxes).
This PR will also change how the gun assembly works: it will no longer spawn full guns, but frames that you will have to complete, and by doing this we finally, finally discard the obsolute generic gun parts that are inexplicably still in.
Recycling and crafting cost has been adjusted so you wont be able to duplicate plasteel: crafting a whole gun for scrach will cost 5 more plasteel than you get by recycling the gun (17 for pistols/smgs, 25 for rifles/shotguns), as well as preventing constant recycle to make more assemblies and re-roll the frame. At the same time, I always think that crafty craftsmen and scavengers should be rewarded: you can recycle the individual parts for the same plasteel it takes to craft an equivalent part, so you can turn a rifle barrel into other rifle barrels, and small mechanisms into other small mechanisms; this also helps since there is **no** indication of what parts go in which gun, and wasting plasteel because you did not know what part went into your gun is bad. 

This expansion is as much my work as handy's (probably more his, considering all he had to put up with).

## Why It's Game For The Good
Old bad, new good, gun smthing cool and parts are cooler.

## Changelog
:cl:
add: Added the ability to craft gun parts in the crafting station
tweak: Changed the gun assembly to print frames instea of whole guns
/:cl:


code/datums/craft/gun_parts.dm
```diff
@@ -258,27 +261,28 @@
 	icon_state = "barrel_35"
 	generic = FALSE
 	bad_type = /obj/item/part/gun/barrel
-	matter = list(MATERIAL_PLASTEEL = 4)
+	matter = list(MATERIAL_PLASTEEL = 8,)
```
zombiemimic, line 264: Does this comma work here?

zombiemimic: or in other matter lists?

maikilangiolo: Forgot to test the recycling of parts. Yeah, fixing commas. All the crafting works, tested each single part

zombiemimic: glad to hear.
Looks good, otherwise

maikilangiolo:
Will eventually have to fix the description of the crafting station. it will be a kilometer long list of required materials, but without Egun parts, its fine for now

code/game/machinery/crafting_station.dm
```diff
@@ -96,8 +104,88 @@
...
+							required_resources = materials_hbarrel
+							items_to_spawn = list("" = /obj/item/part/gun/barrel/antim)
+
+				if("Mechanisms")
+					choice = input(user, "Which mechanism do you want to craft?") as null|anything in list("Small mechanisms", "Large mechanisms", "Machinegun mechanism")
+					switch(choice)
+						if("Small mechanisms")
+							required_resources = materials_smechanism
+							choice = input(user) as null|anything in list("Pistol mechanism", "Revolver mechanism", "SMG mechanism")
+							to_chat(user, "first mechanism choice")
```
hyperioo, line 144:
```suggestion
```
Honk

hyperioo (changes requested):

hyperioo (approved):

PR #7561, opened 2022-09-19
When where u when assembly was kill

maikilangiolo:
## About The Pull Request
Kills the assembly because it accepts every gun/part item, including *any* modular gun mod which can be printed for as cheap as 6 plastic or wood. 

## Why It's Good For The Game
Critical oversight (we didn't know it would happen), and it's the quickest fix I got i can do on mobile. We still have to rework how gun crafting works, we may removed the option to craft guns that aren't makeshift like this. 

## Changelog
:cl:
del: deleted the gun assembly crafting recipe. Rest in piss
/:cl:


Alliostra:
I was at discord posting on feedback when gitbah ring

"Assembly is kil"

"no"

TheShown911:
Maybe just check the type of the part that is being inserted instead of removing this for no major reason.

TheShown911:
That's very sad, but lacking the ability to code something is not an excuse to remove another thing from the game.

maikilangiolo:
I need to make a separate PR for it. Humon, which is still a dev del, said to remove "ex novo" gun crafting, but instead make more frames spawn in maint. Instead of spending X plasteel to make Y guns buckled to a chair, you would go and find gun frames in maint and assemble them.

I still haven't decided whether to do as he says, or not. I do like gun crafting, but scavenging from frames is also fun and when there was no gun crafting, there were less issues of people grinding mech and make very rare and powerful guns with ease: STS, lmgs, c20r, etc etc.

But this is a fix to an exploit, for the time being.

hyperioo:
I'll merge it to fix the exploit but I hope to see that PR you're talking about so that gun frame fabrication is not lost forever. Restricting it to "makeshift" gun could be a nice way yeah.

maikilangiolo:
That's already in. You craft makeshift frames

PR #7562, opened 2022-09-20
[Critical] Fix of frame assembling

maikilangiolo:
## About The Pull Request
Did not know all vars had to be lists even with one single item. oops

EDDIT: see the comment for the steel barrel and plasteel issue. ALSO fixed obrez guns NOT inheritng the parent barrel (only used in the makeshift bolt gun)

## Why It's Good For The Game
its bad becase now people can craft guns instead of being bugged (guns are scary)

## Changelog
:cl:
fix: fixed frame crafting
/:cl:


TheShown911 (approved):

maikilangiolo:
ok really fixed now.
NOTE: THe special steel variant of the .20 and .30 bolt guns is now for the momet unusedd because it only caused issues. FOR NOW, bolt guns cost 8 plasteel to print (4 less than the tiniest pistols) to avoid transmutating a 10 steel gun into an 8 plasteel barrel, since all bolt guns drop the plasteel barrels.

fyi, steel barrels could NOT be used to craft other .30 and .20 guns, making it a buggy special barrel that could not be interchanged between guns. the bolt gun mechanism is unchanged and still recycles for only steel

PR #7566, opened 2022-09-21
Help Those That Help Themselves

Mycah142:
## About The Pull Request

Added check for id_scan flip before cruciform checks to the NeoTheology vendor.

## Why It's Good For The Game

The Guild has orders that require church items that are locked within the church, or behind the Theomat that checks for only the faithful. An aspiring techie can now try to hack the NT Vendor for the items stealthily.

This will also lead into future changes that I intend where ritual books need more availability to people outside the church.

## Changelog
:cl:
tweak: Added id_scan check to the NeoTheology vendor before it checks for cruciform.
/:cl:

### DIFF
--- a/code/game/machinery/vending.dm
+++ b/code/game/machinery/vending.dm
@@ -1492,7 +1492,7 @@
 	
 		var/mob/living/carbon/human/H = user
 
-		if(scan_id == !1) //I have no idea why == 0 didn't work. It just wouldn't. But this did. I don't know why. Thanks Jeff Jefferson.
+		if(!scan_id)
 			bingo = TRUE
 
 		else if(is_neotheology_disciple(H))
--- a/code/game/machinery/vending.dm
+++ b/code/game/machinery/vending.dm
@@ -1492,7 +1492,7 @@
 	
 		var/mob/living/carbon/human/H = user
 
-		if(scan_id == !1)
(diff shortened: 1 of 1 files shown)


code/game/machinery/vending.dm
```diff
@@ -1489,8 +1489,13 @@
 /obj/machinery/vending/theomat/proc/check_NT(mob/user)
 	var/bingo = FALSE
 	if(ishuman(user))
+	
 		var/mob/living/carbon/human/H = user
-		if(is_neotheology_disciple(H))
+
+		if(scan_id == !1) //I have no idea why == 0 didn't work. It just wouldn't. But this did. I don't know why. Thanks Jeff Jefferson.
```
TheShown911, line 1495:
```suggestion
		if(!scan_id) //I have no idea why == 0 didn't work. It just wouldn't. But this did. I don't know why. Thanks Jeff Jefferson.
```
Maybe this should work?

Mycah142: I tried that initially and it didn't work, as id_scan is defined as 1 initially.

Maybe I tried it while it was lower in the if_else sequence when nothing was working.

I can try again tomorrow morning when I have time.

Mycah142: It did work. Thank you. I likely had it further down in the if_else statements which gave me trouble. I'm not coding with anything other than what comes to me in dreams.

PR #7573, opened 2022-09-22
As shrimple as that: fixes the misinformative item UI

maikilangiolo:
## About The Pull Request
No more "15%" armor on BP with a red bar, no more 42 damage wintermute, no more useless wound multiplier shown. Items properly display their stats, and the max values were tweaked so that they are color coded properly like before.
image
image
image
image

ISSUES:
this happens with some guns, I think only does that have a 1.5 wound mulipier 
image

(and this is the formula as is)
image

## Why It's Good For The Game
About fucking time. Players *should not* be forced to do math or have to codedive for stats, or have armor displayed as a %.

## Changelog
:cl:
fix: fixed armor and gun UI being hard to understand or lying
/:cl:


TheShown911 (approved):
Maybe now i will know what any of those numbers meant...

SirRichardFrancis:
Nah, you won't.

maikilangiolo:
> Nah, you won't.

is this a code issue? humon helped me fix the mistake, I removed the formula from the template file and made the gun.dm have the formula

SirRichardFrancis:
> is this a code issue?

It's balance PRs being meaningless issue.

githubuser4141:
Better then how it is currently, IMO the numbers look out of place to represent what the bars are saying? But that'd require a UI overhaul so 🦐 
Would it be possible to add a small description of WTH those numbers and terms actually mean? Percentages in the old system were straightforward; a bullet that does 50 damage with 40% armor ofc does 30 damage. "10 armor" VS 30 damage & 15 wound multiplier doesn't really compute

maikilangiolo:
The armor explain is simple, but rosa didn't include it in the devlog and humor didn't include in the changelog. 
Each point of armor reduces one point of damage, and armor is divided by AP (1 AP doesn't pierce armor, 0.8 is 20% worse vs armor and 2 al halves armor).

After you reduce bullet damage by armor, you multiply it by the wound multiplier but, I have to be honest, it's most code magic that allows us to fine tune bullet damage.

maikilangiolo:
> Better then how it is currently, IMO the numbers look out of place to represent what the bars are saying? But that'd require a UI overhaul so 🦐 Would it be possible to add a small description of WTH those numbers and terms actually mean? Percentages in the old system were straightforward; a bullet that does 50 damage with 40% armor ofc does 30 damage. "10 armor" VS 30 damage & 15 wound multiplier doesn't really compute

Also old armor **wasn't** straightforward. 40% armor did reduce damage by 40%. If a gun had 0 AP (there were no guns with 0 AP) 40% armor reduced damage by 4 and then by 40%. So a 26 damage bullet (with 0 AP) would have been 26-4 = 22, -40% which is 13.2, and a 31 damage bullet would have been 16.2: so in this case 40% actually reduced damage by 50%.

It gets worse when you take into account AP. .25 was 25 damage 10 AP. So 40% would reduce damage by 3 (26-3=23) and then reduce it by 30% (23-30%=16). Or we can use the .20 bullet, which was 21 damage 25 AP. So against 40% armor, that's 21-1.5=19.5, then 19.5-15%=16.5. 

Armor wasn't straightforward before. It used a complex and hidden (from the UI but also general knowledge) system that had a different reduction based on the bullet you got hit by.

PR #7576, opened 2022-09-22
Spaghetti code: we found a new bug

maikilangiolo:
## About The Pull Request
Ok so, due to some shitcode, if a gun **did not** specify the penetration_multiplier, even if it was 0 (defauly caliber), it caused guns to have essentially double the AP. I guess I know why the molly did so much damage @MLGTASTICa (clueless)
The only guns that still did not specify were the STS and clarrissa (and child anne)

## Why
Fix!!!!!!!!

## Changelog
:cl:
fix: fixed spaghetti with guns that did not specify ap multplier, the sts and clarissa. whoops
/:cl:


Firefox13 (changes requested):
изображение
All guns default to multiplier of `1`, since it's defined in `code\modules\projectiles\gun.dm`
Change that line to zero too and keep changes in other files for clarity.

Firefox13 (approved):
great now we just have to replace all `multiplier` with `modifier` but w/e

PR #7586, opened 2022-09-25
Fixed the NT Cloning bug

BlackCat-055:
## About The Pull Request

Neotheology Cloning pumps out a generic potato man instead of your character's original appearance, which sort of sucked.

So I fixed that and now basically all of your character's apperance is copied over, code-wise technically so should the TTS seed (how your character sounds) as well, but I cant test it in local server so it has to be tested live.

Thanks to @Pink-Chink for helping with the skin tone part of the appearance code

## Why It's Good For The Game

A bug is fixed and people can keep their appearances as NT followers, not really life important, but it is a QoL imo

## Changelog
:cl:
fix: fixed the NT not being able to copy all of appearance bug
code: Did some new code on the cloning machine and cruciform module to allow for this
/:cl:

### DIFF
--- a/code/modules/core_implant/cruciform/machinery/cloning.dm
+++ b/code/modules/core_implant/cruciform/machinery/cloning.dm
@@ -213,6 +213,15 @@
 			occupant.real_name = R.real_name
 			occupant.b_type = R.b_type
 			occupant.age = R.age
+			occupant.h_style = R.h_style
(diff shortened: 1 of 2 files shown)


DiaFRAME444:
Praise be.

Mycah142:
Amen

hyperioo (approved):

PR #7592, opened 2022-09-25
Fix wrong icon_file check

hyperioo:
## About The Pull Request

Fix those runtimes

```
[15:25:15] Runtime in asset_list_items.dm,410: bad icon
  proc name: register (/datum/asset/simple/design_icons/register)
  src: /datum/asset/simple/design_ico... (/datum/asset/simple/design_icons)
  call stack:
  /datum/asset/simple/design_ico... (/datum/asset/simple/design_icons): register()
  /datum/asset/simple/design_ico... (/datum/asset/simple/design_icons): New()
  get asset datum(/datum/asset/simple/design_ico... (/datum/asset/simple/design_icons))
  Assets (/datum/controller/subsystem/assets): Initialize(null)
  Master (/datum/controller/master): Initialize(10, 0, null)

[15:25:15] Runtime in asset_list_items.dm,451: bad icon
  proc name: register (/datum/asset/simple/craft/register)
  src: /datum/asset/simple/craft (/datum/asset/simple/craft)
  call stack:
  /datum/asset/simple/craft (/datum/asset/simple/craft): register()
  /datum/asset/simple/craft (/datum/asset/simple/craft): New()
  get asset datum(/datum/asset/simple/craft (/datum/asset/simple/craft))
  Assets (/datum/controller/subsystem/assets): Initialize(null)
  Master (/datum/controller/master): Initialize(10, 0, null)
```

## Changelog
:cl: Hyperio
fix: Fix wrong icon_file check
/:cl:

PR #7593, opened 2022-09-25
[CI] Add placeholder sprites that are used as initial

hyperioo:
## About The Pull Request

CI is angry because some sprites do not exist. In practice they're never used because the ammo magazines use initial() + "xx" depending on how much they are filled.

## Changelog
:cl: Hyperio
add: Add placeholder sprites for some ammo types
/:cl:


TheShown911:
Fails CI 🤣

hyperioo:
Should be better now that #7592 is merged, but now CI is not finished /tableflip

hyperioo:
Yeah it seems it's gonna throw a CI fail for all sprites in the game whose initial() is never used and just meant to be used with a suffix

PR #7597, opened 2022-09-26
Hyperburst Fix

budgetcommander:
## About The Pull Request

The high-delay rapid burst is now both high-delay and rapid, instead of just rapid. The burst firerate is honestly a bit insane, but because of the massive delay between bursts, it's really just spectacle.

## Why It's Good For The Game

Hyperburst is back, baby! It was broken after the gun rework.

## Changelog
:cl:
fix: The sol's hyperburst actually fires fast again.
/:cl:


Humonitarian:
Should be enough if you put it at 1.4 I believe, same as Molly hyperburst

maikilangiolo:
Probably better not to be 0.5. If it's got such a massive delay, no real point using it. Molly has the "updated" high delay value

Humonitarian:
> Probably better not to be 0.5. If it's got such a massive delay, no real point using it. Molly has the "updated" high delay value

Also note guns shoot faster on server than local. Don't question why. I don't know either.

budgetcommander:
Molly's hyperburst is underwhelming.

Humonitarian:
> Molly's hyperburst is underwhelming.

You could lower that down to 1.1, was meant to be slightly faster than Sol's (more bullets / burst make it last long)

PR #7598, opened 2022-09-26
Adds placeholder aberrant organ research

Pink-Chink:
## About The Pull Request

After every 3rd flesh cube, the disgorger will say a line and unlock a new design in the organ fabricators. This only works if the organ fabricators are in the area. When constructed organ fabricators will check for a disgorger in the area and copy its researched designs.

## Why It's Good For The Game

More self-sufficient progression, more gooder.

## Changelog
:cl:
add: Added placeholder aberrant organ research
/:cl:

### DIFF
--- a/code/modules/aberrants/organs/machinery/disgorger.dm
+++ b/code/modules/aberrants/organs/machinery/disgorger.dm
@@ -31,6 +31,7 @@
 
 	// Lazy research placeholder
 	var/spits_until_unlock = 3
+	var/datum/research/knowledge
 	var/list/designs_to_unlock = list(
 		/datum/design/organ/organ_mod/parenchymal_large,
 		/datum/design/organ/teratoma/special/chemical_effect,
@@ -80,6 +81,8 @@
 /obj/machinery/reagentgrinder/industrial/disgorger/Initialize()
 	. = ..()
 	spit_target = get_ranged_target_turf(src, dir, spit_range)
+	knowledge = new /datum/research(src)
+	knowledge.known_designs = list()
 
 /obj/machinery/reagentgrinder/industrial/disgorger/InitCircuit()
 	if(!circuit)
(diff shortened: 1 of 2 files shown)


code/modules/aberrants/organs/machinery/organ_fabricator.dm
```diff
@@ -38,6 +38,13 @@
 	. = ..()
 	files = new /datum/research(src)
 
+	// Remove when actual organ research is made
+	for(var/disgorger in get_area_all_atoms(get_area(src)))
+		if(istype(disgorger, /obj/machinery/reagentgrinder/industrial/disgorger))
+			var/obj/machinery/reagentgrinder/industrial/disgorger/D = disgorger
+			for(var/design in D.knowledge.known_designs)
+				files.AddDesign2Known(design)
```
SirRichardFrancis, line 46:
```suggestion
	for(var/obj/machinery/reagentgrinder/industrial/disgorger/D in get_area_all_atoms(get_area(src)))
		for(var/design in D.knowledge.known_designs)
			files.AddDesign2Known(design)
```

code/modules/aberrants/organs/machinery/disgorger.dm
```diff
@@ -179,6 +232,35 @@
...
+		"To know a thing well, know its limits; Only when pushed beyond its tolerance will its true nature be seen." = 1,	//
+		"You do not take from this universe. It grants what it will." = 1,							// Dune Messiah
+		"Belief can be manipulated. Only knowledge is dangerous." = 1,								//
+		"You can\'t stop a mental epidemic. It leaps from person to person across parsecs." = 1,	//
+		"..." = 31,
+		"...?" = 31,
+		"...!" = 31
+		))
+
+	for(var/mob/O in hearers(src, null))
```
SirRichardFrancis, line 256: hearers() meant to return mobs only, so type check is unnecessary.
```suggestion
	for(var/mob/O as anything in hearers(src, null))
```

SirRichardFrancis (commented):
I don't see the point of doing typeless search in loops, if both comparison and type assignment are made anyway.
Same thing exactly as searching for type, but double to code.

Pink-Chink:
I was told that typeless search was faster regardless. Something about DM being better than C type loops when done this way.

Pink-Chink:
Might as well to keep things consistent.

hyperioo (approved):

PR #7547, opened 2022-09-17
AMR good: changes to .60 and buffs to AMR

maikilangiolo:
## About The Pull Request
.60 stats got nerfed, since there is talk of a future HMG using .60 and it shouldnt do AMR damage. to compensate, amr got 2x damage mod and extra damage so its slightly less shit (seriously, buckshot was better than the AMR). Hitscan was given to the AMR by setting the step delay to 0, as per humon's suggestion.

PR in preparation of a new makeshift gun PR, and needs #7546 to work since refactor of gun assembly. 

## Why It's The Game
AMR no more shit, .60 can be used in other guns without being too broken.

## Changelog
:cl:
balance: rebalanced .60 stats. AMR stats changed accordingly and buffed
/:cl:

PR #7603, opened 2022-09-26
Adds variable scope zoom + AMR improvements + bugfixes

Alliostra:
## About The Pull Request

The AMR and valkyrie can now switch between 2 different levels of zoom when using the scope.

Before, the AMR received a bonues 0.2 damage multiplier when zoomed out. This value has been doubled to make it add the same bonus damage percentage as before, since in https://github.com/discordia-space/CEV-Eris/pull/7547 the damage_multiplier of the AMR was set to 2, so logically the bonus damage when scopes must also be doubled.

When zoomed out at max level, the AMR gets 0.4 bonus damage multiplier, at the lower zoom level it gets 0.2

The shellshock can now fit a scope and bayonet again, because that was removed by accident before

The AMR no longer has a scope slot, because it already has a scope by default.

Some changes to action buttons, which can now be used for more things (use them!)

Huge thanks to Garen and SCPR for helping me with removing a critical bug.

Please mention in the changelog that people should yell at me if they find any new bugs with scopes, Rosa.
## Why It's Good For The Game

AMR cool, bugfixes good.

## Changelog
:cl:
add: the valkyrie and AMR can now switch between 2 different levels of zoom
tweak: increased the bonus damage AMR gets when using the scope
fix: fixed AMR having a scope slot
fix: fixed shellshock having no scope and bayonet slot
/:cl:


maikilangiolo:
> efore, the AMR received a bonues 0.2 damage multiplier when zoomed out. This value has been doubled to make it add the same bonus damage percentage as before

This will make the AMR do 150 damage per shot.

Humonitarian:
Resident recoil coder.
Remove the damage buff, it makes no sense. Make the bonus to recoil scale with zoom level instead. Thanks.

maikilangiolo:
The extra damage just bumps it from 72 damage to

Alliostra:
> Resident recoil coder. Remove the damage buff, it makes no sense. Make the bonus to recoil scale with zoom level instead. Thanks.

If you want to argue realism, or "logic", then it's due to overpenetration at short range. people should be encouraged to use the scope.

TheShown911 (approved):

TheShown911:
Conflicts!

PR #7606, opened 2022-09-26
Reorders all common oddities in oddities.dm to make more sense

DiaFRAME444:
## About The Pull Request
Prepares this document for another PR I'm planning to make later down the line.
General order should be Singular Combat Stat > Singular Work Stat > Dual Combat Stat > Dual Mixed >, and so on.
Also fixes description on the observer book.

## Why It's Good For The Game
I tried to mess with this file earlier, but since everything was out of order and ended up constantly jumping towards different points.
I also believe the reorganized state of this file reveals some things about our current oddity design and layout.

## Changelog
:cl: 
tweak: The oddities.dm file has been reordered.
spellcheck: Observer book description changed.
fix: Bear math oddity is now a subtype of /obj/item/oddity/common/, which allows it to spawn in maintenance.
/ :cl:

### DIFF
--- a/code/game/objects/items/oddities.dm
+++ b/code/game/objects/items/oddities.dm
@@ -53,15 +53,39 @@
 	bad_type = /obj/item/oddity/common
 	spawn_blacklisted = FALSE
 
-/obj/item/oddity/common/blueprint
-	name = "strange blueprint"
-	desc = "There's no telling what this design is supposed to be. Whatever could be built from this likely wouldn't work."
-	icon_state = "blueprint"
+/obj/item/oddity/common/lighter
+	name = "rusted lighter"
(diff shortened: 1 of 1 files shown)


Kurgis:
Looks good.

Firefox13:
If you're actually sorting them then add comments on when certain type of sort starts.

Firefox13:
Even if it does nothing for playerside changelogs are important for bookkeeping.

DiaFRAME444:
> Looks good, though you could argue that oddities can be ordered by alphabet, or ordered by their general palette/description theme as well.

This seems to be the most helpful layout.
Do correct me if I'm wrong.

> If you're actually sorting them then add comments on when certain type of sort starts.

Done

> Even if it does nothing for playerside changelogs are important for bookkeeping.

Fixed

hyperioo (approved):

PR #7618, opened 2022-09-27
Backstabbing

Alliostra:
## About The Pull Request

Adds backstabbing. Backstabbing happens when you stab someone in the back or stab someone who is prone. Backstabbing deals extra damage.
When attacking from the 3 tiles behind a person that are adjacent to them, a backstab happens.

The reason for why prone people are always backstabbed is mostly just that you can't really see which side someone's "back" is if you haven't memorized it, and having to move "behind" someone who is on the ground to backstab is kinda weird.

All knives except for the cleaver and the meathook can backstab (cleaver and meathook would deal too much damage if they could)
Regular knives deal double damage when backstabbing, daggers deal 2,5x damage, to create some distinction of daggers and knives, and to compensate for the lower damage of daggers.

damage values are of course up to discussion. If deemed necessary, damage values for stabbing in the back and stabbing while prone could be turned into different values.

also tweaks the description of the NT scourge to be more informative.

## Why It's Good For The Game

More reason to use knives good backstabbing cool. 

## Changelog
:cl:
add: knives can now perform backstab attacks. prone people are always backstabbed.
/:cl:

### DIFF
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -138,10 +138,19 @@ avoid code duplication. This includes items that may sometimes act as a standard
 		target.IgniteMob()
 
 	var/power = force
+	var/backstabbed = FALSE
+	if(can_backstab && backstab_multiplier)
+		var/attack_dir = 0
+		var/vulnerable_dirs = target.dir
+		attack_dir = get_dir(get_turf(user), get_turf(target))
+		if((attack_dir && (attack_dir & vulnerable_dirs))||target.lying)
+			power*= backstab_multiplier
+			backstabbed = TRUE
+
 	if(ishuman(user))
 		var/mob/living/carbon/human/H = user
 		power *= H.damage_multiplier
 //	if(HULK in user.mutations)
 //		power *= 2
-	target.hit_with_weapon(src, user, power, hit_zone)
+	target.hit_with_weapon(src, user, power, hit_zone, backstabbed)
 	return
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -78,6 +78,8 @@
 
 	//Damage vars
 	var/force = 0	//How much damage the weapon deals
+	var/can_backstab = FALSE //Whether or not this item can backstab
+	var/backstab_multiplier = 2 //force multiplier for backstabbing
 	var/embed_mult = 0.5 //Multiplier for the chance of embedding in mobs. Set to zero to completely disable embedding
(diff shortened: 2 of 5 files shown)


ErisMeleeEnjoyer:
team fortress 2

ErisMeleeEnjoyer:
daggers should have extra AP but 1.5x damage increase instead

DiaFRAME444:
This transforms a spike tool mod from 4 damage to 10. Then add sharpened edge for +15% damage.

Imagine being grabbed and then killed because you a vagabond got 2 swings of 2.5x damage on you.

Alliostra:
> Multiply dagger damage by 2.5x, then add spike for +10 (4 x 2.5 = 10) damage, then add sharpened edge for +15% damage.
> 
> Imagine being deleted because you a vagabond got a quick grab on you.


> 

I see your point, what values would you consider more appropriate? and no, I don't think this stacks with throat slitting, but I have to test that first

DiaFRAME444:
I mean my biggest problem is that taking spessman's "direction" into account is silly. People can rotate around incredibly fast.
Servers where I think its okay to have this kinda stuff are Lifeweb where you can't rotate 10 times per second, so not being able to see behind yourself is fine.

Alliostra:
> I mean my biggest problem is that taking spessman's "direction" into account is silly. People can rotate around incredibly fast. Servers where I think its okay to have this kinda stuff are Lifeweb where you can't rotate 10 times per second, so not being able to see behind yourself is fine.

Well, shields also work with directions, and when running in a straight line, such as when fleeing, you can't really turn around for longer than a split second most of the time.

DiaFRAME444:
Eh. Well, disable backstabs if you're grabbing the person in question? This feels like the most obvious exploit given daggers already have throat slitting.

Firefox13:
I don't know if this is a good idea. We don't have FOV (***do not add them***) and this is going to end up like TF2 chat "how the fuck you backstabbed me" and way too much turning around

Humonitarian:
> > I mean my biggest problem is that taking spessman's "direction" into account is silly. People can rotate around incredibly fast. Servers where I think its okay to have this kinda stuff are Lifeweb where you can't rotate 10 times per second, so not being able to see behind yourself is fine.
> 
> Well, shields also work with directions, and when running in a straight line, such as when fleeing, you can't really turn around for longer than a split second most of the time.

Rejecting the shield as a precedent for directional combat. Shields are already questionable and among the list of "things we should find an alternative for"

Alliostra:
y'all boring as hell.

PR #7620, opened 2022-09-27
[CI] CI fixes related to asset cache

hyperioo:
## About The Pull Request

Too many items have a default icon_state that is never used in practice without a prefix (added at Initialization) so this check was gonna fail for tens of items since the default sprite is not in .dmi

For instance "boltgun" sprite does not exist, because in game it's always either "boltgun_closed" or "boltgun_open".

Add a default build path for circuit design so that asset system does not derp.

Also add an additional check in get asset url proc to handle the cases in which there is no asset.

## Changelog
:cl: Hyperio
server: Comment CI sprite check for designs
add: Add a default build path for circuit designs
add: Add a check in get asset url to handle null cases
/:cl:


hyperioo:
image

TheShown911 (approved):

TheShown911:
Fugg it.

PR #7623, opened 2022-09-28
Lasblender buff

assortedbeads:
## About The Pull Request
Lasblender has slightly better damage, pen and no initial offset now.
Description changed to remove the accuracy malus mention.

Furher dies in 4 bursts.

The gun was overnerfed the first time and now it's a waste of materials, it's still pretty bad now but not as bad.
Does 10.5 damage with 1 pen, so 84 if you get pegged with every shot from a full burst.
## Changelog
:cl:
add: Added Lasblender crafting recipe to the guns section
del: Removed Lasblender crafting recipe from the weapons section
tweak: Lasblenders description has been updated
balance: Lasblender does has a 0.35 damage mod and a 1 pen mod now, up from 0.28 and default
balance: Lasblenders initial offset is gone
/:cl:

### DIFF
--- a/code/datums/craft/recipes/guns.dm
+++ b/code/datums/craft/recipes/guns.dm
@@ -95,6 +95,20 @@
 		list(QUALITY_ADHESIVE, 15, 70)
 	)
 
+/datum/craft_recipe/gun/lasersmg
+	name = "Lasblender"
+	result = /obj/item/gun/energy/lasersmg
+	steps = list(
+		list(/obj/item/gun/projectile/automatic/atreides, 1),
+		list(QUALITY_WELDING, 10, "time" = 30),
+		list(CRAFT_MATERIAL, 6, MATERIAL_PLASTEEL, "time" = 10),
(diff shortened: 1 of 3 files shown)


DiaFRAME444:
Removing the initial offset makes it less chaotic, though.
Maybe only reduce the recoil/initial offset instead of removing it, keep the damage buff, and increase its firerate?

assortedbeads:
> Removing the initial offset makes it less chaotic, though. Maybe only reduce the recoil/initial offset instead of removing it, keep the damage buff, and increase its firerate?

Increasing the firerate is a bit of a slippery slope since it already fires fast for an uzi and has laser hitscan, so making it fire even faster would mean you could full auto someone to death quite easily.

While removing the initial offset doesnt make it more chaotic, it's annoying to have offset to begin with when taking stiffness or just poor vig into account. You should at least hit what you're aiming for on the first shot.

TheShown911 (approved):
**216** DAMAGE

PR #7625, opened 2022-09-28
OS generator and turrets

Pink-Chink:
## About The Pull Request

Adds some One Star machinery for away missions. They won't be present in the game without admemery, but it's for people who want to design away maps.

Fusion Microgenerator
- A portable generator that slowly ramps up power output until it hits 300 kW without the need for fuel. Needs MEC or COG to understand and operate. Requires OS parts or it will fail.

Turrets (Gauss and Pulse)
- Turrets that will return fire when shot, even if the target is out of normal range. When destroyed or disabled, it can be disassembled/unanchored. A reconstructed turret will only shoot roaches, spider, and golems. Can only be constructed with OS parts.

## Why It's Good For The Game

Content

## Changelog
:cl:
add: Added OS generator and turrets
/:cl:


ValoTheValo:
POG

code/modules/onestar/os_turret.dm
```diff
@@ -0,0 +1,306 @@
...
+	update_icon()
+
+	var/list/potential_targets = list()
+	var/mob/nearest_valid_target
+	var/nearest_valid_target_distance
+
+	if(should_target_players)
+		potential_targets = (GLOB.player_list & SSmobs.mob_living_by_zlevel[(get_turf(src)).z])
+	else
+		potential_targets = (GLOB.superior_animal_list & SSmobs.mob_living_by_zlevel[(get_turf(src)).z])
```
SirRichardFrancis, line 71: Doesn't this object have it's own `.z`? Which is always the same as it's turf's.

code/modules/onestar/os_turret.dm
```diff
@@ -0,0 +1,306 @@
...
+
+	var/def_zone
+
+	if(istype(target, /mob/living/carbon/human))
+		def_zone = pick(
+			organ_rel_size[BP_CHEST]; BP_CHEST,
+			organ_rel_size[BP_GROIN]; BP_GROIN,
+			organ_rel_size[BP_L_ARM]; BP_L_ARM,
+			organ_rel_size[BP_R_ARM]; BP_R_ARM,
+			organ_rel_size[BP_L_LEG ]; BP_L_LEG,
```
SirRichardFrancis, line 223:
```suggestion
			organ_rel_size[BP_L_LEG]; BP_L_LEG,
```

code/modules/onestar/os_turret.dm
```diff
@@ -0,0 +1,306 @@
...
+			nearest_valid_target_distance = distance_to_target
+		else
+			if(distance_to_target > nearest_valid_target_distance)
+				nearest_valid_target = L
+				nearest_valid_target_distance = distance_to_target
+	
+	if(nearest_valid_target)
+		try_shoot(nearest_valid_target)
+	
+	return
```
SirRichardFrancis, line 101:
```suggestion
```

SirRichardFrancis (commented):
New turret kind, existing in parallel with other turrets? Alright.
You may want to replace "old" One Star turrets on map and get rid of their code then.

code/modules/onestar/os_turret.dm
Pink-Chink: Yeah, this was a lazy copy pasta from something else. I’ll change it before merge.

Pink-Chink:
New type because the current one is a bit bloated for what I wanted to accomplish.

SirRichardFrancis (approved):
>Total runtimes: 158

There are runtimes at roundstart... None related to this PR, however.

Pink-Chink:
I'm here for a good time not a runtime.

TheShown911 (approved):

PR #7629, opened 2022-09-29
Makes prosthetics examine text work properly

Iglpigl:
## About The Pull Request

Makes prosthetics examine text work properly, instead of showing stuff two times.
No more this:
17-Sep-22_21-59-00

## Why It's Good For The Game

Bug bad.

## Changelog
:cl:
fix: prosthetics examine text now works properly for all prosthetics
/:cl:


Iglpigl:
Nope, they are not for sale anymore, maybe they were when elar named them but now they are not.

Pink-Chink:
Either myself or the original coder took em off because they weren’t working with the stations. I don’t remember all the details. Regardless, trade stations can set a custom name for a good. So, there wouldn’t be any issues with item names.

Pink-Chink (approved):

PR #7633, opened 2022-09-29
Slug Buff (rework?)

budgetcommander:
## About The Pull Request

Reworks slugs to have lower wound value, but higher base damage.

## Why It's Good For The Game

This makes slugs better against armour than buckshot. As they currently exist, slugs, despite their somewhat superior armour penetration, still lose out against buckshot save for fringe cases due their damage not being enough. These reworked slugs will now handily outdamage buckshot against 15 armour, slightly win over them at 10 armour, and drastically fall behind against unarmoured targets.

## Changelog
:cl:
balance: Slugs now have lower wound value and higher base damage,
/:cl:


Humonitarian:
@maikilangiolo has the statistics on why you're wrong.

PR #7636, opened 2022-09-29
Adjusted Slug Speed

budgetcommander:
## About The Pull Request

Slugs are no longer slightly slower than other bullets.

## Why It's Good For The Game

It's a really strange decision to make slugs slower. I'm not sure what the reasoning was.

## Changelog
:cl:
balance: Slug step delay reduced from 1.1 to the default value of 1
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/bullettypes.dm
+++ b/code/modules/projectiles/projectile/bullettypes.dm
@@ -246,7 +246,6 @@ There are important things regarding this file:
 	damage_types = list(BRUTE = 25)
 	armor_divisor = 1
 	knockback = 1
-	step_delay = 1.1
 	style_damage = 25
 	recoil = 8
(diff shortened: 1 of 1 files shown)


maikilangiolo:
Need to make all obrez guns guns have 1.1 step delay, since they're 1.2 right now. Or make all sawn offs 1.2 since slugs aren't 1.1 anymore.

Humonitarian:
> It's a really strange decision to make slugs slower. I'm not sure what the reasoning was.

A shotgun slug travels approximately at 480 meters per second.
A 7mm magnum travels at 880 meters per second.
A .458 magnum travels at 652 meters per second.
A .303 rifle round travels at 800 meters per second.

You see the pattern?

Firefox13:
I guess it's one of those "realistic features" we have.
So rare these days.

PR #7648, opened 2022-10-01
Fix double overlays on sheaths

ErisMeleeEnjoyer:
## About The Pull Request

gunsheath no more(this must be a feature)
@SirRichardFrancis localtest lied to us apparently
## Why It's Good For The Game

~~it's not~~ fix

## Changelog
:cl:
fix: no more gun overlay on sheaths
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/storage/holster.dm
+++ b/code/game/objects/items/weapons/storage/holster.dm
@@ -355,13 +355,13 @@
 		overlays += image('icons/inventory/pockets/icon.dmi', "revolver_layer[contents.len]")
 
 /obj/item/storage/pouch/holster/belt/sheath/update_icon()
+	..()
 	cut_overlays()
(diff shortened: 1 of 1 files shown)


SirRichardFrancis (commented):
Look. You're cutting overlays that isn't there.
Then add sword overlay, which is fine.
Then calling pistol holster proc, which also adds pistol overlay.
So you'll end up with the same thing again, have to either reorder proc calls inside update_icon() or not inherit sheath from pistol holster.

ErisMeleeEnjoyer:
my head is bign't

SirRichardFrancis (approved):
That should do it. Not the cleanest code ever, but if it works - it works. ¯\\\_(ツ)\_/¯

hyperioo (approved):

PR #7651, opened 2022-10-02
Switch stalker position in armory dungeon room

hyperioo:
## About The Pull Request

Move the position of a stalker in the "armory" dungeon room. It was causing runtime at roundstart due to airflow trying to move the mob too soon, before its movement variables were all properly initialized

image

## Changelog
:cl: Hyperio
tweak: Switch stalker position in armory dungeon room
/:cl:

### DIFF
--- a/maps/submaps/dungeon_rooms/rooms/armory.dmm
+++ b/maps/submaps/dungeon_rooms/rooms/armory.dmm
@@ -149,7 +149,7 @@
 /area/crawler)
 "U" = (
 /mob/living/carbon/superior_animal/stalker,
-/turf/simulated/floor/tiled/derelict/red_white_edges,
+/turf/simulated/floor/tiled/dark/brown_platform,
 /area/crawler)
 
 (1,1,1) = {"
@@ -181,7 +181,7 @@ f
 k
 a
 x
-f
+U
 A
 a
 "}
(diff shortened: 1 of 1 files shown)


Kurgis:
Airflow occurring at all during roundstart is an issue in of itself. Though, since it would of course require someone to go through maps, and replace the majority of flooring with variable edited subtypes without gas present. Wouldn't be an enviable task by any means. Especially since I vaguely recall years ago there being an issue with airless floor subtypes with construction/deconstruction bugs.

SirRichardFrancis:
> Airflow occurring at all during roundstart is an issue in of itself.

Not quite. Gas simulation should not happen before initialization.
Nearby tiles having different pressure is silly, but not the real problem.
Mob AI also runs before initialization is finished, that's why we have mice with uninitialized stats stepping on traps.

TheShown911 (approved):

hyperioo:
> Airflow occurring at all during roundstart is an issue in of itself. Though, since it would of course require someone to go through maps, and replace the majority of flooring with variable edited subtypes without gas present. Wouldn't be an enviable task by any means. Especially since I vaguely recall years ago there being an issue with airless floor subtypes with construction/deconstruction bugs.

Yeah, it sounds like a painful thing to dig into

PR #7653, opened 2022-10-02
[DNM] Mob optimizations le epic part 3

MLGTASTICa:
## About The Pull Request
Reworks how mobs are looked up and triggered
TODO:
(Done) add a chunk system to actually not check the entire z-level for the mobs themselves, and also not iterate through thousands of items / etc
Add a component for triggering mobs every X tiles
Remove sanity ticker mob-triggering
Make mobs not process at all under certain circumstances.

## Changelog
:cl:
add: Added new things
del: Removed old things
tweak: tweaked a few things
balance: rebalanced something
fix: fixed a few things
soundadd: added a new sound thingy
sounddel: removed an old sound thingy
imageadd: added some icons and images
imagedel: deleted some icons and images
spellcheck: fixed a few typos
code: changed some code
refactor: refactored some code
config: changed some config setting
admin: messed with admin stuff
server: something server ops should know
/:cl:


SirRichardFrancis:
Finally, minecraft in SS13.
Why not use areas instead?

MLGTASTICa:
> Finally, minecraft in SS13. Why not use areas instead?

Area are single-tons on a turfs. inbuilt in byond, can't really use the without overriding areas to be 8x8
(Their contents are also not filtered and theres a whole lot of objects to check in the first place)

SirRichardFrancis:
> Area are single-tons on a turfs. inbuilt in byond, can't really use the without overriding areas to be 8x8 (Their contents are also not filtered and theres a whole lot of objects to check in the first place)

Why do you need "chunks" to be 8x8 exactly? Mobs keeping track of their area, might as well let areas keep track of mobs.

MLGTASTICa:
> > Area are single-tons on a turfs. inbuilt in byond, can't really use the without overriding areas to be 8x8 (Their contents are also not filtered and theres a whole lot of objects to check in the first place)
> 
> Why do you need "chunks" to be 8x8 exactly? Mobs keeping track of their area, might as well let areas keep track of mobs.

i don't need chunks to be 8x8 exactly , but its a number that keeps the mob count small for the map size we currently employ.
(Also , if you check areas , areas include EVERYTHING in the turfs they own, because areas can only own turfs)

SirRichardFrancis:
> i don't need chunks to be 8x8 exactly , but its a number that keeps the mob count small(per chunk) for the map size we currently employ. (Also , if you check areas , areas include EVERYTHING in the turfs they own, because areas can only own turfs)

That doesn't matter, mobs already keep track of their area, just let them record their presence in the area datum.
Similar, if not the same, thing would have to be done for chunks, but why add a layer of complexity?
Use existing stuff, it's fit for the job.

SaltyOldBull:
this PR will affect the local trout population

MLGTASTICa:
Too much of the mob framework is far too poorly handled for this to be viable for now.

PR #7657, opened 2022-10-04
Adjustments to maint generated oddities

DiaFRAME444:
## About The Pull Request
- Many oddities now have different stat layouts... (Single-statters now exist for every stat, photo oddities were reworked to have VIG-focused layout, dual stats now cover every ROB-TGH-VIG combo, and so on)
- Oddities now roll from 2 to Max, instead of 1 to max.

This is a watered down version of what I wanted to do but oh well.

## Why It's Good For The Game
Maint oddities are now more useful in general. People can actually focus down a specific stat layout instead of being given maybe "ROB/TGH/VIG all in combat" or "BIO/COG/MEC all in work".

## Changelog
:cl:
balance: Oddities in maint have slightly more diverse stat layouts. You may just find one that better fitting for your intended job or build.
balance: Maint oddities will no longer roll 1s on their RNG stat rolls.
/:cl:

### DIFF
--- a/code/game/objects/items/oddities.dm
+++ b/code/game/objects/items/oddities.dm
@@ -37,7 +37,7 @@
 	if(oddity_stats)
 		if(random_stats)
 			for(var/stat in oddity_stats)
-				oddity_stats[stat] = rand(1, oddity_stats[stat])
+				oddity_stats[stat] = rand(2, oddity_stats[stat])
 		AddComponent(/datum/component/inspiration, oddity_stats, perk)
 
 /proc/get_oddity_perk()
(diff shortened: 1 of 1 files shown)


Firefox13 (commented):
Long ago stats minimum on oddities was indeed 2.
I don't really dig the seemingly random stats changing though, but it looks okay

DiaFRAME444:
> Long ago stats minimum on oddities was indeed 2.

Weh. Getting +1 to a stat each level up is a pretty much unnoticeable effect, so I can't figure out why that was changed.

> I don't really dig the seemingly random stats changing

There was some thought put into it, even if it was mostly "There's 4 oddities with this exact same stat layout, I'm gonna recycle them into something else", combined with a few moves so oddity stats somewhat fit their item (Syndicate Lighter and Old ID still make sense as ROB and TGH oddities, but photo doesn't, so things get shifted around to make photo the VIG oddity)

hyperioo (approved):

PR #7662, opened 2022-10-05
RIG armor buff

Humonitarian:
## About The Pull Request

Buffs all RIG armor across the board, rescue suit RIG has less slowdown

Further buffs might be necessary, but due to how overly effective we have made AP, it walks a fine line between being worthless against a Tosshin and being indestructible against an atreides.

## Why It's Good For The Game

Rework made RIGs not as effective as they should be, this alleviates it

## Changelog
:cl:
 balance: RIG armor values increased
 balance: Rescue Suit RIG slowdown decreased
/:cl:

### DIFF
--- a/code/modules/clothing/spacesuits/rig/rig.dm
+++ b/code/modules/clothing/spacesuits/rig/rig.dm
@@ -30,9 +30,9 @@
 
 	// These values are passed on to all component pieces.
 	armor = list(
-		melee = 7,
-		bullet = 5,
-		energy = 5,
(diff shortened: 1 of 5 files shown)


Rexdyer1:
Cool. Quick question, though, I've noticed that the blood red rig suit is much worse than the hazard and combat rigsuit. the combat one makes sense, but having it be worse in two functions compared to a hazard security one confuses me.

Pink-Chink (approved):

PR #7668, opened 2022-10-07
Certain scrap ammo is no longer superior to normal ammo

DiaFRAME444:
## About The Pull Request

Fixes a error that was discussed weeks ago and then nobody got up to fix.

## Why It's Good For The Game
Scrap ammo is supposed to be worse than normal ammo, but easier to produce. No idea what happened here.

## Changelog
:cl:
balance: Fixed two cases where scrap ammo was superior to normal ammo.
/:cl:


TheShown911 (approved):

Humonitarian (approved):
I didn't even notice these

PR #7675, opened 2022-10-07
Add check for division by 0 in block_damage

hyperioo:
## About The Pull Request

Avoid division by 0 when there is just no damage to block.

```
[20:50:02] Runtime in projectile.dm,638: Division by zero 
```

## Changelog
:cl: Hyperio
add: Add check for division by 0 in block_damage
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -635,10 +635,7 @@
 		on_impact(A)
 		qdel(A)
 
-	if(!dmg_total)
-		return 0
-	else
-		return dmg_remaining / dmg_total
+	return dmg_total ? (dmg_remaining / dmg_total) : 0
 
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/projectile.dm
```diff
@@ -635,7 +635,10 @@
 		on_impact(A)
 		qdel(A)
 
-	return dmg_remaining / dmg_total
+	if(!dmg_total)
+		return 0
+	else
+		return dmg_remaining / dmg_total
```
SirRichardFrancis, line 641:
```suggestion
	return dmg_total ? (dmg_remaining / dmg_total) : 0
```

hyperioo:
Thanks for the suggestion, I sometimes forget the ternary operator

PR #7682, opened 2022-10-10
Resprite shoes

Firefox13:
## About The Pull Request

Resprites most **shoes** on humans, doesn't touches item states

## Why It's Good For The Game

They fit better in our color palette

shoening2
(upper line in first image are chaincuffs, you can use handcuffs on orange shoes to make people walk slower)
shoening1
All shoes except rainbow I've switched it from red and green (they aren't even near in rainbow) to red and orange

## Changelog
:cl:
imageadd: new onmob shoe sprites
/:cl:


Firefox13:
Brown shoes are probably way too dark I'll fix it later

Firefox13:
изображение

hyperioo (approved):
Ok codewise (spoiler: there is no code)

Kurgis:
Nice. Red shoes could have more brightness, right now they fit better with dark red clothing. I know it aligns with its previous level of brightness, but it could really use it I think. Good to see the shoes coming into line with uniforms instead of only just reinforced shoes.

Firefox13:
изображение
I've made red shoes a bit lighter, I've also resprited two different unused boots (they got caught up in a crossfire)
изображение
"tourist"
изображение
"mime" which is just white shoes resprite

PR #7602, opened 2022-09-26
Knuckle dusters

ErisMeleeEnjoyer:
## About The Pull Request
adds six variants of knuckle dusters - gloves that increase your punch damage when worn

steel knuckle dusters - increase damage by 5, made from steel

silver knuckle dusters - increase damage by 5, made from silver, slightly increase your style

plasteel knuckle dusters - increase damage by 10, made from plasteel

golden knuckle dusters - increase damage by 10, made from plasteel and gold, increase your style by 2

platinum knuckle dusters - increase damage by 15, made from platinum and spike toolmod(plasteel), increase your style by 2

weighted knuckle gloves - not really knuckle dusters, but instead knuckle gloves from drip vendor with plasteel powder in them; increase damage by 10, made from plasteel, style is the same as knuckle gloves, sprite is also the same, as well as the name - inconspicious thing, isn't it?

image

all of the above can be crafted in 'clothing' category of craft menu.

all of these use a new human define - punch_damage_increase, which can be used in various implants, bionics and mutations in future contributions.

## Why It's Good For The Game

drip. punch. stronk.

How are these balanced? Quite simple - hard-to-get materials, as well as no armor(except knuckle gloves, but you have to pay cash money) or armor penetration whatsoever. Might fuck over style users, but an operative that is covered in BP/riot armor? Good luck. You will get shot.

## Changelog
:cl: Kegdo
add: knuckle dusters
add: weighted knuckle gloves
/:cl:

PR #7683, opened 2022-10-10
Make golden knuckles use plasteel instead of platinum

Firefox13:
## About The Pull Request

#7602 desc mentions that they are made out of plasteel and gold
besides you can make vastly better knuckle dusters with just platinum
изображение
изображение

## Why It's Good For The Game

bugfix

## Changelog
:cl:
balance: Golden knuckle dusters now require plasteel+gold instead of platinum+gold
/:cl:


ErisMeleeEnjoyer:
there are two more oversights I've forgotten about. I'll make a fix in a minute.

PR #7630, opened 2022-09-29
[BOUNTY] GYM

ErisMeleeEnjoyer:
Fuck you, leather man

## About The Pull Request

- [x] Insight stacking removal, rest points 
Removed insight stacking, added an option to gain a rest point when rest finishes, individual objectives tweaked accordingly.
Copied #6895(doubled rewards for resting, basically)

- [x] Dumbbells
Added dumbbells. Pure fluff, it's just a hammer. Yeah, laser-guided dumbbells.

- [x] Cooldown for structures(perks)
Added the "Overextertion" and "Dimmed reason" perks that work as cooldowns for rest point using structures. Overextertion reduces TGH by 15 for 3 minutes, Dimmed reason reduces COG by 15 for 3 minutes. While these perks are active, the structures cannot be used.

- [x] Protein shakes
Added the regular and commercial protein shakes. Regular is created by mixing milk and protein, commercial is bought-only in the gym vendor. Regular protein shake reduces overextertion duration by 3/second, commercials reduces it by 2/second.

- [x] Structures
Added the gym structures that use rest points for permanent stat gain, or just give temporary stats. Added gym vendor. Added gym tickets.

- [x] Sprites
I luv infrared baron and his sprites

- [x] Mapping
Big thanks to @Firefox13 for mapping.

image

image

## Why It's Good For The Game

I am the boss of this gym, and I get paid to make one a TRUE gym

## Changelog
:cl: Kegdo, INFRARED_BARON's sprites, Firefox13's mapping
del: Holodeck
add: DA GYM
add: rest points
add: dumbbells
del: insight stacking
/:cl:

PR #7689, opened 2022-10-11
Gym fixening

ErisMeleeEnjoyer:
#7630 with fixes
also fixed finish_rest a bit so that the menu won't pop up in an instant in case you are in danger(which is a skill issue tbh)


ErisMeleeEnjoyer:
I did check this on local(i fucking hate testing sanity code), and it seems to be working; however, since this touches sanity, I recommend putting this on test, just to be sure that I didn't miss anything.

SirRichardFrancis:
Wanna note that #7630 was reverted because it was suspected in making server lag.
Server continued to lag like crazy after this revert, however.

ErisMeleeEnjoyer:
> Wanna note that #7630 was reverted because it was suspected in making server lag. Server continued to lag like crazy after this revert, however.

it was reverted because it had runtimes and critical bugs, not lags. Machinery is responsible for most of unnedeed processing and therefore lag

ErisMeleeEnjoyer:
Applied fixes to bugs that were found during testing

TheShown911 (approved):
Should be good if there weren't problems on the test server.

PR #7692, opened 2022-10-11
Drink effects pt1

null-Aurelian:
## About The Pull Request

Adds various effects to alcoholic drinks. Benefits are supposed to be thematic with the descriptions or names of the drinks. Effects are intended to be narrow in scope or require specific triggers so not to overlap too strongly with normal stims.

This is a draft to get feedback on ideas and to poll for other drink effects.

## Why It's Good For The Game

Makes it so that club has more reason to make difficult drinks and for patrons to order drinks specific to their perks and context, even if they do not have a desire for those drink flavors.

## Changelog
:cl:
change: Devil's Kiss drink grants a speed boost if body temperature is above 350K
change: Demon's Blood drink grants a 10% burn damage resist if body temperature is above 350K
change: Aloe grants double sanity gain on ingest if the drinker has burn damage
change: Amasec grants +3 Toughness if the drinker has the Gunslinger perk
/:cl:


Firefox13:
Sbiten/Moonshine could give TGH bonus to technomancers/alcoholics
Toxins Specials could give COG bonus for scientists cuz' bomb
Beepsky Smash maybe should not affect IH personnel and buff them too

Just don't add so it also checks antags because muh valid check

null-Aurelian:
I can do that, but what is the reason for Sbiten? As far as I know it doesn't have any theming for alcoholics or technomancers.
And yeah valid checks would be a dumb effect.

null-Aurelian:
If that is the case I'll probably tie Hooch and Moonshine to the Alcoholic perk
Technomancers should probably get some of the more interesting drinks

code/modules/reagents/reagents/food-Drinks.dm
```diff
@@ -1878,6 +1888,11 @@
 	glass_center_of_mass = list("x"=16, "y"=2)
 	taste_tag = list(TASTE_SPICY,TASTE_DRY)
 
+/datum/reagent/alcohol/demonsblood/affect_ingest(mob/living/carbon/M, alien, effect_multiplier)
+	. = ..()
+	if(M.bodytemperature > 350)
```
SirRichardFrancis, line 1893: 77°C body temperature? Lethal temperature is 44°C.
38-39°C is severe discomfort, could use that instead.
```suggestion
	if(M.bodytemperature > 311)
```

code/modules/reagents/reagents/food-Drinks.dm
```diff
@@ -1894,6 +1909,11 @@
 	glass_center_of_mass = list("x"=16, "y"=8)
 	taste_tag = list(TASTE_BITTER,TASTE_DRY,TASTE_SLIMEY)
 
+/datum/reagent/alcohol/devilskiss/affect_ingest(mob/living/carbon/M, alien, effect_multiplier)
+	. = ..()
+	if(M.bodytemperature > 350) //Sbiten or on fire because devil
+		M.add_chemical_effect(CE_SPEEDBOOST, 0.1) //No actual downsides outside massive body temp so the effect is weak, 1/6 of hyperzine
```
SirRichardFrancis, line 1915:
```suggestion
	if(M.on_fire)
		M.add_chemical_effect(CE_SPEEDBOOST, 0.1) //No actual downsides outside massive body temp so the effect is weak, 1/6 of hyperzine
```

SirRichardFrancis (changes requested):

code/modules/reagents/reagents/food-Drinks.dm
null-Aurelian: Trigger temperature is now 311, though I would like to point out most of the warming drinks will be able to trigger this

code/modules/reagents/reagents/food-Drinks.dm
null-Aurelian: Thanks, I didn't know this was a function we had!

code/modules/reagents/reagents/food-Drinks.dm
SirRichardFrancis: Current body temp handling is super ancient and poorly done, increasing temperature to crazy numbers sometimes, and in general making no sense.
If anything, we should improve that system, not adjust other features to it's nonsensical behavior.

SirRichardFrancis (approved):

code/modules/reagents/reagents/food-Drinks.dm
```diff
@@ -1603,7 +1603,7 @@
 /datum/reagent/alcohol/amasec/affect_ingest(mob/living/carbon/M, alien, effect_multiplier)
 	. = ..()
 	if(M.stats.getPerk(PERK_GUNSLINGER))
-		M.stats.addTempStat(STAT_TGH, 3, STIM_TIME, "Amasec")
+		M.stats.addTempStat(STAT_TGH, STAT_LEVEL_BASIC*effect_multiplier, STIM_TIME, "Amasec") //Improves Toughness by ~15 if Gunslinger
```
SirRichardFrancis, line 1606:
```suggestion
		M.stats.addTempStat(STAT_TGH, STAT_LEVEL_BASIC, STIM_TIME, "Amasec") //Improves Toughness by 15 if Gunslinger
```
Effect multiplier does silly math all the time and varies a lot. Let's not have 3.12561561 or 43.1555 stat buffs from this.
Solid and stable 15 is good.

code/modules/reagents/reagents/food-Drinks.dm
```diff
@@ -2090,6 +2090,11 @@
 	glass_desc = "You've really hit rock bottom now... your liver packed its bags and left last night."
 	taste_tag = list(TASTE_BITTER,TASTE_DRY)
 
+/datum/reagent/alcohol/hooch/affect_ingest(mob/living/carbon/M, alien, effect_multiplier)
+	. = ..()
+	if(M.stats.getPerk(PERK_ALCOHOLIC))
+		M.add_chemical_effect(CE_PAINKILLER, 15) //Acts as Inaprovaline painkiller if alcoholic
```
SirRichardFrancis, line 2099: Alcohol already provides painkiller effect, depending on it's strength.
With formula of `max(35 - (strength / 2), 1))` and strength of hooch of 1, we'll get 34.5 painkiller effect.
This adds 15 on top, to a whooping 49.4, same as tramadol.

SirRichardFrancis (changes requested):

code/modules/reagents/reagents/food-Drinks.dm
```diff
@@ -1733,7 +1743,10 @@
 
 /datum/reagent/alcohol/beepsky_smash/affect_ingest(mob/living/carbon/M, alien, effect_multiplier)
 	..()
-	M.Stun(2)
+	if(M.stats.getPerk(PERK_SURVIVOR))
+		M.stats.addTempStat(STAT_ROB, STAT_LEVEL_BASIC * effect_multiplier, STIM_TIME, "BEEPSKY SMASH") //SMASH
```
SirRichardFrancis, line 1747:
```suggestion
		M.stats.addTempStat(STAT_ROB, STAT_LEVEL_BASIC, STIM_TIME, "BEEPSKY SMASH")
```

code/modules/reagents/reagents/food-Drinks.dm
SirRichardFrancis:
```suggestion
		M.stats.addTempStat(STAT_TGH, STAT_LEVEL_BASIC, STIM_TIME, "Amasec")
```
Also, please don't use value of constant in a comment, those may change, and nobody will go trough every comment to update values. Might cause some confusion later down the line.
This code is descriptive enough.

null-Aurelian: Fixed to STAT_LEVEL_BASIC

code/modules/reagents/reagents/food-Drinks.dm
null-Aurelian: Fixed to STAT_LEVEL_BASIC

code/modules/reagents/reagents/food-Drinks.dm
null-Aurelian: I see, I didn't realize that. Changed effect to heal OxyLoss at -0.3, half strength of Tricord. I didn't have any other ideas for what the effect could be, so this is kinda a placeholder until someone gets an actually good idea.

code/modules/reagents/reagents/food-Drinks.dm
SirRichardFrancis: Point about referencing other values in comments still stands.
Values are changing often, thanks to contributors very concerned with game balance.
It makes sense to reference other chemicals with similar effects in a PR, as a justification for changes made, but rarely (actually never) needed in code comments. Capiche?

null-Aurelian: okay, will stop adding in comments.

Deal5:
Good stuff: I like the gameyfying drinks so people can actually think what they want to drink, IH preferring beepskies when choosing makes good impression for a Bartender that will make it for them, and people who are looking at this interaction

Neutral: Not all drinks seem to have effects, looking at changes. It's a good step forward towards arcade shit, still.

Complain: Potential issue - battle drinks. If people will start gulping a drink for a battle instead of stims, this may and probably will look weird and will solidify sweating ass playstyle to kill someone. 

Potential but unlikely solution - long-lasting effects (+optionally not OP effects, so the drinks are the pleasant, but that will not be your primary focus in game), so people don't have an incentive to save a drink for emergency moments and instead drink it NOW, so for the next 10 minutes they don't need to worry about anything and won't need to sweat their ass

More likely solution - make drinks turn into shit liquid if noone drinks them for [x] minutes, it was fresh, now it isn't, better drink it NOW.

PR #7696, opened 2022-10-11
Seize means of fatfingering

Firefox13:
## About The Pull Request

Adds a confirmation panel when you press "Summon stash" button
I've tried also adding a notification that someone summoned it but I couldn't fucking figure it out, I'll appreciate help with that
## Why It's Good For The Game

People won't accidentally summon stash anymore.

## Changelog
:cl: Firefox13, SirRichardFrancis
add: Added confirmation panel when summoning excelsior stash
add: Added notification to all excelsior members when stash is summoned
/:cl:

PR #7710, opened 2022-10-15
Revert "Seize means of fatfingering"

Firefox13:
Reverts discordia-space/CEV-Eris#7696

For some reason it works fine on local but not on live server.


Firefox13:
I'll fix it
Later
:tm:

SirRichardFrancis:
Think guy who reported stash missing just messed something up. Code of spawning stash is the same.

Firefox13:
I just tested it on live. It does not work.

ValoTheValo:
> Think guy who reported stash missing just messed something up. Code of spawning stash is the same.

it doesnt. about 3 more missing stash spawns happened since then. not one successful and fox tested it on live and it didn't work

PR #7711, opened 2022-10-16
Block 1583 BYOND build from connecting

Firefox13:
## About The Pull Request

so yeah this version has (popular?) cheats and it's not the latest one so it's better to be on the safe side

## Why It's Good For The Game

???

## Changelog
:cl:
config: 1583 BYOND build blocked from connecting due to security issues
/:cl:


Alliostra:
Literally 1583

PR #7713, opened 2022-10-17
Psychiatrist Stop-Gap Changes

Kurgis:
## About The Pull Request
Psychiatrist's Office is guaranteed to always spawn a bottle of whiskey for making Mindwipe, and they also now start with a vial of Psilocybin on their person. Psychiatrist may now access Xenoflora and Medical Equipment, both things they need in order to produce sanity-gain chems. They begin with ten more Vigilance so they aren't busy breaking down too much in order to help anyone.

zrestaadw

## Why It's Good For The Game
They can at least pursue a bit more options from just their office.

## Changelog
:cl:
tweak: Map changes to Psychiatrist's Office, guaranteed to always spawn a bottle of whiskey for making Mindwipe.
balance: Psychiatrist may now access Xenoflora and Medical Equipment, they start with a vial of Psilocybin on their person, and they begin with ten more Vigilance.
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/glass/beaker.dm
+++ b/code/modules/reagents/reagent_containers/glass/beaker.dm
@@ -40,14 +40,14 @@
 		var/mutable_appearance/label = mutable_appearance(icon, label_icon)
 		add_overlay(label)
 
+//// Subtypes ////
+
 /obj/item/reagent_containers/glass/beaker/cryoxadone
 	preloaded_reagents = list("cryoxadone" = 30)
 	spawn_blacklisted = TRUE
 
 /obj/item/reagent_containers/glass/beaker/sulphuric
 	preloaded_reagents = list("sacid" = 60)
-
-//// Subtypes ////
 
 /obj/item/reagent_containers/glass/beaker/large
 	name = "large beaker"
--- a/code/modules/reagents/reagent_containers/glass/beaker.dm
+++ b/code/modules/reagents/reagent_containers/glass/beaker.dm
@@ -40,6 +40,12 @@
 		var/mutable_appearance/label = mutable_appearance(icon, label_icon)
 		add_overlay(label)
 
+/obj/item/reagent_containers/glass/beaker/cryoxadone
+	preloaded_reagents = list("cryoxadone" = 30)
+	spawn_blacklisted = TRUE
+
+/obj/item/reagent_containers/glass/beaker/sulphuric
+	preloaded_reagents = list("sacid" = 60)
 
 //// Subtypes ////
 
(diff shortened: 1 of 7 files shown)


code/modules/reagents/reagent_containers/glass/beaker.dm
```diff
@@ -40,6 +40,12 @@
 		var/mutable_appearance/label = mutable_appearance(icon, label_icon)
 		add_overlay(label)
 
+/obj/item/reagent_containers/glass/beaker/cryoxadone
+	preloaded_reagents = list("cryoxadone" = 30)
+	spawn_blacklisted = TRUE
+
+/obj/item/reagent_containers/glass/beaker/sulphuric
+	preloaded_reagents = list("sacid" = 60)
```
SirRichardFrancis, line 48: This probably was moved up here by accident, subtypes belong under `//// Subtypes ////` comment, I believe.

SirRichardFrancis (changes requested):
Didn't notice initially that psychiatrist is now a scientist instead of a doctor, been changing item paths and stuff... Anyway, would appreciate screenshots of map changes.

code/modules/reagents/reagent_containers/glass/beaker.dm
SirRichardFrancis: And there are a lot of "objects with only changed contents" and they are all "somewhere right under their parent", but as you may see, they are also under `//// Subtypes ////` section, with actually distinct beaker types being on top, with "reagent subtypes" being under those.
I don't see why cryoxadone and sacid beakers should be an exception.

Kurgis:
> Suggestion that should be fairly easy to do and would make a good bit of flavorful sense: A job perk, "calming presence" or something similar that restores sanity around the shrink?

This is a good suggestion, perhaps just giving the Psychiatrist the Calming Presence oddity perk at round-start.

Firefox13:
This would be a job perk and it shouldn't be on oddities.

Firefox13:
Perhaps something more unique, such as increased insight gain rate for people around if they have high sanity?

Pink-Chink (approved):

code/modules/reagents/reagent_containers/glass/beaker.dm
TheShown911: Agreed.

Kurgis:
Ready for review.

code/modules/reagents/reagent_containers/glass/beaker.dm
Kurgis: Made requested changes.

PR #7718, opened 2022-10-19
Nanite timer fix

Pink-Chink:
## About The Pull Request

Nanites would gib when their parent Kraftwerk roach would die, but they would be qdel'd before the gib() proc finishes. I changed it to death() since it already leaves behind an oil spill. Plus, the gib() proc only makes blood splatters.

## Why It's Good For The Game

Bug fix

## Changelog
:cl:
fix: Fixed nanite death via parent roach death
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/superior_animal/roach/types/nanites.dm
+++ b/code/modules/mob/living/carbon/superior_animal/roach/types/nanites.dm
@@ -97,12 +97,13 @@
 	..()
 	parent = nuparent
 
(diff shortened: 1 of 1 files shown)


SirRichardFrancis (approved):

SirRichardFrancis:
Review speedrun, two lines of code edition.

Pink-Chink:
Any% WR

TheShown911 (approved):

PR #7720, opened 2022-10-19
Buffs NT Faith Shield cruciform upgrade

DiaFRAME444:
## About The Pull Request
Buffs Faith's Shield to 20% brute and burn protection.

## Why It's Good For The Game
The Cruciform IS competing with Moebius mutations and artificial organs, or robotic limbs. And that free bicardine mutation is really, really strong. 
I think free Bicardine+Tricord is still better than Faith's Shield actually, but this covers fully burn damage and the aforementioned combo only less-than-half-covers it, so lets call it even.

I could be wrong but it was really hard to feel the difference in-game, even at +100 TGH (And thus 200 HP instead of 100~)

## Changelog
:cl:
balance: NT Faith's Shield damage reduction just got stronger.
/:cl:

### DIFF
--- a/code/modules/core_implant/cruciform/upgrades.dm
+++ b/code/modules/core_implant/cruciform/upgrades.dm
@@ -106,7 +106,7 @@
 	desc = "This upgrade will slightly increase follower resistance to physical and burn injuries from any source."
 	icon_state = "faiths_shield"
 	matter = list(MATERIAL_BIOMATTER = 50, MATERIAL_GOLD = 5, MATERIAL_PLASTEEL = 10)
-	var/shield_mod = 0.1
+	var/shield_mod = 0.2
 
 /obj/item/cruciform_upgrade/faiths_shield/OnInstall(var/disciple, var/_cruciform)
 	..()


Firefox13:
(almost) All mutations have a price (NSA) whereas cruciform upgrade don't have a comparable one.

garrulousTeamotei:
> The main advantage of Neotheology is literally being able to be cloned. That's the big one.

And that cloning will cause clone damage if not done by an antag inquisitor or the Preacher. Clone damage healable by who? Moebius.

And MoeMed has the resusicator. Which can be used as long as the target has been dead for less than 15 minutes, not including with the use of the craftable stasis bag. And all expedited by Moebius' suit sensor system.

Firefox13:
> And that cloning will cause clone damage if not done by an antag inquisitor or the Preacher. 

Wow that's fucked up
let's fix this

Firefox13:
#7721

DiaFRAME444:
Look, I'm just saying that if someone at +100 TGH has 200 effective HP, then Faith's Shield will upgrade that to 220 HP. Now 240.

Might be underbuffed actually but whatever.

TheShown911 (approved):
I think NT needs more power overall and 20% damage reduction is not that much considering all robotic limbs already have it (or did have it, before the 1000 armor reworks).

PR #7721, opened 2022-10-19
Allow acolytes to clone without genetic damage

Firefox13:
## About The Pull Request

makes acolyte clone people without genetic damage
I have **NO** fucking idea why it's `is_inquisidor` blame russian coders
## Why It's Good For The Game

I find it really fucked up.

## Changelog
:cl:
balance: Acolytes now can clone people without genetic damage
/:cl:

### DIFF
--- a/code/modules/core_implant/cruciform/rituals/machinery.dm
+++ b/code/modules/core_implant/cruciform/rituals/machinery.dm
@@ -35,10 +35,8 @@
 
 	if(pod.start())
 		var/damage_modifier = 1
-		if(is_inquisidor(user) || is_preacher(user))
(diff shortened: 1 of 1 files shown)


Iglpigl:
Firstly there is absolutely no reason for this to be made, secondly it will screw up NT and make them actually do something expect clean and grow plants, thirdly, one and possibly more cloning changes are direct violation to contributing guidelines, you should try reading them.
https://wiki.cev-eris.com/Content_GuidelinesEn

Firefox13:
A different solution to this problem would be more ways to fix genetic damage, as currently there are only very few of them:
borer
cryoxadone/clonexadone
rezadone

Perhaps when we add more ways we'll return to this issue

TheShown911 (approved):
I think this is fine, cloning is CBT enough as is.

PR #7685, opened 2022-10-11
adds tgui (now real!)

LetterN:
## About The Pull Request
Adds tgui (properly) this time
**YOU CAN IGNORE `tgui/` FOLDER WHEN REVIEWING**

image

## Why It's Good For The Game
mmm fast ui yes

## Changelog
:cl:
code: Add tgui, change changelog view to tgui
code: add verb_manager so that verbs wont cause overtimes
/:cl:

PR #7728, opened 2022-10-21
Revert "adds tgui (now real!)"

TheShown911:
Reverts discordia-space/CEV-Eris#7685

Our only hope.


DiaFRAME444:
Darkness falls...

PR #7748, opened 2022-10-31
Mob qdel fix

Pink-Chink:
## About The Pull Request

Cleans up the parallax and HUD in mob Destroy().

## Why It's Good For The Game

Performance

## Changelog
:cl:
fix: Cleans up the parallax in mob Destroy()
/:cl:

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -4,12 +4,12 @@
 	GLOB.living_mob_list -= src
 	GLOB.mob_list -= src
 	unset_machine()
-	qdel(hud_used)
+	QDEL_NULL(hud_used)
+	QDEL_NULL(parallax)
 	if(client)
 		for(var/atom/movable/AM in client.screen)
 			qdel(AM)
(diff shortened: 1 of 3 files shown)


code/modules/mob/mob.dm
```diff
@@ -4,12 +4,12 @@
 	GLOB.living_mob_list -= src
 	GLOB.mob_list -= src
 	unset_machine()
-	qdel(hud_used)
+	QDEL_NULL(hud_used)
+	QDEL_NULL(parallax)
```
SirRichardFrancis, line 8: Wait a moment. Parallax is only created on client login, right? So if player's body is destroyed at the moment of their death, they will lose parallax until reconnect?

Pink-Chink: Hmm. I have a local test up right now. The parallax does contain a reference to the parent mob, which is causing a hard del.

Pink-Chink: I didn't lose the parallax in testing.

PR #7749, opened 2022-11-01
Make TTS receive auth keys automatically

SirRichardFrancis:
## About The Pull Request

Apparently bearer tokens are short-lived, so we'll request new one each round from login-password hash.
I can't tell how long hash will last though.
Do not merge until changes to config have been made.

## Why It's Good For The Game

Less round without TTS. Gray wouldn't have to change config often.

## Changelog
:cl:
code: Removed TTS maintenance busywork.
/:cl:

### DIFF
--- a/code/controllers/configuration.dm
+++ b/code/controllers/configuration.dm
@@ -223,7 +223,7 @@ GLOBAL_LIST_EMPTY(storyteller_cache)
 	var/webhook_url
 	var/webhook_key
 
-	var/tts_bearer // Token we use to talk with text-to-speech service
+	var/tts_key // Login and password that we use to generate tts_bearer
 	var/tts_enabled // Global switch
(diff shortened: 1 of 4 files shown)


GrayRachnid:
We only have to change it once a month(after payment goes through) is there a need to look it up every month? Would it lead to any annoying side effects?

SirRichardFrancis:
> We only have to change it once a month(after payment goes through) is there a need to look it up every round?

It could be done some other way, like only requesting new token if old one is not working anymore, but that's approach that requires extra work and that I can not really test properly, since there is third-party server involved.
I certainly wouldn't bother doing that if it turns out you're cool with manually editing the config.

> Would it lead to any annoying side effects?

There is no side effects I'm aware of, but I obviously can't guarantee anything, what NovelAI will or will not do is completely out of my control.

PR #7762, opened 2022-11-04
New Drugs v0.5

Pink-Chink:
## About The Pull Request

Adds crystal dream (roach meth) and paroin (spider heroin) along with their convoluted recipes. While they can be made with maint chems, Guild can bypass part of their recipe via the new industrial chemical supply trade station, AGTB Arctor (replaces the oddity trade station). These have constant metabolism enabled, which means they shouldn't get stuck in your system like most chems. Wiki info on recipes forthcoming. New perks are weak and barebones at the moment, but there will be content related to them in the future.

Crystal Dream effects:
- Temporarily lose 15 MEC, 70 TGH, 55 ROB, and 15 VIG
- Painkiller, jitteriness, and hallucinations
- Toxic damage
- Constant damage to heart, liver, lungs, kidneys, and blood vessels
- On overdose at 15 units, you gain a perk that grants a permanent boost to your stats: 10 MEC, 5 TGH, 5 ROB, 5 VIG

Paroin effects:
- Temporarily lose 15 VIG and 15 TGH
- Hallucinations, large slowdown, and confusion
- Toxic damage
- 1% chance each reagent tick to grow a spider egg inside a random limb
- On overdose at 15 units, you gain a perk that grants a permanent boost to your stats: 10 BIO and 5 COG

Both drugs are ingest-only. Injections do nothing.

## Why It's Good For The Game

Oddities should be junk loot.

## Changelog
:cl:
add: Added drugs Crystal Dream and paroin
add: Added BIDON presets for industrial chems and drugs
add: Added trade station Arctor
add: Added drug-related perks Dreamer and Eighth Eye
del: Removed Garbaj
/:cl:


DiaFRAME444:
> (replaces the oddity trade station)

Well

> Oddities should be junk loot

Yeah, you're right.
But right now maintenance depletes of oddities at highpop, so I feel like that should be fixed first.

githubuser4141:
I love the failure stages for the drug recipes, it adds immersion. The requirement to OD to gain positive effects besides sanity feels a bit forced imo, but I haven't used sanity gains much so 🤷 ? Maybe there could be some minor stat gains for casual use?

Pink-Chink:
> But right now maintenance depletes of oddities at highpop

Fixing junk is a process. Removing the oddity station is more about Guild having easy access to something it shouldn’t.

> The requirement to OD to gain positive effects besides sanity feels a bit forced imo

The OD requirement is extreme for what you get out of it. In the future, I’ll be adding some things that interact with the perks you gain from the drugs that will make it worth the effort and time.

DiaFRAME444:
Sudden thought: Can you end up rolling either of those perks if you're a NT Last Shelter spawn?

> Fixing junk is a process. Removing the oddity station is more about Guild having easy access to something it shouldn’t.

I still disagree with the order of operations here but now I realize I'm probably the only one that orders oddities from that station when I can't find one in maint, so I guess it doesn't matter for most of the playerbase.

Pink-Chink:
I don’t think you can. From what I can tell, you get a random oddity perk.

It’s just how it is with open-source dev.

Humonitarian:
> I don’t think you can. From what I can tell, you get a random oddity perk.
> 
> It’s just how it is with open-source dev.

You can get any perk, including faction ones like codespeak. At least, based on experience.

Voldirs:
Wow, that's a lot of reactions, I wonder how MIRC handles those and did you try to get those new chems via local because even chem heater isn't instant and that's a MESS of temperature ranges

Humonitarian:
I would give them an insight gain bonus when drank, maybe a single boost once alongside the perk.

Pink-Chink:
> Wow, that's a lot of reactions, I wonder how MIRC handles those and did you try to get those new chems via local because even chem heater isn't instant and that's a MESS of temperature ranges

The idea is that you need to be precise, even with professional tools. The chem heater makes things a lot easier as long as you use it the right way. We'll see how it looks on MIRC.

Pink-Chink:
> I would give them an insight gain bonus when drank, maybe a single boost once alongside the perk.

The sanity gain boosts insight by quite a bit. I don't actually know how it works, but I was leveling up pretty quickly.

PR #7769, opened 2022-11-06
More Relevant Salaries

BlackjackIsTired:
## About The Pull Request

Reduces the time between salaries and reduces the salary amount themselves by half.

## Why It's Good For The Game

Makes salaries more relevant in gameplay for the average crewmember.

## Changelog
:cl:
Salaries paid at 30 minutes instead of 1 hour.
Salaries slashed in half.
/:cl:

### DIFF
--- a/code/__DEFINES/economy.dm
+++ b/code/__DEFINES/economy.dm
@@ -2,8 +2,8 @@
 
 //Hourly wage defines
 #define WAGE_NONE			0	//Unpaid. Assistant, captain, and those who sell things
-#define WAGE_LABOUR_DUMB	450 //Dumb jobs anyone can do. Janitor, actor, cargotech, etc
-#define WAGE_LABOUR			600 //The standard wage that most people get
-#define WAGE_LABOUR_HAZARD	750	//Hazard pay. For miners and IHOperatives
-#define WAGE_PROFESSIONAL	900	//The wage for educated professionals. Doctors, scientists, etc
-#define WAGE_COMMAND		1200	//Wage paid to command staff, generally regardless of department
+#define WAGE_LABOUR_DUMB	225 //Dumb jobs anyone can do. Janitor, actor, cargotech, etc
+#define WAGE_LABOUR			300 //The standard wage that most people get
+#define WAGE_LABOUR_HAZARD	375	//Hazard pay. For miners and IHOperatives
+#define WAGE_PROFESSIONAL	450	//The wage for educated professionals. Doctors, scientists, etc
+#define WAGE_COMMAND		600	//Wage paid to command staff, generally regardless of department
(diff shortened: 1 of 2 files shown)


SirRichardFrancis (approved):

Humonitarian:
Make sure this does not affect starting cash, which to my knowledge is based on these wages.

SirRichardFrancis:
Hmm, that's actually correct for most jobs.

Have to change line 5 in wages.dm to correct that.

Firefox13:
Screenshot_20221108-102322_Firefox
Change this line too since it's half hourly. Or just exclude world "hourly". same goes for all other mentions of it, I can't search for them right now (phonecoding)

code/game/jobs/wages.dm
```diff
@@ -1,8 +1,8 @@
 //Determines starting account balance
 /datum/job/proc/one_time_payment(var/custom_factor = 1)
 	if (initial_balance != -1)
-		return round (initial_balance * RAND_DECIMAL(0.7, 1.3))
-	return round(wage * RAND_DECIMAL(1.5, 3.5))
+		return round (initial_balance * RAND_DECIMAL(1.4, 1.7))
+	return round(wage * rand(1.5 * 10, 1.5 * 10) / 10)
```
Pink-Chink, line 5:
```suggestion
		return round (initial_balance * rand(14, 17) / 10)
	return round(wage * rand(15, 35) / 10)
```
The proc rand() only produces integers and RAND_DECIMAL() doesn't work like we'd expect (or at all). For line 5, this is a better way to achieve what I'm assuming you intended.

Humonitarian: round(wage * rand(1.5 * 10, 1.5 * 10) / 10) should work the same as the adjusted RAND_DECIMAL

but in both cases a better idea would be actually fixing RAND_DECIMAL, and using that for both

Pink-Chink: My point was that rand(1.5 * 10, 3.5 * 10) is identical in output to rand(15, 35), but with more operations. The latter is better practice. Also, I'd rather a RAND_DECIMAL() fix be its own PR.

Humonitarian: > My point was that rand(1.5 * 10, 3.5 * 10) is identical in output to rand(15, 35), but with more operations. The latter is better practice. Also, I'd rather a RAND_DECIMAL() fix be its own PR.

👍

Pink-Chink (approved):

TheShown911:
Shouldn't starting cash calculations be doubled now?

code/game/jobs/wages.dm
```diff
@@ -1,8 +1,8 @@
 //Determines starting account balance
 /datum/job/proc/one_time_payment(var/custom_factor = 1)
 	if (initial_balance != -1)
-		return round (initial_balance * RAND_DECIMAL(0.7, 1.3))
-	return round(wage * RAND_DECIMAL(1.5, 3.5))
+		return round (initial_balance * rand(14, 17) / 10)
+	return round(wage * rand(15, 35) / 10)
```
Pink-Chink, line 5:
```suggestion
	return round(wage * rand(30, 70) / 10)
```
Doubling wage multiplier since wages are halved. Should be the same as it was pre-PR, but with the range fixed.

PR #7771, opened 2022-11-07
Happy Meal Change

Kurgis:
## About The Pull Request
Makes some changes to the McRonalds trade station, to the benefit of kitchen workers primarily.

## Why It's Good For The Game
Thou shalt not ignore the kitchen... as much.

## Changelog
:cl:
tweak: Changed the Happy Meal to include some different items, and a soda.
balance: You may no longer buy whole cakes from the McRonalds trade beacon, they instead come in smaller slices.
balance: A few cooked items such as jelly burgers and sandwiches now count as cooked for gathering insight.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/storage/boxes.dm
+++ b/code/game/objects/items/weapons/storage/boxes.dm
@@ -761,10 +761,11 @@
 /obj/item/storage/box/happy_meal/New()
 	. = ..()
 	var/list/things2spawn = list(
-		/obj/item/reagent_containers/food/snacks/sliceable/plaincake,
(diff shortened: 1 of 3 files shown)


Pink-Chink:
Should we change the trade station price of the slices from 300 to something cheaper like 60? Also, changes to cooked foods should be in a separate PR, but it's a small PR so it's fine.

DiaFRAME444:
The point is that if you buy food from this station, its gonna be way more expensive than using the club.

Pink-Chink (approved):

Rexdyer1:
300 is a bit much still. 80?

PR #7772, opened 2022-11-07
Let me clean it!

Kurgis:
## About The Pull Request
Microwaves accepts more items for cleaning. That being soap bars, and the rag.

## Why It's Good For The Game
Too often a Club Worker will hit the microwave with a bar of soap only to be frustrated, they will then print a spray bottle, go through the effort of acquiring space cleaner to fill it, and also face that it _also_ doesn't work. There's still many issues and bugged behavior with this object, but this is at least one step.

## Changelog
:cl:
tweak: You may now clean microwaves if they get dirty with soap bars or a dry/damp rag.
/:cl:

### DIFF
--- a/code/game/machinery/kitchen/microwave.dm
+++ b/code/game/machinery/kitchen/microwave.dm
@@ -109,7 +109,7 @@
 		return
 
 	else if(src.dirty==100) // The microwave is all dirty so can't be used!
-		if(istype(I, /obj/item/reagent_containers/spray/cleaner)) // If they're trying to clean it then let them
+		if(istype(I, /obj/item/reagent_containers/spray/cleaner) || istype(I, /obj/item/soap) || istype(I, /obj/item/reagent_containers/glass/rag)) // If they're trying to clean it then let them
 			user.visible_message( \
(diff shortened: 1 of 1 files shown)


code/game/machinery/kitchen/microwave.dm
```diff
@@ -109,7 +109,7 @@
 		return
 
 	else if(src.dirty==100) // The microwave is all dirty so can't be used!
-		if(istype(I, /obj/item/reagent_containers/spray/cleaner)) // If they're trying to clean it then let them
+		if(istype(I, /obj/item/reagent_containers/spray/cleaner) || istype(I, /obj/item/soap) || istype(I, /obj/item/reagent_containers/glass/rag)) // If they're trying to clean it then let them
```
Pink-Chink, line 112:
```suggestion
		if(istype(I, /obj/item/soap) || istype(I, /obj/item/reagent_containers/glass/rag)) // If they're trying to clean it then let them
```
Might as well remove the spray cleaner object check since it only works with bottles that were mapped in.

Pink-Chink (approved):
Based and bread-pilled

PR #7773, opened 2022-11-07
Burrow GC fix

Pink-Chink:
## About The Pull Request

More cleanup for burrows to reduce the number of hard deletes.

## Why It's Good For The Game

Performance

## Changelog
:cl:
fix: Burrows remove themselves from global lists on Destroy() to reduce hard dels
/:cl:


Pink-Chink:
Live testing.

PR #7774, opened 2022-11-07
/mob/living GC fix

Pink-Chink:
## About The Pull Request

Stat holder datums clean up perks, which may be a cause of human hard dels. /mob/living objects clean up their static overlay, which may be the cause of image hard dels (unsure).

## Why It's Good For The Game

Performance

## Changelog
:cl:
fix: Stat holders clean up perks on Destroy()
fix: /mob/living objects clean up their static overlays on Destroy()
/:cl:


Pink-Chink:
Live testing.

PR #7776, opened 2022-11-08
Oldificator GC fix

Pink-Chink:
## About The Pull Request

Stops oldification from removing bullets from a magazine due to the missing bullets being hard deleted. The oldficator (sic) component hold a reference to all variables on its parent object, which is bad. The component should be reworked at a later date.

## Why It's Good For The Game

Performance. This should take care of the bulk of roundstart lag.

## Changelog
:cl:
tweak: Oldified guns, magazines, and bullets clean up their oldification components on Destroy()
tweak: Oldification no longer removes bullets from magazines
/:cl:


SirRichardFrancis (approved):
// TODO: Fix later

TheShown911 (approved):

PR #7782, opened 2022-11-09
Adds vipr to golden disk

Iglpigl:
## About The Pull Request

Adds vipr to golden disk it already has a golden sprite, license  cost of 6 used like of similar guns on that disk.

## Why It's Good For The Game

Fixing a mistake I made with an earlier PR

## Changelog
:cl:
fix: vipr is not on gold paint disk
/:cl:


DiaFRAME444:
Vipr is a low quality weapon so it could potentially cost 3.

Iglpigl:
You could say the same for atreides which is also 6.

Pink-Chink (approved):
7000 credits for two Viprs isn't gonna break anything.

PR #7800, opened 2022-11-11
Rewrites embed code to make more sense

Alliostra:
## About The Pull Request

Items have an embed multiplier that makes it more/less likely than normal for that item to embed. Unfortunately, the way this was calculated caused to the problem that if a weapon used to have a negative base chance to embed (without the embed multiplier calculated in), then even having an insanely high embed multiplier would never cause the weapon to embed. This is now fixed, and for the majority of weapons, everything stays the same. Only weapons that had a non-standard embed mult now behave differently. 
Those weapons that are designed to embed such as throwing knives, the ritual dagger or the displacement dagger are now much more likely to embed. 
Weapons that had a higher embed multiplier before (saws including chainsaws, pickaxes including drills and jackhammers , the meat hook and the fire axe)  should have roughly the same chance of embedding now than they did before, but leaving the chances completely the same is impossible, as the embed chance also varies with damage and robustness.

Also, of the changes seem weird at first, keep in mind that the default embed mult has been changed from being 0.5 to being 1.

## Why It's Good For The Game

Items that were intended to embed by having high embed multipliers often could not. The embed chance calculation now works more intuitively.

## Changelog
:cl:
balance: changes to weapon embedding, so weapons that were designed to embed, but could not due to a bad formula now have a much higher chance of embedding.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/carbon_defense.dm
+++ b/code/modules/mob/living/carbon/carbon_defense.dm
@@ -43,8 +43,8 @@ true, and the mob is not yet deleted, so we need to check that as well*/
 
 		//The user's robustness stat adds to the threshold, allowing you to use more powerful weapons without embedding risk
 		embed_threshold += user.stats.getStat(STAT_ROB)
-		var/embed_chance = (damage - embed_threshold)*I.embed_mult
-		if (embed_chance > 0 && prob(embed_chance))
+		var/embed_chance = (damage*I.embed_mult - embed_threshold)/2
+		if(embed_chance > 0 && prob(embed_chance))
 			src.embed(I, hit_zone)
 
 	return TRUE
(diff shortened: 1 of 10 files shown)


code/modules/mob/living/carbon/carbon_defense.dm
```diff
@@ -43,8 +43,9 @@ true, and the mob is not yet deleted, so we need to check that as well*/
 
 		//The user's robustness stat adds to the threshold, allowing you to use more powerful weapons without embedding risk
 		embed_threshold += user.stats.getStat(STAT_ROB)
-		var/embed_chance = (damage - embed_threshold)*I.embed_mult
-		if (embed_chance > 0 && prob(embed_chance))
+		var/embed_chance = (damage*I.embed_mult - embed_threshold)/2
+		to_chat(world, "[embed_chance]")
```
SirRichardFrancis, line 47: Bruh.

Alliostra: ah, yes. Thank you.

TheShown911 (approved):

PR #7806, opened 2022-11-13
Lights no longer process

MLGTASTICa:
## About The Pull Request
Removes around 600+ objects from machinery processing

## Why It's Good For The Game
Less CPU spent on useless stuff that is already handle by APCs

## Changelog
:cl:
code: Lights no longer double their power draw.
/:cl:

### DIFF
--- a/code/modules/power/lighting.dm
+++ b/code/modules/power/lighting.dm
@@ -636,20 +636,6 @@
 				broken()
 	return
 
-//blob effect
-
-
-// timed process
-// use power
-
-#define LIGHTING_POWER_FACTOR 20		//20W per unit luminosity
-
-
-/obj/machinery/light/Process()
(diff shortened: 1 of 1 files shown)


SirRichardFrancis (approved):
Imo that's better be delegated to APCs, rather than entirely removed, but not a big deal.

MLGTASTICa:
> Imo that's better be delegated to APCs, rather than entirely removed, but not a big deal.

All power consumption should be handled by APCs , with their draw being modified with signals sent from machinery whenever they're on and off and with the APC keeping a queue of machinery to provide power to , with power_change being called when said power status changes ,but we are in 2022 on eris which is built on bay 🤮  code

Pink-Chink (approved):

PR #7809, opened 2022-11-14
System override timer reduction

Kurgis:
## About The Pull Request
Makes the Malf AI's System Override reasonably achievable. It should no longer be expected to take ~2 hours to self-destruct the ship.

Cost for system override reduced to help mid-round Malf AIs.

Adjusted when command announcements trigger for both lore, and the fact there's no real metagame rules.

## Why It's Good For The Game
The AI blowing up the ship finally has the needed closure of the round. The AI spends the time murdering people, only for it to sputter out into a "glorious death" scenario. Even on low-pop of ~5 players as Malf AI, in a 4 hour round, I was still unable to successfully perform override. That was years ago, and was when I realized this role was dead end garbage with little ground to explore as it currently stood. This isn't exploring anything, it's just one step to getting to the best part of a Malf AI round, it ending. This time, hopefully with an attempt at closure.

## Changelog
:cl:
balance: Malfunctioning AIs may now get closer to their goal of destroying the ship with a faster system override command, one which doesn't take two hours to complete.
/:cl:

### DIFF
--- a/code/game/gamemodes/malfunction/newmalf_ability_trees/tree_networking.dm
+++ b/code/game/gamemodes/malfunction/newmalf_ability_trees/tree_networking.dm
@@ -32,7 +32,7 @@
 
 /datum/malf_research_ability/networking/system_override
 	ability = new/datum/game_mode/malfunction/verb/system_override()
-	price = 2750
+	price = 2000
 	name = "System Override"
 
 // END RESEARCH DATUMS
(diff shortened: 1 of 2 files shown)


SirRichardFrancis:
Aren't you effectively removing this antag in the meantime? #7810

Firefox13 (approved):

PR #7813, opened 2022-11-14
Fix Jive comms

Firefox13:
## About The Pull Request
Removes NONVERBAL flag from it. As much as I can say, those two flag (NONVERBAL and SIGNLANG) shouldn't be used together and Jive is more of subtle motions rather than moaning slightly and pointing fingers (which is kind of non-verbal)

## Why It's Good For The Game
Fixes people talking on Jive through radio

## Changelog
:cl:
tweak: Removed NONVERBAL flag from Jive, in return it's much more like a sign language and people can't talk on it through comms
/:cl:

### DIFF
--- a/code/game/objects/items/devices/radio/radio.dm
+++ b/code/game/objects/items/devices/radio/radio.dm
@@ -281,6 +281,9 @@ var/global/list/default_medbay_channels = list(
 	//  Fix for permacell radios, but kinda eh about actually fixing them.
 	if(!M || !message) return 0
 
+	if(speaking && (speaking.flags & (NONVERBAL|SIGNLANG))) // so we don't speak sign language over radio
+		return 0
+
 	//  Uncommenting this. To the above comment:
 	// 	The permacell radios aren't suppose to be able to transmit, this isn't a bug and this "fix" is just making radio wires useless. -Giacom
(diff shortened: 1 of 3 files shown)


Firefox13:
I'd like to see if it fucks with TTS in any way

DiaFRAME444:
> flags = SIGNLANG | NO_STUTTER

Wait, can this be used to activate the Cruciform rituals while in pain?

Humonitarian:
> > flags = SIGNLANG | NO_STUTTER
> 
> Wait, can this be used to use Cruciform rituals while in pain?

Yeah, totally.

SirRichardFrancis (commented):
NONVERBAL flag only indicates that people are using sign language to those who don't know that language, and garbles messages over radio.
This will turn Jive speaker into telepaths, as none of their communication would be narrated with hand gestures, that's it.
I fail to see how it would prevent Jive speakers from using radio. Have you tested this?

Firefox13:
I've fixed Jive appearing on comms (FOR SURE THIS TIME).
About it becoming a pure SIGNLANG - I've also made so if language has either flag (SIGNLANG or NONVERBAL, former in our case) it will emote.

Pink-Chink (approved):
Seemed to work fine in test.

PR #4661, opened 2020-02-25
Adds a Hammer

Dentist2904:
## About The Pull Request
 It is a basic tool that was still missing

## Details

<details>
  <summary>A hammer that can be build or found and used for the tool function hammering, also includes a powered hammer</summary>

</details>

## Screenshots

<details>
  <summary>Expand</summary>

hammer

</details>

PR #7822, opened 2022-11-15
Combat tool mod for hammers.

DiaFRAME444:
## About The Pull Request
Changes stats and name of flat surface, now its a toolmod made for hammers.

## Why It's Good For The Game
Flat surface didn't add anything to the game and was barely taken. This change lets hammers shine in combat, they didn't have the appropriate toolmods to compete with swords or hypersaws before.

## Changelog
:cl:
tweak: Flat surface toolmod reworked into platsteel surface toolmod. 
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/tools/mods/mod_types.dm
+++ b/code/game/objects/items/weapons/tools/mods/mod_types.dm
@@ -578,20 +578,21 @@
 	I.prefix = "sanctified"
 
 /obj/item/tool_upgrade/augment/hammer_addon
-	name = "Flat surface"
+	name = "Platsteel surface"
 	icon_state = "hammer_addon"
-	desc = "An attachment that fits on almost everything, that gives a simple flat surface to employ the tool for hammering."
+	desc = "An hammer reinforcement made of platsteel, a hardier material for ."
 	matter = list(MATERIAL_PLASTEEL = 3, MATERIAL_STEEL = 2)
-	rarity_value = 20
+	rarity_value = 30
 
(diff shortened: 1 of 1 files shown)


Firefox13 (changes requested):
If you really want to make a separate mod that increases damage for hammers, then make one and don't remove and reuse existing one.
Call it something like "force ehnancer" while in reality it's just some weights that you attach to hammerhead to make it hurt more. Needs a sprite, too.

DiaFRAME444:
#4661
Hammers have been in the game for 2 years partially with the purpose of being weapons and still nobody has fixed them. There was a recent change to make them stronger wielded, but without tool mods they're still suboptimal.
Fix it. Get a spriter. Something.

Humonitarian:
> Yeah, I don't think the flat surface mod has EVER really been used. Adding a quality to an existing tool at a shitty level? It's just not something worth doing. Personally I think there needs to be more than one good combat mod for hammers to really make up the difference, but one is a good start.

Actually, you all are wrong. I had this scenario before - power shovel modded to hell with 900% speed and 30 precision. I added the flat surface, didn't really impact the stats. Suddenly I had a tool that was effectively a maxed out hammer, alongside the usual shovel qualities.

Flat surface is powerful, its only flaw is the lack of uses for hammers.

More useful interactions, such as breaking down bolts on airlocks. That would make it viable.

Alliostra:
I also support this the reasons stated above (unironically)

DiaFRAME444:
Darn, I forgot this thing was here and accidentally deleted the repository.

PR #7828, opened 2022-11-18
Fixes melee bullshit

Alliostra:
## About The Pull Request

"I died for your sins" my ass, fix your bugs responsibly.

~~Firstly , being on help intent no longer deals no damage to humans. Yeah.~~

~~Secondly,  for weapons with double tact, you will no longer raise your weapon if you click on items that are directly adjacent to you, unless you have harm intent on. For the non-coding folk, Items exclude turfs (floors and walls) and mobs. This should remove the brunt of non-attacking interactions, such as placing items on a table, requiring double tact to use, while practically not affecting combat.~~

~~With harm intent, you will double tact regardless of what you click on, if you want to be absolutely sure you don't even miss 1% of that precious gamer DPS. Again, this is only for weapons with double tact.~~

~~This should be a good tradeoff for making double tact not cancer to interact with (who am I kidding).~

Moved to https://github.com/discordia-space/CEV-Eris/pull/7847

~~Thirdly, the energy sword and energy blade are now bulky again, so you can't put them into your backpack, but they don't have double tact despite being bulky (the marvels of coding!). This also allows future coders to make bulky and huge weapons not require double tact, just through the switch of a single variable ;)~~

Moved to https://github.com/discordia-space/CEV-Eris/pull/7848

Also, you can now attack the floors again when wielding an item that usually would only do a wide swing attack. I also removed how hitting the floor increases your click cooldown, let's not punish players for choosing the obvious intent for fighting, yeah?

During a wide swing, the damage dealt is not the same for every tile (bet you didn't know!)
The tile in the middle of the wide swing always deals 80% of the original damage, while the tile the swing begins deals full damage, and the one where it ends deals only 60% of damage. I switched this around because it makes more sense that way.

Oh, also fixes a bug where embedding a weapon while perfoming a wide swing attack would boost the damage of the weapon permanently. The fix for this one is fucking awful (because the code that causes it is awful), and will probably cause other problems, especially in the future. This bug is caused by wide swing attacks dealing different amounts of damage depending on the tile, and it may be wiser to just remove that to fix this bug, but I leave this up to maintainers.

and when a weapon embeds during a wide swing, it will no longer attack when it should not anymore due to already being embedded.

I recommend putting this on the test server first and tell people to find out if funky stuff is going on with double tact. I did test this (presumably more than the melee rework was tested), but since this changes clicking code, you can never be too safe.

## Why It's Good For The Game

I hate that I have to explain this, but I hate more that I'm the one who has to fix this, so here we go.

No damage on help intent is simply a bruh moment. Not only will countless people, especially newbies ~~die to roaches~~ be at a huge disadvantge in their first melee fight just because they didn't know about this, but it also makes more people walk around without help intent  in order to be prepared, so no more letting people pass through you. If you accidentally decap someone with your fully modded battle axe that's a skill issue, and even if we needed a "safety" for melee weapons, it should not be tied into the intent system, which is used for fucking everything.

## Changelog
:cl:
tweak: when performing a wide strike, you will now deal less damage on the first two tiles you hit, instead of at the last two tiles you hit.
fix: you can now hit floors with wielded large items again.
fix: having a weapon embed while doing a wide swing will no longer permanently raise the weapon's damage.
fix: when a weapon embeds during a wide strike, it will no longer hit the rest of the tiles.

/:cl:

### DIFF
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -48,35 +48,34 @@ avoid code duplication. This includes items that may sometimes act as a standard
 	return A.attackby(src, user, params)
 
 //Returns TRUE if attack is to be carried out, FALSE otherwise.
-/obj/item/proc/double_tact(mob/user, atom/atom_target, adjacent)
+/obj/item/proc/double_tact(mob/user, atom/atom_target)
+	if(user.a_intent == I_HELP)//no damage on help intent, no need for raising
+		return TRUE
 	if(atom_target.loc == user)//putting stuff in your backpack, or something else on your person?
 		return TRUE //regular bags won't even be able to hold items this big, but who knows
-	if(w_class >= ITEM_SIZE_BULKY && !abstract && !istype(src, /obj/item/gun) && !no_double_tact)//grabs have colossal w_class. You can't raise something that does not exist.
-		if(!adjacent || istype(atom_target, /turf) || istype(atom_target, /mob) || user.a_intent == I_HURT)//guns have the point blank privilege
-			if(!ready)
-				user.visible_message(SPAN_DANGER("[user] raises [src]!"))
-				ready = TRUE
-				var/obj/effect/effect/melee/alert/A = new()
-				user.vis_contents += A
-				qdel(A)
-				var/unready_time = world.time + (10 SECONDS)
-				while(world.time < unready_time)
-					sleep(1)
-					if(!(ready))
-						user.vis_contents -= A
-						return FALSE
-					if(!(is_equipped()))
-						ready = FALSE
-						user.vis_contents -= A
-						return FALSE
-				user.visible_message(SPAN_NOTICE("[user] lowers \his [src]."))
-				ready = FALSE
-				user.vis_contents -= A
-				return FALSE
-			else
-				ready = FALSE
-				return TRUE
+	if(w_class >= ITEM_SIZE_BULKY && !abstract && !istype(src, /obj/item/gun))//grabs have colossal w_class. You can't raise something that does not exist.
+		if(!(ready))						//guns have the point blank privilege
+			user.visible_message(SPAN_DANGER("[user] raises [src]!"))
+			ready = TRUE
+			var/obj/effect/effect/melee/alert/A = new()
+			user.vis_contents += A
+			qdel(A)
+			var/unready_time = world.time + (10 SECONDS)
+			while(world.time < unready_time)
+				sleep(1)
+				if(!(ready))
+					user.vis_contents -= A
+					return FALSE
+				if(!(is_equipped()))
+					ready = FALSE
+					user.vis_contents -= A
+					return FALSE
+			user.visible_message(SPAN_NOTICE("[user] lowers \his [src]."))
+			ready = FALSE
+			user.vis_contents -= A
+			return FALSE
 		else
+			ready = FALSE
 			return TRUE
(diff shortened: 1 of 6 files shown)


assortedbeads:
You still kill roaches on help atm

Alliostra:
> You still kill roaches on help atm

yeah, it seems like the no damage on help was only a thing for humans. What else did I expect.

Alanii:
> You still kill roaches on help atm

Unless the mob type is nonhostile like pets, help intent does damage against mobs in general.

Unless they fucked intents.

code/_onclick/item_attack.dm
```diff
@@ -213,6 +216,13 @@ avoid code duplication. This includes items that may sometimes act as a standard
 //Area of effect attacks (swinging)
 /obj/item/proc/tileattack(mob/living/user, turf/targetarea, modifier = 1)
 	var/original_force = force
+	var/original_unwielded_force = force
+	if(!wielded)
+		return
+	if(force_wielded_multiplier)
+		original_unwielded_force = force/force_wielded_multiplier
+	else
+		original_unwielded_force = force/1.3
```
Pink-Chink, line 225:
```suggestion
	if(!wielded)
		return
	var/original_force = force
	var/original_unwielded_force = force_wielded_multiplier ? force / force_wielded_multiplier : force / 1.3
```
Is it supposed to return null if the weapon isn't wielded? If it is, you could probably clean this up further. If not, this proc may not work as intended.

code/_onclick/item_attack.dm
```diff
@@ -238,7 +248,12 @@ avoid code duplication. This includes items that may sometimes act as a standard
 		force = original_force
 		return
 	attack(target, user, user.targeted_organ)
-	force = original_force
+	if(wielded)
+		force = original_force
+	else
+		force = original_unwielded_force
```
Pink-Chink, line 254:
```suggestion
	force = wielded ? original_force : round(original_unwielded_force, 1)
```
Rounding to catch any rounding errors.

Pink-Chink:
1) Atomize, please. Wide swing change and embed bug fix are fine. See 2) for the rest.
2) If you're serious about fixing this, I think you should go hard on it. Adding more checks won't make it more stable and I can already see a few areas where we can make things simpler code-wise.

code/_onclick/item_attack.dm
Alliostra: yeah, it's supposed to check for this, so if the weapon embeds in someone on the first hit of a wide swing attack, the rest of the swing attack doesn't deal damage.

Pink-Chink: Right, if it leaves your hands it becomes unwielded. Probably fine as is in the suggestion, then.

code/game/turfs/simulated/floor_attackby.dm
```diff
@@ -38,7 +38,6 @@
 				take_damage(I.force*I.structure_damage_factor, I.damtype)
 			else
 				visible_message(SPAN_DANGER("[user] ineffectually hits [src] with [I]"))
-			user.setClickCooldown(DEFAULT_ATTACK_COOLDOWN*1.75) //This longer cooldown helps promote skill in melee combat by punishing misclicks a bit
```
Firefox13, line 41: This means you can spam-click tiles to destroy them

Firefox13 (approved):

PR #7833, opened 2022-11-20
Thrusters back-fire has a bigger range

MLGTASTICa:
## About The Pull Request
Makes the thruster output have a range between 1 - 8 depending on thrust power
Makes the thruster output be hot if the temperature is above 1000k , instead of it being only when its combustible and above 400K

## Why It's Good For The Game
Fighting near thrusters is more interesting
## Changelog
:cl:
balance: Thruster output range is now between 1-8 tiles instead of just 1
balance: Thruster output is hot if the gasses inside have above 1000K , and not if its combustible and above 400K anymore.
/:cl:


code/modules/overmap/ships/engines/thermal.dm
```diff
@@ -99,12 +99,20 @@
...
 		T.assume_air(removed)
-		new/obj/effect/engine_exhaust(T, exhaust_dir, air_contents.check_combustability() && air_contents.temperature >= PLASMA_MINIMUM_BURN_TEMPERATURE)
+		// if its 1000k hot
+		new/obj/effect/engine_exhaust(T, exhaust_dir, air_contents.temperature >= 1000, (range > 1))
+		for(var/i=1,i < range,i++)
+			T = get_step(T,exhaust_dir)
+			if(!T)
+				break
+			sleep(0.5)
+			new/obj/effect/engine_exhaust(T, exhaust_dir, air_contents.temperature >= 1000, (i != range - 1))
```
Pink-Chink, line 115: It would be better if this was a callback timer. Also, based on some reading, sleep() and spawn() will truncate decimal values.
```suggestion
			addtimer(CALLBACK(src, .proc/extend_plume, T, exhaust_dir, air_contents.temperature >= 1000, i != range - 1), i - 1, TIMER_UNIQUE)

/obj/machinery/atmospherics/unary/engine/proc/extend_plume(turf/T, exhaust_dir, is_flame, is_midsection)
	new/obj/effect/engine_exhaust(T, exhaust_dir, is_flame, is_midsection)
```

Pink-Chink:
image

Apparently the thrusters fire in sequence, but it looks neat.

MLGTASTICa:
> image
> 
> Apparently the thrusters fire in sequence, but it looks neat. My suggestion with the timers will make the plumes start at the same time.

commited ,thought timers didn't have more accuracy then sleep.

Pink-Chink:
It's not about the accuracy. I'm trying to have people use sleep and spawn less since byond procs tend to cause a lot of headaches. Timers aren't a perfect solution, but they do work. In this case, thrusters only fire when the nav buttons get pressed and I like how it makes the thruster fire all at once.

Pink-Chink (approved):
Will keep an eye on the effect this has on the timer subsystem. If my suggestion ends up being a bad call, we'll go back to your original version with sleep(0.5).

PR #7834, opened 2022-11-20
Weak windows now have FLIMSY resistance , and can no longer be dislocated by non-damaging hits

MLGTASTICa:
## About The Pull Request
Weak windows now have FLIMSY resistance
Any window that is not reinforced can no longer be dislocated if the hit does not damage it

## Why It's Good For The Game
Bug fix + slightly tougher windows so they can't be broken with a lighter

## Changelog
:cl:
balance: windows now have a resistance of 2 
fix: windows being dislocated by non-damaging hits
/:cl:


Pink-Chink (approved):
Seemed fine. Windows weren't extra beefy or anything.

PR #7835, opened 2022-11-20
Windows now take double damage from hammers , normal from any tool with digging , excavation or shoveling quality , and a quarter from everything else.

MLGTASTICa:
## About The Pull Request
Makes windows take double damage from any tool with hammering quality , normal for any with digging/excavation/shoveling , and a quarter for everything else

## Why It's Good For The Game
Buff hammering tools  + a knife shouldn't be that good at breaking a window.
## Changelog
:cl:
balance: Hammering quality tools deal double damage to windows
balance: Any tool with excavation,digging or shoveling quality deals normal damage to windows
balance: Any tool without the aforementioned qualities deals a quarter of their damage.
/:cl:

### DIFF
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -383,23 +383,26 @@
 				if(hit(I.force*I.structure_damage_factor*2) && health <= 10)
 					set_anchored(FALSE)
 					step(src, get_dir(user, src))
+				..()
 				return
 			// excavation and drilling, guaranteed to do damage.
 			if(QUALITY_DRILLING)
 				hit(I.force*I.structure_damage_factor)
+				..()
 				return
 			if(QUALITY_EXCAVATION)
 				hit(I.force*I.structure_damage_factor)
+				..()
 				return
 			if(QUALITY_DIGGING)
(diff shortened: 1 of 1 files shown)


DiaFRAME444:
Shouldn't structure damage to windows just be handled by the already existing structure_damage_factor variable? This means they basically get triple or quadruple damage.

Alliostra:
yeah, the fire axe will now deal 9 damage vs a window. I support the idea of just handling this through the structure damage factor.

garrulousTeamotei:
Vagabond crowbar buff.

DiaFRAME444:
I'm pretty sure a greatsword would also easily break through a window, so this is a weird PR.
What's the point? Can't be for balance because anyone can still easily break windows, can't be for realism because its the opposite of that, can't be for balance because its balanced like shit.

Pink-Chink:
I think the resistance change in your other PR is sufficient.

Kurgis:
There's no user input regarding this, the information about tool qualities affecting glass structure damage should be provided in some way in-game.

Firefox13:
Counterpoint:
изображение

PR #7837, opened 2022-11-20
Hotfix for projectiles deleting mobs

Pink-Chink:
## About The Pull Request

This may or may not fix the bug, but it should be done regardless.

## Why It's Good For The Game

Bug fix

## Changelog
:cl:
fix: Adds cleanup in /obj/item/projectile Destroy()
/:cl:


ValoTheValo:
based pr

PR #7839, opened 2022-11-21
Some fireaxe love

Alliostra:
## About The Pull Request

The Fire axe has received a small damage and penetration buff (from 33 damage wielded to 34 damage wielded and from 1.6 AP to 1.8 AP), since it was pretty shit for a weapon that doesn't fit into a bag.

The fire axe cabinet can now also be unlocked by an ID with head or medical access without smashing the glass.

## Why It's Good For The Game

No longer neglected

## Changelog
:cl:
add: the fire axe cabinet can now be opened by an ID with head or medical access.
balance: slightly buffed fire axe damage
/:cl:

### DIFF
--- a/code/game/objects/structures/fireaxe_cabinet.dm
+++ b/code/game/objects/structures/fireaxe_cabinet.dm
@@ -87,7 +87,7 @@
 		if(has_access(list(), req_one_access, ID.GetAccess()))
 			toggle_lock(user)
 		else
-			to_chat(user, SPAN_NOTICE("Insufficient access"))
+			to_chat(user, SPAN_NOTICE("You try to unlock the cabinet, but nothing happens."))
 	if(istype(O, /obj/item/tool/fireaxe))
 		if(open)
 			if(fireaxe)
--- a/code/game/objects/structures/fireaxe_cabinet.dm
+++ b/code/game/objects/structures/fireaxe_cabinet.dm
@@ -10,7 +10,7 @@
 	var/unlocked
 	var/shattered
 	var/obj/item/tool/fireaxe/fireaxe
-	req_access = list(access_moebius, access_heads)
+	req_one_access = list(access_moebius, access_heads)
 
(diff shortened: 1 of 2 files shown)


DiaFRAME444:
Should the Fireaxe really be so strong? I'm pretty sure its better than the NT Greatsword now.

assortedbeads:
Axes do tend to be stronger than swords

Alliostra:
> Should the Fireaxe really be so strong? I'm pretty sure its better than the NT Greatsword now.

The Fireaxe deals 2 less damage than the greatsword (both wielded), and has slightly worse AP too. It also occupies the slot usually reserved for a big gun, and deals a puny 10 damage when not wielded.

DiaFRAME444:
> The Fireaxe deals 2 less damage than the greatsword (both wielded)

The way the multiplier works means a massive damage jump when a Spike toolmod is added. Still wasn't a top tier weapon just yet at 3.3x, but 3.7x and AP buff might make it so.
I guess a more fair comparsion would be the Halberda.

Alliostra:
> > The Fireaxe deals 2 less damage than the greatsword (both wielded)
> 
> The way the multiplier works means a massive damage jump when a Spike toolmod is added. Still wasn't a top tier weapon just yet at 3.3x, but 3.7x and AP buff might make it so. I guess a more fair comparsion would be the Halberda.

Good point.

code/game/objects/structures/fireaxe_cabinet.dm
```diff
@@ -10,6 +10,7 @@
 	var/unlocked
 	var/shattered
 	var/obj/item/tool/fireaxe/fireaxe
+	req_one_access = list(access_moebius, access_heads)
```
Firefox13, line 13: I remember seeing a fire axe in technomancers locker room too

Pink-Chink:
image

Pink-Chink (approved):

PR #7844, opened 2022-11-23
Changes world.fps to match config fps

Pink-Chink:
## About The Pull Request

Changes world.fps to 40. As far as I can tell, this will only affect SStimer bucket length and the discrepancy in tick lag/FPS at roundstart when subsystems call PreInit() before the MC changes the FPS to 40 when it calls Initialize(). This should probably get scrutinized for safety. Local test confirmed that SStimer did not reset at roundstart with this change.

## Why It's Good For The Game

Makes the default FPS value match our config.

## Changelog
:cl:
tweak: Changed world.fps to match config fps (40)
/:cl:

### DIFF
--- a/code/world.dm
+++ b/code/world.dm
@@ -17,4 +17,4 @@
 	view = "15x15"
 	hub = "Exadv1.spacestation13"
 	cache_lifespan = 0	//stops player uploaded stuff from being kept in the rsc past the current session
-	fps = 20
+	fps = 40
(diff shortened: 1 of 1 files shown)


Pink-Chink:
Getting a lot of tick drift on local with this.

Pink-Chink:
Tick drift may have been a PC skill issue. Gonna need some eyes on this.

PR #7845, opened 2022-11-23
Window grab interaction rework

MLGTASTICa:
## About The Pull Request
Getting thrown into a window has the following effects depending on your toughness and size, and perks
If you have the ass of concrete perk , your damage done to the window is doubled and you only take 5 brute
Else it uses the following equation
victimToughness = victim toughness , this is sanitized to not be 0
victimDivisior = 100 if victim toughness is positive , -(100 - victimtoughness) if negative 
if victimTougheness * mob_size / toughnessDiv < windowResistance * health_ratio , then they will get 5 hallos , get weakened for 2 ticks , and get
40 * (1 - vicTgh/tghDiv) * healthRatio damage,  with an armor_div of 0.5 (doubled armor)
else if the opposite is true , they will get the same equation but with 20 damage and no armor effectiveness buff

Smashing someone against a window will break your grab instantly
Smashing someone against a window now gives you the DEFAULT_ATTACK_COOLDOWN (the same one as punches)
calculations used for effects
grabberRob = grabbers.Rob (sanitized to not be 0) + targetTgh(if target has negative rob), same applies for TargetTgh , with negative grabber rob buffing your Tgh
skillRatio =  grabberRob / targetTgh clamped between 0.3 and 3
toughTarget = target having ass of concrete perk
depending on the grab power it has diferent effects 
for a passive grab , you will get rounded(10 * skillRatio * healthRatio) damage , which is divided by 3 with the ass of conc perk, and with an armor div of 0.5, along with a -25 debuff to vigilance unless you have the ass of concrete perk that lasts 8 seconds
The damage dealt to the window itself is your mob_size * skill_ratio / window_resistance and multiplied by 2 if you are tough

for a aggresive grab , if the skill ratio is above 2, and the target is not weakened or tough , they will be weakened for 1 tick.
They will receive -34 to their vigilance for 12 seconds or 6 if they are tough.
Damage dealt to window is the same as last one but multiplied by 1.5 
Damage dealt to the player is the same as last one , but with 20 instead of 10 as the starting point and an armor divisor of 0.4

for neck grabs
targets gets weakened for round(3 * skillRatio) ticks,  divided by 3 if they are tough
The damage equation is the same ,but instead of 20 they have 30 as the starting point , with an armor div of 0.3
The vig debuff is equal to -50 , for 10 seconds if they are tough , or 20 if not.
the damage to the window is the same as the last equation , but multiplied by 2 and not 1.5

## Why It's Good For The Game
Makes window combat more tactical
## Changelog
:cl:
balance : Reworked window smash-grab and thrown interactions for players.
code: Updated some of the structure code to modern standards.
/:cl:


hyperioo:
Need feedback of players about all these values since I have no idea how they are compared to other sources of damage

code/game/objects/structures/window.dm
```diff
@@ -250,46 +244,98 @@
...
+		// if you less tougher and less sized than the window itself and its health , you are more likely to suffer more
+		if(victimToughness * M.mob_size / toughnessDivisor < windowResistance * healthRatio)
+			M.adjustHalLoss(5)
+			M.Weaken(2)
+			// 40 in worst case, 10 with 15 melle armor
+			M.damage_through_armor(40 * (1 - victimToughness/toughnessDivisor) * healthRatio, BRUTE, body_part, ARMOR_MELEE, sharp = FALSE, armor_divisor = 0.5)
+		else
+			M.adjustHalLoss(3)
+			// 20 in worst  case , 5 with 15 melle armor
+			M.damage_through_armor(20 * (1 - victimToughness/toughnessDivisor) * healthRatio, BRUTE, body_part, ARMOR_MELEE, sharp = FALSE)
```
hyperioo, line 334: If the victim has more than 100 thoughness (the current max is 150) then they will take negative damage because of `(1 - victimToughness/toughnessDivisor)` since `thoughnessDiviser = 100`

hyperioo (changes requested):
Negative damage

code/game/objects/structures/window.dm
```diff
@@ -250,46 +244,98 @@
...
-		M.Weaken(1)
-	M.damage_through_armor(rand(7,10), BRUTE, body_part, ARMOR_MELEE)
+	// being super tough has its perks!
+	if(!M.stats.getPerk(PERK_ASS_OF_CONCRETE))
+		var/victimToughness = M.stats.getStat(STAT_TGH, FALSE)
+		victimToughness = victimToughness ? victimToughness : 1
+		var/windowResistance = resistance ? resistance : 1
+		var/healthRatio = health/maxhealth
+		// you shall suffer for being negative on toughness , it becomes negative so it cancels the negative toughness
+		var/toughnessDivisor = victimToughness > 0 ? 100 : -(100 - victimToughness)
```
hyperioo, line 324:
```suggestion
		var/toughnessDivisor = victimToughness > 0 ? STAT_VALUE_MAXIMUM : -(STAT_VALUE_MAXIMUM  - victimToughness)
```

hyperioo:
Ah damn, we can't apply suggestion because fork was deleted, I'll do another PR

hyperioo (approved):

PR #7849, opened 2022-11-25
Makes borers actually able to copy stats

greggbot:
## About The Pull Request

So apparently, the borer's "Read Mind" ability was checking if the host did NOT have stats, rather than checking if they did.
That's obviously dumb and unintended, so I fixed it.

I fully tested this. Worked as intended, stats were copied, messages regarding copied stats and increased brain/sanity damage work fine, no runtimes or anything nasty.

## Why It's Good For The Game

In all honesty it's probably not, I fully expect this to have balance issues, but it's a bug so I fixed it.

## Changelog
:cl:
fix: Borers can now actually copy stats as intended.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/borer/borer_powers.dm
+++ b/code/modules/mob/living/simple_animal/borer/borer_powers.dm
@@ -438,7 +438,7 @@
 		return
 
 	var/list/copied_stats = list()
-	if(!host.stats)
+	if(host.stats)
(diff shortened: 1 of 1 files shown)


DiaFRAME444:
They're bugged don't mind it.

SirRichardFrancis (approved):

Pink-Chink (approved):

PR #6416, opened 2021-08-10
HICU: Huge Integrated Circuit Update

Doster-d:
## About The Pull Request

This pull request is port from Onyxbay, TG, Baystation circuit code. 

Old circuits code are not removed, but disabled.

Added new systems to optimization circuit stuff, like:
- Circuit Components. Fires do_work proc only if tick have some space for it.
- Circuits. Initialize circuits and generate key for XOR encryption 

Added circuit copy/pasting (cloning).

Added many circuits so descriptions these circuits will be at least like "Война и Мир".

Added/updated related stuff for circuits, like fonts, css files.

Updated interface for assemblies, now it look like interface of Visual Studio Code.

Potentially fixes old bugs with circuits.

## Why It's Good For The Game

The old circuits code are... Outdated.
The PR's circuits are good, contains numerous bug fixes.

## Changelog
:cl:
add: Circuit Component subsystem to optimization circuit do_work proc.
add: Circuit subsystem to optimization initialization stuff.
add: New numerous circuits.
add: circuits cloning ability.
tweak: Updated interface for assemblies, now it look like interface of Visual Studio Code.
del: Disabled old circuit subsystem.
soundadd: added some new sound for circuits, like tripple beep sound.
soundadd: added HEV suit sound from HL2. (Circuits only)
soundadd: added AI sounds from System Shock. (Circuits only)
admin: Added secret to disable/enable circuit proc do_work firing.
config: Added config var to disable circuit cloning
/:cl:

Please check MMI tank, PAI holder and AI holder circuits, I can miss some mistakes.

PR #7852, opened 2022-11-27
Remove integrated circuits

SirRichardFrancis:
## About The Pull Request

This PR effectively reverts #6416 and #912 (circuits were added twice, so we have that code partially duplicated), removing over 17 thousands lines of pretty much useless code.

## Why It's Good For The Game

A while back I wanted that feature a lot, but now, after over a year since it have been added, I have to admit that it didn't work out. At all.

We have rather outdated version with little content, and most of the stuff that is present - just does not work on our build.
Even very simple constructs like a flashlight is bugged, let alone more complex circuits like chemistry or weapon related stuff.

There is also very few things that are worth making with circuits. Even if Eris were to update and fix everything that is wrong about this feature, completely rooting out exploits and bugs (good luck with that), there is still hardly any reason to have circuits.
They are only encouraging people to commit hours-long autism in far corner of Moebius lab, which is far from intended gameplay.

TL;DR: Feature is broken and requires too much time and effort to "fix". Even then it's not worth the trouble.

## Changelog
:cl:
del: integrated circuits
/:cl:


pexel5:
nooo, my implants that give me my meth on command, just not that!

SirRichardFrancis:
> nooo, my implants that give me my meth on command, just not that!

Get an organoid.

Pink-Chink:
Heads up, there will be a merge conflict.

hyperioo (approved):

PR #7854, opened 2022-11-27
Remove hallucination door shocking

SirRichardFrancis:
## About The Pull Request

image

image

## Why It's Good For The Game

Annoying stuff.

## Changelog
:cl:
del: hallucination door shocking
/:cl:


MLGTASTICa:
replace it with one that gives you agony.

Humonitarian:
> replace it with one that gives you agony.

Seconded, make it check for insuls too (cheap one working too regardless of statistics), because if you think you're shielded, you don't hallucinate getting shocked

Pink-Chink (approved):
Agony-only hallucination shocks sound fine, but we should discuss that in it's own PR. The request was for complete removal.

PR #7856, opened 2022-11-28
Replaces getter procs with vars in superior animal status updates

Pink-Chink:
## About The Pull Request

Apparently, proc calling is very expensive in byond. Superior animal's handle_cheap_regular_status_updates() was one of the more costly procs used by superior mobs. Based on a local test, replacing the getter procs for damage with the damage vars cut the proc's CPU time cost in half. I doubt we'll see a noticeable improvement in performance, but every little bit counts.

## Why It's Good For The Game

Performance

## Changelog
:cl:
code: Replaced damage getter procs with damage vars in superior animal status updates
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/superior_animal/superior_animal.dm
+++ b/code/modules/mob/living/carbon/superior_animal/superior_animal.dm
@@ -288,7 +288,7 @@
 	weakened = max(weakened-3,0)
 
 /mob/living/carbon/superior_animal/proc/handle_cheap_regular_status_updates()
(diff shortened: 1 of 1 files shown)


SirRichardFrancis (approved):

MLGTASTICa:
Its not apparently , its obvious. 8 extra instructions per call really. for each function call.
In a perfect world , we would not have these handled at all in a process tick , they would be checked every time damage is dealt.

Alliostra:
In a perfect world, procs like these wouldn't exist.
But this is not a perfect world.

Firefox13 (approved):
In a perfect world there would be no roaches

PR #7869, opened 2022-11-30
Xenoarcheology removal

sgtmind:
First pass of the xenoarchealogy removal. Am leaving behind a couple items which have use outside of this.

## About The Pull Request

The removal of xenoarchaeology and all its components in order to make way for something better, or at least with far less spaghetti code. (Only reason I am doing this a second time is because I couldn't update my branch)

## Why It's Good For The Game

Honestly, had an awful experience with a random artifact spawning once, and I realized that xenoarcheaology adds nothing to the game. In addition, it is a bunch of spaghetti code

## Changelog
:cl:
del: Removed xenoarcheology
/:cl:


ValoTheValo:
same as before. fuck xenoarch

sgtmind:
What I am likely going to require help with is removing the part where it drops the artifact from mining.

sgtmind:
How would I deal with the error of M.GetDrilled not being fouund?

code/modules/mining/ore.dm
```diff
@@ -113,11 +113,4 @@
 /obj/item/ore/Initialize(mapload)
 	. = ..()
 	pixel_x = rand(0,16)-8
-	pixel_y = rand(0,8)-8
```
SirRichardFrancis, line 116: Maybe keep that one.

SirRichardFrancis (changes requested):
You certainly have to keep `/turf/simulated/mineral`, but with most of the code removed.
GetDrilled() proc should stay, it's the one that makes ore drop.
Look where it's called, and make sure those calls are preserved, otherwise miners may have a problem.
Also please keep `/obj/structure/boulder` as a placeholder object, I'll remove it at a later date. Removal now would create annoying merge conflict.
`/datum/talking_atom` also should be kept, but moved elsewhere.
Location of `/obj/item/talkingcrystal` would be a good place for it.

sgtmind:
Should be good, did a test, ore seems to drop, no artifacts within the code as far as I can see.

SirRichardFrancis (commented):
`/obj/structure/boulder` and "excavation" should go too.

code/modules/mining/ore.dm
```diff
@@ -113,11 +113,4 @@
...
-
-/obj/item/ore/attackby(obj/item/W, mob/user)
-	if(istype(W,/obj/item/device/core_sampler))
-		var/obj/item/device/core_sampler/C = W
-		C.sample_item(src, user)
-	else
-		return ..()
+	pixel_y = rand(0,8)-8
```
hyperioo, line 116:
```suggestion
	pixel_y = rand(0,8)-8

```
Newline

code/modules/reagents/prefilled.dm
```diff
@@ -44,4 +44,4 @@
 	preloaded_reagents = list("fuel" = 60)
 
 /obj/item/reagent_containers/glass/beaker/silicon
-	preloaded_reagents = list("silicon" = 60)
+	preloaded_reagents = list("silicon" = 60)
```
hyperioo, line 47:
```suggestion
	preloaded_reagents = list("silicon" = 60)

```
Newline

code/modules/mining/ore.dm
hyperioo: It is kept just a few lines under

hyperioo:
> `/obj/structure/boulder` and "excavation" should go too.

Why? Boulder seems to be still use when stuff is being excavated for ores

SirRichardFrancis:
> Why? Boulder seems to be still use when stuff is being excavated for ores

And what purpose does it serve?

sgtmind:
I am still not sure how to fix it

SirRichardFrancis:
It's no big deal, could just let it be for now.

SirRichardFrancis (approved):

hyperioo (approved):

vode-code:
> It's no big deal, could just let it be for now.

false, the result is the production of invulnerable boulders.

PR #7886, opened 2022-12-03
Vendors can use alternate currencies + OS vendor and OS cash placeholders

Pink-Chink:
## About The Pull Request

Gives vendors functionality to accept alternative currencies. This function must be enabled by defining an alternative currency object path in the vendor definition, which will also prevent the use of credits. Several placeholder OS vendors and OS cash have been added for use in future OS content. 

## Why It's Good For The Game

Framework for future content

## Testing

- Purchased an item from OS vendor with OS cash.
- Attempted to use spacecash (our regular credits) in alt currency enabled vendor.
- Attempted to swipe ID to pay in alt currency enabled vendor.
- Attempted to use OS cash in a regular vendor.

image

## Credits

OS vendor sprites by Infrared Baron
OS coin sprites by bobob9000

## Changelog
:cl:
add: Added placeholder OS vendors and OS cash
code: Added alternative currency behavior to vending machines
/:cl:


SirRichardFrancis:
Coins with eyes? Think OS could do cooler than that.
Something like social credit implants maybe.
Or at least credit chips.

code/game/machinery/vending.dm
```diff
@@ -357,32 +356,39 @@
...
-			handled = 1
+		var/paid = FALSE
+		var/handled = FALSE
+
+		if(alt_currency_path)
+			if(istype(I, alt_currency_path))
+				paid = pay_with_item(I, user)
+			else
+				var/atom/movable/AM = alt_currency_path
+				to_chat(user, SPAN_WARNING("This vending machine only accepts [initial(AM.name)] as currency."))
```
TheShown911, line 373: Can you take a var from an un-instanced path, i thought you could only do it by using initial?

TheShown911 (approved):
Lore nerds can clean up descriptions later.

code/game/machinery/vending.dm
```diff
@@ -566,6 +569,28 @@
...
+	var/amount_to_spend = currently_vending.price
+
+	if(istype(I, /obj/item/stack))
+		var/obj/item/stack/S = I
+		if(S.amount >= amount_to_spend)
+			S.use(amount_to_spend)
+			should_qdel = FALSE		// Don't qdel a stack with remaining charges
+		else
+			to_chat(user, SPAN_WARNING("\icon[I] That is not enough money."))
+			return FALSE
```
hyperioo, line 584:
```suggestion
	if(istype(I, /obj/item/stack))
		var/obj/item/stack/S = I
		if(S.amount >= amount_to_spend)
			S.use(amount_to_spend)
			if(S.amount)
				should_qdel = FALSE		// Don't qdel a stack with remaining charges
		else
			to_chat(user, SPAN_WARNING("\icon[I] That is not enough money."))
			return FALSE
	else
			return FALSE
```
You just still delete the stack if the person has exactly enough money since there is 0 remaining.
Also unction is not supposed to be called with a non-stack right? Better be safe by returning FALSE otherwise the message will pop and item will be deleted.

code/modules/onestar/vending.dm
```diff
@@ -0,0 +1,46 @@
...
+	vendor_department = DEPARTMENT_OFFSHIP
+	alt_currency_path = /obj/item/stack/os_cash
+
+/obj/machinery/vending/one_star/Initialize()
+	. = ..()
+	set_light(1.4, 1, COLOR_LIGHTING_CYAN_BRIGHT)
+	earnings_account = department_accounts[DEPARTMENT_OFFSHIP]
+
+/obj/machinery/vending/one_star/guns
+	desc = "A vendor of the One Star variety. This one sells firearms of the One Star variety."
```
hyperioo, line 21: There has to be a better way to word that to avoid repetition.
Something like "A One Star firearms vendor" "A One Star food vendor" and so on

hyperioo (changes requested):

code/modules/onestar/vending.dm
Pink-Chink: These are placeholders for Valo to rewrite.

hyperioo: Ok

code/game/machinery/vending.dm
```diff
@@ -566,6 +569,31 @@
...
+		var/obj/item/stack/S = I
+		if(S.amount >= amount_to_spend)
+			S.use(amount_to_spend)
+			if(S.amount)
+				should_qdel = FALSE		// Don't qdel a stack with remaining charges
+		else
+			to_chat(user, SPAN_WARNING("\icon[I] That is not enough money."))
+			return FALSE
+	else
+			return FALSE
```
hyperioo, line 587:
```suggestion
		return FALSE
```

hyperioo:
My bad, put a wrong indent in my suggestion

PR #7912, opened 2022-12-10
Fixes robot HUD not updating on module storage

the-og-gear:
## About The Pull Request
Fix the robot HUD not updating on module storage

## Why It's Good For The Game
It's very slightly less confusing if you store a module then immediately equip a new one, so it won't look like your newly equipped module is selected. If this description is confusing it's because the issue is confusing and I'm tired as fuck.

## Testing
I loaded up the game, spawned in as a borg. Then I equipped a module, activated it, then stowed it while active. Additionally checked for runtimes to ensure the logs aren't literally fucking dying.

## Changelog
:cl:
fix: Stowing borg modules no longer leaves the inventory slot highlighted
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/inventory.dm
+++ b/code/modules/mob/living/silicon/robot/inventory.dm
@@ -51,7 +51,7 @@
 		module_state_3:loc = module
 		module_state_3 = null
 		//inv3.icon_state = "inv3"
-	for (var/obj/screen/inv in src.HUDinventory)
+	for (var/obj/screen/inv as anything in HUDinventory)
 		inv.update_icon()
 	update_robot_modules_display()
 	updateicon()
--- a/code/modules/mob/living/silicon/robot/inventory.dm
+++ b/code/modules/mob/living/silicon/robot/inventory.dm
@@ -51,7 +51,7 @@
 		module_state_3:loc = module
 		module_state_3 = null
 		//inv3.icon_state = "inv3"
-	for (var/obj/screen/inv as anything in HUDinventory)
+	for(var/obj/screen/inv as anything in HUDinventory)
 		inv.update_icon()
 	update_robot_modules_display()
 	updateicon()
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/silicon/robot/inventory.dm
```diff
@@ -51,6 +51,8 @@
 		module_state_3:loc = module
 		module_state_3 = null
 		//inv3.icon_state = "inv3"
+	for (var/obj/screen/inv in src.HUDinventory)
+		inv.update_icon()
```
SirRichardFrancis, line 55:
```suggestion
	for(var/obj/screen/inv as anything in HUDinventory)
		inv.update_icon()
```

the-og-gear: 🤔
Not as versed in byond optimization. Performance improvement by just taking anything it gives me, yeah?

SirRichardFrancis: For typed loops BYOND will do an `istype()` check, and will skip a list entry if it's type does not match an iterator's type.
But if we know for sure that list only contains datums of a specific type (which it does in our case) - we can skip those type checks with `as anything` keyword.
It's not much in the way of performance gain, but I don't see why neglect it.
As for `src.`, it's simply redundant and should be avoided as per Eris code contribution guidelines.

the-og-gear: Ah, I see. Didn't know it'll skip the type check with `as anything`
Interesting, definitely going to watch for that in the future.

code/modules/mob/living/silicon/robot/inventory.dm
```diff
@@ -51,6 +51,8 @@
 		module_state_3:loc = module
 		module_state_3 = null
 		//inv3.icon_state = "inv3"
+	for (var/obj/screen/inv as anything in HUDinventory)
```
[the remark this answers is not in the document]
the-og-gear: lmao fair. didn't notice the lack of space in the previous suggestion

Pink-Chink (approved):

PR #7917, opened 2022-12-13
Handmade Gun Crafting Rework 

driecg36:
## About The Pull Request

Changes most handmade guns to be cheaper, faster, and easier to craft. Recipes give full guns again instead of frames. Gave new names to the handmade guns that didn't have any as well as make gun names in the crafting window consistent. Did not change any gun values except making the zip-gun small. 

## Why It's Good For The Game

Makeshift guns are very under-utilized at the moment. One of the main reasons in my opinion is how much effort you actually have to put in to making them, which is especially ridiculous for the truly awful guns (like makeshift pistol and slide shotty). 

Currently, you need to do the following: 
6-10 crafting steps>Acquire parts and board for crafting bench>build all the other gun parts>assemble gun>gun
Now:
4-6 crafting steps>gun

The very purpose of these guns is to be easily accessible but worse alternatives to professionally made guns. They aren't nearly powerful enough to be worth the current amount of effort required unless you are specifically doing a gimmick so they are barely ever used. Having them not only take significantly less time to craft but also not be tethered to needing a crafting bench should hopefully encourage people to use these guns a lot more. 

## Testing

Ran a private server, all crafting recipes worked, and names showed up properly.

## Changelog
:cl:
add: New names for the already existing guns.
tweak: Homogenized gun names in the crafting window for better organization/consistency. 
balance: Handmade guns are much easier and faster to craft. 
spellcheck: Gun Description typos. 
/:cl:


Mycah142:
Makeshift guns would see alot more use if normal, "better" guns weren't a dime a dozen.

driecg36:
> Makeshift guns would see alot more use if normal, "better" guns weren't a dime a dozen.

As much as I agree, that's a much harder change (that's been attempted many many many times) than just "make handmade guns cheap." I can't fix the entire balance of the economy/junk loop but I can at least make under-used options more appealing.

Alliostra:
just remove junk piles

Humonitarian:
Ideally makeshift gun names should be scifi references, if we follow the existing naming scheme.
Most current HM guns are named after Foundation characters, for example.

Additionally, reminder that most makeshift guns can have their barrel swapped out for a different caliber. If you're up for it, an interaction that lets you swap barrels wholesale would be pretty cool.

Considering the recipes now produce full guns, make sure disassembling them into parts does not dupe materials.

driecg36:
> Ideally makeshift gun names should be scifi references, if we follow the existing naming scheme.
Most current HM guns are named after Foundation characters, for example.

Additionally, reminder that most makeshift guns can have their barrel swapped out for a different caliber. If you're up for it, an interaction that lets you swap barrels wholesale would be pretty cool.

Considering the recipes now produce full guns, make sure disassembling them into parts does not dupe materials.

Considering the HM "Motherfucker" I was under the impression that makeshift guns had a wider naming scheme. Especially considering they aren't real designs but just things cobbled together, I thought more out there names would be appropriate. 

I may take a swipe at barrel refitting later (maybe disassembling the gun and using a hammer on the barrel? something like that.) At the moment it is no more effort to refit barrels with this than it was previously so I'm fine with leaving it as is for the moment if you are. 

As to the duping items thing, I will take a look at that in a moment.

driecg36:
you could already dupe mats with most of the makeshift guns :mitosis:
handmade pistol frame gave you plasteel :mitosis:

driecg36:
@Humonitarian I think that should be all the dupes fixed.

maikilangiolo:
> you could already dupe mats with most of the makeshift guns

No you couldn't. The frames recycled for the exact same materials it costed to make, and the full gun recycled for the frame + the parts used to build. 
The makeshift pistol frame giving plasteel is an oversight

code/modules/projectiles/guns/projectile/revolver/handmade.dm
```diff
@@ -1,12 +1,12 @@
 /obj/item/gun/projectile/revolver/handmade
-	name = "handmade revolver"
-	desc = "Handmade revolver, made from gun parts. and some duct tap, will it even hold up?"
+	name = "HM REV \"Cartel\""
+	desc = "A handmade revolver built with junk, duct tape, and dubious engineering skills. Will it even work?"
```
Firefox13, line 3:
```suggestion
	desc = "A handmade revolver built with junk, duct tape and dubious engineering skills. Will it even work?"
```

Firefox13 (approved):
Should be okay...

PR #7918, opened 2022-12-14
Adds a subsystem for handling stuns , weakening and paralyzing of mobs

MLGTASTICa:
## About The Pull Request
Separates stunning , weakening and paralyzing from mob/Life to its own subsystem
## Why It's Good For The Game
Stunning , weaking and paralyzing can not be adjusted for in ingame-time unlike medicine , healing and other factors,  as such
these should be on max priority and handled with maximum accuracy since they can ruin someone's ruind

## Testing
Not tested yet , still to do

## Changelog
:cl:
add: Added new things
del: Removed old things
tweak: tweaked a few things
balance: rebalanced something
fix: fixed a few things
soundadd: added a new sound thingy
sounddel: removed an old sound thingy
imageadd: added some icons and images
imagedel: deleted some icons and images
spellcheck: fixed a few typos
code: changed some code
refactor: refactored some code
config: changed some config setting
admin: messed with admin stuff
server: something server ops should know
/:cl:


SirRichardFrancis (commented):
Idea is 100% reasonable, but I don't like the implementation at all.
Would elaborate later. Also it doesn't pass integration tests.

SirRichardFrancis (changes requested):
My problems with this PR:
1) It replaces simple variable checks with proc calls (a lot), which are stupidly expensive in DM.
You've acknowledged it in #7856 yourself, yet here we are, repeating old mistakes.
2) It removes a combat proc in a "performance" PR, without even mentioning it.
3) It is done in overcomplicated manner, you juggle weakref datums, creating and deleting new status effect datums, making lists for them, introducing new procs, but none of that is needed.
Only thing we need to do, is to loop though mobs, see if they have active effects, and decrement related variable once. Then check if it's 0, and call related `update_whatever()` proc, if one is needed.
4) It doesn't even work.

MLGTASTICa:
> My problems with this PR:
> 
> 1. It replaces simple variable checks with proc calls (a lot), which are stupidly expensive in DM.
>    You've acknowledged it in #7856 yourself, yet here we are, repeating old mistakes.
> 2. It removes a combat proc in a "performance" PR, without even mentioning it.
> 3. It is done in overcomplicated manner, you juggle weakref datums, creating and deleting new status effect datums, making lists for them, introducing new procs, but none of that is needed.
>    Only thing we need to do, is to loop though mobs, see if they have active effects, and decrement related variable once. Then check if it's 0, and call related `update_whatever()` proc, if one is needed.
> 4. It doesn't even work.

1.Proc calls are not stupidy expensive in DM, they might be slow , but for the context we are working in , it isn't a matter to worry about
2. The proc was removed due to being redundant with the new functionality
3. Sorting and caching the effects of each mob is not needed , but it is better for searching performance, weakrefs are necesarry due to signals not being up to latest TG standards ( and not providing a reference when something gets deleted) and as to not prevent deletion of mobs , along with the system being compliant with ECS and compatible with a wide array of targets.
4.  It doesn't work cause its not finished yet , guh

Pink-Chink:
Might as well get the ball rolling since I have a bit of time to look at this.

Pink-Chink:
First things first: variables, types and methods should be snake case (contribution guidelines).

code/controllers/subsystems/processing/statuseffects.dm
```diff
@@ -0,0 +1,188 @@
...
+		return FALSE
+	for(var/datum/statusEffect/effect as anything in SSstatusEffects.affectedMobs[ref(target)])
+		if(effect.identifier == effectType)
+			if(!(effect.flags & SE_FLAG_UNIQUE))
+				return effect
+			return_list.Add(effect)
+	return return_list
+
+
+/datum/statusEffect
```
Pink-Chink, line 114: Does this need to be derived from `/datum`? If all you need are the vars below, it would function fine as its own `/status_effect` object.

code/_onclick/click.dm
```diff
@@ -104,7 +104,7 @@
 		CtrlClickOn(A)
 		return 1
 
-	if(stat || paralysis || stunned || weakened)
+	if(stat || hasStatusEffect(src, SE_PARALYZED) || hasStatusEffect(src, SE_STUNNED) || hasStatusEffect(src, SE_WEAKENED))
```
Pink-Chink, line 107: Knowing that something as lightweight as getter procs were causing issues in roach mob processing, I don't see this being a good idea. I wonder if it would be better to use bitflags to indicate whether a mob has a status or not.

CocaColaTastesGood:
Are proc calls really that slow? Signals like ONMOVE call like a billion times in some cases.

MLGTASTICa:
> Are proc calls really that slow? Signals like ONMOVE call like a billion times in some cases.

They're not , its the frequency at which they are called.

MLGTASTICa:
Closing this for now due to the head repository being deleted, not being complete , and the implementation not being the best as it is right now.

PR #7931, opened 2022-12-21
Full auto cleanup and burst fire fix

Pink-Chink:
## About The Pull Request

Fixes an issue where fire delay would set a shorter click cooldown than burst delay, causing certain guns to be able to fire multiple bursts at once. Includes Humonitarian's shooting loop cleanup.

## Why It's Good For The Game

Bug fixes and cleanup

## Testing

- Live test

## Changelog
:cl:
tweak: Click cooldown from firing uses both burst delay and fire delay
code: Changed the shooting loop to use sleep() instead of spawn(), uses Humonitarian's logic
code: Moved a define to weapons.dm
/:cl:

### DIFF
--- a/code/__DEFINES/weapons.dm
+++ b/code/__DEFINES/weapons.dm
@@ -54,6 +54,9 @@
 #define STRUCTURE_DAMAGE_POWERFUL    	2
 #define STRUCTURE_DAMAGE_DESTRUCTIVE 	3
 
+// Weapon minimum fire_delay
+#define GUN_MINIMUM_FIRETIME 1.1 // 110 MS , ~9 shots per second.
+
 //Quick defines for fire modes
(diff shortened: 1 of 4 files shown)


code/__DEFINES/weapons.dm
```diff
@@ -54,6 +54,9 @@
 #define STRUCTURE_DAMAGE_POWERFUL    	2
 #define STRUCTURE_DAMAGE_DESTRUCTIVE 	3
 
+// Weapon minimum fire_delay
+#define GUN_MINIMUM_FIRETIME 1.1 // 110 MS , ~9 shots per second.
```
Humonitarian, line 58:
```suggestion
#define GUN_MINIMUM_FIRETIME 1.1 // 110 MS , ~9 shots per second, 545 RPM
```
Should help clarify the breakoff point where RPM becomes a cosmetic.

PR #7938, opened 2022-12-23
Caps all stats to a max of 150

MLGTASTICa:
## Why It's Good For The Game
Easier to handle balancing , standardizes the definition of high stats across the codebase 

## Testing

https://youtu.be/D675HU21oK8

## Changelog
:cl:
balance: All stats are now capped at a maximum of 150.
/:cl:

### DIFF
--- a/code/datums/mob_stats.dm
+++ b/code/datums/mob_stats.dm
@@ -194,7 +194,10 @@
 			return SM
 
 /datum/stat/proc/changeValue(affect)
-	value = value + affect
+	if(value + affect > STAT_VALUE_MAXIMUM)
+		value = STAT_VALUE_MAXIMUM
+	else
+		value = value + affect
 
 /datum/stat/proc/getValue(pure = FALSE)
 	if(pure)
@@ -210,7 +213,10 @@
 			. += SM.value
 
 /datum/stat/proc/setValue(value)
-	src.value = value
+	if(value > STAT_VALUE_MAXIMUM)
+		src.value = STAT_VALUE_MAXIMUM
+	else
+		src.value = value
 
 /datum/stat/proc/copyTo(var/datum/stat/recipient)
 	recipient.value = getValue(TRUE)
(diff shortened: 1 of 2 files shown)


assortedbeads:
Oh nice, cheers for making this.

Hopefully it's the first step to fixing the stat maxxing mentality that people have, and reducing the combat potential robustness has

code/datums/mob_stats.dm
```diff
@@ -210,7 +213,10 @@
 			. += SM.value
 
 /datum/stat/proc/setValue(value)
-	src.value = value
+	if(value > STAT_VALUE_MAXIMUM)
+		src.value = STAT_VALUE_MAXIMUM
+	else
+		src.value = value
```
SirRichardFrancis, line 219:
```suggestion
		value = STAT_VALUE_MAXIMUM
	else
		value = value
```

SirRichardFrancis (changes requested):

code/datums/mob_stats.dm
CocaColaTastesGood: why not just use CLAMP

SirRichardFrancis: You mean `min()`?

MLGTASTICa: because i feel fancy

CocaColaTastesGood: I'm guessing there should be a minimum stat value as well.

SirRichardFrancis: Stats can go negative, that's ok and intended, as far as I know.

greggbot: Yeah, but a negative cap can't hurt, right? Inverse of the cap should be fine, so -150. I don't expect we'll ever want to debuff anyone that much, nevermind more.

CocaColaTastesGood: If someone was able to reliably lower their stats you could probably get fucky behavior as well.

hyperioo (commented):
In the videos it goes up to 190 with menace, so it's not a hard cap?

hyperioo (approved):

PR #7941, opened 2022-12-24
Borer Changes

CocaColaTastesGood:
## About The Pull Request
Cleans up borer code and fixes possibly quite a few bugs.

## Testing

https://user-images.githubusercontent.com/47264839/209574269-847ce6f0-8a7e-447c-8e05-b5ece423e49b.mp4
image

## Changelog
:cl:
fix: Borers no longer get soft locked by their host dying while assuming control.
fix: Non-humanoid carbons can be controlled again.
fix: Borer's no longer persist in a host's implants/contents after leaving.
/:cl:


EgorDinamit (commented):
There was a bug with borers infesting simple(superior) mobs, like roaches, and permanently losing all abilities(this persisted even after roach died, was gibbed or borer was teleported out).
I only skimmed through the code and might've missed it, but it'd be nice if you fixed this issue here.

CocaColaTastesGood:
> There was a bug with borers infesting simple(superior) mobs, like roaches, and permanently losing all abilities(this persisted even after roach died, was gibbed or borer was teleported out). I only skimmed through the code and might've missed it, but it'd be nice if you fixed this issue here.

I disabled the ability to infect non carbons completely, or at least I believe I did. Borers were never made to handle non-humanoids. I might end up adding some support for them. I'm mostly waiting for my first borer pull to be merged so I don't have to handle conflicts mid-way through.

CocaColaTastesGood:
> I'd just like to note that infecting roaches USED to work, and is a neat tool, but I imagine it'd be hell to get working again, yeah. Definitely did work at some point though.

I'll fix it then, I'm currently stuck trying to figure out how to fix the soft lock on death though.

EgorDinamit:
I would personally love to rework borers entirely, but we should leave it for the next PR.
If roaches are carbons, then we should simply make sure all checks are iscarbon and not ishuman in place of putting abilities.
As for the "death softlock" - add a signal listening to the target's death/destruction.

CocaColaTastesGood:
> I would personally love to rework borers entirely, but we should leave it for the next PR. If roaches are carbons, then we should simply make sure all checks are iscarbon and not ishuman in place of putting abilities. As for the "death softlock" - add a signal listening to the target's death/destruction.

I tried that, death signals only are sent to oviewers, which doesn't seem to include borers inside of their body. I was able to fix it though.

code/modules/mob/living/simple_animal/borer/borer_powers.dm
```diff
@@ -674,3 +642,5 @@
 
 		to_chat(H, SPAN_WARNING("Your nose begins to bleed..."))
 		H.drip_blood(1)
+
+#undef DEFAULT_INFESTATION_DELAY
```
TheShown911, line 646: No newline.

TheShown911 (approved):

TheShown911:
But integration failures...

CocaColaTastesGood:
> But integration failures...

What does that mean?

code/modules/mob/death.dm
```diff
@@ -76,6 +76,11 @@
 	for(var/obj/item/implant/carrion_spider/control/C in src)
 		C.return_mind()
 
+	// Notifies borer of host death.
+	var/mob/living/simple_animal/borer/B = has_brain_worms()
+	if(B?.controlling)
+		B.host_death()
```
hyperioo, line 82: `has_brain_worms` returns `FALSE` instead of `null` when there is no borer so it runtimes

```
[18:13:00] Runtime in death.dm,81: Cannot read 0.controlling
  proc name: death (/mob/proc/death)
  src: the space carp (/mob/living/simple_animal/hostile/carp)
  src.loc: the underplating (77,11,13) (/turf/simulated/floor/plating/under/airless)
  call stack:
  the space carp (/mob/living/simple_animal/hostile/carp): death(null, "dies!", "You have died.")
  the space carp (/mob/living/simple_animal/hostile/carp): death(null, "dies!")
  the junk tractor beam generato... (/obj/jtb_generator): generate edge()
  the junk tractor beam generato... (/obj/jtb_generator): populate z level()
  the junk tractor beam generato... (/obj/jtb_generator): generate junk field()
  Misc (/datum/controller/subsystem/misc): build junk field()
  Misc (/datum/controller/subsystem/misc): Initialize(null)
  Master (/datum/controller/master): Initialize(10, 0, null)
```

hyperioo (changes requested):
@CocaColaTastesGood CI failure comes from this

code/modules/mob/death.dm
```diff
@@ -76,6 +76,12 @@
 	for(var/obj/item/implant/carrion_spider/control/C in src)
 		C.return_mind()
 
+	// Notifies borer of host death.
+	if(has_brain_worms())
+		var/mob/living/simple_animal/borer/B = has_brain_worms()
+		if(B?.controlling)
+			B.host_death()
```
hyperioo, line 83:
```suggestion
	var/mob/living/simple_animal/borer/B = has_brain_worms()
	if(B?.controlling)
		B.host_death()
```
Sorry I should have been more clear, it's cleaner by having has_brain_worms() return `null` instead of a weird function that returns either an object or FALSE. We want either `borer/null`  or `TRUE/FALSE`. I made a PR for that #7967 and will merge yours after.

hyperioo (approved):

hyperioo:
To be merged after #7967 so that Integration Tests don't fail

hyperioo:
Would have to be rebased for CI to succeed. I'll sepuku if it breaks everything.

PR #7953, opened 2022-12-29
Laser Swords Do Burn Damage Now

assortedbeads:
## About The Pull Request
Makes all laser swords

- Craftable Laser Sabre
- Antag Laser Sword
- Implanted Laser Sword
- Rig Laser Sword

Do burn damage when switched on.

## Why It's Good For The Game

Laser swords are inspired by light sabres, so it felt weird that they did brute damage like you were being stabbed with a knife.

It is inadvertently a buff since burn; are more painful, can cause infections easier, are resisted by less armor.
But the laser swords tended to ignore most armor anyway.

## Testing

- Launched Local
- Spawned in a tech
- Spawned in disciples
- Spawned in each sword type
- Attacked with switched off inhands to confirm those still did brute
- Switched on and attacked with activated laser weapons till the targeted limb flashed away into ashes
- Examined and used a health scanner to confirm burn damage was being done
- Did the same for the implanted and rig versions

## Changelog
:cl:
tweak: laser swords do burn damage now
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/melee/energy.dm
+++ b/code/game/objects/items/weapons/melee/energy.dm
@@ -24,6 +24,7 @@
 	active = 1
 	force = active_force
 	throwforce = active_throwforce
+	damtype = BURN
 	sharp = TRUE
 	edge = TRUE
 	w_class = active_w_class
@@ -38,6 +39,7 @@
 	active = 0
 	force = initial(force)
 	throwforce = initial(throwforce)
+	damtype = initial(damtype)
 	sharp = initial(sharp)
 	edge = initial(edge)
 	w_class = initial(w_class)
@@ -187,6 +189,7 @@
 	icon_state = "blade"
 	force = WEAPON_FORCE_ROBUST //Normal attacks deal very high damage - about the same as wielded fire axe
 	armor_divisor = ARMOR_PEN_MAX
+	damtype = BURN
 	sharp = TRUE
 	edge = TRUE
 	anchored = TRUE    // Never spawned outside of inventory, should be fine.


DiaFRAME444:
> It is inadvertently a buff since burn; are more painful, can cause infections easier, are resisted by less armor.

Brute and burn both generate 1 agony per point of damage.
Damage resistance should be determined by armor type, in this case always melee armor.

Its true that burns cause infections more easily but generally brute damage has bigger downsides like bleeding and internal damage. So to correct what you've said, this is technically a nerf.

DiaFRAME444:
Actually, could you make it so energy swords check energy armor instead of melee?

assortedbeads:
> Actually, could you make it so energy swords check energy armor instead of melee?

Possibly? I would have to see if I can find where that's handled

assortedbeads:
> I think a nice middle ground would be having it do burn, with no bleeding wounds (cauterization) and less bone breaking, but with additional armor pen over other weapons. Changing the armor type is certainly an option, but I think a simple penetration bonus is fine and makes a meaningful distinction between energy/physical melee.

I dont think it breaks bones at all as burn honestly? At least I didnt notice bones being broken during the test

DiaFRAME444:
> I think a nice middle ground would be having it do burn, with no bleeding wounds (cauterization) and less bone breaking

Burn damage doesn't do those things.

> but I think a simple penetration bonus is fine and makes a meaningful distinction between energy/physical melee.

The least amount of armor piercing a mainline (Commonly used, referenced) melee weapon has is 1.6. The Hypersaw has 1.6, the junkblade has 1.6, the advanced sledgehammer has 2, the NT Spatha has 2, the NT Halberd has 3, the Moesci Katana ignores armor. I don't remember where Captain's Sabre is stored but its nanomonecular edge was said to be very armor piercing.
The exception is the Big Wrench which has just 1, but its a pretty weak melee weapon overall.
Ergo, it would be more distinctive if energy melee was actually crap at piercing armor. Not that I recommend that, just saying.

>> Actually, could you make it so energy swords check energy armor instead of melee?
> Possibly? I would have to see if I can find where that's handled

Nice

Firefox13:
изображение

Firefox13:
Our burn damage cannot cleanly cut off limbs (unlike in Star Wars) so it will lose that style.
Howeverm making it dealing burn and giving some other bits will make it an unique choice of melee.

Humonitarian:
> Our burn damage cannot cleanly cut off limbs (unlike in Star Wars) so it will lose that style. Howeverm making it dealing burn and giving some other bits will make it an unique choice of melee.

Iirc it can so long it is sharp

assortedbeads:
> Iirc it can so long it is sharp 

These swords do still have the sharp flag, but I wasnt able to decap or delimb when I tested.

Limbs all got ashed

Firefox13:
It ashes limbs instead of cutting them off, yeah. That's not fancy.

Alliostra:
should allow them to shave heads, that would be based

tristain01:
For one I think this is an intresting switch up for.the swords, looking foward to testing the.balance

hyperioo (commented):
Yeah the fact that the whole limb is ashed instead of cut is weird. You would have to dig a bit in the code to see if it's possible to do something that has "brute properties" yet still apply burn damage

Kurgis:
I always assumed these were beams of contained voodoo-~~magic~~-_science_ plasma instead of a laser, which explains the brute damage, the plasma is exploding the contacted areas. They'd just be called "laser swords" as that are what they are most popularly known as.

Humonitarian:
> I always assumed these were beams of contained voodoo-~magic~-_science_ plasma instead of a laser, which explains the brute damage, the plasma is exploding the contacted areas. They'd just be called "laser swords" as that are what they are most popularly known as, not necessarily strictly true of how they function.

Plasma weapons exist too and cause burn damage though, not that your justification doesn't work

Kurgis:
> Plasma weapons exist too and cause burn damage though, not that your justification doesn't work

I was referring to the state of matter, not the magical purple rock. Like a super-heated gas, that's contained in the shape of a sword through gravity/forcefield technology.

driecg36:
Ashing limbs with a laser sword is pretty cool IMO, even if its not the same as star wars

Humonitarian:
> I was referring to the state of matter, not the magical purple rock.

That is what plasma guns are, atm. They require no plasma to produce or use, they just shoot hot balls, presumably heated gas.

Firefox13:
Given we get #7999 I have no issues with it now. Very cool and unique.

Firefox13 (approved):

PR #7955, opened 2022-12-31
Remove RNG hit chance at close range

ValoTheValo:
## About The Pull Request
Kills RNGesus, Removes the accuracy debuff you get at close range for some reason.  why was this even added?
(curse you github for not letting me embed images) https://cdn.discordapp.com/attachments/534648497861951488/1057496131455680573/TacticalCombat.png

## Why It's Good For The Game
having a random miss chance at close range generally was counterintuitive and is annoying to deal with. without it combat should feel less loose and random and more tight and skillbased

## Testing
i compiled then dumped 9 mags into a test dummy

## Changelog
:cl:
del: Removed accuracy debuff at close range
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -279,7 +279,7 @@
 		recoil = leftmost_bit(recoil) //LOG2 calculation
 	else
 		recoil = 0
-	distance = distance <= 3 ? 5 - max(1,distance) : leftmost_bit(distance)
+	distance = leftmost_bit(distance)
 
 	def_zone = ran_zone(def_zone, 100 - (distance + recoil) * 10)
 
@@ -288,7 +288,6 @@
 	var/hit_mod = 0
 	switch(target_mob.mob_size)
 		if(120 to INFINITY)
-			hit_mod = -6
 		if(80 to 120)
 			hit_mod = -4
 		if(40 to 80)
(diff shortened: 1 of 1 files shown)


code/modules/projectiles/projectile.dm
```diff
@@ -288,7 +288,6 @@
 	var/hit_mod = 0
 	switch(target_mob.mob_size)
 		if(120 to INFINITY)
-			hit_mod = -6
```
Humonitarian, line 291: What's up with this one?

ValoTheValo: from what i could tell that was what was generating the random missses. im not really a coder so i may be wrong

Humonitarian: this one increases the hit chance if a mob is bigger than Iriska (megafauna and mechs)

ValoTheValo: > this one increases the hit chance if a mob is bigger than Iriska (megafauna and mechs)

so a negative hit mod indicates decreased chance to hit. wierd

ValoTheValo: reverted that part

Humonitarian (approved):

hyperioo (commented):
> Removes the accuracy debuff you get at close range for some reason. why was this even added?

I guess the idea was that when something is close to you, if it moves one meter, it's a big change from where you are aiming. If it's 10 meters away, then it barely moves and you don't have to adjust your aim that much.

ValoTheValo:
> > Removes the accuracy debuff you get at close range for some reason. why was this even added?
> 
> I guess the idea was that when something is close to you, if it moves one meter, it's a big change from where you are aiming. If it's 10 meters away, then it barely moves and you don't have to adjust your aim that much.

Still just makes combat generally more annoying to do.

hyperioo (approved):

PR #7964, opened 2023-01-11
Microfusion gen buff

driecg36:
## About The Pull Request

Buffed Microfusion gen. Increased output by a factor of ten (300KW to 3MW) and increased charging speed by a factor of 100 (500W per tick to 10KW per tick) 

## Why It's Good For The Game

Considering how difficult it is to get the microfusion generator, it was rather underwhelming in its previous state, being able to barely power a single department and maybe half another. This takes it to about 1/4 the standard power output of the SM so that the reward for this piece of tech is less negligible. 

## Testing

image

Yup.

## Changelog
:cl:
balance: Buff Onestar Microfusion Generator
/:cl:

### DIFF
--- a/code/modules/onestar/generator.dm
+++ b/code/modules/onestar/generator.dm
@@ -7,8 +7,8 @@
 	power_gen = 0						// Watts output per power_output level
 	var/can_generate_power = FALSE
 	var/is_circuit_fried = FALSE		// Doesn't actually fry anything, maybe it should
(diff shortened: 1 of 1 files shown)


ValoTheValo:
Based PR

driecg36:
> Based PR

Only took me like 3 months to remember you told me I could do this

Pink-Chink (approved):
Max power might be a bit too high, but I think it'd be better to see this thing get used first and adjust the values from there, if needed.

PR #7974, opened 2023-01-17
Space weather tweaks and fixes

Pink-Chink:
## About The Pull Request

- Moves the parallax image logic off of update(), which is called each move.
- Changes chance-based space weather triggers to tick-based.
- Reduces the storyteller weight of all space weather events, making them much less frequent. Also, limits each space weather event to one per round.
- Commented out photon vortex until a more performant solution is coded.

## Why It's Good For The Game

Minor optimizations. Space weather is supposed to be splashy, but the novelty wears off fast when we see them multiple times in a round with repeats.

## Testing

- Force spawned each event.

## Changelog
:cl:
del: Removed photon vortex from the pool
tweak: Space weather behavior is now tick-based
tweak: Space weather event weight reduced, limited to one per round for each event
code: Parallax icon logic moved to event subsystem
code: Spell-checked a variable
/:cl:

### DIFF
--- a/code/_onclick/hud/parallax.dm
+++ b/code/_onclick/hud/parallax.dm
@@ -13,11 +13,13 @@ GLOBAL_VAR_INIT(random_parallax, pick("space0", "space1", "space2", "space3", "s
 /obj/parallax/New(mob/M)
 	owner = M
 	owner.parallax += src
+	SSevent.all_parallaxes += src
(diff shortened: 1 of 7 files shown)


TheShown911 (approved):
I think they should be able to trigger more than once per round but good anyways.

PR #7976, opened 2023-01-19
NT Scourge buff

TheLeopold:
## About The Pull Request
Reduces NT Scourge's size to make it single-tap when folded.

## Why It's Good For The Game
Making faction-specific sword that takes 225 Armament points to get the disk for, better than a junkblade that is made roundstart sounds reasonable to me. Also look at its sprite, it's literally the same size of a Gladius, only makes sense to be as fast.

## Testing
Spawned Scourge, disciple, hit disciple with a sword, extended sword, hit disciple with a sword. Works as expected.

## Changelog
:cl:
tweak: NT Scourge now smaller
balance: NT Scourge is single-tap when folded
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/nt_melee.dm
+++ b/code/game/objects/items/weapons/nt_melee.dm
@@ -100,7 +100,7 @@
 	var/agony = 20
 	var/agony_extended = 45
 	var/stun = 0
-	w_class = ITEM_SIZE_BULKY
+	w_class = ITEM_SIZE_NORMAL
 	price_tag = 1000
 	matter = list(MATERIAL_BIOMATTER = 50, MATERIAL_STEEL = 5, MATERIAL_PLASTEEL = 2)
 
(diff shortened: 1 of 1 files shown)


Firefox13 (changes requested):
Use `no_double_tact` variable instead

TheLeopold:
> Use `no_double_tact` variable instead

So it's intended to be bulky when folded?
Asking because of the sprite. As I said, sprite of Scourge is really similar to Gladius. And Gladius is ITEM_SIZE_NORMAL
Screenshot_1

Firefox13:
You can't exactly make swords look very different in 32x32.

TheLeopold:
But you can make one sword clearly bigger than other
изображение

Firefox13:
OK
I dunno then

Firefox13 (approved):

PR #7983, opened 2023-01-22
Stiffness and obscuration removal

Iglpigl:
## About The Pull Request

Removes(comments out) stiffness and obscuration.

## Why It's Good For The Game

Stiffness creator is fine with it's removal, there are better ways to balance armors.
Screenshot_1

## Testing

Everything works as it should on local(no runtimes, both are fully removed, nothing unrelated should be changed).

## Changelog
:cl:
del: Removed stiffness and obscuration
/:cl:

### DIFF
--- a/code/__DEFINES/items_clothing.dm
+++ b/code/__DEFINES/items_clothing.dm
@@ -218,7 +218,7 @@
 #define STORY_MALFUNCTION "Malfunction"
 
 //Recoil for suits
-/*define LIGHT_STIFFNESS 0.5
+/*#define LIGHT_STIFFNESS 0.5
 #define MEDIUM_STIFFNESS 1
 #define HEAVY_STIFFNESS 1.5*/
 
--- a/code/__DEFINES/items_clothing.dm
+++ b/code/__DEFINES/items_clothing.dm
@@ -228,9 +228,9 @@
 #define HEAVY_SLOWDOWN 1
 
 //Offset for helmets
-#define LIGHT_OBSCURATION 2
+/*#define LIGHT_OBSCURATION 2
 #define MEDIUM_OBSCURATION 3
-#define HEAVY_OBSCURATION 5
+#define HEAVY_OBSCURATION 5*/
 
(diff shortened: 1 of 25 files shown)


Humonitarian:
My concern would be RIGs being too weak, but right now they are heavily overbuffed due to the issues we had with agony. So no need to worry about that one.

As for crusader, light slowdown gonna make them probably the best department armor, albeit pretty fair considering the buffs the Moebius voidsuit gets with this as well.

Eitherways if any armor becomes unbearably unbalanced after this, we can nerf it later.

Humonitarian (approved):

Firefox13:
изображение
What do you mean by this? Are there any runtimes? Is the stiffness gone absolutely? This is not how you describe testing

Firefox13:
Also why did you use /* and */ instead of // for single line commenting out.

ValoTheValo:
Very based

ValoTheValo:
Also if you want the system disabled you'd also need to comment out obscurity on helmets. It's essentially the same thing but for helmets

SaltyOldBull:
holy mother of based

Iglpigl:
> Also if you want the system disabled you'd also need to comment out obscurity on helmets. It's essentially the same thing but for helmets

Addressed.

Iglpigl:
Also all of your concerns addressed too, Firefox.

hyperioo (commented):
I see no real point in leaving that commented in the code. If the stiffness/obscuration game mechanics is scrapped and replaced by an adjustment of the slowdown, then we can just remove everything.

If someone ever wants to put it back, they will be able to retrieve everything just by reverting the PR.

Iglpigl:
> I see no real point in leaving that commented in the code. If the stiffness/obscuration game mechanics is scrapped and replaced by an adjustment of the slowdown, then we can just remove everything.
> 
> If someone ever wants to put it back, they will be able to retrieve everything just by reverting the PR.

Taken care of.

PR #8000, opened 2023-02-05
Makes human mobs be a high-priority process , misc code improvements

MLGTASTICa:
## About The Pull Request
Humans are high-priority on mob processing , they process before every mob now.
Added framework for additional control over how stuff is processed.
Weather no longer processes , since it did nothing ?
Machinery is now a processing SS , instead of a normal one.

## Why It's Good For The Game
Humans no longer lag , more control is good , weather  just lagged a bit , and machinery should've always been a processing SS

## Testing
Human mobs were tested along with machinery and space weather.  Humans always ticked while mobs were prone to missing as per usual. 
No functionality changes for machinery and spce weather

## Changelog
:cl:
fix: Fixed  humans lagging to extreme levels
refactor: SSweather no longer processes.
refactor: SSmachines is now a processing subsystem 
code: Framework for more control over Subsystem processing handling
/:cl:


code/controllers/subsystems/machines.dm
```diff
@@ -47,37 +46,26 @@ SUBSYSTEM_DEF(machines)
...
-/datum/controller/subsystem/machines/stat_entry()
+/datum/controller/subsystem/processing/machines/stat_entry()
 	..("M:[processing.len]|PN:[powernets.len]")
 
-/datum/controller/subsystem/machines/fire(resumed = FALSE)
+/datum/controller/subsystem/processing/machines/fire(resumed = FALSE)
 	if (!resumed)
 		for(var/datum/powernet/powernet as anything in powernets)
 			powernet.reset() //reset the power state.
 		src.currentrun = processing.Copy()
```
Pink-Chink, line 56:
```suggestion
```

Pink-Chink:
Rad storms still work with the SSweather changes?

code/controllers/subsystems/machines.dm
MLGTASTICa: copy is needed since processing works off reducing list length gradually.

MLGTASTICa:
> Rad storms still work with the SSweather changes?

yes , the SS did nothing in its own processing.

code/controllers/subsystems/machines.dm
Pink-Chink: It's not needed here since the parent proc does the same thing.

Pink-Chink:
Also, looks like tgui.bundle.js was changed.

code/controllers/subsystems/machines.dm
MLGTASTICa: fair point

MLGTASTICa:
> Also, looks like tgui.bundle.js was changed.

its TGUI updating libraries i believe,  Feel free to remove the change if you can find it since my git can't open it and show the modification.

hyperioo (approved):

hyperioo:
I checked and js changes are related to smelter interface so it should be fine

PR #8003, opened 2023-02-08
Buckshot fix

Humonitarian:
## About The Pull Request

Fixes a bug related to buckshot. Cannot confirm if this fixes all buckshot related issues, but it fixes the biggest one.

The exact bug:
Bullets lose damage due to armor calculations when penetrating humans. This affects buckshot, per pellet. Since buckshot works in stacks, each pellet affected the stack individually, causing most shots to lose all of their damage before they could even attempt to hit the player.

## Why It's Good For The Game

Buckshot sucks due to bugs crippling it.

## Testing

Compiled, I could fire a buckshot and it did damage.

## Changelog
:cl:
fix: fixed a bug related to buckshot
/:cl:

### DIFF
--- a/code/modules/mob/living/living_defense.dm
+++ b/code/modules/mob/living/living_defense.dm
@@ -135,8 +135,10 @@
 	// Returns if a projectile should continue travelling
 	if(return_continuation)
 		var/obj/item/projectile/P = used_weapon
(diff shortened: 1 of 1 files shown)


Humonitarian:
While this fixes an issue related to buckshot, they still seem to not penetrate armor properly. Further investigation is currently underway.

TheShown911 (approved):

PR #7770, opened 2022-11-06
Armour Protec (Armour reduces pain more.)

BlackjackIsTired:
## About The Pull Request

Raised pain resistance given by armour from 60% to 75%. (No more near paincrits from a single .40 rubber against bulletproof armour.)

## Why It's Good For The Game

Currently, the balance between Style and Armour is in a poor state leaning towards Style, this change should hopefully lessen the power difference as it will be harder to paincrit people wearing high amounts of armour.

(Free to discuss in comments if 80% is too high or too low, I personally feel like 85% is too high and 75% is too low. However regardless, I think this is a necessary change to make armour a more solid choice when considering your defensive options.)

## Changelog
:cl:
Changed pain block from 60% to 75%
/:cl:

PR #8008, opened 2023-02-21
Reverts through armor agony nerf

DiaFRAME444:
## About The Pull Request

Reverts #7770

## Why It's Good For The Game

The reason that #7770 was merged was because there was a huge misunderstanding in through-armor agony. 
A (now fixed) bug made it so if a attack was reduced to 0 damage, it'd resort to dealing agony damage to you according to your melee armor value, so 15 melee armor meant roaches could obliterate you in a few swings. People thought this was proof through armor agony was overtuned. That wasn't the case, and this new value is too low, so TGHmaxing makes through armor agony useless.

## Testing
No testing done for this one.

## Changelog
:cl:
balance: pain gained from armor blocking bullets is now 40% of the original damage. Should make shotguns viable again.
/:cl:

### DIFF
--- a/code/modules/mob/living/living_defense.dm
+++ b/code/modules/mob/living/living_defense.dm
@@ -1,4 +1,4 @@
-#define ARMOR_HALLOS_COEFFICIENT 0.25
+#define ARMOR_HALLOS_COEFFICIENT 0.4
 
 
 //This calculation replaces old run_armor_check in favor of more complex and better system


Humonitarian (approved):
A one-line change of a define, balance-wise approved.

assortedbeads:
So what does this mean for getting shot?

Will people go down quicker to bullet pain even in heavy armor? Cause they get knocked out pretty quick already.

Also why is it tested to see how quick they're getting koed compared to before the change?

DiaFRAME444:
It was always 0.4, so this should cause no issue.

Pink-Chink (approved):

PR #8013, opened 2023-03-01
Maintshrooms follow original design, now with 75% less cancer

SingingSpock:
## About The Pull Request

Fixes the code for maintshroom spreading, removing a block of code that was letting it spread through shutters, windows, and closed airlocks when that was not the intended effect. Maintshrooms will now only grow in places a human can move, which includes tables and low walls but not closed airlocks or lockers. Spacevines will still spread .through doors (the original point of the code), but not windows. Additionally adds a check to plant migration to check if the tile is over 0.5 lumens and stop the spread if so, matching all other maintshroom growth sources.

## Why It's Good For The Game

Maintshrooms were never *intended* to go through solid objects, as evidenced by the comments on the file. However, CanPass was always returning true, even in cases it wasn't intended to, and every case that it *should* return true Enter was already covering. With this fix, maintshrooms will still spread through open space, z-levels, and burrows, inevitably (but more slowly) consuming the maintenance tunnels if not combated, but should have far less purchase outside of them and not constantly invade the living spaces of the ship without any way to counter them beyond massively intensive extermination campaigns.

## Testing

- Shrooms don't go through closed airlocks or shutters (Check)
- Shrooms do go through open airlocks and shutters (Check)
- Shrooms don't go through windows (Check)
- Shrooms do go through low walls without windows (Check)
- Spacevines do still go through nonwelded doors (Check)
- Shrooms don't burrow travel to well lilt areas (Check)

## Changelog
:cl:
fix: Maintshrooms no longer treat solid objects as a mere suggestion
fix: Maintshrooms now check if the other end of a burrow is too bright for it to grow before spreading
/:cl:

### DIFF
--- a/code/controllers/subsystems/migration.dm
+++ b/code/controllers/subsystems/migration.dm
@@ -388,6 +388,7 @@ This proc will attempt to create a burrow against a wall, within view of the tar
 	while (i < plantspread_burrows_num && sorted.len)
 		var/obj/structure/burrow/C = sorted[1] //Grab the first element
 		sorted.Cut(1,2)//And remove it from the list
+		var/turf/simulated/T = get_turf(C)
 
 
 		//It already has plants, no good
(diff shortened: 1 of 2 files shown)


Alliostra:
SOVLLESS

TheShown911 (approved):

PR #8017, opened 2023-03-06
Minor bugs

SingingSpock:
## About The Pull Request

There's a few minor bugfixes and two tweak. Firstly, two mapped areas simply didn't work: a tiny maintenance room had its APC not connected to power, and a disposal and scrubber pipeline was entirely missing a segment. Secondly, the public storage welding and water tanks were moved a bit to a location less likely to get hit with random coins. Thirdly, an interface name and description were added to the rigsuit storage compartment so it doesn't show up as a "generic upgrade". Fourthly, the Gunnery Sergeant is moved up to WAGE_PROFESSIONAL because they're an experienced NCO and 2IC of Ironhammer and deserve to be paid on par with the specialist and investigator, rather than the exact same as the operatives.

Screenshot for the APC fix (The APC was originally on the same tile as the tech loot): 
image

Screenshot for the tank moving (the loot now in the corner was taken from the upper corner, total loot in the room is unchanged): 
image

Screenshot for the piping fix (the pipe used to be one higher, run into the wall above the airlock, and not have anything below the wall. To avoid making pipes under walls, I just moved the whole thing down):
image

## Why It's Good For The Game

Bugs are bad, people getting paid what they're worth is good.

## Testing

Compiled and ran it, all things work as intended.

## Changelog
:cl: SingingSpock
fix: Disposal/atmos piping above the gym no longer runs into a wall.
fix: A maintenance APC in a tiny room on deck 1 is actually placed where its wire is rather than never charging.
tweak: Tanks in public tools are protected from stray bullets/coins by a windoor.
fix: Harsuit storage module now has a name in the hardsuit interface rather than being generic
tweak: Gunnery Sergeant gets paid on par with the specialist/investigator rather than an operative.
/:cl:

### DIFF
--- a/code/game/jobs/job/security.dm
+++ b/code/game/jobs/job/security.dm
@@ -71,7 +71,7 @@
 	supervisors = "the Ironhammer Commander"
 	selection_color = "#a7bbc6"
 	department_account_access = TRUE
-	wage = WAGE_LABOUR_HAZARD
+	wage = WAGE_PROFESSIONAL
 	also_known_languages = list(LANGUAGE_NEOHONGO = 100)
 
 	outfit_type = /decl/hierarchy/outfit/job/security/gunserg
--- a/code/modules/clothing/spacesuits/rig/modules/storage.dm
+++ b/code/modules/clothing/spacesuits/rig/modules/storage.dm
@@ -2,6 +2,8 @@
 /obj/item/rig_module/storage
 	name = "internal storage compartment"
 	desc = "A storage container designed to be installed in a RIG suit. Allows a few items to be stored inside"
+	interface_name = "internal storage"
+	interface_desc = "A compartment within the suit allowing a few items to be stored."
 
 	var/obj/item/storage/internal/container
 	w_class = ITEM_SIZE_BULKY
--- a/maps/CEVEris/_CEV_Eris.dmm
+++ b/maps/CEVEris/_CEV_Eris.dmm
@@ -13590,11 +13590,6 @@
 /obj/structure/catwalk,
 /obj/structure/table/rack,
 /obj/spawner/pack/tech_loot,
-/obj/machinery/power/apc{
-	dir = 1;
-	name = "North APC";
-	pixel_y = 28
-	},
 /obj/machinery/camera/network/fourth_section{
 	dir = 8
 	},
@@ -14403,6 +14398,10 @@
 	d1 = 16;
 	d2 = 0;
 	icon_state = "16-0"
+	},
+/obj/machinery/power/apc{
+	name = "South APC";
+	pixel_y = -28
 	},
 /turf/simulated/floor/plating/under,
 /area/eris/maintenance/section4deck5starboard)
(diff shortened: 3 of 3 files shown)


Mycah142:
This should be atomized separately, so if there is any issue with, for example, the RIG defines, the sarge pay increase isn't also reverted.

Firefox13:
> fix: A couple map errors have been fixed.

Changelog is made to list all changes, no matter how small each one should be mentioned separately for better bookkeeping.

thevandie:
> Soulless. I don't like this.

I suggested it. I was going to add it myself but spock was making a map change, so we decided to add it. Humon gave me the go-ahead for doing it, we had an extensive conversation in #discussion about it.

Firefox13:
Just move the fuel tank then

Firefox13:
изображение
By "compiled and run" do you mean you just launched the server with new code or you looked at each change and made sure it works?
(it should be the latter case)

thevandie:
I don't think there's a spot in public tool storage that's covered in any way without being at the mercy of a vendor coin. One way windows are cool and fix the problem in an easy manner. I don't see what's soulless about that.

Firefox13:
Fuel tanks exist for two purposes:
* Refill welding tools
* Get shot
If it happens *that* much the issue is in the vendors which don't get repaired. Don't turn it into a hugbox

thevandie:
There's a difference between shooting the tank and sneezing on it, right now it can explode at the slightest touch
With windows yeah it'll stop a coin, but if a firefight breaks out it's gonna explode
It's just the lightest protection possible to stop the lightest of attacks, it's still volatile as hell, this just balances chaos with fun
Too much chaos and it's just unfun. The tanks exploding at a coin toss is just unfun.

maikilangiolo:
> right now it can explode at the slightest touch

Nah it explodes when it's shot, not touched, and neither when something is thrown at it.

thevandie:
Metaphors for exaggeration, we should really contact this in #discussion

SingingSpock:
@Mycah142 The commits are, in fact, atomized, and reverts can be done on the commit level.

@Firefox13 I have updated the CL to account for all the changes made. As for the testing, it is in fact the latter, I don't trust things I haven't looked at personally. As for the windoor change, it happens way too often as a result of just simple randomness rather than player agency. A minimum-level event should not be regularly setting off a bomb, and getting technomancers to fix anything that isn't an actual breach is like pulling teeth on this server. It could be moved, but this is a more elegant solution that still allows for it to be deliberately used as a weapon with a small amount of effort.

Firefox13 (changes requested):
I don't like the glass solution for the exploding fuel tank (I don't see it as an issue at all).
Please either remove it or move the tanj but don't add a literal hugbox for it.

Firefox13:
Also give map screenshots we don't have map preview bot

SingingSpock:
There you are

ValoTheValo:
also atomize your PRs i beg you

hyperioo (approved):
Seems the matters have been addressed. But please next time make several PRs instead of map fixes mixed with a pay change and a rig fix. You made separate commits but when we merge they're squashed so we cannot easily revert them individually.

PR #8018, opened 2023-03-11
Block people from leaving/entering shuttles through their ceiling

hyperioo:
## About The Pull Request

Prevent people from directly breaking into merc shuttle from the ceiling when it does not land on the upmost zlevel (since shuttle has no roof). Easy solution with a simple check. A better solution would be to spawn an opaque roof on top of the shuttle by that would required manipulating turfs to spawn the roof then return them to space when shuttle moves.

Also applies to other shuttles (but for now the problem never arises with them because their roof is covered and planets only have 1 zlevel).

Also prevent the bug where people could go inside the shuttle from under it.

## Why It's Good For The Game

Merc shuttle is not an open bar anymore.

## Testing

With a jetpack turned on, going up from inside the shuttle, then going down from above the shuttle.

image

## Changelog
:cl: Hyperio
add: Block people from leaving/entering shuttles through their ceiling
/:cl:

### DIFF
--- a/code/modules/multiz/movement.dm
+++ b/code/modules/multiz/movement.dm
@@ -81,6 +81,12 @@
 		to_chat(src, SPAN_WARNING("\The [destination] blocks you."))
 		return FALSE
 
+	// Prevent people from going directly inside or outside a shuttle through the ceiling 
+	// Would be possible if the shuttle is not on the highest z-level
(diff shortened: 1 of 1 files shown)


MLGTASTICa:
not the best fix , but then a proper one would require adding another layer between each z-level , so understandable i guess.

Pink-Chink (approved):

PR #8029, opened 2023-03-13
Erismed fixes and adjustments

Pink-Chink:
### Adjustments
- Burn damage will always transfer at least 25% damage to internal organs.
- Damage to limbs will only be deducted when the transferred damage creates/progresses a wound.
- Oxy damage doesn't contribute to pain. Organ wounds cause pain as a side effect and organ damage to certain organs can causes oxy damage. It doesn't make sense to make organ wounds apply pain from two sources.
- Unbreakable type bones don't break. Slime people are the only ones that use these bones and this change is needed to allow them to die.
- Blood loss organ wounds progression time increased to 1 minute per level instead of 18 seconds.
- Oxy damage from damaged organs reduced by 50% to 75% depending on the case.
- Oxygenation effect from Dexalin and Dexalin Plus don't stack with other effects.
- Glass Widow radiation damage reduced to 25 from 100.
- Atomik radiation damage reduced to 7.5 from 30.
- Demolecularisor laser (used by Prototype: biological demolecularisor) radiation damage reduced to 10 from 100.
- Demolecularisor laser clone damage increased to 12 from 5.
- Gestrahlte melee radiation damage reduced to 10 from 40.
- Gestrahlte bile projectile radiation damage reduced to 5 from 15.
- Malignant tumors can be removed with a cutting tool.
- Lexorin duration significantly reduced. 0.5u is consumed per tick instead of 0.01u.
- Stasis prevents suffocation/oxy damage.
- Toxins don't use the effect multiplier for their base toxin effect.
- Aranecolmin only applies rot wound once.
- Pararein strength reduced to 2 from 8.
- Created blood oxygenation modifier for heart processes. Previously, the chemical effect of Dexalin and Dexalin Plus were unused. This allows them to be used to aid in preventing the negative effects of blood loss. The new blood volume calculation in heart processing is:
  - `(current blood / species blood max) * ((heart efficiency / 100) + 0.4 * oxygenation effect - 0.2 * blood clotting effect)`

### Fixes
- Carrion core won't take damage while inside a human.
- Blood loss wounds won't stack and restoring blood levels will treat them.
- Brains don't suffer blood loss. Oxy loss already handles that behavior. Also, this may handle the issue where resuscitator wouldn't revive a body that should otherwise be revivable.
- Removed legacy lung rupture wound.
- Humans should no longer have a pulse when dead.
- Fixed runtime related to liver and kidney processes.
- Organ wound treatment does the progress bar and stat check before consuming charges. UI will update on a successful treatment.
- Fixed an issue where still live organs wouldn't heal after a wound was treated.
- Fixed an issue where succumb wouldn't let you die if you were too close to death.

## Why It's Good For The Game

Bug fixes and balancing

## Testing

Live test

## Changelog
:cl:
tweak: Burn damage will always transfer at least 25% damage to internal organs
tweak: Damage to limbs will only be deducted when the transferred damage creates/progresses a wound
tweak: Oxy loss doesn't contribute to pain
tweak: Unbreakable type bones don't break
tweak: Blood loss organ wounds progression time increased to 1 minute per level instead of 18 seconds
tweak: Oxy damage from damaged organs reduced by 50% to 75% depending on the case
tweak: Oxygenation effect from Dexalin and Dexalin Plus don't stack with other effects
tweak: Stasis prevents suffocation/oxy damage
tweak: Aranecolmin only applies rot wound once
tweak: Created blood oxygenation modifier for heart processes
tweak: Malignant tumors can be removed with a cutting tool
tweak: Toxins don't use effect multiplier for base toxin effect
balance: Glass Widow rad damage reduced to 25 from 100
balance: Atomik radiation damage reduced to 7.5 from 30
balance: Demolecularisor laser (used by Prototype: biological demolecularisor) radiation damage reduced to 10 from 100
balance: Gestrahlte melee radiation build-up reduced to 10 from 40
balance: Gestrahlte bile projectile radiation damage reduced to 5 from 15
balance: Lexorin duration significantly reduced, 0.5u is consumed per tick instead of 0.01u
balance: Pararein strength reduced to 2 from 8
fix: Carrion core won't take damage while inside a human
fix: Blood loss wounds won't stack and restoring blood levels will treat them
fix: Brains don't suffer blood loss
fix: Humans should no longer have a pulse when dead
fix: Fixed runtime related to liver and kidney processes
fix: Organ wound treatment does the progress bar and stat check before consuming charges, UI will update on a successful treatment
fix: Fixed an issue where still live organs wouldn't heal after a wound was treated
fix: Fixed an issue where succumb wouldn't let you die if you were too close to death
del: Removed legacy lung rupture wound
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -167,7 +167,7 @@ default behaviour is:
 
 /mob/living/verb/succumb()
 	set hidden = TRUE
-	if (health < 0) // Health below Zero but above 5-away-from-death, as before, but variable
+	if (health < 0)
 		adjustOxyLoss(health + maxHealth * 2) // Deal 2x health in OxyLoss damage, as before but variable.
 		health = -maxHealth
 		to_chat(src, "\blue You have given up life and succumbed to death.")
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -167,9 +167,9 @@ default behaviour is:
 
 /mob/living/verb/succumb()
 	set hidden = TRUE
-	if ((src.health < 0 && src.health > (5-src.maxHealth))) // Health below Zero but above 5-away-from-death, as before, but variable
-		src.adjustOxyLoss(src.health + src.maxHealth * 2) // Deal 2x health in OxyLoss damage, as before but variable.
-		src.health = src.maxHealth - src.getOxyLoss() - src.getToxLoss() - src.getFireLoss() - src.getBruteLoss()
+	if (health < 0) // Health below Zero but above 5-away-from-death, as before, but variable
+		adjustOxyLoss(health + maxHealth * 2) // Deal 2x health in OxyLoss damage, as before but variable.
+		health = -maxHealth
 		to_chat(src, "\blue You have given up life and succumbed to death.")
 
 
--- a/code/game/machinery/adv_med.dm
+++ b/code/game/machinery/adv_med.dm
@@ -244,7 +244,6 @@
 		"dermaline_amount" = H.reagents.get_reagent_amount("dermaline"),
 		"blood_amount" = round((H.vessel.get_reagent_amount("blood") / H.species.blood_volume)*100),
 		"disabilities" = H.sdisabilities,
-		"lung_ruptured" = H.is_lung_ruptured(),
 		"external_organs" = H.organs.Copy(),
 		"internal_organs" = H.internal_organs.Copy(),
 		"species_organs" = H.species.has_process, //Just pass a reference for this, it shouldn't ever be modified outside of the datum.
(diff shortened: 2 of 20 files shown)


Humonitarian:
/obj/item/gun_upgrade/barrel/isotope_diffuser

Make sure to nerf this too

code/modules/mob/living/living.dm
```diff
@@ -167,9 +167,9 @@ default behaviour is:
 
 /mob/living/verb/succumb()
 	set hidden = TRUE
-	if ((src.health < 0 && src.health > (5-src.maxHealth))) // Health below Zero but above 5-away-from-death, as before, but variable
-		src.adjustOxyLoss(src.health + src.maxHealth * 2) // Deal 2x health in OxyLoss damage, as before but variable.
-		src.health = src.maxHealth - src.getOxyLoss() - src.getToxLoss() - src.getFireLoss() - src.getBruteLoss()
+	if (health < 0) // Health below Zero but above 5-away-from-death, as before, but variable
```
hyperioo, line 170:
```suggestion
	if (health < 0)
```

hyperioo (approved):

PR #8046, opened 2023-03-20
Disks smartfridge

TheLeopold:
## About The Pull Request
Disks smartstorage. Needs to be mapped in, but Valo tweaking map, I can't make him wait another week.
изображение

## Why It's Good For The Game
qol. Reduces disks clutter.

## Testing
Spawned, filled, emptied

## Changelog
:cl:
add: Added disks smartstorage
imageadd: added filled states for disk storage
/:cl:


MLGTASTICa:
Why do the disks need to be cold

Humonitarian:
Why do the disks need to be cold

TheLeopold:
Why do the disks need to be cold

Loldaddy5676:
Why do the disks need to be cold

code/game/machinery/kitchen/smartfridge.dm
```diff
@@ -35,6 +35,24 @@
...
+	desc = "When you need disks fast!"
+	icon_state = "smartfridge"
+	icon_fill10 = "diskfridge-fill10"
+	icon_fill20 = "diskfridge-fill20"
+	icon_fill30 = "diskfridge-fill30"
+
+/obj/machinery/smartfridge/disks/accept_check(var/obj/item/O as obj)
+	if(istype(O,/obj/item/computer_hardware/hard_drive/portable))
+		return 1
+	return 0
```
hyperioo, line 52:
```suggestion
		return TRUE
	return FALSE
```

hyperioo (approved):

Humonitarian (changes requested):
Make it not require the same amount of power a fridge requires.

Pink-Chink:
Fine for now. We should think about getting a non-fridge looking sprite in the future.

PR #8056, opened 2023-03-24
Ore Trade Ship "Recoll" Adjustment

Mycah142:
## About The Pull Request
Adjusted Trade Ship Recoll's offers to only accept precious or exotic ore, and doubling their pay rate to compensate the removal of far easier to locate ore. Also changed ship description to match new details.

## Why It's Good For The Game
The guild, and miners, have no real reason to sell refined materials like plasteel to the ship denizens. So most don't even bother to refine a single thing, and just dump a crate of ore on the ground by the trade beacons without a second thought. This meant that currency had a greenbacked-like system where the cost of items being sold player to player were either flippant, or based on the trade ships which are purposefully overpriced.

By forcing basic mats to be smelted instead of the raw ore exported, this creates a more goldback-like system, where currency could be far easier to base on the cost of materials that will be more available.

The doubling of the remaining leftover ore may be reduced in the future, entirely reliant on the playerbase behavior.

## Testing
image

## Changelog
:cl:
tweak: tweaked trade ship to encourage more player material trading
/:cl:

### DIFF
--- a/code/modules/trade/datums/trade_stations_presets/1-common/material_refine.dm
+++ b/code/modules/trade/datums/trade_stations_presets/1-common/material_refine.dm
@@ -1,6 +1,6 @@
 /datum/trade_station/mat_refinery
 	name_pool = list(
-		"RS \'Recoll\'" = "Refinery Ship \'Recoll\': \"We accept all types of ores and we sell refined materials at cheap prices, as well!\""
+		"RS \'Recoll\'" = "Refinery Ship \'Recoll\': \"We accept precious or exotic ores and we sell refined materials in bulk, as well!\""
 	)
 	icon_states = list("htu_frigate", "ship")
 	uid = "materials"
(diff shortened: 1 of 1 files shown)


DiaFRAME444:
I implore you, if you're gonna do this please change it so you buy materials in stacks of 60 and halve their price.

maikilangiolo:
Guild mains on suicide watch

Mycah142:
> I implore you, if you're gonna do this please change it so you buy materials in stacks of 60 and halve their price.

The entire purpose of this is to have miners make materials more available to the ship at far cheaper cost than the trade ships. If I do that, it just removes miners as a whole.

thevandie:
> Good, we don't need bay leftovers.
So you're suggesting we remove both shuttles, the miner role, miner clothes, every tool designed for mining, and repurpose the miner equipment room? Just because it is tangentially related to bay?

Mycah142:
> Now disable all exports.

Vagabonds cry

Humonitarian (commented):
I'm not sure I am a fan of these changes. While it will make basic materials more affordable, the high tier ones like gold will become much less worth selling directly to crew.

Mycah142:
> I'm not sure I am a fan of these changes. While it will make basic materials more affordable, the high tier ones like gold will become much less worth selling directly to crew.

I originally was going to leave gold and silver out of the offers as well but realized the rarity of these ore, and of platinum meant that the export ores I had left were just not common enough for Miner to even make any offers for the guild.

PR #8058, opened 2023-03-25
Revert "recycling matter oversight"

ValoTheValo:
Reverts discordia-space/CEV-Eris#8037

why is it good for the game:
cruciforms are now currently melted by the bioreactor and broken down into components. this happens because they were given materiel value

### DIFF
--- a/code/datums/autolathe/basic_stuff.dm
+++ b/code/datums/autolathe/basic_stuff.dm
@@ -39,7 +39,7 @@
 	build_path = /obj/item/light/bulb
 
 /datum/design/autolathe/misc/ashtray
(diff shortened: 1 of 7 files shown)


ValoTheValo:
actually ill make a PR just to remove it from cruciform. not other stuff

TheLeopold:
but cruciforms **had** material valie! It was 1 stell 1 glass inherited from implants

ValoTheValo:
> but cruciforms **had** material value! It was 1 stell 1 glass inherited from implants

they didnt get recycled by bioreactor

TheLeopold:
I think it's because I added biomatter

PR #8064, opened 2023-03-27
Neotheology Taint

Humonitarian:
## About The Pull Request

Implements NeoTheology taint. A quality that everything containing biomatter possesses, and is a requirement for many effects of the church.

## Why It's Good For The Game

Lore

## Testing

Compiled and ran locally, used variable viewer to confirm everything was in order.

## Changelog
:cl:
add: Biomatter taint
/:cl:


code/modules/mob/living/carbon/slime/life.dm
```diff
@@ -228,7 +228,7 @@
 
 				if(ishuman(L)) //Ignore slime(wo)men
 					var/mob/living/carbon/human/H = L
-					if(H.species.name == "Slime")
+					if(H.species.name == SPECIES_SLIME || H.species.name == SPECIES_TAINTED)
```
hyperioo, line 231: So everyone with a cruciform will be untargeted by slimes?

hyperioo (commented):
Can you describe a bit more the content of the PR?

Why do dummy mannequins get godblood mutation? 'they're only admin spawnable right?)

What's with the delayed gib?

Also changing tainted people to a new species might have a lot of edge effects, like for instance:
```
/datum/antag_contract/item/blood/check_contents(list/contents)
	var/list/samples = list()
	for(var/obj/item/reagent_containers/C in contents)
		var/list/data = C.reagents?.get_data("blood")
		if(!data || data["species"] != SPECIES_HUMAN || (data["blood_DNA"] in samples))
			continue
		samples += data["blood_DNA"]
		if(samples.len >= count)
			return TRUE
	return FALSE
```
means tainted NT people cannot fill this contract anymore.

code/modules/mob/living/carbon/slime/life.dm
Humonitarian: Depends on taint level

Humonitarian:
Those are edge case scenarios when taint triggers, specifically caused by cloning under the wrong astrological conditions.

Humonitarian:
Due to issues that came up during testing the patch, I'll have to redo this differently

Humonitarian:
Codebase should be ready for Taint attempt 2 after erismed 4

SirRichardFrancis (commented):
Biomatter taint sounds like a very interesting idea, but only if there's more to it than making churchies look like Simpsons characters.
Seriously, this is underwhelming feature-wise and despite what the "Why It's Good For The Game" sections says, I don't see how this expands on the lore.
Need screenshots for visual changes too.
The code is alright.

Humonitarian:
> Biomatter taint sounds like a very interesting idea, but only if there's more to it than making churchies look like Simpsons characters. Seriously, this is underwhelming feature-wise and despite what the "Why It's Good For The Game" sections says, I don't see how this expands on the lore. Need screenshots for visual changes too. The code is alright.

>He doesn't know

SirRichardFrancis:
> He doesn't know

So... that's a meme? Damn.

Humonitarian:
> > He doesn't know
> 
> So... that's a meme? Damn.

There's a design doc for this, but it is way more complicated than turning NeoTheology into old-school Lego figures lol

It's not related to Taint either, Taint would be for NT rituals. Biomatter golems are designed to be a lategame product of one of the Tenets (expansion to NT's RND)

PR #8069, opened 2023-03-29
Actual Image GC fix

Pink-Chink:
## About The Pull Request

When blood overlays on items are created, they are saved in a global list so we only need to create the overlay once. When items are deleted, they call qdel() on their reference to the overlay. This has two problems: 1) the overlay var  is intended to be a reference and 2) the overlay has to be re-created when another item of the same type gets bloody.

- Fixes an issue where an image reference was being deleted when it should have been nulled

## Why It's Good For The Game

Performance

## Testing

Test server

## Changelog
:cl:
fix: Fixes an issue where an image reference was being deleted when it should have been nulled
/:cl:


ValoTheValo:
does this mean image stash notes will finally return?

Pink-Chink:
I don't know what that is

ValoTheValo:
stashes used t

> I don't know what that is

there used to be versions of stashes that were images instead of cords but they caused lag and sometimes broke in a bad way so got commented out forever ago

PR #8086, opened 2023-04-06
[Ready]Codebase optimizations part 3: SSchunks strikes back

MLGTASTICa:
## About The Pull Request
Implements a SSchunks subsystem for handling fast search of hearers , signal receivers , mobs and sanity damagers
still heavily WIP since im adding support for containerization
Needs the SSjamming PR as a framework since it added codebase-related signals.

What does it do 
Replaces all targetLookups for all AI mobs from using GLOB.MOB_list to SSchunks,  reducing the average lookups from 200 to a median of 5 or 10.
Replaces all searches for GLOB.radio_listeners / GLOB.hearing_objs with SSchunks.hearers , which should cut down from checking 600 radios at once to just 20-30 in the nearby chunks.
Also removed a few spawns / sleeps in mob code , so that they don't lock up the SSmobs.

## Why It's Good For The Game
Massively cuts down searching times + less code to execute

## Testing
Walked around , said funny words,  tested visibility , fought roaches n mobs

## Changelog
:cl:
refactor: Refactored how all AI-mobs lookup for their targets, to cut down on lag.
refactor: Refactor how all visible message, said messages search for hearers , to cut down on lag.
/:cl:


code/game/machinery/jammers.dm
```diff
@@ -0,0 +1,147 @@
+#define OVERKEY_JAMMER_ON "JAM_ON"
+#define OVERKEY_JAMMER_OFF "JAM_OFF"
+
```
Penwin0, line 3: Is this file supposed to be here at all?

Penwin0: Seems rather out of scope of the PR

code/datums/components/overlay_manager.dm
```diff
@@ -0,0 +1,40 @@
+/// Manages overlays based off a key
+// IT USES ONLY STRING KEYS!
+
```
Penwin0, line 3: Same here, seems weird to have this in this PR

Penwin0 (commented):
There's a lot of random shit cluttering this PR that has nothing to do with the new SS, atomization would have been ideal

MLGTASTICa:
> There's a lot of random shit cluttering this PR that has nothing to do with the new SS, atomization would have been ideal

Read the PR description.

Penwin0:
> Read the PR description.

Which part? I'm just saying this is going to be a pain in the ass to properly review for the poor maintainer that has to, and it needn't have been so.

MLGTASTICa:
> > Read the PR description.
> 
> Which part? I'm just saying this is going to be a pain in the ass to properly review for the poor maintainer that has to, and it needn't have been so.

The part where it mentions it uses the SSjamming PR as a part due to the signals added there.

TheShown911 (approved):
What if we just live test it, what could go wrong?

MLGTASTICa:
> Contributor

It was already tested on test-server..

Firefox13:
Live-testing inbound

PR #8092, opened 2023-04-07
Revert "Allows tables to be flipped while mapping"

Pink-Chink:
Reverts discordia-space/CEV-Eris#8091

I'll PR the correct version another time. Note: `flipped` is not binary. If `flipped` is less than 0, the table cannot be flipped.

image


Firefox13 (approved):
We *don't* really need this at the moment and it's breaking things.

Pink-Chink:
Integration failure was that dumb mouse hitting a trap

ValoTheValo:
> We _don't_ really need this at the moment and it's breaking things.

it was a fix for tables flipped via sdmm acting strangely in my PR. but its just a minor bug not worth breaking tables

PR #8095, opened 2023-04-09
Removes mob hit by bullet and X fires Gun messages

MLGTASTICa:
## About The Pull Request
 Removes all messages about being hit by a bullet and firing a gun
They create unnecesarry lag and strain on the server and lock people during combat
## Why It's Good For The Game

Prevent lockups during combat , especially from LMGs
We also got audio , visual and pain to show when you get hit by something , plenty enough.

## Testing
Went on local , pew pewed , no messages , still got visuals . audio , pain

## Changelog
:cl:
del: Removed mob hit by bullet and mob fires gun messages.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -475,7 +475,7 @@
 		playsound(user, fire_sound_silenced, 15, 1, -5)
 	else
 		playsound(user, fire_sound, 60, 1)
-
+		/*
 		if(reflex)
 			user.visible_message(
 				"<span class='reflex_shoot'><b>\The [user] fires \the [src][pointblank ? " point blank at \the [target]":""] by reflex!</b></span>",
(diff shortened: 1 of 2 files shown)


Humonitarian:
One step closer to chatbox-free Eris

code/modules/projectiles/gun.dm
```diff
@@ -475,7 +475,7 @@
 		playsound(user, fire_sound_silenced, 15, 1, -5)
 	else
 		playsound(user, fire_sound, 60, 1)
-
+		/*
```
Penwin0, line 478: Why keep it commented? Is this standard practice in this repo?

MLGTASTICa: kinda , and also to let anyone turn it back on if they get the idea to do so. for downstreams , etc.

TheShown911 (approved):
Yeah, I think people can realize they are being shot without the messages.

PR #8098, opened 2023-04-10
Removes commented/unused code and corrects a comment

Penwin0:
## About The Pull Request
Deletes a file that had "// TODO: Remove this from the game" at the top, filled with unused/commented code.
Deleted a list randomly declared at the top of a file that had nothing to do with it and was never referenced.
Corrects a comment to point you to where supplypacks actually are.

## Why It's Good For The Game
Guidelines say commented code is bad, and I agree, it was already marked for deletion.
I like when comments don't like.

## Testing
Compiled and looked around a bit, nothing with cargo is broken (which is shouldn't be, because this theoretically doesn't touch it)

## Changelog
:cl:
add: Added new things
del: Removes commented/unused code previously marked for deletion
fix: fixes a misleading comment in code
/:cl:


hyperioo (approved):
Thanks for the cleaning 👍

PR #8101, opened 2023-04-12
New litany: Search

TheLeopold:
## About The Pull Request
_(yay, another NT buff from me, who could have expected)_
Adds knock-off of inquisitor's Scrying litany for finding lost people: Search. All it does is tells the user the location of selected disciple. (similar to Entreaty)
Can be used by acolytes. Requires facing EOTP. Costs 25 energy. So it can't be used to reliably track murderer or someone hiding. It's also different from suit sensors or that removed Eye miracle, in that you have to be already looking for said person.
Target doesn't receive any notification, because 1984.

## Why It's Good For The Game
Gives NT real way of finding dead disciples, without giving suit sensors console to church or disabling the counterplay (keep low profile/move, remove cruciform/throw body in space)

## Testing
used litany, used litany facing EOTP, selected disciple.
I wasn't able test if it works with dead disciples, as I believe dead people needs ckey to be displayed. 

## Changelog
:cl:
add: Added new acolyte's litany: Search
/:cl:

PR #8111, opened 2023-04-19
Ultimate last shelter buff

TheLeopold:
## About The Pull Request
(Shitcode alert!
Please point where I made mistakes or forgot something)
If no ghost signed up for last shelter, it teleports random detached cruciform to self.
To teleport, cruciform should not be carried by person, be inside cruciform reader, BSDM or machines that can destroy crucifrm such as destructive analyzer or recycling vendor.

Added proc to check if atom is inside one selected things

Speacial thanks to Humon, Handy, and MLGTASTICa.

## Why It's Good For The Game
When I saw last shelter first time I assumed it works exactly like that (some people too). This adds expected behavior to shelter. After all if it can teleport cruciform from a deeps space why can't it teleport from the ship?

## Testing
cruciform on the floor, floor/bag, floor/bag/box, locker/bag/box, in hand, in pocket, in my bag, in my bag/box, locker/my hands, locker/my bag/box, every blacklisted machine

## Changelog
:cl:
balance: last shelter will teleport existing cruciforms if no ghost found
code: added is_inside proc
/:cl:


code/__HELPERS/_lists.dm
```diff
@@ -328,6 +328,11 @@
 		//Results will have a length of quality
 	return results
 
+//Pick a random element from the list.
+/proc/pick_random(list/L)
+	if(L.len)
+		var/picked = rand(1,L.len)
+		. = L[picked]		//Cut is far more efficient that Remove()
```
MLGTASTICa, line 335: You can just use pick(L) , it will automatically pick a random item from the list , no need to make a custom function.

code/game/objects/items/devices/last_shelter.dm
```diff
@@ -72,7 +72,9 @@ GLOBAL_DATUM(last_shelter, /obj/item/device/last_shelter)
 /obj/item/device/last_shelter/proc/get_cruciform()
 	var/datum/mind/MN = request_player()
 	if(!MN)
-		return FALSE
+		var/obj/item/implant/core_implant/CI = pick_n_take(lost_cruciforms)
+		CI.forceMove(get_turf(src))
+		return CI
```
MLGTASTICa, line 77: You might want to check if its in an inventory or being held. ForceMoving a object while its in an inventory causes a shadow-item to remain (and gets stuck in the inventory/hand , letting people teleport it to themselves on drop or throw remotely)

Firefox13:
This was proposed before and was deemed very OP.

TheLeopold:
I'm trying to make less
for instance it won't teleport cruciforms that are currently carried by someone, so counterplay is still there

DiaFRAME444:
> This was proposed before and was deemed very OP.

This can't be OP given it'll never come up in the real game.

TheLeopold:
How do I get rid of Dusty?!

TheLeopold:
_Anyway. It works, even if summons coder demon in the process._
Now I can explain why it's not OP, and in fact can only rarely happen.
Because if NT is being killed, the murderer will either leave the body as is (which won't allow teleportation) or remove cruciform and destroy it, and at no point in time cruciform can teleport, because it's either in inventory or in machine which can destroy cruciform.
The only circumstances of it teleporting cruciform of murdered disciple is
A. Murderer for some reason decided to hide cruciform, instead of destroying it
B. Trigger perfectly timed between ejection of cruciform and picking it up. (will never happen in real game, trust me)

ValoTheValo:
Good PR, last shelter now doesnt feel half assed and worthless

hyperioo (approved):
I guess we can try it out with current limitations.

hyperioo (commented):
After discussion with other members of staff, it seems to have little use considering your new Search litany #8101 to find NT corpses.

hyperioo (changes requested):
But feel free to make another PR for the `is_inside` proc, it can be useful.

TheLeopold:
Well that's true. Its only use is to find lost cruciforms.
Which is stated in its description...

Mycah142:
> After discussion with other members of staff, it seems to have little use considering your new Search litany #8101 to find NT corpses.

It cannot find corpses without cruciforms. So if roaches eat a body, leaving just the ID and cruciform, Search does nothing.

TheLeopold:
The main and whole point of this was to make it easier to find a detached cruciform. (I still have that gibbed techno flashback)
When I added Search, I treated only as a temporary solution to finding people before I can properly do this PR. Because it was far easier to implement and I wasn't sure if this will even work.
Now that it works I can delete Search and make Last Shelter rip cruciforms out of dead people. Or just delete Search.
Intact corpses have suit sensors, after all.

TheLeopold:
Nevermind. I just burnt out. Opening it in case you change your mind.

PR #8120, opened 2023-04-19
Increase shrapnel removal damage.

MLGTASTICa:
## About The Pull Request
Increases shrapnel removal damage multiplier from 0.1 to 1
Fail multiplier increased from 0.3 to 1.2

## Why It's Good For The Game
Katana with 50 melle damage dealt a whopping 5 brute on shrapnel removal, and 15 on fail.
We also have throwing knifes/kitchen knifes with abysmally low damage for removing shrapnel with minimal risk.

## Testing
Tested locally , shrap removal ran fine.

## Changelog
:cl:
balance: Shrapnel removal multiplier changed from 0.1 on succes to 1 , and from 0.3 to 1.2 on fail
/:cl:

### DIFF
--- a/code/modules/surgery/organic.dm
+++ b/code/modules/surgery/organic.dm
@@ -392,7 +392,7 @@
 	)
 	var/obj/item/shrapnel = locate(/obj/item/material/shard/shrapnel) in organ.implants //will succeed regardless
 	organ.remove_item(shrapnel, user, FALSE)
-	organ.take_damage(tool.force * 1.2, sharp=TRUE)
+	organ.take_damage(tool.force * 1, sharp=TRUE)
 
 //Cauterizing a wound to stop bleeding
 /datum/surgery_step/close_wounds
--- a/code/modules/surgery/organic.dm
+++ b/code/modules/surgery/organic.dm
@@ -383,7 +383,7 @@
 	)
 	var/obj/item/shrapnel = locate(/obj/item/material/shard/shrapnel) in organ.implants
 	organ.remove_item(shrapnel, user, FALSE)
-	organ.take_damage(tool.force, 0, sharp=TRUE) //So it's a bad idea to remove shrapnel with a chainsaw
+	organ.take_damage(tool.force * 0.5, 0, sharp=TRUE) //So it's a bad idea to remove shrapnel with a chainsaw
 
 /datum/surgery_step/remove_shrapnel/fail_step(mob/living/user, obj/item/organ/external/organ, obj/item/tool)
 	user.visible_message(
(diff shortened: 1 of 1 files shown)


DiaFRAME444:
This guy just pushed multiple salt PRs.
If there ever was a time to stop merging whatever gets PR'd, that time is now.

Firefox13:
изображение
So we're back to square one

Alliostra:
cumshoe_chainsaw

thevandie:
Can we ban this guy? Like 3-5 salt prs???

thevandie:
Shrapnel removal is actually broken. The damage scales with your bio stat while also making it less likely for you to fail, so 90 bio stat could decap you instantly but almost never fail, and 5 bio stat would scratch you and always fail. But even on a fail the Shrapnel is always removed, it's if it does damage or not.

You should fix that instead of just increasing the damage overall.

MLGTASTICa:
> Shrapnel removal is actually broken. The damage scales with your bio stat while also making it less likely for you to fail, so 90 bio stat could decap you instantly but almost never fail, and 5 bio stat would scratch you and always fail. But even on a fail the Shrapnel is always removed, it's if it does damage or not.
> 
> You should fix that instead of just increasing the damage overall.

This is not a thing? Not sure where you got this idea.

TheShown911 (commented):
How about some middle ground, maybe 0.5 on success and 1 on fail?

hyperioo (changes requested):
☝️

code/modules/surgery/organic.dm
```diff
@@ -383,7 +383,7 @@
 	)
 	var/obj/item/shrapnel = locate(/obj/item/material/shard/shrapnel) in organ.implants
 	organ.remove_item(shrapnel, user, FALSE)
-	organ.take_damage(tool.force * 0.1, 0, sharp=TRUE) //So it's a bad idea to remove shrapnel with a chainsaw
+	organ.take_damage(tool.force, 0, sharp=TRUE) //So it's a bad idea to remove shrapnel with a chainsaw
```
hyperioo, line 386:
```suggestion
	organ.take_damage(tool.force * 0.5, 0, sharp=TRUE) //So it's a bad idea to remove shrapnel with a chainsaw
```

code/modules/surgery/organic.dm
```diff
@@ -392,7 +392,7 @@
 	)
 	var/obj/item/shrapnel = locate(/obj/item/material/shard/shrapnel) in organ.implants //will succeed regardless
 	organ.remove_item(shrapnel, user, FALSE)
-	organ.take_damage(tool.force * 0.3, sharp=TRUE)
+	organ.take_damage(tool.force * 1.2, sharp=TRUE)
```
hyperioo, line 395:
```suggestion
	organ.take_damage(tool.force * 1, sharp=TRUE)
```

PR #8123, opened 2023-04-20
Raises GYM tickets price from 50 to 200

MLGTASTICa:
## About The Pull Request
Raises the GYM tickets price from 50 to 200

## Why It's Good For The Game
The machines give between 15 and 20 points for the wanted combat stat, which is the amount which a overwhelming-upper high oddity would give. It being available for just 50 credits really doesn't stimulate any vaga-economy , so this should help with better motivating vagabonds (and anyone in general) to do jobs.

## Testing
Ran a test locally , vendor showed the price as 200 creds

## Changelog
:cl:
balance: Gym tickets now cost 200 credits , up from 50 credits.
/:cl:

### DIFF
--- a/code/game/machinery/vending.dm
+++ b/code/game/machinery/vending.dm
@@ -1939,7 +1939,7 @@
 		)
 
 	prices = list(
-		/obj/item/gym_ticket = 50,
+		/obj/item/gym_ticket = 200,
 		/obj/item/tool/hammer/dumbbell = 90,
 		/obj/item/reagent_containers/food/drinks/protein_shake = 150,//a total ripoff
 		/obj/item/reagent_containers/food/drinks/energy = 200,


Humonitarian:
Gym is for gymbros, not vagabonds.

Vagabonds find oddities. Gym is for those who don't, aka. jobbies.

Also gonna note all oddities give double stats.
Your average noble oddity outdoes gym, even if you don't have the freedom of choice.

DiaFRAME444:
I agree with this change but now I must comment because everyone is wrong.

> The machines give between 15 and 20 points for the wanted combat stat, which is the amount which a overwhelming-upper high oddity would give. 

The gym is subpar and outranked by the Techno-Tribalism Enforcer and Corpus Deus due to doubled oddity gains. Its only benefit is its modularity, due to being able to pick a singular stat every level up.

> Pay to Win

Do you know what that means?

> It being available for just 50 credits really doesn't stimulate any vaga-economy 
> Vagabonds find oddities. Gym is for those who don't, aka. jobbies.

Gym tickets are a very regular maintenance spawn, meaning vagas often get this for free.

ValoTheValo:
> Gym is for gymbros, not vagabonds.
> 
> Vagabonds find oddities. Gym is for those who don't, aka. jobbies.

you should note that gym tickets are a super common maint spawn right now

hyperioo (approved):

code/game/machinery/vending.dm
```diff
@@ -1939,7 +1939,7 @@
 		)
 
 	prices = list(
-		/obj/item/gym_ticket = 50,
+		/obj/item/gym_ticket = 200,
```
TheShown911, line 1942:
```suggestion
		/obj/item/gym_ticket = 2000000,
```

PR #8126, opened 2023-04-21
Custom Seed Packet Labels

Mycah142:
## About The Pull Request
Allows using a pen to relabel seed packets.

## Why It's Good For The Game
Helps a botanist keep track of seed variants by renaming them using their own system, which facilitates staggered generations of seeds that organize smart in the current dumbvendors.

## Testing
image

## Changelog
:cl:
add: Added pen labels for seed packets.
/:cl:

### DIFF
--- a/code/modules/hydroponics/seed_packets.dm
+++ b/code/modules/hydroponics/seed_packets.dm
@@ -282,3 +282,18 @@ var/global/list/plant_seed_sprites = list()
 
 /obj/item/seeds/thaadra
 	seed_type = "thaadra"
+
+/obj/item/seeds/attackby(obj/item/I, mob/user, params)
+	if(istype(I, /obj/item/pen))
+		var/new_name = input(user, "What would you like to label the seed packet?", "Tape labeling") as null|text
+		if(isnull(new_name)) return
+		new_name = sanitizeSafe(new_name)
+		if(new_name)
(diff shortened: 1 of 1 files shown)


TheShown911 (approved):

hyperioo (commented):
How do you wipe off the label? new_name cannot be null because of the check on line 289

Mycah142:
> How do you wipe off the label? new_name cannot be null because of the check on line 289

That is a good question. I assume similarly with how it's done via the data disk I stole the code from, although with Data Disk, the original "tag" is blank and thus starts null, so for this I assume renaming it the original name "wipes" the tag.

Considering the name is overwritten by doing this regardless of what you enter in the field, I guess it's an oversight.

hyperioo:
Well, you can test it and fix it then. 
sidecat

Maybe just adding something like `&& new_name != ""` is enough

Mycah142:
> Well, you can test it and fix it then. sidecat
> 
> Maybe just adding something like `&& new_name != ""` is enough

[dreamseeker_sfpVv6af7t.webm](https://user-images.githubusercontent.com/11076040/233801129-7bd53ccd-af4a-4669-b570-4b6f51c9dadd.webm)
m

Current functionality.

hyperioo:
Noice!

PR #8130, opened 2023-04-21
Replace scrap shotgun shells with standard ones in crafting

koghs:
## About The Pull Request

Made crafting shotgun ammo with ammo kits and crafting station produce standard quality ammo instead of scrap quality ammo

## Why It's Good For The Game

Raises availability of good quality ammo for under-performing firearm type (shotgun), appeals to common sense and realism (hand-loading shotgun shells is famously quite easy)

## Testing

Tested, change works as intended

image
image

## Changelog
:cl:
tweak: Crafting shotgun ammo with ammo kits and crafting station now produces standard quality ammo instead of scrap quality ammo
/:cl:


MLGTASTICa:
Vagabonds shouldn't be able to craft pristine ammo , theres a reason we have gun disks and autolathes.
Perhaps consider locking this behind a cognition check or more advanced materials (steel , plastic,  etc)

koghs:
> Vagabonds shouldn't be able to craft pristine ammo , theres a reason we have gun disks and autolathes. Perhaps consider locking this behind a cognition check or more advanced materials (steel , plastic, etc)

Special material requirements would overcomplicate ammo crafting while not achieving much (getting non-rare resources as a vagabond is very trivial) 
Cognition already determines how much ammunition you can craft
And paywalling shotgun ammo crafting behind something like "gunsmith" trait would make it completely useless
Although now I see that ability to craft non scrap ammo that has actual monetary value might be problematic by itself

Humonitarian:
The issue is shotgun ammo underperforming.

I'm not against this PR, but modulars guns are currently in the process of rebalancing how calibers compare to eachother in a major way, which might fix this issue as well.

For this reason, I'm not approving this PR either.

PR #8139, opened 2023-04-25
Disposable gun:   ...   Svalinn!

TheLeopold:
## About The Pull Request
Turned Svalinn into disposable gun
Comes preloaded with unremovable NT Spark 300 cell
Costs 18 biomatter, 9 steel to print. Loses 5 biomatter on recycling
Removed powercell from svalinn disk
Greatly reduced price of svalinn
Reduced price of NT energy pack

May or may not be unbalanced and too strong for its new cost. Or too cheap for its power. Balance is a timid beast and I'm not a numbers guy.
Tell me if you got better suggestions.

## Why It's Good For The Game
Disposable guns are cool. ~~Martyr was cool.~~ Only disposable gun is slaugh-o-matic.
I found it easier to just carry multiple preloaded svalinns than reload one. (in current state)
Church is Apple and you should buy new iGun when old one runs out of charge.

## Testing
printed, shot, recycled, tried to insert cell, remove cell, recharge svalinn

## Changelog
:cl:
tweak: removed cell from svalinn disk
balance: now svalinn is disposable
balance: reduced the price of NT energy pack crate
/:cl:


code/modules/cargo/packs.dm
```diff
@@ -201,7 +201,7 @@ var/list/all_supply_groups = list("Operations","Security","Hospitality","Enginee
 				/obj/item/gun/energy/taser,
 				/obj/item/gun/energy/nt_svalinn,
 				/obj/item/gun/energy/nt_svalinn)
-	cost = 4500
+	cost = 4000
```
Humonitarian, line 204: I think you can drop this to like 500 or less

TheLeopold: Make it 500? You mean 2 svalinns costed 4000?

Humonitarian (approved):
We talked about this on Discord, and I approve these changes.

a few QOL like making sure it fits in tactical belts would be handy too.

Mycah142:
We've gone full circle and reintroduced the poop gun.

Fun fact, custodian and acolyte lockers contain cells because they used to have disposable guns before melee.

Humonitarian:
> What if we just return Martyr with a better sprite?

Would overlap with the svalinn. It would take ingenious design to make both work at once, perhaps making one pretty much lategame

Humonitarian:
> > Would overlap with the svalinn
> > Would it? The Martyr was primarily a non-lethal pain inflicting plasma weapon with a generally ineffective one time use lethal plasma round
> > The Svalinn is a entirely lethal weapon with hitscan capabilities.

All lethal weapons work as ghetto nonlethals, and all nonlethals combined with a sword work as ghetto lethals

ValoTheValo:
just readd the martyr but with a non bad sprite this time, i know someone made a better sprite at some point

TheLeopold:
I don't know how related this is to the current state, but, I guess, it's something to consider.
изображение

Mycah142:
The Martyr was introduced before the Slaught-O-Matic and was "just worse compared to real guns". The SOM also has more bullets so it feels better as a disposable.

Humonitarian:
If someone were to sprite a new Martyr, either we add the martyr to replace the svalinn and make the svalinn the "energy avasarala" for lategame NT, or make the martyr a sort of single shot pipe gun for midgame NT.

Hopekz:
tumblr_inline_o9wkulQ0Gl1u977gz_400 (1)

Humonitarian (approved):

PR #8142, opened 2023-04-25
Metabolism changes

RikuTheKiller:
## About The Pull Request

This makes the volume of a reagent irrelevant when calculating how quickly it should metabolize. Previously, it was taking the overdose threshold into account and the math for that was based on volume, leading to cases where even though the metabolization rates were accounted for, the reagents wouldn't run out at the same time. This only applies to the bloodstream.

It also makes blood circulation affect ingested reagents, since reagents from the stomach end up in the bloodstream anyways and the absorption of reagents from the colon is assisted by blood flow to the area as well.

Also, I removed a safety check that caused reagents with a metabolism of 0 to still metabolize and synaptizine to metabolize twice as fast while in the stomach. Reaching 0 is practically impossible, not to mention managing to do it triggers a check in metabolization code that skips reagent effects entirely.

I've also removed REAGENTS_MIN_EFFECT_MULTIPLIER and REAGENTS_MAX_EFFECT_MULTIPLIER as they're not actually needed. Why? Because reagents in the stomach work just fine and they aren't clamped by these whatsoever.

Also, loads of code refactoring in the modified procs, reading them no longer gives you brain damage.

Bleeding is now affected by heart efficiency. Like, a lot.

## Why It's Good For The Game

It makes metabolization code easier to read and fixes seemingly correct reagent ratios not running out at the same time.

It also doesn't make much sense for a reagent to be processed faster if there's more of it gameplay-wise, overdoses on their own are intended to be the only reason to avoid taking lots of the same reagent.

Also, high and low metabolisms both carry significant upsides and downsides. Since these pretty much balance out, limiting them seems irrelevant considering how insane this codebase is by nature.

## Testing

Caused no abnormalities in a few basic tests. I can't really test everything since this has very high edge-case potential, make sure to TM it first as people may find unwanted exploits like infinite healing or something.

## Changelog
:cl:
balance: Chemicals no longer take their volume into account when processing. This should lead to far more predictable reagent behaviour.
balance: Removed all limits on reagent metabolization modifiers. This means giving yourself 10 hearts is even funnier now.
add: Having a highly efficient heart may cause elevated bleeding, exercise caution.
/:cl:

### DIFF
--- a/code/__DEFINES/chemistry.dm
+++ b/code/__DEFINES/chemistry.dm
@@ -31,9 +31,6 @@
 
 #define REAGENTS_OVERDOSE 30
 
-#define REAGENTS_MIN_EFFECT_MULTIPLIER 0.2
-#define REAGENTS_MAX_EFFECT_MULTIPLIER 2.5
-
 #define CHEM_SYNTH_ENERGY 3000 // How much energy does it take to synthesize 1 unit of chemical, in Joules.
 
 #define CE_STABLE			"stabilization"		// Inaprovaline
--- a/code/modules/organs/blood.dm
+++ b/code/modules/organs/blood.dm
@@ -85,6 +85,7 @@
 					blood_max += W.damage * WOUND_BLEED_MULTIPLIER
 		if (temp.open)
 			blood_max += OPEN_ORGAN_BLEED_AMOUNT  //Yer stomach is cut open
+		blood_max *= get_organ_efficiency(OP_HEART) //Hey check out how hard I can bleed
 
 	// bloodclotting slows bleeding
 	if(chem_effects[CE_BLOODCLOT])
--- a/code/modules/reagents/reagents.dm
+++ b/code/modules/reagents/reagents.dm
@@ -69,20 +69,12 @@
 		return consumed_amount_human(M, alien,location) // Since humans should scale off livers , hearts and stomachs
 	var/removed = metabolism
 	if(location == CHEM_INGEST)
-		if(ingest_met)
-			removed = ingest_met
-		else
-			removed = removed/2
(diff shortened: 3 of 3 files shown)


hyperioo (approved):
Good to deploy on test server just in case

Humonitarian:
> since reagents from the stomach end up in the bloodstream anyways

This is missing from Changelog

Humonitarian:
Do reagents such as Dylovene and Soporific still metabolize faster the more there is in the bloodstream?

The overdose threshold was used as a system to purge them from the body faster, so you can't inject 100 units for permanent effects.

RikuTheKiller:
> Do reagents such as Dylovene and Soporific still metabolize faster the more there is in the bloodstream?
> 
> The overdose threshold was used as a system to purge them from the body faster, so you can't inject 100 units for permanent effects.

That should be addressed via giving stuff like dylo actual ODs, not via a band-aid that affects everything.

RikuTheKiller:
> > since reagents from the stomach end up in the bloodstream anyways
> 
> This is missing from Changelog

Why would I add something I simply used as reasoning and reference to the changelog?

Humonitarian:
> > > since reagents from the stomach end up in the bloodstream anyways
> > 
> > 
> > This is missing from Changelog
> 
> Why would I add something I simply used as reasoning and reference to the changelog?

I'm sorry, I thought you were implementing that from other codebases.
Our reagents metabolize within the stomach, without moving to the bloodstream.

RikuTheKiller:
> I'm sorry, I thought you were implementing that from other codebases. Our reagents metabolize within the stomach, without moving to the bloodstream.

I'm implementing it later.

PR #8143, opened 2023-04-26
Advanced and normal slime mutation toxins no longer instantly convert people

MLGTASTICa:
## About The Pull Request
You need 10 units and 40 seconds to convert someone to a slime person
You need 15 units and 60 seconds to covnert someone to a slime
The process is now done with messages ,warning the target of the transformation progress.

## Why It's Good For The Game
Insta-transformations are cringe overall.
## Testing
Tested locally , works  fine.

## Changelog
:cl:
balance: Slime mutation toxin now needs 10 units and 40 seconds to convert
balance: Advanced slime mutation toxin now needs 15 units and 60 seconds to convert.
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/toxins.dm
+++ b/code/modules/reagents/reagents/toxins.dm
@@ -591,6 +591,7 @@
 				if(20)
 					to_chat(M, SPAN_DANGER("Your skin starts burning from water!"))
 					M.adjustHalLoss(10)
+					M.visible_message("Emanates clouds of steam as the water from their skin starts boiling! ", range = 7)
 					var/datum/effect/effect/system/smoke_spread/smoke = new()
 					smoke.set_up(3, 0 , get_turf(M), FALSE)
 					smoke.start()
--- a/code/modules/reagents/reagents/toxins.dm
+++ b/code/modules/reagents/reagents/toxins.dm
@@ -521,6 +521,7 @@
 
 /* Transformations */
 
+#define UNITS_TO_CONVERT_TO_SLIMEPERSON 10
 /datum/reagent/toxin/slimetoxin
 	name = "Mutation Toxin"
 	id = "mutationtoxin"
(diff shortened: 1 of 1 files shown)


assortedbeads:
Anyway to make it respect gas masks?

RikuTheKiller:
> Anyway to make it respect gas masks? Also it shouldnt work on fbp IMO

Definitely shouldn't work on fbps. Making it retain prosthetics for humans would be neat as well.

Firefox13 (changes requested):
Remove complete slime mutation toxin.
While slimepersoning people forcefully is a big stretch for mechanics and lore literally nuggetting them is definitely not something we need.
You can replace it in xenobio code with the slimeperson toxin (in lower quantity or something I don't remember bay xenobio code)

Firefox13:
If we go "improve not remove" you could make advanced slime mutation toxin just change the color (its type) of the *slime* mob to a random one.

RikuTheKiller:
> If we go "improve not remove" you could make advanced slime mutation toxin just change the color (its type) of the _slime_ mob to a random one.

How about turning a slimeperson into a slime
The timers would compound
And having both in one container should revert the advanced to normal toxin

Firefox13:
As long as sliming deletes cruciform it still has to go.
The warning isn't a dealbeaker either. I don't think we even need slimepeople tbh.

RikuTheKiller:
> As long as sliming deletes cruciform it still has to go. The warning isn't a dealbeaker either. I don't think we even need slimepeople tbh.

Just make it not delete it. Plop it out, maybe?

assortedbeads:
> > As long as sliming deletes cruciform it still has to go. The warning isn't a dealbeaker either. I don't think we even need slimepeople tbh.
> 
> Just make it not delete it. Plop it out, maybe?

That would need a lot more reworking then, cause right now it
- breaks cruciforms
- turns fbps
- pierces closed voidsuits
- pierces gas masks (?)

I dont mind slime people but turning people into a simplemob just sucks for the person on the opposite end

RikuTheKiller:
> That would need a lot more reworking then, cause right now it
> 
> * breaks cruciforms
> * turns fbps
> * pierces closed voidsuits
> * pierces gas masks (?)
> 
> I dont mind slime people but turning people into a simplemob just sucks for the person on the opposite end

K fix it, i might do it

Mycah142:
> As long as sliming deletes cruciform it still has to go. The warning isn't a dealbeaker either. I don't think we even need slimepeople tbh.

This is still a progressive step in the right direction, and an improvement over the current situation.

Jakubek54:
> As long as sliming deletes cruciform it still has to go. The warning isn't a dealbeaker either. I don't think we even need slimepeople tbh.
I agree completely. Even if you get rid of the base slime there's still the issue of Slimeperson transformation completely fucking over all Cruciform users. The mechanic also isn't fun enough to justify being kept around in such a broken state.

PR #8160, opened 2023-05-09
Fixes for prosthetics and delimbing

AltHit:
Factions

## About The Pull Request

Fixes for some eldritch bugs with prostetics and one fresh ErisMed bug
New Erismed check verifies that limb is broken before tearing it off, which was a rare case for burn damage because it does not break limbs. Now this check is skipped for burn weapons (laser and plasma guns), so now they melt properly.
Also we restored armor values that were intended for all prosthetics by default, but it was forgotten somehow.
Most importantly prosthetics now use same conditions for limb separation as humans do, instead of instantly beheading you after 3 pistol shots.

## Why It's Good For The Game

Now everything works as our forefathers intended

## Testing

Dozens of crippled and brutally butchered humans and FBPs shown that changes are in fact working

## Changelog
:cl:
add: Armor values for prosthetics returned to their rightful place
balance: Indirectly changes balance by making prosthetics playable (probably)
fix: Limbs now melt from burn weapons again
fix: Prosthetics now do not delimb after ridiculously low damage
/:cl:


code/modules/organs/external/damage.dm
```diff
@@ -91,19 +91,28 @@
...
+						else if((amount + prev_brute) >= max_damage * DROPLIMB_THRESHOLD_DESTROY)
+							droplimb(FALSE, DROPLIMB_BLUNT)
+						else if((amount + prev_brute) >= max_damage * DROPLIMB_THRESHOLD_TEAROFF)
+							droplimb(FALSE, DROPLIMB_EDGE)
+						else if(BP_IS_ROBOTIC(src) && edge_eligible && (amount + prev_brute) >= max_damage * DROPLIMB_THRESHOLD_EDGE)
+							droplimb(TRUE, DROPLIMB_EDGE)
+						else if(BP_IS_ROBOTIC(src) && (amount + prev_brute) >= max_damage * DROPLIMB_THRESHOLD_DESTROY)
+							droplimb(FALSE, DROPLIMB_BLUNT)
+						else if(BP_IS_ROBOTIC(src) && (amount + prev_brute) >= max_damage * DROPLIMB_THRESHOLD_TEAROFF)
+							droplimb(FALSE, DROPLIMB_EDGE)
```
TheShown911, line 106: What do the extra `BP_IS_ROBOTIC(src)` checks do, shouldn't it be already covered by the earlier checks?

code/modules/organs/external/damage.dm
```diff
@@ -91,19 +91,28 @@
...
 				if(BURN)
 					if(edge_eligible && (amount + prev_burn) >= max_damage * DROPLIMB_THRESHOLD_EDGE)
 						droplimb(TRUE, DROPLIMB_EDGE_BURN)
 					else if((amount + prev_burn) >= max_damage * DROPLIMB_THRESHOLD_DESTROY)
 						droplimb(TRUE, DROPLIMB_BURN)
+					else if(BP_IS_ROBOTIC(src) && edge_eligible && (amount + prev_burn) >= max_damage * DROPLIMB_THRESHOLD_EDGE)
+						droplimb(TRUE, DROPLIMB_EDGE_BURN)
+					else if(BP_IS_ROBOTIC(src) && (amount + prev_burn) >= max_damage * DROPLIMB_THRESHOLD_DESTROY)
+						droplimb(TRUE, DROPLIMB_BURN)
```
TheShown911, line 115: Here too.

code/modules/organs/external/damage.dm
AltHit: Done. They are indeed were unnessesery. Commented those if anybody in the future will deside to create separate conditions for FBP's

code/modules/organs/external/damage.dm
AltHit: Those too.

hyperioo:
You can remove the commented lines, they will have the git history if they ever want to check how it was before.

TheShown911 (approved):

PR #7642, opened 2022-09-30
Removes home system and faction fields from the crew records program as they were always showing up empty

Iglpigl:
## About The Pull Request

Removes home system and faction fields from the crew records program as they were always showing up empty and Nanako ain't returning.

## Why It's Good For The Game

Empty fields bad.

## Changelog
:cl:
del: Removed two fields from the crew records console which were empty anyways
/:cl:

PR #8168, opened 2023-05-13
Revert "Removes home system and faction fields from the crew records program as they were always showing up empty"

Firefox13:
Reverts discordia-space/CEV-Eris#7642
Improve do not remove


Firefox13:
But seriously as far as I remember this broke something in records display.

And is fixable...

TheShown911 (commented):
If it's not tested it doesn't work.

Firefox13:
Yeah I know I'll get to this some other day.

Kurgis:
Hm, you can try plugging the Home System into character's Background selection. Faction's field is completely stripped though from character set-up, and I can't think of anything suitable that replaced it.

PR #8172, opened 2023-05-14
Nerfs nuka cola, hyperzine, and leg augmentation

Humonitarian:
## About The Pull Request

Reduces the effectiveness of nuka cola and hyperzine, makes nuka cola irradiate the consumer.

Muscle/nerve efficiency now gives diminishing returns, requiring significantly more effort for significantly less payout.
Little effort gives more payout than it used to, however.

## Why It's Good For The Game

Chems were overpowered, nuka cola was regularly exploited to gain an unfair advantage.
Diminishing augmentation good for the lowly consumer, bad for the powergamer.

Nerfs requested by Fox.

## Testing

Augmented character walked visibly faster when modded, but felt slower than it used to.

## Changelog
:cl:
tweak: nuka cola irradiates you
balance: nuka cola, hyperzine nerfed
balance: augmenting your leg with more muscle gives diminishing returns
/:cl:

### DIFF
--- a/code/modules/organs/external/_external.dm
+++ b/code/modules/organs/external/_external.dm
@@ -354,8 +354,8 @@
 	var/muscle_eff = owner.get_specific_organ_efficiency(OP_MUSCLE, organ_tag)
 	var/nerve_eff = max(owner.get_specific_organ_efficiency(OP_NERVE, organ_tag),1)
 	muscle_eff = (muscle_eff/100) - (muscle_eff/nerve_eff) //Need more nerves to control those new muscles
-	. += max(-(muscle_eff), MAX_MUSCLE_SPEED)
-
+	if(muscle_eff)
+		. += 0.6 * muscle_eff / (muscle_eff + 0.4) // Diminishing returns with a hard cap of 0.6 and soft cap of 0.5
 	. += tally
 
 /obj/item/organ/external/proc/is_nerve_struck()
--- a/code/modules/organs/external/_external.dm
+++ b/code/modules/organs/external/_external.dm
@@ -355,7 +355,7 @@
 	var/nerve_eff = max(owner.get_specific_organ_efficiency(OP_NERVE, organ_tag),1)
 	muscle_eff = (muscle_eff/100) - (muscle_eff/nerve_eff) //Need more nerves to control those new muscles
 	if(muscle_eff)
-		. += 0.6 * muscle_eff / (muscle_eff + 0.4) // Diminishing returns with a hard cap of 0.6 and soft cap of 0.5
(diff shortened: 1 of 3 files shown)


Jakubek54:
Salt will lead us to truth.

Humonitarian:
> Salt will lead us to truth.

So severe this time project lead went out of their way to ask for this. You guys really didn't hold back today.

ValoTheValo:
Extremely good PR

MemedHams:
You may want to replace the nuka-cola in energy drinks with hyperzine- or rename them to point out the rad risk ICly

MLGTASTICa:
We should be using an equation that makes the boosts be based on turfs per second , instead of arbitrary numbers. Aka hyperzine should only give a 20% boost based on base turfs per second instead of -0.5 tally.
Nuka cola also still has no downside for FBPs , which can freely chug it

Humonitarian:
> We should be using an equation that makes the boosts be based on turfs per second , instead of arbitrary numbers. Aka hyperzine should only give a 20% boost based on base turfs per second instead of -0.5 tally. Nuka cola also still has no downside for FBPs , which can freely chug it

Is there a proc to check if someone's an FBP?

MLGTASTICa:
> > We should be using an equation that makes the boosts be based on turfs per second , instead of arbitrary numbers. Aka hyperzine should only give a 20% boost based on base turfs per second instead of -0.5 tally. Nuka cola also still has no downside for FBPs , which can freely chug it
> 
> Is there a proc to check if someone's an FBP?

if the chest is robotic,  they're a FBP.

thedarkcrusader:
I think a much better solution  to have chems or organs that  interact  with the muscles  to directly  lower  the effectiveness  and  or damage  the muscles   while  bloodstream  or when there overexerting themselves  via continuous   movement. I think  it'll be interesting

thevandie:
> I think a much better solution to have chems or organs that interact with the muscles to directly lower the effectiveness and or damage the muscles while bloodstream or when there overexerting themselves via continuous movement. I think it'll be interesting

Code it then
Cause we need this fix asap
If you want it to be more mechanically interesting you can code it next week.

Firefox13:
Cobbychem? Everyone hated it.

No please.

lolman360:
> I think a much better solution to have chems or organs that interact with the muscles to directly lower the effectiveness and or damage the muscles while bloodstream or when there overexerting themselves via continuous movement. I think it'll be interesting

hyperzine causing muscle damage on ending metabolism (with amount of muscle damage scaling on how long it's been in your system) doesn't... sound that bad honestly
if it's a moderate dose it can be healed relatively easily (at least now that internal wounds can be healed with chems) but if you keep yourself on it for 30 minutes you're going to lose all your muscles which will be very funny

TheShown911 (approved):
It's so over.

PR #8192, opened 2023-05-28
Fix downstream grenade duping bug

WeNeedMorePhoron:
## About The Pull Request
This fix the bug as of 1 year ago on Eris where firing a grenade launcher at certain angle into wall lead to server crashing and duping. 

Can't replicate this anymore on current Eris code but replicated it downstream while porting it over on old code. This happen if you fire a grenade launcher at certain angle between two walls.

Can't go wrong with not firing grenade effect when qdeleted.

## Why It's Good For The Game
Prevent possible crash

## Testing

- Spawned living/carbon/human

- Fired 12 rounds of frag and stings into the wall. Nothing wrong 
- Also tested downstream

## Changelog
:cl:
fix: Fixed a possible crash / lagfest when grenade launcher are fired at certain angles
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile/projectilegrenades.dm
+++ b/code/modules/projectiles/projectile/projectilegrenades.dm
@@ -12,13 +12,18 @@
 
 /obj/item/projectile/bullet/grenade/Move()	//Makes grenade shells cause their effect when they arrive at their target turf
 	if(get_turf(src) == get_turf(original))
+		if(QDELETED(src)) return
+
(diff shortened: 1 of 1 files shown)


Humonitarian:
I recall we actually fixed this bug by changing grenade behaviour. Can't recall which PR, but I'll keep an eye out for it.

WeNeedMorePhoron:
Got it. Closing the PR. We will have it downstream with a sync soon.

PR #8193, opened 2023-05-28
Shotgun ammo packets and short magazine

maikilangiolo:
## About The Pull Request
Adds a 40-round ammo packet and an 8-round magazine for shotguns. Bumps the shotgun ammo box to 160 rounds (4 times the package capacity, which is in line with rifles: 60 rnd packets and 240 (60x4) rnd boxes), and the drum to 16 rounds. Both got a size upgrade in line with existing examples: the shotgun drum is normal sized like the .20 and .30 drums, and the shotgun ammo box is bulky like the rifle ammo boxes.
packets
bullet

## Why It's Good For The Game
8 round magazine is too small for magazine fed shotguns. Tube fed shotguns have the advantage of being able to be topped off, while magazine shotguns need to swap mag, empty or not, if they want to reach full capacity. loose ammo shells are also easier to carry vs magazines, and you can grab a handful of shells to reload in a pinch instead of having to reload the magazine.
At the same time, the addition of ammo packets means tube fed shotguns keep their advantage of easy ammo carry, and means they dont bite the dust compared to, say, revolvers which have essentially the same ammo capacity (i.e. kammerer is 5 rounds, DB is 2, and regulator 9, and a revolver is 6 to 8 shots) but have packets to carry bulk ammo. 30 rounds for the boxes was too small, so a bulky ammo box fits 160 shells, or you can carry the smaller packets around for some less ammo. I chose 40 rounds in the packets because it is both a nice, round number and a multiple of 8, which corresponds with the short magazine capacity, so each packet reloads full magazines instead of some full and one partial magazine, in line with all other packets and magazines.

## Testing
Immagine 2023-05-29 115410

## Changelog
:cl:
add: added 40-round shotgun ammo packets
add: added a short, 8-round shotgun magazine
balance: shotgun ammo boxes are now bulky and fit 160 shells
balance: shotgun drums are now normal sized and fit 16 shells
fix: fixed the confusing name for slugs in the trade stations
balance: hellcat now only sells shotgun ammo packets. unlock higher ammo stations for the boxes
balance: the crafting station can craft scrap ammo shotgun boxes for 30 points each. Only high MECH characters can craft them.
/:cl:


assortedbeads:
>shotgun buffs

I used to pray for days like this

maikilangiolo:
> Code is fine but i'm not sure we need 160 shotgun shells in a packet.

The 160 one isn't a packet, it's a box. It's bulky, so just like the 240 round rifle box

PR #8198, opened 2023-05-30
Brace rework

Humonitarian:
## About The Pull Request

Removes Kegdo brace, implements stationary bracing.

Recoil is massively reduced while shooting without movement, scaling with the size of your gun.
Recoil provides a slowdown, but guns no longer stun your movement.

Bipods now activate when stationary.

Includes balance patch to Takeshi as it made such an easily avaliable item way too powerful out-of-the-box.

## Why It's Good For The Game

More accuracy, more combat depth.

## Testing

I shot a variety of guns, ran around while holding them. Worked fine.

Major balance changes, live testing required.

## Changelog
:cl:
del: Removed bracing on tables
fix: Foregrip gives foregrip modifier
balance: Standing still improves accuracy when shooting, especially with larger guns
del: Removed recoil gained from movement
balance: Shooting guns no longer stuns your movement
balance: Recoil provides a scaling slowdown
balance: Takeshi stats nerfed, LMG recoil increased
/:cl:

### DIFF
--- a/cev_eris.dme
+++ b/cev_eris.dme
@@ -1935,7 +1935,6 @@
 #include "code\modules\mob\living\carbon\human\human_movement.dm"
 #include "code\modules\mob\living\carbon\human\human_organs.dm"
 #include "code\modules\mob\living\carbon\human\human_powers.dm"
-#include "code\modules\mob\living\carbon\human\human_recoil.dm"
 #include "code\modules\mob\living\carbon\human\human_species.dm"
 #include "code\modules\mob\living\carbon\human\inventory.dm"
 #include "code\modules\mob\living\carbon\human\life.dm"
--- a/code/__DEFINES/weapons.dm
+++ b/code/__DEFINES/weapons.dm
@@ -99,10 +99,11 @@
 
 #define VIG_OVERCHARGE_GEN 0.05
 
-#define EMBEDDED_RECOIL(x)     list(1.3 *x, 0  *x, 0  *x )
-#define HANDGUN_RECOIL(x)      list(1.15*x, 0.1*x, 0.6*x )
-#define SMG_RECOIL(x)          list(1   *x, 0.2*x, 1.2*x )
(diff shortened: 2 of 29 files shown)


Faryaa:
As we have a separate HUD button for holding weapons with both hands - why won't we have a separate button for bracing? Which would brace your weapon on nearby suitable surface and prevent movement untill you unbrace (or move on Help or Harm intent specifically)

Humonitarian:
> As we have a separate HUD button for holding weapons with both hands - why won't we have a separate button for bracing? Which would brace your weapon on nearby suitable surface and prevent movement untill you unbrace (or move on Help or Harm intent specifically)

Bracing is gonna be based on standing still, rather than aiming

Humonitarian:
> With this PR, it is possible for the recoil stacked from going full dakka to hamper the user's movement so much that they will be on equal speed terms with T*keshi players.


There was a bug that prevented FA weapons from run and gunning. I fixed that, but as it involves the removal of a key slowdown mechanic, I replaced that with the slowdown you see.

> A solution to this would be to make a simple size/caliber check for guns in **both** hands; if the caliber is low enough, like .35 or .25, the recoil slowdown would be either much less, or completely ignored.

It does scale with how big the weapon is. An SMG provides much less slowdown than a battle rifle.

> Same for melee, if this applies to every item and not only guns. This would also render cheesing Tak*shi by shooting and then switching to offhand to ignore gun-slowdown impossible.

Anything that isn't a gun provides the same slowdown as being emptyhanded, which is a fraction of what most guns have.
Emptyhanded -> least slowdown
HMG -> most slowdown



Overall your criticism is valuable, though testing on the testserver would give us a much more clear picture, as I could only test on local so far.

ValoTheValo:
honestly just kill bracing, no further changes needed. it was a salt mechanic that was terribly implemented

hyperioo (approved):

PR #8202, opened 2023-06-07
[RTM] [DO NOT MERGE] Sanity drains slower

ValoTheValo:
## About The Pull Request
due to #8000 sanity drain became very very quick, it would only take a second or two to nuke your sanity if you were in a slightly dirty room. this wasnt the purpose of that PR just a side effect of reducing moblag. so this PR adjusts the sanity drain modifier to make it go slower. will require tweaking based on actual live server conditions so requires test merge and testing

## Why It's Good For The Game
less obsurd sanity drain speed. you wont need to smoke and drink 24/7 anymore

## Testing
it compiled fine but needs testing on live server to see lag effects

## Changelog
:cl:
balance: Sanity drains slower now
/:cl:


Humonitarian:
//How often is the onLife() triggered and by how much are the effects multiplied

Read Valo, read!

ValoTheValo:
> Time to get hit by x2 sanity drain causing me to go insane in a single tick

okay i halved it. oops

DiaFRAME444:
It still does nothing except drive up lag and make sanity drain more modular Valo
Do you not understand
Read the code comment, its right there

ValoTheValo:
> It still does nothing except drive up lag and make sanity drain more granular Valo Do you not understand Read the code comment, its right there

This is what I was told by a maintainer when I asked awhile ago. Code comment is how fast sanity processes? No?

Screenshot_20230610-170705-182

DiaFRAME444:
Okay, I'm going to interpret it for you. 
"How often is the onLife() triggered and by how much are the effects multiplied"

Lets break it into two parts.
"How often is the onLife() triggered"
So, you know the thing where monkeys were slowing down the game because of their sanity? Yeah, processing sanity is heavy. Not just for mobs, but for humans too.
This variable allows us to slow it down from checking sanity EVERY mob life() tick, to every OTHER life() tick. Its a way to halve the lag generated by it.

"and by how much are the effects multiplied"
But that brings us to a problem. If sanity is being checked only every other tick, its updating half as fast. How do we deal with that?
So this variable ALSO multiplies sanity loss. If you slow it down to every two ticks, then in the second tick people will gain/lose twice as much sanity.

tl;dr: Close this PR.

ValoTheValo:
> Okay, I'm going to interpret it for you. "How often is the onLife() triggered and by how much are the effects multiplied"
> 
> Lets break it into two parts. "How often is the onLife() triggered" So, you know the thing where monkeys were slowing down the game because of their sanity? Yeah, processing sanity is heavy. Not just for mobs, but for humans too. This variable allows us to slow it down from checking sanity EVERY mob life() tick, to every OTHER life() tick. Its a way to halve the lag generated by it.
> 
> "and by how much are the effects multiplied" But that brings us to a problem. If sanity is being checked only every other tick, its updating half as fast. How do we deal with that? So this variable ALSO multiplies sanity loss. If you slow it down to every two ticks, then in the second tick people will gain/lose twice as much sanity.
> 
> tl;dr: Close this PR.

Not closing the PR but I'll try and make a different fix. Like just making it process every other tick and not doubling damage

ValoTheValo:
actually i am closing the PR to redo it

PR #8216, opened 2023-06-21
[DNM] [TEST] Erismed 4 tweaks

Humonitarian:
## About The Pull Request

Makes Erismed 4 significantly more lethal, generally encourages organs to die more often. Adds a lot of helpful defines.

Changes include:
* Reduction in the health of most small organs
* Creation of defines for such health values
* Change to fracture code, so instead of fractures being required before organ damage, the damage is instead directed to the bones until they break
* Limb internal organ lists now associate organ with their size for weight-based targeting
* Toughness mutation makes the bones consider all organs, not just vitals

## Why It's Good For The Game

You need less magazines to kill someone.

## Testing

I shot a lot of humanmobs and counted how many is needed to kill. TTK was decreased significantly, but killing from a distance (2+ tiles) takes several magazines, or several minutes of bleeding.

## Changelog
:cl:
balance: small organs have less health, large organs have more
balance: organs are more likely to be hit the larger they are
balance: damage no longer disappears if your bones aren't broken, bones now only protect vital organs
tweak: toughness mutation makes bones actually protect your organs
/:cl:


hyperioo (approved):

hyperioo:
Merging after tests on test server

PR #8219, opened 2023-06-22
Quickequip hotfix

Humonitarian:
## About The Pull Request

Makes quickequip unscope the player. Unforeseen consequences can be possible, but are quite unlikely.

## Why It's Good For The Game

Fixes a goofy bug.

## Testing

Attempted to equip and unequip Tosshin, both scoped and unscoped. Worked as intended, could not fire the gun and the scope did not stay on when the Tosshin was put in the suit slot or back.

## Changelog
:cl:
fix: "e" to quickequip no longer allows firing guns from within storage
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/inventory.dm
+++ b/code/modules/mob/living/carbon/human/inventory.dm
@@ -14,8 +14,6 @@ This saves us from having to call add_fingerprint() any time something is put in
 	var/target_slot = get_quick_slot(I)
 	if(I.pre_equip(usr, target_slot))
 		return
-	//if(!I.try_transfer(target_slot, usr))
-	//	quick_equip_storage(I)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/inventory.dm
```diff
@@ -14,7 +14,9 @@ This saves us from having to call add_fingerprint() any time something is put in
 	var/target_slot = get_quick_slot(I)
 	if(I.pre_equip(usr, target_slot))
 		return
-	if(!I.try_transfer(target_slot, usr))
+	//if(!I.try_transfer(target_slot, usr))
+	//	quick_equip_storage(I)
```
hyperioo, line 18:
```suggestion
```
No point in leaving commented code, we have git history for that

hyperioo (approved):

PR #8254, opened 2023-08-11
No Trash Nun

Mycah142:
## About The Pull Request
The Medicus Robes were not intended to spawn in maint. The space suit I built it off of, the helmet is already blacklisted, but the suit wasn't, and I didn't realize that.

## Why It's Good For The Game
The only thing worse than a vagabond, is a vagabond cosplaying as a nun. Begone, thot.

But no really it's a hotfix.

## Testing
Compiled, tested locally, and I looked into a single clothing trash pile and no nun came at me with a gun. Mission successful.
dreamseeker_JSNh2clbJe

## Changelog
:cl:
fix: Removed unintended trash catholics.
/:cl:

### DIFF
--- a/code/modules/clothing/spacesuits/miscellaneous.dm
+++ b/code/modules/clothing/spacesuits/miscellaneous.dm
@@ -41,6 +41,7 @@
 	item_state = "nt_habithat_visor"
 	desc = "Protective helmet meant more to safeguard against disease, retrofit to also be spaceworthy."
 	style = STYLE_HIGH //Very low defenses, but looks better than a normal spacesuit
+	spawn_blacklisted = TRUE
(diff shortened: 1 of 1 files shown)


SirRichardFrancis:
Looked into a trash pile? To see if something spawned, where and in what quantities use SDQL query in debug tab.
Type something like `select /obj/item/clothing/head/space/medicus` and see what comes out.

Mycah142:
> Looked into a trash pile? To see if something spawned, where and in what quantities use SDQL query in debug tab. Type something like `select /obj/item/clothing/head/space/medicus` and see what comes out.

That is very good to know, thank you, I'll update the testing.

Firefox13 (approved):

PR #8258, opened 2023-08-13
Grenade launcher shells' materials

TheLeopold:
## About The Pull Request
Added missing materials to stinger, NT frag, heatwave shells.

I just eyeballed values, as I had nothing to compare to, so they may be bad, give better.

## Why It's Good For The Game
Fix

## Testing
Recycled each shell, printed printable.

## Changelog
:cl:
fix: missing materials for some grenade launcher's shells
/:cl:


Humonitarian:
Rubber grenades should probably contain plastic, considering they are armed with rubber balls.

Also, it might be a good idea to give material costs to the non-craftable grenades as well so they can be recycled. Not necessary, though.

hyperioo (approved):

PR #8262, opened 2023-08-19
Rework E-armblade

Mycah142:
## About The Pull Request
Directly removes Double Tact from the Energy Armblade cybernetic without sacrificing it's BULKY size, which has defensive capabilities while using the BLOCK hotkey.

## Why It's Good For The Game
Why should a quick e-shank require two clicks? It does 6 more damage than the wrist shank (a makeshift item) but has zero modifiability. The wrist shank does more damage with a single mod, without double tact. At the very least, such an item that isn't modifiable should have more advantages than something assembled from actual junk.

## Testing
Compiled, launched, swung my energy armblade into floor tiles without double tact.

## Changelog
:cl:
balance: Removed Double Tact from the Energy Armblade cybernetic.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/melee/energy.dm
+++ b/code/game/objects/items/weapons/melee/energy.dm
@@ -240,6 +240,7 @@
 			spawn(1) if(src) qdel(src)
 
 /obj/item/melee/energy/blade/organ_module //just to make sure that blade doesnt delet itself
+	w_class = ITEM_SIZE_SMALL //Same as other armblades, it shouldn't be a bulky item requiring double-tact
 	cleanup = FALSE
 	init_procees = FALSE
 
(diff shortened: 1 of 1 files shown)


Mycah142:
brave_zsl9hNRHFa
No idea how this is my problem.

SirRichardFrancis:
> No idea how this is my problem.

It really isn't.

SirRichardFrancis:
Forgot to clarify, integration tests fail due to these roundstart runtimes:
image
image
Basically mice running all over maintenance are setting off traps before all necessary stuff is initialized.

Firefox13 (changes requested):
Use `no_double_tact` or however it is called variable to remove double tact when it is specifically required to remove double tact from an item that shouldn't or when it doesn't have sense for it to have double tact.
Double tact.

Mycah142:
> Use `no_double_tact` or however it is called variable to remove double tact when it is specifically required to remove double tact from an item that shouldn't or when it doesn't have sense for it to have double tact. Double tact.

So you still want it to have the blocking ability of other Bulky items?

Firefox13:
I mean, it's an armBLADE

Firefox13 (approved):

PR #8268, opened 2023-08-29
Make forcing events logged

Firefox13:
Logging, security, I dunno how many teleports *someone* spawned *that one* round. I'd like to. I can't test it because of how logging works clientside, so I just opted in to make a change on web.

## Why It's Good For The Game
important things logging = good

## Testing
can't

## Changelog
:cl:
admin: When an admin forces an event it's now logged.
/:cl:


Firefox13:
oopsie daisy

PR #8271, opened 2023-09-01
Less acolytes, more custodians, tiny map fix

TheLeopold:
## About The Pull Request
Increasing custodian slots 2->4.
Decreasing acolyte slots 4->2 to increase their importance through scarcity.

Also fixed missing downward pipes behind guild shop while I'm at it.

Map changes:
изображение
изображение
vents

## Why It's Good For The Game
Makes acolytes more important (hopefully). Will make ship a bit cleaner (hopefully).
Also fix.

## Testing
Respawned many times to fill both slots. Checked closets' contents, crawled through both pipes.

## Changelog
:cl:
tweak: tweaked a few things
balance: Increased Custodian slots 2->4; Decreased Acolyte slots 4->2.
fix: fixed missing downward pipes near guild
/:cl:

### DIFF
--- a/code/game/jobs/job/church.dm
+++ b/code/game/jobs/job/church.dm
@@ -74,8 +74,8 @@
 	department = DEPARTMENT_CHURCH
 	department_flag = CHURCH
 	faction = "CEV Eris"
-	total_positions = 4
-	spawn_positions = 4
+	total_positions = 2
+	spawn_positions = 2
 	supervisors = "the NeoTheology Preacher"
 	selection_color = "#ecd37d"
 	also_known_languages = list(LANGUAGE_LATIN = 100)
@@ -165,8 +165,8 @@
 	department = DEPARTMENT_CHURCH
 	department_flag = CHURCH
 	faction = "CEV Eris"
-	total_positions = 2
-	spawn_positions = 2
+	total_positions = 4
+	spawn_positions = 4
 	supervisors = "the NeoTheology Preacher"
 	selection_color = "#ecd37d"
 	//alt_titles = list("Custodian","Sanitation Technician")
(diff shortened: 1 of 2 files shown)


ValoTheValo:
could just make both slots have 3.  so its even

TheLeopold:
I disagree and think they shouldn't be even as acolytes are higher in hierarchy, they more like a Lieutenant or Sergeant of IH (you can tell I don't play IH).

code/game/jobs/job/church.dm
```diff
@@ -165,8 +165,8 @@
 	department = DEPARTMENT_CHURCH
 	department_flag = CHURCH
 	faction = "CEV Eris"
-	total_positions = 2
-	spawn_positions = 2
+	total_positions = 4
```
Firefox13, line 168: How about we set it to 3 instead, same as agrolyte?
1 preacher -> 2 acolytes -> 3 custodians and 3 agrolytes

TheLeopold: https://discord.com/channels/255035529085583360/265411250341543936/1147369213590245417

Mycah142: > How about we set it to 3 instead, same as agrolyte? 1 preacher -> 2 acolytes -> 3 custodians and 3 agrolytes

To say here what I said on discord:

As an agrolyte main, there is barely room for two agrolytes who know what they're doing. A third without more physical room to divvy up responsibilities would be hell on earth. I'd rather take the hit to total department count.

null-Aurelian: Adding my two credits: Technically agrolytes are also best positioned to work the bioreactor, but again if they know what they are doing it isn't so far fetched to see them solo the entire downstairs area. Extra custodians makes sense and if you wanted to I'd say move a slot from the agrolytes to custodian or acolyte roles.

TheLeopold: > if you wanted to I'd say move a slot from the agrolytes to custodian or acolyte roles.

I do want to swap agrolyte slot to custodian, but that would make 5 custodians which is kinda overkill in my opinion.
I think I'll leave it as it is right now and change things depending on how it goes.

hyperioo (approved):
Let's see how it goes, it's a small change to tweak if necessary

PR #8284, opened 2023-09-04
Massively lower the amount of powercells in moebius robotech vendor

Firefox13:
## About The Pull Request
Now instead of 12(!) 5000L batteries (3 silver each) they only get one 1000L (still 3 silver)

## Why It's Good For The Game
Moebius R&D is usually in a need of materials, but not in silver - because they just recycle all these useless powercells and have silver enough for the whole round. This, I believe, is an oversight from bay where these cells just didn't matter as much as they do now. Cope. Go buy some from guild.

## Testing
изображение

## Changelog
:cl:
balance: Robotech Deluxe now only has one 1000L large cell instead of twelve 5000L
/:cl:


Firefox13:
Spit it out.

hyperioo (approved):

PR #8287, opened 2023-09-06
Suez Nerf

Mycah142:
## About The Pull Request
Disabled the ability to dual wield (fire two at the same time while in harm intent) for stun revolvers, and increased the cost per shot from 50 to 100. Seems folk have caught on that dual wielding Suez, modifying the ever loving shit out of it, create department deleters. If this isn't enough, I'll revert it and look into making them unmodifiable, which will incredibly retard their potential.

## Why It's Good For The Game
Cheap and easy ways to delete entire swaths of people might be fun for the person with the guns, but not everyone else.
7y53gj

## Testing
Yepp it's a gun, it shoots and I can only shoot one at a time even on harm intent. I don't know what mods people use to make them into weapons of mass deletion intimately, so I can't recreate the carnage players have been doing.

## Changelog
:cl:
balance: Removed Stunrevolver Dual Wield, doubled energy cost.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy/stun.dm
+++ b/code/modules/projectiles/guns/energy/stun.dm
@@ -35,9 +35,9 @@
 	icon_state = "stunrevolver"
 	item_state = "stunrevolver"
 	fire_sound = 'sound/weapons/Gunshot.ogg'
-	can_dual = TRUE
+	can_dual = FALSE
 	origin_tech = list(TECH_COMBAT = 3, TECH_MATERIAL = 3, TECH_POWER = 2)
-	charge_cost = 50
+	charge_cost = 100
 	matter = list(MATERIAL_PLASTEEL = 12, MATERIAL_WOOD = 6, MATERIAL_SILVER = 2)
 	price_tag = 1500
 	suitable_cell = /obj/item/cell/small
@@ -45,6 +45,7 @@
 	projectile_type = /obj/item/projectile/energy/electrode
 	init_recoil = HANDGUN_RECOIL(1)
 	serial_type = "NT"
+
 
 /obj/item/gun/energy/stunrevolver/moebius
 	name = "ML SP \"Suez\""	//Ersatz name


garrulousTeamotei:
Maybe instead of disallowing the dual wielding of a pistol, maybe make it so the Tesla Overdrive Chip reduces not just the damage, but stun "damage" as well.

Mycah142:
> 

Cuz that sounds like alot of work.

Firefox13 (changes requested):
> forbidding dual-wielding a REVOLVER

No. Look into the projectile it shoots. Perhaps replace it with either deafult stunbeam or nerf it to be in line with it.
But keep the charge cost increase.

Mycah142:
> > forbidding dual-wielding a REVOLVER
> 
> No. Look into the projectile it shoots. Perhaps replace it with either deafult stunbeam or nerf it to be in line with it. But keep the charge cost increase.

The main issue I believe is the rapidity it can come across. An ultimate fix is to rework the HALOSS on how the Overcharge mod works or whatever, but it only has a damage multiplier of .66, and no multiplier exists in defines for pain.

So if I wanted to do a *real* fix, I'd have to create a whole new multiplier specifically for HALOSS.

If that's what is wanted, then it's out of my capabilities currently.

Firefox13:
> The main issue I believe is the rapidity it can come across.
Then reduce the firing speed? Since it's a revolver, it shouldn't be fast-firing.

Firefox13:
ENERGY revolver, mind you. That's even worse.

null-Aurelian:
Can someone verify if I am reading the Tesla mod right? 0.66 damage multiplier, 3 fire rate multiplier and 0.5 energy cost means you squeeze 4 times the number of attacks per cell?
Could the mod itself have the energy cost upped?

Firefox13:
Good thinking, mod itself may be the issue.

null-Aurelian:
At the very least I think it could stand to be a x1 energy cost modifier. Kneecapping the mod because of the insanity of the otherwise meh stun revolvers feels bad for plasma or laser weapons. 

That or try to bring both mods to be 'sidegrades' with x1 damage compared to stock values.

Mycah142:
> Good thinking, mod itself may be the issue.

That's what I said.

Mycah142:
> The main issue I believe is the rapidity it can come across. An ultimate fix is to rework the HALOSS on how the Overcharge mod works or whatever, but it only has a damage multiplier of .66, and no multiplier exists in defines for pain.
> 
> So if I wanted to do a _real_ fix, I'd have to create a whole new multiplier specifically for HALOSS.
> 
> If that's what is wanted, then it's out of my capabilities currently.

This is the real fix, and as stated, it is beyond my current time abilities. So I will close this PR until I come up with time to make a whole new multiplier.

PR #8291, opened 2023-09-07
Revert "Leather satchels are stylish"

pexel5:
## About The Pull Request

Revert of "Leather satchels are stylish" PR.

## Why It's Good For The Game

Since style no longer affects combat, it became much less valuable and I think no longer warrants reduction of leather satchel's storage capacity.
+1 style is hardly beneficial now, but reduced capacity on already small storage is a big deal.
I (as well as some other players) prefer this satchel for it's appearance, but by choosing it we're put at a disadvantage for effectively no gain. Probably shouldn't be that way.

## Testing
Tested. This fine.
image
image
image
## Changelog
🆑 
balance: Returned storage capacity to leather satchel
/🆑

### DIFF
--- a/code/game/objects/items/weapons/storage/backpack.dm
+++ b/code/game/objects/items/weapons/storage/backpack.dm
@@ -324,10 +324,8 @@
 
 /obj/item/storage/backpack/satchel/leather
 	name = "leather satchel"
-	desc = "A fancy satchel made with fine leather. While fancy, it has less storage space than most other satchels."
+	desc = "A fancy satchel made with fine leather."
 	icon_state = "satchel_leather"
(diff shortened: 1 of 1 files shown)


Firefox13:
> Changes too small to warrant testing, this code worked perfectly fine before

Yeah, no. Test it.

thevandie:
I don't like Firefox, however
Load it up and make sure the game doesn't crash
I've fucked up spelling on things more times than I can count, causing errors, which stops the game from working

Mycah142:
I once made a PR to change the offer values through the guild offer system and managed to delete a floor tile straight into space.

No PR is too small.

PR #8298, opened 2023-09-11
Randomize mid-round role appearance

null-Aurelian:
## About The Pull Request

This pull request randomizes human bodies spawned for Excelsior conscripts, Last Shelter disciples and Carrion body generation. Values are bounded to fairly narrow values to avoid creating overly eye-catching bodies (ie no neon hair). This also adds a function to humans to randomize appearances and TTS seed usage within the limits of the human's gender.

## Why It's Good For The Game

Adds variety to the mid-round device-based spawns so they aren't all mooks (white, bald, male) in appearance, unlikely as it is that anyone will see them spawned.

## Testing

Tested to ensure that appearances were agnostic of previous spawns by repeatedly teleporting out and spawning Excelsior conscripts. Got about a 60/40 split on conscript genders over 10ish spawns.
Verified that genders properly changed certain attributes (female first names, option for males to have facial hair).
**Additional testing**: Tested by running generate_body procs on carrion cores.
**Additional testing**: Tested by using last shelter cruciforms in cloner.
**Additional testing**: Verified that appearance is set once and that repeated use of the last shelter cruciform does not generate a new appearance.

## Changelog
:cl:
tweak: randomizes body creation for various spawns
code: adds the randomize_appearance() proc in appearances.dm for humans
/:cl:


SirRichardFrancis (commented):
Did TTS part work as well? I see no mention of that in testing section.
Also need screenshots of resulting conscripts. I suspect they may have a severe case of anime hair, need to verify.

null-Aurelian:
The TTS testing didn't seem to trigger unfortunately, so I made sure that the list of voices will include the default voices for the gender as a failsafe. If it does work, all it means is that the default voices have a slightly higher weight for picking which we can live with I think. Also linking a clip of the testing, ran into a couple of warnings where the body gets qdel'd as part of the teleport and when adding a new character to an extinct faction.
[https://streamable.com/qnlh7m](url)

SirRichardFrancis:
Frankly, haircuts look pretty much the way I was afraid they would.
image image
Consider making several color presets, instead of picking a completely random color.

Also what was a runtime error there? What code does it point at?

null-Aurelian:
alright, I'll curate the lists then. The runtime errors were for calling destroy proc multiple times when teleporting a person out and when adding a member of excelsior to a null faction reference.
`Debug Server: "Pausing execution (reason: Runtime(\"/obj/item/organ/external/head destroy proc was called multiple times, likely due to a qdel loop in the Destroy logic\"))"
`
`Debug Server: "Pausing execution (reason: Runtime(\"Cannot execute null.add member().\"))"
`

null-Aurelian:
hair color should be more limited now. If the conscript wants to set their hair they can very well use a mirror or the revs can have a hand mirror available for touching up.

Mycah142:
Carrion regenerate new body?

null-Aurelian:
> 
can you explain? Does the carrion regeneration use the default body?

radarseas:
> > 
> 
> can you explain? Does the carrion regeneration use the default body?

yeah it also does default everything just like _shelter/conscripts_

Firefox13 (commented):
Honestly we just need an appearance randomizer and use it everywhere it is... "used"

null-Aurelian:
Well I guess I can try to figure out where to put this so it is accessible by all those classes as a function. Maybe in the human.dm file?

null-Aurelian:
Additional testing: Tested by running generate_body procs on carrion cores.
Additional testing: Tested by using last shelter cruciforms in cloner. 
Additional testing: Verified that appearance is set once and that repeated use of the last shelter cruciform does not generate a new appearance.

Firefox13 (approved):

code/modules/mob/living/carbon/human/appearance.dm
```diff
@@ -173,3 +173,27 @@
...
+	else
+		change_facial_hair(pick(GLOB.facial_hair_styles_list)) //pick a random facial hair
+		tts_voices += TTS_SEED_DEFAULT_MALE //Failsafe voice
+	real_name = species.get_random_name(gender) //set name according to gender
+	//Set voice based on gender
+	for(var/i in tts_seeds) //from tts_seeds, add voices that match gender tag to list tts_voices
+		var/list/V = tts_seeds[i]
+		if((V["gender"] == gender || V["category"] == "any") && (V["category"] == "human" || V["gender"] == "any")) //cribbed from /code/modules/client/preference_setup/general/01_basic.dm
+			tts_voices += i
+	tts_seed = pick(tts_voices) //pick a random voice from list tts_voices
```
hyperioo, line 199:
```suggestion
	tts_seed = pick(tts_voices) //pick a random voice from list tts_voices

```
Newline

null-Aurelian: fixed

Issue #7994, opened 2023-01-28
Supermatter shards don't arrive with a crate


radarseas:
so as it currently stands, ordering an SM shard from guild, for 12000 **sends an entire shard**, without a crate.

if someone doesn't know how an SM works by activation from a single dusting, _this can slightly derail or end an entire round_ also its nearly impossible to move out of the guild, due to this.

there isn't any warning on the order menu for this, so I'm presuming its unintentional.

radarseas:
also if it helps for issue reproduction, this was off of the guild departmental account.

CocaColaTastesGood:
Adding a warning for carts with high risk items would probably help.

Mycah142:
> Maybe we should change the SM to not dust people, just burn + irradiate them. SM dusting almost isn't ever a fun mechanic even for traitors, getting round removed because you misclicked or lagged isn't ever fun.

This is a terrible idea. SM dusting is the best shit ever and I want my SM disposal chute I build round start to enable me to make a vagabond powered ship until I reroute all trash bins to point at the SM.

Trash ship is a good ship.

assortedbeads:
> Maybe we should change the SM to not dust people, just burn + irradiate them. SM dusting almost isn't ever a fun mechanic even for traitors, getting round removed because you misclicked or lagged isn't ever fun.

We should make the SM dust not only on contact but if you are thrown into it as well (iirc it doesnt rn)

null-Aurelian:
> > Maybe we should change the SM to not dust people, just burn + irradiate them. SM dusting almost isn't ever a fun mechanic even for traitors, getting round removed because you misclicked or lagged isn't ever fun.
> 
> We should make the SM dust not only on contact but if you are thrown into it as well (iirc it doesnt rn)

Tested manually throwing a person into the SM, dusting works in that case. Might be a issue when thrown by a device ie disposals

PR #8306, opened 2023-09-14
SM now comes in crates

null-Aurelian:
Per issue #7994, quick check for SM shard and crates the order if SM is present in order

## About The Pull Request

Per issue #7994, QoL change to put the SM from the Zayra in a crate. We do want guild techs, even if it is funny when they run into the SM they ordered.

## Why It's Good For The Game

Minor QoL fix so people aren't upset about dusting themselves by accident.

## Testing

Tested by ordering a single SM from Zayra
Tested by ordering an SM with other items from Zayra

## Changelog
:cl:
tweak: added a check to creating an order from the trade beacon so that orders containing an SM shard are crated on spawn.
/:cl:

### DIFF
--- a/code/controllers/subsystems/trade.dm
+++ b/code/controllers/subsystems/trade.dm
@@ -446,7 +446,7 @@ SUBSYSTEM_DEF(trade)
 	var/obj/structure/closet/secure_closet/personal/trade/C
 	var/count_of_all = collect_counts_from(shopList)
 	var/price_for_all = collect_price_for_list(shopList)
-	if((isnum(count_of_all) && count_of_all > 1) || shopList.Find("/obj/machinery/power/supermatter") > -1)
+	if((isnum(count_of_all) && count_of_all > 1) || shopList.Find("/obj/machinery/power/supermatter") > 0)
 		C = senderBeacon.drop(/obj/structure/closet/secure_closet/personal/trade)
 		if(is_order)
 			C.locked = TRUE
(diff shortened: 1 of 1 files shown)


Mycah142:
Guild is now literally unplayable.

null-Aurelian:
Just have the maintainers close the PR.
Skill issue

Humonitarian:
It's over. The guild has fallen. Millions must dust.

Mycah142:
It's gover.

null-Aurelian:
"Guild to Desk!!!1!" x30
meanwhile: Guild tech, dusted

TheShown911 (approved):
Literally 0 soul, next thing they are gonna do is forbid people from spawning without a brain.

code/controllers/subsystems/trade.dm
```diff
@@ -446,7 +446,7 @@ SUBSYSTEM_DEF(trade)
 	var/obj/structure/closet/secure_closet/personal/trade/C
 	var/count_of_all = collect_counts_from(shopList)
 	var/price_for_all = collect_price_for_list(shopList)
-	if(isnum(count_of_all) && count_of_all > 1)
+	if((isnum(count_of_all) && count_of_all > 1) || shopList.Find("/obj/machinery/power/supermatter") > -1)
```
hyperioo, line 449: From byond manual

> The Find proc could be used for that purpose, but if you don't care about the position of the item in the list, in is more convenient and efficient. The expression is 1 if the item exists in the list and 0 if not. 

Your `> -1` will always be true ?

null-Aurelian: I didn't realize, I just thought it would return the position and -1 if it doesn't exist like a normal language. I'll fix this

null-Aurelian: change has been applied, searches for >0 rather than >-1

hyperioo: > like a normal language

This is BYOND we're dealing with my good sir!

PR #8310, opened 2023-09-16
fix pacman nukes

Firefox13:
## About The Pull Request
Adds hard caps on pacman explosions by adding `min(`.

## Why It's Good For The Game
They can explode the entire Eris. I'm not sure if parent explode act numbers are good (they're only used by excel diesel generator, which isn't very explosive), shouldn't be *too* bad*

## Testing
изображение
изображение
изображение
изображение

## Changelog
:cl:
balance: added caps to pacmans explosion power
/:cl:

### DIFF
--- a/code/modules/power/port_gen.dm
+++ b/code/modules/power/port_gen.dm
@@ -80,7 +80,7 @@
 			stat &= ~EMPED
 
 /obj/machinery/power/port_gen/proc/explode()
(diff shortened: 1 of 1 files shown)


garrulousTeamotei:
Soul of Eris is being ripped to shreds. Piece by Piece.

PR #8315, opened 2023-09-18
Multistructure power requirement change

Humonitarian:
## About The Pull Request

The bioreactor and biogenerator no longer require power to work, putting them inline with all other NeoTheology technology. (The biogenerator *produces* power, why did it need it to work?)

## Why It's Good For The Game

Bugfix good

## Testing

image

## Changelog
:cl:
fix: bioreactor and biogenerator no longer need power
/:cl:


Humonitarian:
I've been notified most other NT machines use power. Drafting this.

PR #8317, opened 2023-09-19
Make resuscitator no longer work on synthetic hearts and work only on inject

Firefox13:
## About The Pull Request
Adds a `return` on `affect_ingest` for resuscitator and adds a check on if the heart is robotic, returns if that is true.

## Why It's Good For The Game
In cool movies when main heroes need to "revive" someone they give him adrenaline - not with a pill but with an injection, which is incredibly cool and reviving people with a cardiac stimulant "pill" doesn't make sense to me.

I also do not think it should affect robotic hearts due to chemical incompatibility. Yes, this is a FBP nerf. And anyone else who takes a robotic heart.

## Testing

https://github.com/discordia-space/CEV-Eris/assets/57810301/0ed5a8ba-f35e-4c9b-bce8-5943e74bf28e

## Changelog
:cl:
balance: Resuscitator now needs to be injected to work
balance: Resuscitator no longer revives people with robotic hearts
/:cl:

### DIFF
--- a/code/modules/reagents/reagents/other.dm
+++ b/code/modules/reagents/reagents/other.dm
@@ -542,8 +542,10 @@
 	affects_dead = TRUE
 	reagent_type = "Medicine"
 
-/datum/reagent/resuscitator/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
+/datum/reagent/resuscitator/affect_ingest(mob/living/carbon/M, var/alien, effect_multiplier)
+	return // since it's a "cardiac stimulant" it shouldn't really work unless injected
 
+/datum/reagent/resuscitator/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
 	if(ishuman(M))
 		var/mob/living/carbon/human/H = M
 		var/obj/item/organ/internal/vital/heart/heart = H.random_organ_by_process(OP_HEART)
@@ -552,7 +554,8 @@
 			if(prob(30))
 				to_chat(H, SPAN_DANGER("Your heart feels like it's going to tear itself out of you!"))
 		if(H.stat == DEAD)
-			H.resuscitate()
+			if(!BP_IS_ROBOTIC(heart) // neither it should work on robotic hearts, chemistry n' stuff
+				H.resuscitate()
 
 /datum/reagent/resuscitator/overdose(mob/living/carbon/M, alien)
 	. = ..()
--- a/code/modules/reagents/reagents/other.dm
+++ b/code/modules/reagents/reagents/other.dm
@@ -549,13 +549,14 @@
 	if(ishuman(M))
 		var/mob/living/carbon/human/H = M
 		var/obj/item/organ/internal/vital/heart/heart = H.random_organ_by_process(OP_HEART)
+		if(BP_IS_ROBOTIC(heart)) // neither it should work on robotic hearts, chemistry and stuf
+			return
 		if(heart)
 			heart.damage += 0.5
 			if(prob(30))
 				to_chat(H, SPAN_DANGER("Your heart feels like it's going to tear itself out of you!"))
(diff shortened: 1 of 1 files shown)


assortedbeads:
>cant resus fbps
If you're going to make them unrevivable by normal means you need to either give them more advantages for _permanent death_ or a way to be brought back that is fbp only

TooFewSecrets:
> If you're going to make them unrevivable by normal means you need to either give them more advantages for _permanent death_ or a way to be brought back that is fbp only

by far the most common forms of death in my experience are bleeding out and liver damage and fbps are already immune to both, leaving ion spam (usually results in braindeath) and acute magdumping/slashing to the head (ditto)
self-surgery with no painkiller requirement using common tools on top of that means fbps already barely ever die outside of being murdered, and when you're decapped usually the guy will bsdm your head anyway so resus is irrelevant

code/modules/reagents/reagents/other.dm
```diff
@@ -542,11 +542,15 @@
...
 
-/datum/reagent/resuscitator/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
+/datum/reagent/resuscitator/affect_ingest(mob/living/carbon/M, var/alien, effect_multiplier)
+	return // since it's a "cardiac stimulant" it shouldn't really work unless injected
 
+/datum/reagent/resuscitator/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
 	if(ishuman(M))
 		var/mob/living/carbon/human/H = M
 		var/obj/item/organ/internal/vital/heart/heart = H.random_organ_by_process(OP_HEART)
+		if(BP_IS_ROBOTIC(heart)) // neither it should work on robotic hearts, chemistry and stuf
```
MLGTASTICa, line 552: this is for bodyparts not for internal organs ( cause we dont have a robotic check for organs).  Btw metabolizes fast so rolling for the random organs until it hits a organic heart will be kinda dumb for people who have both organic and inorganic hearts)

pexel5:
I don't think it's worth doing this, FBP is weak now and if we introduce this, it won't hurt to introduce some special defibrillator for FBP/humans with mechanical hearts that would allow them to be resuscitated (and also don't forget, if the heart is damaged so much that it can't be repaired, you will need to find either donor, or fuck with nt with a request to make organs that no one ever does). But as for the pills, I agree, it's better to leave the injection.

DiaFRAME444:
I was gonna say "Finally a drawback to being a FBP-Mutant-Cruciformed" but I realized that if you're cruciformed you can be revived anyways
Powergamers still winning

code/modules/reagents/reagents/other.dm
Firefox13: No, `BP_IS_ROBOTIC`, `BP_IS_ASSISTED` and `BP_IS_ORGANIC` is used both for internal and external body parts.
изображение
изображение

Firefox13:
> either give them more advantages

> I don't think it's worth doing this, FBP is weak now

TooFewSecrets gave enough explanation why they don't need to be buffed alongside this nerf

Firefox13:
> if you're cruciformed you can be revived anyways

Takes significantly more time and if NT won't murder you when they realize you're godblood you honestly deserve the chance to be resurrected as a meatsack

ValoTheValo:
does this work on assisted hearts?

Firefox13:
@ValoTheValo Yes it does.

(there is a way to check if the heart is assisted but it was just faster do it through icon)
https://github.com/discordia-space/CEV-Eris/assets/57810301/830589ab-6b86-47d8-be54-9d0280f3756f

hyperioo (approved):

hyperioo:
Could be improved by looking through all hearts insted of picking a random one, for people having several organic/robotic hearts.

PR #8318, opened 2023-09-19
Fix qdel of implants when slimed

null-Aurelian:
Fixes the qdel effect on implants and prosthesis after processing advanced slime mutation toxin When processed all implants and prosthetic components are dropped FBPs just puke up the toxin.
Adds function that prosthetic limbs are dropped when transformed into a slime person, unless an FBP

## About The Pull Request

Fixes the TODO where implants and prosthetic external organs are deleted after being transformed by the advanced slime mutation toxin. All implants and prosthetic organs are dropped at the location of the transformation. 

- In cases where the victim has no organic parts (FBPs) they instead vomit up the contents of their stomach to reduce processing effects as quickly as possible.
- In the case of cruciform holders, the victim is gibbed but the cruciform drops with their soul intact so they can be revived.

## Why It's Good For The Game
Implants and prosthetic body parts are no longer lost to the void when a person is slimed.

## Testing
Tested by

- Injecting a cruciformed human with advanced mutation toxin and then reviving them via the Clone pod to ensure that appearances were persisted
- Injecting an augmented human with either slime mutation or advanced slime mutation toxin and ensuring that all prosthetic limbs and organs dropped after the transformation
- Injecting a human with an excelsior implant with either regular or advanced slime mutation toxin and ensuring that the implant dropped after transformation
- Injecting a monkey
- Ingesting advanced mutation toxin as an FBP using Asters parts
- Ingesting advanced mutation toxin as an FBP using Moebius parts
- Ingesting advanced mutation toxin as an FBP using Homebrew parts

## Changelog
:cl:
balance: FBPs can no longer be turned into grey slimes. Cruciformed humans are gibbed but not transformed.
fix: Implants and prosthetics are not deleted by transforming into a grey slime or a slimeperson. They will instead be dropped on the ground.
code: added a dropped() implementation for prosthetic limbs
/:cl:

### DIFF
--- a/code/modules/organs/external/subtypes/robotic.dm
+++ b/code/modules/organs/external/subtypes/robotic.dm
@@ -38,6 +38,10 @@
 /obj/item/organ/external/robotic/removed()
 	deactivate(emergency=TRUE)
 	..()
+	
+/obj/item/organ/external/robotic/dropped()
+	removed()
+	..()
 
 /obj/item/organ/external/robotic/setBleeding()
 	return FALSE
--- a/code/modules/reagents/reagents/toxins.dm
+++ b/code/modules/reagents/reagents/toxins.dm
@@ -532,8 +532,11 @@
 /datum/reagent/toxin/slimetoxin/affect_blood(mob/living/carbon/M, alien, effect_multiplier)
 	if(ishuman(M))
 		var/mob/living/carbon/human/H = M
-		if(H.species.name != SPECIES_SLIME)
+		if(H.species.name != SPECIES_SLIME && !H.isSynthetic()) //cannot transform if already a slime perosn or lack flesh to transform
 			to_chat(M, SPAN_DANGER("Your flesh rapidly mutates!"))
+			for(var/obj/item/W in H) //Check all items on the person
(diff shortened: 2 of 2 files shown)


null-Aurelian:
Ah shit, I don't know how to deal with these integration test failures involving runtimes.

null-Aurelian:
@Firefox13, Unintended side effect was found; if the victim has a prosthetic brain but a fleshy torso they will die and become a ghost before the transformation goes into effect as the prosthetic brain is considered an implant. A slime/slime person will still form. Should this be a bug or a feature?

Firefox13:
Honest? A prosthetic brain shouldn't be a thing in the first place.

null-Aurelian:
Sounds like a vote for feature then.

hyperioo (approved):

hyperioo:
> @Firefox13, Unintended side effect was found; if the victim has a prosthetic brain but a fleshy torso they will die and become a ghost before the transformation goes into effect as the prosthetic brain is considered an implant. A slime/slime person will still form. Should this be a bug or a feature?

If you got a prosthetic brain and drink body mutating stuff, it can make sense that the toxin does not affect the brain and eject it while the rest of the body transforms.

PR #8339, opened 2023-10-07
Add K.A.R.L mining tool

hyperioo:
## About The Pull Request

Add K.A.R.L - Kinetic Acceleration Reconfigurable Lodebreaker. Three of them are available at roundstart inside the Hulk. More can be purchased at 2500 a piece from Aster's Solnishko Trade Station.

image

Sprites by Granodd 

## Why It's Good For The Game

_Following Holvr's mining rework:_

In order to properly emphasise the exploration part of the update, and to allow the miners to spelunk to their heart's content, we will give them THE TOOL. The tool is a miner’s best friend and lifeline, and should be cherished.

It’s a dual-purpose mining tool and weapon, which can reconfigure between a ranged and a melee state. A brief list of its features:
* Ranged mode is far less effective in an atmosphere than it is in a near-vacuum. This is so the miner’s don’t just start the round with an amazing all-purpose gun. It’s supposed to be great for mining and killing monsters.
* Melee mode has a toggleable energy blade, which makes it more damaging and a far faster mining tool.
* It runs on M class cells, which can be recharged by pumping the tool in ranged mode, or mining/attacking with it in melee OFF state melee mode. This is crucial for sustainability (we want the players to continue exploring and spelunking, instead of having to come to base for spare cells).
* Sports a modest Prying tool quality and a laughable Cutting quality when in OFF melee state, and a modest Welding quality in ON melee state, to help with navigating man-made POIs.

**In ranged mode**, the tool operates like a semi-automatic energy rifle. The weapon can fire semi-automatically as long as it has the charge to do so (think like the Cassad). To recharge the weapon while it’s in this form a pump is provided on the underside of the device.  Using the weapon in hand either by clicking on it or pressing Z will engage a series of mini-progress bars, automatically pumping the gun and continuing to do so until it can gain no more charge. This process can be interrupted by moving, putting the weapon away, among other things that would interrupt the process. The weapon can be pump charged with one hand, but it will take longer than pumping it while it’s wielded. The damage in a vacuum would be equivalent to something like a normal laser or mid-line assault rifle. When the weapon is used in atmosphere, the damage is more equivalent to a small arms pistol - not negligible, but not powerful enough to provide any significant threat.

**In Melee mode**, the weapon takes on the form of both a pick for mining, and an axe for swinging and dealing damage (all of its damage will be counted as sharp brute damage). This form of the weapon may be balanced to deal the same amount of damage in pressure as it does in vacuum due to it being melee, and not having the advantage of ranged. The melee mode itself has two states: powered and unpowered. When the weapon is not powered, it hits as hard as a normal pickaxe. When the weapon is powered, it will deal more melee damage at the cost of charge as well. Recharging the weapon in this form requires the energy blades to be off, and charge is gained upon hitting things with it in melee, or by mining out walls with it unpowered.

Switching between modes is done with **Unique Action** in object tab or a button that appears in the top left corner of the screen (like a flashlight on/off).

## Testing

* Spawned KARL
* Checked that battery can be swaped in and out.
* Checked that charge is consumed when switched on.
* Checked battery is recharged when used switched off.
* Checked switching between two modes
* Checked damage depending on pressure
* Checked recharging by pumping
* Check that UI button is working to switch mode

## Changelog
:cl: Hyperio
add: Add K.A.R.L mining tool
/:cl:


assortedbeads:
Based....so fucking based

assortedbeads:
Is there anyway to get new ones/more? Can you order them from a station?

Also what are the storage options? Im guessing suit storage only?

Mycah142:
When it's merged I might add it to the Aster's trade ship under Mining Tools.

Mycah142:
Also, suggestion: instead of yet another new verb, why not make the Unique Action be *spin_in_hand* or whatever it is? Or, better yet, UI button like flashlights?

hyperioo:
> Is there anyway to get new ones/more? Can you order them from a station?
> 
> Also what are the storage options? Im guessing suit storage only?

Storage options are the same than a standard pickaxe `slot_flags = SLOT_BELT` and `w_class = ITEM_SIZE_BULKY`

I'll add an option to buy some from Aster ship as Mycah suggested.

hyperioo:
> Also, suggestion: instead of yet another new verb, why not make the Unique Action be _spin_in_hand_ or whatever it is? Or, better yet, UI button like flashlights?

Yeah we were wondering if we should put it on unique action or the existing spin verb (just for hotkey purpose even the user isn't really spinning)

thevandie:
ROCK AND STONE!
FOR KARL!
(Doktor is NOT gonna be pleased with this, considering he fucking hate DRG for some reason)

Mycah142:
> > Also, suggestion: instead of yet another new verb, why not make the Unique Action be _spin_in_hand_ or whatever it is? Or, better yet, UI button like flashlights?
> 
> Yeah we were wondering if we should put it on unique action or the existing spin verb (just for hotkey purpose even the user isn't really spinning)

Or spin in hand to recharge, as activating/deactivating something is usually associated with Z (see: guns with multiple fire modes). I didn't think about this til now.

hyperioo:
Z is already used to turn the axe energy blade on/off.

Spin Held Item is not really reusable cleanly, it's a user verb that display a bunch of stuff before actually calling the `hand_spin` of the item

Mycah142:
Yeah I noticed that too, my little spin attack for whips uses the spininhand verb and the text is annoying as shit. I planned on redoing it so the text was dependent on what was spun, with the things having custom text flag string whatevers. But obviously that isn't the case right now.

master832:
Tool information button also switches weapon modes on my LAN.

hyperioo:
> Tool information button also switches weapon modes on my LAN. I didn’t notice it right away, it seemed like it was intended. It's worth renaming the button.

Thanks for telling me, I'll fix it. Just need to add a check in
```
/obj/item/tool/karl/ui_action_click(mob/user, actiontype)
	toggle_karl_mode(user)
```
because I believe it is called when you click the info UI button.

thevandie:
> Yeah I noticed that too, my little spin attack for whips uses the spininhand verb and the text is annoying as shit. I planned on redoing it so the text was dependent on what was spun, with the things having custom text flag string whatevers. But obviously that isn't the case right now.

Make spinning stuff in your hand be tied to a hotkey
Like autoequip
It'd make life so much easier

hyperioo:
Tying spinning verb to a hotkey would have to be done in another PR to keep stuff relatively atomistic. The best solution would be to tweak how spin works because right know if I put KARL mode change on spinning, it will spam the chat with "XXX spins the KARL in his hand" or some stuff like that.

PR #8343, opened 2023-10-23
Mapping disk smartfridges

TheLeopold:
## About The Pull Request
Finally mapping in disc smartfridges I added a year ago.

## Why It's Good For The Game
Convenient<sup>tm</sup> storage.

## Testing
cargo
club
cruch
engi
hammer
public
sci

## Changelog
:cl:
add: Added disk smartfridge next to each lathe
/:cl:


thevandie:
Why do the disks need to be cold?

hyperioo:
To cool down the electrons so that data does not leave the disk as fast as at ambient temperature.

PR #8345, opened 2023-10-28
Adjust Evac Voting Values

Mycah142:
## About The Pull Request
Getting folks to stick around as ghosts (either to wait for the next round, or to become a ghost antag, or just observe) is difficult with round duration averages passing the four hour mark. So I gave ghosts lightly more power, and Command/Antags slightly less power.

## Why It's Good For The Game
During rounds where something actually happens, there's usually a big ol' firefight between two factions, a bunch of people die. Sometimes there isn't even an antag involved. The round hit it's climax for most players, many of them dying. Many of the players are satisfied, the story ridden to it's conclusion.

Problem is, that one MEO who hasn't left RnD and that one carrion who has been deep maint diving for 3 hours voted NO to Evac, so now the 13 players who are dead have to sit there and watch these two numbskulls do absolutely nothing because they decided to not participate with other players and just play solo minecraft.

13 dead players is 3.9 voting strength. The MEO and one carrion is 4 voting strength. So two players overcame 13 dead players. And that's at a simple 50/50 split. If we actually take the 60% majority needed to leave into consideration, there would need to be over 18 dead players to out-vote TWO PEOPLE. This is simply madness. It isn't "salt vote" or "ided" at that point. Especially since command staff, once dead, then have a .3 vote regardless.

This isn't the american voting system. Players are players, regardless of whether they're a ghost or not. They deserve to have more of a say.

## Testing
dreamseeker_puneV4Gkwo

## Changelog
:cl:
tweak: Increased dead vote
tweak: Decreased Command/Antag vote
tweak: Balanced Majority Value to 55%
/:cl:

### DIFF
--- a/code/controllers/subsystems/voting/poll_types.dm
+++ b/code/controllers/subsystems/voting/poll_types.dm
@@ -158,7 +158,7 @@
 	name = "Initiate Bluespace Jump"
 	question = "Do you want to initiate a bluespace jump and restart the round?"
 	time = 120
-	minimum_win_percentage = 0.6
+	minimum_win_percentage = 0.55
 	cooldown = 20 MINUTES
 	next_vote = 90 MINUTES //Minimum round length before it can be called for the first time
 	choice_types = list()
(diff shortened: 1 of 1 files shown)


thevandie:
Amen

PR #8375, opened 2023-11-25
Removes painkilling effects from alcohol in blood and makes it metabolize into ethanol

MLGTASTICa:
## About The Pull Request
Alcohols no longer have any painkilling effects whilst in blood , now instead they metabolize into ethanol better.

## Why It's Good For The Game
You could inject 3 different alcohols and get more painkilling than oxycodone.

## Testing
Tested on local

## Changelog
:cl:
balance: Alcohol in blood no longer has painkilling effects , it instead metabolizing into ethanol.
/:cl:


Firefox13 (approved):
sigh

PR #8381, opened 2023-11-30
Egg Rebalance

vode-code:
## About The Pull Request
This PR makes the first few eggs start fertile, and after that eggs are less likely to be fertile depending on the number of chickens. This PR also adds a new "candling" mechanic which allows for progress monitoring of eggs.

## Why It's Good For The Game
Farming Chickens is currently rather difficult, as only one in ten eggs can hatch, and it is invisible which are which; this PR solves those issues both by making eggs start more likely to hatch, and by making it possible to tell if eggs have begun to grow.

## Testing
spawned a chicken, fed it wheat, let the eggs hatch, used an unlit flare to fail to check the levels and used a lit flare on them to check the levels. After allowing flames to check the levels, used an unlit zippo to fail to check the levels and a lit zippo to check the levels.

## Changelog
:cl: Chickenish
add: you can now use a flashlight or candle on an egg to see how grown it is.
tweak: chicken eggs now start guaranteed fertile and less of them are fertile the more chickens exist.
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/food/snacks.dm
+++ b/code/modules/reagents/reagent_containers/food/snacks.dm
@@ -668,6 +668,7 @@
 	center_of_mass = list("x"=16, "y"=13)
 	preloaded_reagents = list("egg" = 3)
 	price_tag = 5
+	description_info = "Can have a flashlight used on it to determine if there is a chick inside."
 
 /obj/item/reagent_containers/food/snacks/egg/afterattack(obj/O as obj, mob/user as mob, proximity)
 	if(istype(O,/obj/machinery/microwave))
@@ -698,6 +699,20 @@
 			return
 		to_chat(user, SPAN_NOTICE("You color \the [src] [clr]"))
 		icon_state = "egg-[clr]"
+	else if (istype(W, /obj/item/device/lighting))
+		var/obj/item/device/lighting/light = W
+		if(!light.on)
+			to_chat(usr, SPAN_WARNING("[W] needs to be turned on to reveal the egg's insides."))
+			return
+		switch(amount_grown)
+			if(0 to 10)
+				to_chat(usr, SPAN_NOTICE("[src] appears to be a normal egg."))
+			if(10 to 50)
+				to_chat(usr, SPAN_NOTICE("[src] contains a spidery red mass."))
+			if(50 to 90)
+				to_chat(usr, SPAN_NOTICE("[src] contains a partially-grown chick."))
+			if(90 to 100)
+				to_chat(usr, SPAN_NOTICE("[src] contains a partially-grown chick.\nYou hear a faint tapping emanating from [src]."))
 	else
 		..()
 
--- a/code/modules/reagents/reagent_containers/food/snacks.dm
+++ b/code/modules/reagents/reagent_containers/food/snacks.dm
@@ -699,22 +699,33 @@
 			return
 		to_chat(user, SPAN_NOTICE("You color \the [src] [clr]"))
 		icon_state = "egg-[clr]"
-	else if (istype(W, /obj/item/device/lighting))
-		var/obj/item/device/lighting/light = W
-		if(!light.on)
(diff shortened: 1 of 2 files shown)


Firefox13:
>  tweak: all chicken eggs are now fertile

And... What if one wants to sell those to kitchen?

vode-code:
> And... What if one wants to sell those to kitchen?

Due to farm_animals.dm lines 256/263, an egg being *anywhere* that is not the floor will kill it, which is the other reason you might need to check the growth progress than wanting to know when it hatches.
if I read the code right, even holding the egg for too long will make it die.

vode-code:
> as a result, simply placing them into a container makes hatching impossible to occur, which solves said issue.

alternatively, simply exceeding the threshold will sterilize all eggs created after that, which allows for easy sale of said eggs regardless of neglect, as eggs currently do not rot.

thevandie:
> as a result, simply placing them into a container makes hatching impossible to occur, which solves said issue.

counterpoint, pull.

githubuser4141:
What if the fertility was opt-in rather then opt out? You can buff an egg's chances of fertility by using a chemical or device on it?

vode-code:
> What if the fertility was opt-in rather then opt out? You can buff an egg's chances of fertility by using a chemical or device on it?

I would prefer obtaining more chickens being easier for the farmer than harder at first, although that idea may be necessary depending on if 100% fertility rate is too much.

Humonitarian:
It seems good, but what about a reducing fertility chance?

The more chickens there are, the less chance of fertile eggs, all the way to 0 once reaching the critical chicken threshold

vode-code:
> It seems good, but what about a reducing fertility chance?
> 
> The more chickens there are, the less chance of fertile eggs, all the way to 0 once reaching the critical chicken threshold
> 
> Edit: Also consider infertile eggs have the added advantage of allowing sale-ready eggs earlier into the round

all eggs become sale-ready once you put them in a box, even so, I have implemented your suggestion for other reasons.

Mycah142:
As an agrolyte main, I never had a problem having both eggs and new chickens. I usually did the following:

Create chairs
Bound chickens to chairs
Fed wheat
Wait a while, probably 30 minutes of doing other shit
There's now a bunch of eggs and new chickens wandering the pen
I use the produce bag to pick up the clearly infertile eggs
Close the new chickens into the airlock to humanely euthanize for purpose of meat

So, while I don't really get why this idea is here, I can see that it's trying to make the Surplus Chickens appear faster, which isn't a bad idea.

I assume the fertility decrease is per chicken? If so, it would mean killin' the named chickens, which is haram and I will have to call the police.

vode-code:
> I assume the fertility decrease is per chicken? If so, it would mean killin' the named chickens, which is haram and I will have to call the police.

the fertility decrease only begins once the tenth chicken has grown, and there are less than ten roundstart chickens. as a result, one does not have to kill any of the starter chickens while maximizing chicken rate.

Humonitarian (approved):
Has my approval regarding the design.

code/modules/reagents/reagent_containers/food/snacks.dm
```diff
@@ -699,7 +700,32 @@
...
+		if(valid)
+			switch(amount_grown)
+				if(0 to 10)
+					to_chat(usr, SPAN_NOTICE("[src] appears to be a normal egg."))
+				if(10 to 50)
+					to_chat(usr, SPAN_NOTICE("[src] contains a spidery red mass."))
+				if(50 to 90)
+					to_chat(usr, SPAN_NOTICE("[src] contains a partially-grown chick."))
+				if(90 to 100)
+					to_chat(usr, SPAN_NOTICE("[src] contains a partially-grown chick.\nYou hear a faint tapping emanating from [src]."))
```
hyperioo, line 726:
```suggestion
					to_chat(usr, SPAN_NOTICE("[src] contains a partially-grown chick.\nYou hear a faint tapping emanating from \the [src]."))
```

PR #8386, opened 2023-12-22
Removes Ammunition Magazine Exploit with minor knockon effect of oldification

vode-code:
## About The Pull Request
This PR changes bullets and the ammo casings that contain them in these ways: both hold their own matter, casings have cleaner code, casings have double the price tag if unexpended, casings have a 10% chance to lose their bullets on being oldified, bullets are immune to oldification, darts are now caseless instead of using broken code to replicate the function, and bullets transfer their first matter to the shrapnel object on human impact.
This PR also reduces the direct cost of many objects that spawn containing bullets, changes the matter list of breacher AMR shells, and allows for items that do have matter but are contained within items that do not have matter to be eaten by autolathes.

## Why It's Good For The Game
Bullets, and the Casings that held them, used to be completely free before this PR.
This was very bad for the economy, as bullets are the most-used consumable within this game.
Most everything else is either making things compatible with these changes, or cleaning up problems found during the creation of this PR.
This PR should improve the economy.

## Testing
spawned havelock and lethal ammo disk, printed one .35 speedloader and one .35 SMG magazine, fired Havelock 36 times, recycled everything I printed, and lost exactly the correct amount of steel for it: 1.8.
fired revolver, maxim, pump shotgun, and rifle, placed the spent shells at the front of the queue, each case had one failure to fire but upon firing again, the appropriate result occurred.
after making caseless cases expended by oldification self-destruct, oldified and fired several box mags of caseless rifle ammunition without finding issues.

## Changelog
:cl: Chickenish
add: it is now possible to lathe-recycle containers that contain items with materials, even when the containers contain no materials.
tweak: Oldified Ammo Casings now rarely spawn spent.
tweak: Breacher Rounds recipe changed from uranium to plasteel.
balance: Bullets now cost materials instead of the magazine holding the entire price.
del: removed bullet inscription feature.
fix: dartguns now less broken!
/:cl:


Humonitarian:
Title is misleading, this PR does more than just remove an exploit.

> casings have a 10% chance to lose their bullets on being oldified

If this jams guns, wait for Firefox approval. Jamming is a rather controversial feature we probably do not want.

Otherwise, good PR and DevDel approval.

vode-code:
> Title is misleading, this PR does more than just remove an exploit.
> 
> > casings have a 10% chance to lose their bullets on being oldified

This is a replacement for the default oldification behaviour, which is to decrease the contained matter by up to 50%.

> 
> If this jams guns
on testing, this did not jam a revolver, a machine gun, pump action shotgun, nor magazine-fed rifle.

vode-code:
turns out dartguns were broken due to improper Qdeling, so I fixed that too.

code/modules/scrap/oldificator.dm
```diff
@@ -412,3 +413,28 @@
...
+/obj/item/projectile/make_old(low_quality_oldification)
+	return // why would the bullet being old change anything?
+
+/obj/proc/pre_old() // defined for compatibility
+	return
+
+/obj/item/ammo_magazine/pre_old(low_quality_oldification = FALSE) // this is needed to allow casings
+	if(!low_quality_oldification) // contained to self-delete
+		for(var/obj/item/ammo_casing/casing in stored_ammo)
+			casing.make_old() // this doesn't technically oldify anything, so can be done here
```
hyperioo, line 440:
```suggestion
			casing.make_old() // this doesn't technically oldify anything, so can be done here

```

code/modules/projectiles/projectile.dm
```diff
@@ -673,6 +673,17 @@
...
 
+/obj/item/projectile/get_matter()
+	. = matter?.Copy()
+	if(isnull(.)) // empty bullets have no need for matter handling
+		return
+	if(istype(loc, /obj/item/ammo_casing)) // if this is part of a stack
+		var/obj/item/ammo_casing/case = loc
+		if(case.amount > 1) // if there is only one, there is no need to multiply
+			for(var/mattertype in .)
+				.[mattertype] *= case.amount
```
hyperioo, line 684: Is it normal that when getting the matter a single projectile, the returned matter is multiplied by the number of projectiles in the stack? Or is gun code so cursed that it is an expected behaviour to have the matter of the whole stack returned?

vode-code: the single projectile is the only projectile in the entire stack, and must be treated as the entire stack to allow lathe code to know how much matter is "in" the stack of casings, as otherwise the "virtual projectiles", which only exist hypothetically, would be ignored.

hyperioo (approved):

PR #8413, opened 2024-01-26
ErisMed fixes Season 1 Episode 2

AltHit:
## About The Pull Request

**Can you hear the visible_message?**
**Can you see the OP_EYES?**
**Can you fix the ORGAN_BROKEN?**
**CAN YOU FEEL MY random_organ_by_process(OP_HEART)?**

Fixed typo in comment
Removed excess whitespace in techno belt line
Fixed people walking without vital organs like they are still inside
Now you actually need to attach organs in order for them to work and by default they are unattached Resuscitator now actually tells you what's wrong during resurection
Also robotist got a new fancy belt instead of toolbox!
Resuscitator now checks if the body is too damaged, as it was intended to do

## Why It's Good For The Game

Removes a critical bug that made you semi-invincible with need to only keep the brain inside your body, I guess this is gud
Gives robotist vital tools in fixing FBP's roundstart and takes rudimentry blue toolbox no-one cared about
Adds actual feedback to resuscitator so doctors shouldn't silently ask themself "WHY are you dead?"
And the fact that organs are need to be attached first is just a neat addition

## Testing

- [x] Check if the robotist belt is spawning correctly
- [x] Check slime and slimehumans spawning correctly
- [x] Check Carrion spawning correctly and without flags that should not be here
- [x] Check FBP's spawning correctly
- [x] Check the cyanide
- [x] Check if the heart is dying fast enough
- [x] Check surgery healing by items
- [x] Check that unhealable eye wound
- [x] Check organs dropped by gibbing
- [x] Check if everyone can see resus messages correctly

## Changelog
:cl:
add: Put multitool and pliers into robotist toolbelt
del: Removed roundstart toolbox from robotist
tweak: Tweaked organs to be unattached by default
fix: Fixed stackable items consumed when used in surgery
fix: Fixed organs counted inside person after removal in some scenarios
fix: Fixed resuscitator ignoring all damage on body upon resurrection
spellcheck: fixed typo in signals.dm and removed one char is belt code
fix: Fixed runtime with 2-step organ wounds
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -4,6 +4,7 @@
 	voice_name = "unknown"
 	icon = 'icons/mob/human.dmi'
 	icon_state = "body_m_s"
+	status_flags = REBUILDING_ORGANS //will protect from gibbing before organs are built. flag should be removed by set_species in initialize
 
 	var/list/hud_list[10]
 	var/embedded_flag	  //To check if we've need to roll for damage on movement while an item is imbedded in us.
@@ -1020,6 +1021,10 @@ var/list/rank_prefix = list(\
 	for(var/obj/item/organ/internal/carrion/C in organs_to_readd)
 		C.replaced(get_organ(C.parent_organ_base))
 
+	for(var/obj/item/organ/internal/I in internal_organs)
+		I.status &= ~ORGAN_CUT_AWAY
+		I.handle_organ_eff()
+
 	status_flags &= ~REBUILDING_ORGANS
 	species.organs_spawned(src)
 
@@ -1348,13 +1353,24 @@ var/list/rank_prefix = list(\
 	reset_view(A)
 
 /mob/living/carbon/human/proc/resuscitate()
-	var/obj/item/organ/internal/vital/heart_organ = random_organ_by_process(OP_HEART)
-	var/obj/item/organ/internal/vital/brain_organ = random_organ_by_process(BP_BRAIN)
+	
+	var/obj/item/organ/internal/vital/heart/heart_organ = random_organ_by_process(OP_HEART)
+	var/obj/item/organ/internal/vital/brain/brain_organ = random_organ_by_process(BP_BRAIN)
 
-	if(!is_asystole() && !(heart_organ && brain_organ) || (heart_organ.is_broken() || brain_organ.is_broken()))
+	if((!heart_organ || heart_organ.is_broken()) && (!brain_organ) || brain_organ.is_broken())
+		visible_message(SPAN_WARNING("\The [src] lay still, devoid of any hint of vitality or warmth."))
+		return 0
+
+	if(!heart_organ || heart_organ.is_broken())
+		visible_message(SPAN_WARNING("\The [src] twitches a bit, but their body remains lifeless, unresponsive to any stimulus."))
+		return 0
+
+	if(!brain_organ || brain_organ.is_broken())
+		visible_message(SPAN_WARNING("\The [src] changes is colour, but remains tranquil and utterly still."))
 		return 0
 
 	if(world.time >= (timeofdeath + NECROZTIME))
+		visible_message(SPAN_WARNING("\The [src] shows no signs of reaction, a grim reminder of the boundaries that separate life and death."))
 		return 0
 
 	var/oxyLoss = getOxyLoss()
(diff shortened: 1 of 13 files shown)


code/modules/organs/internal/_internal.dm
```diff
@@ -85,14 +90,25 @@
...
+			owner.internal_organs_by_efficiency[process] -= src
 
 	for(var/proc_path in owner_verbs)
 		verbs |= proc_path
 
 	if(GetComponent(/datum/component/internal_wound/organic/parenchyma))
 		owner.mutation_index++
 
+/obj/item/organ/internal/proc/handle_organ_eff()
+	owner = src.owner
```
hyperioo, line 105: I'm not sure I get what this one is doing. Assign owner to owner?

code/modules/organs/internal/_internal.dm
```diff
@@ -85,14 +90,25 @@
 	for(var/process in organ_efficiency)
 		if(!islist(owner.internal_organs_by_efficiency[process]))
 			owner.internal_organs_by_efficiency[process] = list()
-		owner.internal_organs_by_efficiency[process] += src
+		if(src.is_usable())
```
hyperioo, line 93:
```suggestion
		if(is_usable())
```
`src` can be omitted. Same for the other `src.is_usable()`

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -1348,21 +1353,32 @@ var/list/rank_prefix = list(\
...
+		visible_message(SPAN_WARNING("\The [src] shows no signs of reaction, a grim reminder of the boundaries that separate life and death."))
 		return 0
 
 	var/oxyLoss = getOxyLoss()
 	if(oxyLoss > 20)
 		setOxyLoss(20)
 
 	if(health <= (HEALTH_THRESHOLD_DEAD - oxyLoss))
-		visible_message(SPAN_WARNING("\The [src] twitches a bit, but their body is too damaged to sustain life!"))
+		visible_message(SPAN_WARNING("\The [src] twitches a bit, but their body is too damaged to sustain life."))
```
hyperioo, line 1381: I think the message could be a bit more clear. It's better than nothing but without looking at the code, I'm not sure people will fully understand what the messages indicate in term of missing heart/brain.

code/modules/surgery/organic.dm
```diff
@@ -142,7 +142,7 @@
 	)
 
 	organ.status &= ~ORGAN_CUT_AWAY
-	organ.replaced(organ.get_limb())
+	organ.handle_organ_eff() //organ is attacheds. Refreshing eff. list
```
hyperioo, line 145:
```suggestion
	organ.handle_organ_eff() //organ is attached. Refreshing eff. list
```

code/modules/surgery/organic.dm
```diff
@@ -178,6 +178,7 @@
 		SPAN_NOTICE("You separate [organ.get_surgery_name()] with \the [tool].")
 	)
 	organ.status |= ORGAN_CUT_AWAY
+	organ.handle_organ_eff() //detaTch of organ. Refreshing eff. list
```
hyperioo, line 181:
```suggestion
	organ.handle_organ_eff() //detach of organ. Refreshing eff. list
```

code/modules/mob/living/carbon/human/human.dm
AltHit: > I think the message could be a bit more clear. It's better than nothing but without looking at the code, I'm not sure people will fully understand what the messages indicate in term of missing heart/brain.

I'll redo message part to be more depended on stats of people looking

code/modules/organs/internal/_internal.dm
AltHit: It's nothing more than a messy copypasting from other proc, does absolutely nothing.
You deserve a medal for each time shit like this doesn't get into upstream

Mycah142:
Part of the draft is a fix for eyes having unrecoverable wounds, and med kits being fully consumed without actually healing wounds.

These are a bigger issue than an Adjustment PR as they were introduced by a supposed Eye Damage Bugfix.

I'd like to request these two fixes as their own PR to be ensured as working and merged as soon as they're confirmed to actually fix the now two or three weeks of the issue.

The PR that introduced them honestly should've been reverted immediately, but here we are.

AltHit:
> Check Mycah's message above about the eye wound and medkit issues for your next ErisMed fix season 1 episode 3. 👍

Already done both of them here! Thanks for the review by the way

PR #8434, opened 2024-02-16
Add sound effect to space bears

Twomoon-Github:
## About The Pull Request

Add sound effect to space bear
## Why It's Good For The Game

Make the space bear more lively
being slashed by a bear is more noticable, as before it just stood next to you without moving or making sound, and you only knew you were attacked by checking the chat or seeing your character getting bloodier
## Testing

the mob behave the same as before the change, and the sounds feels good in game.
sadly i didnt understood why OBS refused to record the game audio so i didnt got you a video.

## Changelog
:cl:

soundadd: added sound effects to the space bears

/:cl:


Twomoon-Github:
btw, sounds are taken from amateur videos of actual bears on youtube, and free to use sound effects.

Firefox13 (commented):
Can these be played with perhaps a 1/3 chance? They can get *very* spammy otherwise. Use prob() for that, ask on discord if you need explanation

On unrelated note, pls if a PR is centered around sounds add them into PR description so you don't need to download them to listen.

Twomoon-Github:
> Can these be played with perhaps a 1/3 chance? They can get _very_ spammy otherwise. Use prob() for that, ask on discord if you need explanation
> 
> On unrelated note, pls if a PR is centered around sounds add them into PR description so you don't need to download them to listen.

changed that with the help of handyman, it works, sounds are less common, also the attack now can sometime be a growl too.
sadly i cant record, but basically, its just better, especially since now the attack is either a growl or a roar randomly, out of the 66 % chance (enough for it to notice being attacked as 33% made the sound quite rare, and with the variety it feels even less spammy than at 33% with one sound), for other stance its 33%

hyperioo (approved):

PR #8440, opened 2024-02-25
Connects missing wires

Kurgis:
## About The Pull Request
Small oversight where a downward wire was never connected to a wire below it.
image

## Why It's Good For The Game
Connects missing wires

## Testing
Tested on local, wires are connected.

## Changelog
:cl:
fix: Adds missing wire section within maintenance.
/:cl:

### DIFF
--- a/maps/CEVEris/_CEV_Eris.dmm
+++ b/maps/CEVEris/_CEV_Eris.dmm
@@ -19512,6 +19512,11 @@
 "aWd" = (
 /obj/machinery/atmospherics/pipe/simple/hidden/supply,
 /obj/effect/window_lwall_spawn/smartspawn,
+/obj/structure/cable/green{
+	d1 = 1;
+	d2 = 2;
+	icon_state = "1-2"
+	},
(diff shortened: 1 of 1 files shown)


Firefox13:
Um... Isn't it connected in the different place? Follow the existing wire.

Firefox13:
Rather, the wire from the club. It is connected to the substation.

Kurgis:
> Um... Isn't it connected in the different place? Follow the existing wire.

You are correct, though regular course of gameplay tends to disconnect it, which is how I happened to notice the missing connection here.

Firefox13:
> regular course of gameplay tends to disconnect it

This is not a mapping issue.

Firefox13:
So.

Wires are actually connected (if you do not believe me, I can provide a video later), and redundancy in wires and supply/scrubber pipes isn't usually a thing in mapping. You can look at Eris, at /tg/, at bay and other servers, pipe and wire systems are meant to be easily sabotageable, for better or worse.

Mostly worse.

PR #8444, opened 2024-02-26
Braingym

Twomoon-Github:
## About The Pull Request

Modify the library to have new machines to train cognition, biology, and mechanic.
Those were intended to be created with the gym with their own perk to replace exertion (deemed reason) but were never made

Make energy drink reduce the deemed reason cooldown (the same way protein shake reduce exertion)

oh and also, fixed a light fixture in the club basement that was placed on the door, just moved it to the wall next to the door so it doesnt looks weird.

braingym

## Why It's Good For The Game

Those were planned + give an use to the library instead of being a dark pile of wood used by vagabond for easy money at the recycler, with now machines to train yourself and also some books got added.

## Testing

I have tested: the current gym machines are unchanged and still works the same way.
the braingym machines works the same way as the other gym machine but give deemed reason (reduce cog for a time) and the deemed reason perk is reduced in time when drinking energy drinks.

https://github.com/discordia-space/CEV-Eris/assets/45800756/6ff8a039-959d-46b7-b000-c6569decf254

## Changelog
:cl:
add: Added brain gym machines
sounddel: removed an old sound 
/:cl:


Twomoon-Github:
about the club light, it was like this 
clubwtf
now the light is on the wall just north of the door

MLGTASTICa:
stat power creep.... they all lose meaning i guess.

Twomoon-Github:
> stat power creep.... they all lose meaning i guess.

how so? the gym isnt that powerful, getting a basic odditie works better and is way cheaper to get stats, i just continued what was in the doc in the kot-map and was intended to come with the gym

Firefox13 (changes requested):
Before I look over the code please create a separate 32x32 file for new brain gym machines and, you know it, switch them to 32x32.
We do not have proper support for such big sprites and it creates clipping issues.

Twomoon-Github:
> Before I look over the code please create a separate 32x32 file for new brain gym machines and, you know it, switch them to 32x32. We do not have proper support for such big sprites and it creates clipping issues.

the gym icons use this size, so i just followed it by making the sprites on the same files

Mycah142:
The player mob interacts with the structures, like the gym? Animated?

Twomoon-Github:
> The player mob interacts with the structures, like the gym? Animated?

yes it do, it display an animation, more subtle, but it still place the player on the same tile (thats why you have things like, "0, -10" in the code, so the player doesnt appear right on top of the machine as its smaller and different in look)

Mycah142:
> > The player mob interacts with the structures, like the gym? Animated?
> 
> yes it do, it display an animation, more subtle, but it still place the player on the same tile (thats why you have things like, "0, -10" in the code, so the player doesnt appear right on top of the machine as its smaller and different in look)

With this in mind, Fox, is the larger sprite not equally valid here as the gym?

Firefox13:
It was a mistake to allow first gym sprites to be 32x48. Again, they look glaringly large compared to other objects and create heavy clipping issues. You can shorten existing sprites.

You copied a lot of code from gym machines, instead you should make brain machines their subtype instead of creating another type by tinkering with existing code.

How do animations kook?

Also mapping is just bad.

Kurgis:
> Also mapping is just bad.
It's going to be exceedingly difficult for contributors to pass this bar, as the very foundation of Section Three's layout is subpar in comparison to the surrounding ship.

Firefox13:
Mapping is fixable without changing the room size, the library is just too stretched out  
not the sections problem.

Twomoon-Github:
> It was a mistake to allow first gym sprites to be 32x48. Again, they look glaringly large compared to other objects and create heavy clipping issues. You can shorten existing sprites.
> 
> You copied a lot of code from gym machines, instead you should make brain machines their subtype instead of creating another type by tinkering with existing code.
> 
> How do animations kook?
> 
> Also mapping is just bad.

yes at first i made gym/braingym but then it would give exertion too due to the proc/go_out code block.
making its own type was an easy fix.

i know this isnt gonna be merged and im not going to rework the gym code and sprites, at least i practiced some pixel art :>

also nuh uh my mapping is flawless xd

Mycah142:
HMMMMMM.... yoink

PR #8447, opened 2024-03-03
Massively increases health of stairs

ValoTheValo:
stairs now have a massive amount of health. they shouldn't be getting blown up by a hand grenade anymore

## About The Pull Request
this pull request increases the health of stairs by a large amount, making them functionally immortal. due to an oversight in recursive explosions, stairs were given the default healthbar of 100, making it so any explosion could entirely destroy a staircase and make an area inaccessible. as i do not know how to make staircases immortal, i just gave them a massive amount of health to make them effectively immortal

## Why It's Good For The Game
fixes oversight, staircases being destroyed by a single frag and being unreparable is not good

## Testing
compiled, threw a dozen or so frag grenades at a staircase and it did not get destroyed. also tested with HE and Admin bomb

## Changelog
:cl:
balance: massively increased health of staircases so they don't get destroyed
/:cl:

### DIFF
--- a/code/modules/multiz/structures.dm
+++ b/code/modules/multiz/structures.dm
@@ -9,6 +9,8 @@
 	anchored = TRUE
 	icon = 'icons/obj/stairs.dmi'
 	bad_type = /obj/structure/multiz
+	health = 5000
+	maxHealth = 5000
 	var/istop = TRUE
 	var/obj/structure/multiz/target
 	var/obj/structure/multiz/targeted_by


MLGTASTICa:
Wasn't a oversight , they should be buildable in first place. They were getting deleted before too.

ValoTheValo:
> Wasn't a oversight , they should be buildable in first place. They were getting deleted before too.

it took a maxcap prior to maybe delete one stair. now they take a single frag grenade to delete all four

thevandie:
> Wasn't a oversight , they should be buildable in first place. They were getting deleted before too.

They aren't buildable pal. So that's the problem.

Firefox13 (approved):

Humonitarian (approved):

PR #8450, opened 2024-03-05
Maint decorated: graffitis

Twomoon-Github:
## About The Pull Request

tiny map change to add a few graffitis in maint
## Why It's Good For The Game

Adds more flavor and style to maint, made sure i dont place too much of them, and only in maint and even very isolated part of maints to not overdo it.
made sure to get them only visible from a specific side of the wall using pixel shifting to keep it consistent and not weird
## Testing

see pics:
showcase

## Changelog
:cl:
add: Added some graffis in maint as decoration
/:cl:

### DIFF
--- a/maps/CEVEris/_CEV_Eris.dmm
+++ b/maps/CEVEris/_CEV_Eris.dmm
@@ -13203,6 +13203,9 @@
 /obj/spawner/pack/tech_loot/low_chance,
 /obj/spawner/gun_parts/low_chance,
 /obj/spawner/gun_parts/low_chance,
+/obj/effect/decal/cleanable/graffiti{
+	pixel_y = 33
+	},
 /turf/simulated/floor/tiled/steel,
 /area/eris/maintenance/section2deck4central)
 "aGH" = (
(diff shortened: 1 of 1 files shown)


Firefox13 (commented):
Why pixel offsets are different everywhere?
Do you think you can instead create a random spawner for graffiti instead? So it'll pick one of existing ones instead of them being pre-set.

Twomoon-Github:
> Why pixel offsets are different everywhere? Do you think you can instead create a random spawner for graffiti instead? So it'll pick one of existing ones instead of them being pre-set.

pixel offset are different for if they are north south of the wall etc but also depending on the graffiti, iriska graffiti being bigger than the piss graffiti for example, it doesnt look as good if i give piss the same pixel shift. i just made them to be well placed.

for a graffiti spawner, that might be hard ngl, as it would need to pixel shift it to the nearest wall, if we place it on a wall its like when placing a poster and will be visible from all side. poster are preset on a tile, and pixel shifted manually to the wall in the map file

Firefox13 (approved):
Adding spawners for graffities will require *is possible* but may simply not worth the effort.
Unless Humon or someone else comes up with a good solution this looks good enough, adding a little bit more life to maints.

PR #8464, opened 2024-03-23
Update recipes.dm

Hanz-Moon:
## About The Pull Request
Fixes typo in code which prevents eye protection from stopping flash powder flashes

## Why It's Good For The Game
flash powder no longer flashes through eye protection

## Testing
tested on local by mixing flash powder with different levels of eye protection

## Changelog
:cl:
fix: flash powder no longer goes through eye protection
/:cl:

### DIFF
--- a/code/modules/reagents/recipes.dm
+++ b/code/modules/reagents/recipes.dm
@@ -660,11 +660,11 @@
 	for(var/mob/living/carbon/M in viewers(world.view, location))
 		switch(get_dist(M, location))
 			if(0 to 3)
-				if(M.eyecheck() <= FLASH_PROTECTION_MAJOR)
+				if(M.eyecheck() < FLASH_PROTECTION_MAJOR)
 					M.flash(15, FALSE , FALSE , FALSE)
 
 			if(4 to 5)
-				if(M.eyecheck() <= FLASH_PROTECTION_MAJOR)
+				if(M.eyecheck() < FLASH_PROTECTION_MAJOR)
 					M.flash(0, FALSE , FALSE , FALSE)
 
 /datum/chemical_reaction/emp_pulse


MLGTASTICa:
Its a balance change , not a fix. There is no code comment showing it was never intended to go through people only protected by 1 flash layer.

ValoTheValo:
good PR

Mycah142:
> Its a balance change , not a fix. There is no code comment showing it was never intended to go through people only protected by 1 flash layer.

Why, did you intend on a single form of Flash bypassing protection without ever stating such in any form, so only you knew about it?

Hanz-Moon:
> Its a balance change , not a fix. There is no code comment showing it was never intended to go through people only protected by 1 flash layer.

i was under the impression that since the code was written with a typo it wasn't intended to be so, is there any reason for the code to be there if it's not being used?

SirRichardFrancis:
> Its a balance change , not a fix. There is no code comment showing it was never intended to go through people only protected by 1 flash layer.

I think Moon's point is all about code being silly with checking eye protection.
You see, `M.eyecheck()` meant to return value up to 2 (which is maximum flash protection, like wearing really flash-protective gear or not having eyes that could be flashed), but proc in question flashes a mob even if they have maximum possible protection.

This proc basically makes no sense and flash protection check does literally nothing with `<=` instead of `<`, which I'm sure SPCR mean to use, otherwise this check wouldn't be present at all.

MLGTASTICa:
> > Its a balance change , not a fix. There is no code comment showing it was never intended to go through people only protected by 1 flash layer.
> 
> I think Moon's point is all about code being silly with checking eye protection. You see, `M.eyecheck()` meant to return value up to 2 (which is maximum flash protection, like wearing really flash-protective gear or not having eyes that could be flashed), but proc in question flashes a mob even if they have maximum possible protection.
> 
> This proc basically makes no sense and flash protection check does literally nothing with `<=` instead of `<`, which I'm sure SPCR mean to use, otherwise this check wouldn't be present at all.

My point is that flash protection can and does go above 2. If you're wearing a welding helmet and goggles

SirRichardFrancis:
> My point is that flash protection can and does go above 2. If you're wearing a welding helmet and goggles

Sure, you may stack multiple sources of strong flash protection and proc will return number above 2... but only because there is really no reason to cap this number at it's intended max value, because nothing (aside of this powder apparently) would react differently to protection above maximum.

Protection 2 is what blind mob has, that's implied immunity to being flashed... so what were you thinking?
And this powder only required sunglasses before.

MLGTASTICa:
> > My point is that flash protection can and does go above 2. If you're wearing a welding helmet and goggles
> 
> Sure, you may stack multiple sources of strong flash protection and proc will return number above 2... but only because there is really no reason to cap this number at it's intended max value, because nothing (aside of this powder apparently) would react differently to protection above maximum.
> 
> Protection 2 is what blind mob has, that's implied immunity to being flashed... so what were you thinking? And this powder only required sunglasses before.

Blind mobs have a special check , they don't check for flash protection. Also no , all flashes now require heavy welding helmets, it was changed.

Hanz-Moon:
i'm gonna do more testing with other flash sources and re-label it as a balance change then

SirRichardFrancis:
> Blind mobs have a special check , they don't check for flash protection

image

> Also no , all flashes now require heavy welding helmets, it was changed.

Not keeping up with combat balance changes, but isn't jump from "sunglasses will do" and "that will get even the blind" is a bit too much? Besides, related code comments do indicate that flash protection not meant to go above 2.

MLGTASTICa:
> > Blind mobs have a special check , they don't check for flash protection
> 
> image
> 
> > Also no , all flashes now require heavy welding helmets, it was changed.
> 
> Not keeping up with combat balance changes, but isn't jump from "sunglasses will do" and "that will get even the blind" is a bit too much? Besides, related code comments do indicate that flash protection not meant to go above 2.

fair enough

Firefox13:
> tested on local by mixing flash powder with different levels of eye protection

And what were the test results?

Hanz-Moon:
> > tested on local by mixing flash powder with different levels of eye protection
> 
> And what were the test results?

without the change the only thing that blocks flash powder is welding goggles/hudglasses and welding mask, with the change it's now on the same level as flashbangs, flashes and welding

Humonitarian (approved):

PR #8469, opened 2024-04-05
Rework the mech shield generator to have an internal capacitor, refactor shield absorption code

XElectricX:
## About The Pull Request

Mainly this was just going to be about adding a capacitor mechanic to the mech energy shield but I have also refactored some of the shield and damage code.

- Mech shield now has a power bank that it recharges using the mech battery; every unit of damage taken is a unit of charge depleted
- Mech shield also drains power when online, but only enough to break even from the charging it receives (turn it off to let it recharge to 100%)
- Will shut down if it takes too much damage
- Ballistic shield now deploys and undeploys in 0.5 seconds
- Ballistic shield deployment can be interrupted if moved and the action can no longer be spammed
- Increased audio range on some effects so nearby mobs can hear them too, and removed some unnecessary text messages

TM instead of yolo-merging this or suffer from my balancing choices.
## Why It's Good For The Game

Mech shield no longer directly scales off the power cell. Also streamlined code.

## Testing

Mostly just made sure it could operate, that the shield battery drain works, and it blocks damage. A C20 SMG will break the shield in like 3/4 or 5/6 of the magazine at point blank.

## Changelog
:cl:
balance: Mech shield generator stops scaling off of the power cell, has an internal power bank
refactor: Mech shield code and damage absorption
/:cl:


code/modules/mechs/equipment/combat.dm
```diff
@@ -840,29 +845,34 @@
 		mech.vis_contents.Remove(visual_bluff)
 	QDEL_NULL(visual_bluff)
 
+/obj/item/mech_equipment/shield_generator/installed(mob/living/exosuit/_owner, hardpoint)
+	. = ..()
+	START_PROCESSING(SSobj, src)
```
SirRichardFrancis, line 850: I hate this.

MLGTASTICa: Could be better handled by adding a equipment flag for something processing on the mech/Life() instead. (After powered checks have been done)

TheShown911 (approved):
Code looks good.

PR #8471, opened 2024-04-08
Restrict combat mech users from entering with any suits/armors

XElectricX:
## About The Pull Request

What it says on the tin. Anything in the suit slot will prevent the player from entering a mech. This only affects mechs with combat and reinforced chassis.

## Why It's Good For The Game

Mech pilots should not be able to get away with wearing armor inside a bigger suit of armor.

## Testing

It just works.

## Changelog
:cl:
balance: Cannot board mechs with combat or reinforced chassis if wearing armor.
/:cl:


XElectricX:
> > It just works.
> 
> Please provide a proper list and explanation of tests you have conducted.

Quite literally just entering a mech to see if the restriction works. And it does.

> What armors you generally can and cannot wear? LIGHT_SLOWDOWN restricts mech users to only basic vests, which are really, really bad.

You just answered your question, homie.

assortedbeads:
This pr allows something like the NT void (a high tech combat void) to be used in a mech while a hazard suit with 7, 7, 10 is denied because it has higher slowdown.

Maybe have a second look at individual slow down values because they can vary a lot.

assortedbeads:
> This pr allows something like the NT void (a high tech combat void) to be used in a mech while a hazard suit with 7, 7, 10 is denied because it has higher slowdown.
> 
> Maybe have a second look at individual slow down values because they can vary a lot.

Going further on this, rather than using slowdown maybe you should look as how much armor the voidsuit has and use that as a cut off.

TheLeopold:
Go further, restrict mech pilots from wearing anything in suit slot.
That seems more logical
And I'm dead serious

XElectricX:
> > This pr allows something like the NT void (a high tech combat void) to be used in a mech while a hazard suit with 7, 7, 10 is denied because it has higher slowdown.
> > Maybe have a second look at individual slow down values because they can vary a lot.
> 
> Going further on this, rather than using slowdown maybe you should look as how much armor the voidsuit has and use that as a cut off.

I have looked through the files. It was why I did it with LIGHT to begin with, because even that category has some really extreme armors. But that's a balance issue with the _armors_; not about to touch every single one of them as that's out of scope and I've rarely experienced PvP here, so I'm not qualified.

I was using armor values at first but those are also all over the place. Slowdown seemed to be the best way to generalize.

> Go further, restrict mech pilots from wearing anything in suit slot.
That seems more logical
And I'm dead serious

This was what I was going to do but Wasteland told me the paramedics needed their suits so it'd screw them over. 🤷 

Although since it only affects combat mechs now, might be viable to do.

Humonitarian:
> > It just works.
> 
> Please provide a proper list and explanation of tests you have conducted. What armors you generally can and cannot wear? LIGHT_SLOWDOWN restricts mech users to only basic vests, which are really, really bad.

That's a good thing.

By forcing mech users to use light armors, we prevent mechs from functioning as a second health bar.

Using slowdown however can be considered questionable, it will prevent the use of most if not all space-worthy suits from being worn, and will allow the crusader armor to be worn as well.

code/modules/mechs/mech_interaction.dm
```diff
@@ -201,6 +201,11 @@
 		if(LAZYLEN(pilots) >= LAZYLEN(body.pilot_positions))
 			to_chat(user, SPAN_WARNING("\The [src] is occupied to capacity."))
 			return FALSE
+	if(ishuman(user) && body?.armor_restrictions)	//wear_suit only exists on humans; only bother with checking if the chassis forbids it
+		var/mob/living/carbon/human/enterer = user
+		if(enterer.wear_suit && enterer.wear_suit.slowdown >= MEDIUM_SLOWDOWN)	//If the user is wearing armor, check if it's too bulky
```
Humonitarian, line 206: Using slowdown will permit some of the best pieces of armor - such as crusader armor - to be worn, meaning it will simply encourage mech users to powergame harder and find these meta armors.

An alternative would be the use of a variable defined on suit level.

Humonitarian (changes requested):

code/modules/mechs/mech_interaction.dm
XElectricX: It might, but I would not be a good judge of saying what armor can or cannot be used. Better to just disallow all of them.

TheLeopold:
> This was what I was going to do but Wasteland told me the paramedics needed their suits so it'd screw them over.

If paramedics are the only problem with this, then maybe snowflaking it is the best solution?
As is "you can't wear armor except paramedic suit". Done.

assortedbeads:
> > This was what I was going to do but Wasteland told me the paramedics needed their suits so it'd screw them over.
> 
> If paramedics are the only problem with this, then maybe snowflaking it is the best solution?
> As in "you can't wear armor except paramedic suit". Done.

Paramedics are not the only issue

Humonitarian:
PR will be approved once the following test cases are checked and results provided:

You can use a mining mech without suffocating/dying to pressure
RIG cannot be deployed while in a mech

assortedbeads:
> PR will be approved once the following test cases are checked and results provided:
> 
> You can use a mining mech without suffocating/dying to pressure
> RIG cannot be deployed while in a mech

Heat can also fuck you in mechs last I checked even when closed and atmos on

I still think weak voidsuits should be fine.

githubuser4141:
> PR will be approved once the following test cases are checked and results provided:
> 
> You can use a mining mech without suffocating/dying to pressure RIG cannot be deployed while in a mech

This isn't necessary because only combat/heavy mechs have this restriction, which do in fact work in nil pressure, and you should be able to put a suit on in light, powerlifter, walker, pod chests.

XElectricX:
> > PR will be approved once the following test cases are checked and results provided:
> > You can use a mining mech without suffocating/dying to pressure RIG cannot be deployed while in a mech
> 
> First part isn't necessary because only combat/heavy mechs have this restriction, which do in fact work in nil pressure, and you should be able to put a suit on in light, powerlifter, walker, pod chests.

What this fella said. I've added a proc to the deployment and toggling process in RIGs to check if the wearer is inside a mech that restricts armors. Works without issue as I tested on a powerloader, walker, and a combat mech.

PR #8472, opened 2024-04-08
hcases QoL and bugfix

Hopekz:
## About The Pull Request
QoL: When you open the container by alt clicking, you also look inside of the crate (this is for people that are used to alt clicking containers but also good QoL for anyone in general).

Bugfix: You have to be right next to the crates to open them vs being able to do it from wherever.
You also can't be incapacitated.

## Why It's Good For The Game
Uhm, bugs bad?

## Testing
"Yes"

## Changelog
:cl: Hopek
add: You now look inside hcases when you open them.
fix: You can't open hcases from an infinite range anymore.
fix: You can no longer open hcases when incapacitated.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/storage/hardcases.dm
+++ b/code/game/objects/items/weapons/storage/hardcases.dm
@@ -97,7 +97,7 @@
 		to_chat(user, SPAN_NOTICE("You open the lid of the [src]."))
 		w_class = ITEM_SIZE_BULKY
 		closed = FALSE
-		src.open(user)
+		open(user)
 	else
 		to_chat(user, SPAN_NOTICE("You close the lid of the [src]."))
 		w_class = ITEM_SIZE_NORMAL
--- a/code/game/objects/items/weapons/storage/hardcases.dm
+++ b/code/game/objects/items/weapons/storage/hardcases.dm
@@ -97,7 +97,7 @@
 		to_chat(user, SPAN_NOTICE("You open the lid of the [src]."))
 		w_class = ITEM_SIZE_BULKY
 		closed = FALSE
-		src.open(user)		
+		src.open(user)
 	else
 		to_chat(user, SPAN_NOTICE("You close the lid of the [src]."))
 		w_class = ITEM_SIZE_NORMAL
--- a/code/game/objects/items/weapons/storage/hardcases.dm
+++ b/code/game/objects/items/weapons/storage/hardcases.dm
@@ -74,6 +74,11 @@
 	open_close(user)
 
 /obj/item/storage/hcases/AltClick(mob/user)
+	if(user.incapacitated())
+		to_chat(user, SPAN_WARNING("You can't do that right now!"))
+		return
+	if(!in_range(src, user))
+		return
 
 	var/able = can_interact(user)
 
@@ -92,6 +97,7 @@
 		to_chat(user, SPAN_NOTICE("You open the lid of the [src]."))
 		w_class = ITEM_SIZE_BULKY
 		closed = FALSE
+		src.open(user)		
 	else
 		to_chat(user, SPAN_NOTICE("You close the lid of the [src]."))
 		w_class = ITEM_SIZE_NORMAL


SirRichardFrancis:
Here, you forgot:

```
<!-- Write **BELOW** The Headers and **ABOVE** The comments else it may not be viewable. -->
<!-- You can view Contributing.MD for a detailed description of the pull request process. -->

## About The Pull Request

<!-- Describe The Pull Request. Please be sure every change is documented or this can delay review and even discourage maintainers from merging your PR! -->

## Why It's Good For The Game

<!-- Please add a short description of why you think these changes would benefit the game. If you can't justify it in words, it might not be worth adding. -->

## Testing

<!-- Describe the tests you ran with your addition. It is recommended to add images, videos and step-by-step explanations of conducted testing. -->

## Changelog
:cl:
/:cl:

<!-- Both :cl:'s are required for the changelog to work! You can put your name to the right of the first :cl: if you want to overwrite your GitHub username as author ingame. -->
<!-- You can use multiple of the same prefix (they're only used for the icon ingame) and delete the unneeded ones. Despite some of the tags, changelogs should generally represent how a player might be affected by the changes rather than a summary of the PR's contents. -->
```

Hopekz:
> Here, you forgot:

I'm not doing that

Firefox13:
> > Here, you forgot:
> 
> I'm not doing that

That's the pull request description template we use, please edit your description to match with it.

SirRichardFrancis:
> I'm not doing that

And why is that the case, pray tell? Ego too massive?

SirRichardFrancis:
Lol.

Firefox13:
> if you don't want the fix just close the PR I don't want to play these weird politic games.
> I just code things on my free time I'm not going out of my way to write up paragraphs where I don't need to

It's not a weird political game, it'ss a simple requirement (that was set in stone years ago) to make reviewing and figuring out PRs easier for maintainers and everyone else.

Everyone does this on their free time.

Did you test your code?

Hopekz:
> Did you test your code?

Yes

code/game/objects/items/weapons/storage/hardcases.dm
```diff
@@ -92,6 +97,7 @@
 		to_chat(user, SPAN_NOTICE("You open the lid of the [src]."))
 		w_class = ITEM_SIZE_BULKY
 		closed = FALSE
+		src.open(user)
```
SirRichardFrancis, line 100:
```suggestion
		open(user)
```

code/game/objects/items/weapons/storage/hardcases.dm
```diff
@@ -92,6 +97,7 @@
...
```
TheShown911, line 100:
```suggestion
		open(user)
```

TheShown911:
I can understand that argument and possibly even support it, but 90% of the codebase doesn't follow using src like that. Most experienced coders should be able to pick up on byond proc scope rather fast, while having src.proc there only 10% of the time will be confusing for new coders, which most of our coders are.

Byond isn't like JS where `this` is really important for scope management, so I would like it to for src to stay simple and mostly used when coders want to pass self ref to a proc.

TheShown911 (approved):

Hopekz:
> I can understand that argument and possibly even support it, but 90% of the codebase doesn't follow using src like that. Most experienced coders should be able to pick up on byond proc scope rather fast, while having src.proc there only 10% of the time will be confusing for new coders, which most of our coders are.
> 
> Byond isn't like JS where `this` is really important for scope management, so I would like it to for src to stay simple and mostly used when coders want to pass self ref to a proc.

No worries just wanted to drop my reasoning to show it was intentional.
I appreciate you (:

SirRichardFrancis:
> src was placed intentionally for readability because open only exists in hcases vs being global

Shouldn't project code standards/guidelines take precedence over coder's personal preference?
Like "I always piss on PR templates" and whatnot.

On Eris `src` tends to be used only when it is absolutely necessary, such as when procs with the exact same name exist in both global and local scopes for some reason. Other DM codebases also use `src` very sparingly, as far as I've seen.

Hopekz:
> Shouldn't project code standards/guidelines take precedence over coder's personal preference? Like "I always piss on PR templates" and whatnot.


Nah if I can get what I believe to be best practice merged without someone changing it that's what I'm going to try to do lmao.

Basically code how you want the codebase to look not the way it is.
If I wanted to leave it in the state it was before I touch it I just wouldn't PR to improve it in the first place.

I'm still going to do src on any future PR's I make here like all my other PR's in the past for procs that are in the object and 50% of the time maybe someone won't care enough to change it. (I think I have around 15 here merged that way)

I don't care about what is currently in place I only care about making this place better and the way I do my part is subjective to my own interpretation of what I think is better. This is why we have good maintainers to wrangle coders back in their place if we do something that is not acceptable.

PR #8481, opened 2024-04-30
Mech drill rework

XElectricX:
## About The Pull Request

Summary of changes:

- Time to drill scales with hardness of drill head material: steel 1.7, plasteel 1.1, diamond 0.5 seconds
- Added mining by bumping into rocks (or fixed, but I suspect it never worked)
- If you drill an object or a wall, the drill automatically continues drilling until it is destroyed (super funny, try it out)
- General fixes and improvements to this ancient code

## Why It's Good For The Game

Funny drill go brrr and better code.

## Testing

Drill works as normal. Can mine, dig up sand, drill my through the entire ship, and the timers work accordingly.

## Changelog
:cl:
balance: Mech drill speed scales off material hardness; minor speed adjustments.
fix: Mech drills were not able to mine by bumping into walls.
code: Mech drill code is more organized.
refactor: General mech drill improvements.
/:cl:

### DIFF
--- a/code/modules/mechs/equipment/utility.dm
+++ b/code/modules/mechs/equipment/utility.dm
@@ -434,6 +434,7 @@
 		return
 
 	target.GetDrilled()
+	drill_head.durability -= 1
 	gather_ores(target, user)
 
 /obj/item/mech_equipment/drill/proc/use_drill(atom/target, user, adjacent, show_message)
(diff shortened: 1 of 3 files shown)


Holvr0:
I'm naturally in favour of - as you put it - better code, but why the change from 3 tiles to 1? The whole point of mech drill was that it can grind down whole swaths of terrain at once, as opposed to just 1 tile at a time like a handheld drill. Even if 1 is being dug out faster than 3, it's still not a good change.

At most, it could be a toggleable drilling mode - drill really fast and narrow, or slower and broad - but not a permanent removal of the latter.

TheShown911 (approved):

PR #8482, opened 2024-05-03
Walls 2: Electric Boogaloo

SirRichardFrancis:
## About The Pull Request

This is PR that initially came to be as an alternative way of drawing 3/4 wall sprites,
Intended to make a fancy wall style that one of our downstreams came up with actually functional,
and spriter experience with making new walls better.
For one, time of making a wall style for cyberspace was not fun, to put it mildly, but changes in this PR should make such process a whole lot easier in the future.

This is how downstream church walls look on Eris:
image
(screenshot is somewhat outdated, window corner looks better now)
We won't be having these walls added anywhere with this PR, it just makes them possible, not replaces any textures.

Sprite sheets looks like this now:
image

image

image

image

image

Visualisation of internal logic:
image

While implementing and testing all that, however, I've been constantly reminded by runtime errors that we have infinite loops and some messed up recursion in turf code (unrelated to sprites) and it annoyed me one time too many,

So I've changed a bit of internal logic there, eliminating runtimes, unused variables, and marginally improving performance of the whole thing.
After discussing this with Handyman, I've also went ahead and ~out of spite~ removed material walls, made low walls a subtype of regular walls with full-tile windows baked in (code-wise), and (de)construction process simplified.

~This mostly works, but some parts of wall/turf code are yet to be rewritten (thermite melting and whatnot), new low walls~ ~need to work with atmos better and with 100% reliability, and there are some more optimisations to be made.~

~And as such, this PR is here as a draft, in case if someone wants to check on \~50k diffs here or offer an opinion.~

UPD:

image
This is only one girder now, is worth 5 steel and will create a regular wall with 5 more steel.
Or a low wall, if it's pried first.
Alternatively it could be reinforced with 2 steel rods and turned into a reinforced wall with 5 plasteel.

image
Wrenching the girder will (un)anchor it, but reinforced girder can't be moved.

Disassembling the reinforced wall now requires fewer steps:
Bolt turning, prying, wire cutting, hammering and welding.

image
Hammering tool must be of quality 15 at least.

image
Wire cutting step depends on tool quality much more.

## Why It's Good For The Game

Not having a proc overwrite spaghetti with infinite loops in the code should be pretty good.
Slight performance improvement.
Merging low walls with floor and windows should cull 5000+ objects from the map. Not that big of a deal, but still nice.
Making new walls is much more straightforward process. Not happens often, but still.

Moving on from bay12 procedurally colored material walls might be a tad controversial, even if this feature was used by literally no one, but ultimately for the best too.

## Testing

Lots.

## Changelog
:cl:
tweak: Changed reinforced wall deconstruction steps.
refactor: Refactored part of turf code, including wall.
code: Changed how wall sprites are generated.
/:cl:


SirRichardFrancis:
There are some unrelated fixes in snacks.dm, which should be in a separate PR (and they will be), I just needed this crap to stop runtiming every time I launch a local server.

MLGTASTICa:
this should be a massive increase in explosion performance if it fixes the underlying shitcode of turfs and walls that cause massive tick usage.

code/game/turfs/simulated/wall_types.dm
```diff
@@ -98,63 +83,18 @@
...
-
-/turf/simulated/wall/untinted/onestar_reinforced/New(var/newloc)
-	..(newloc, MATERIAL_STEEL,MATERIAL_STEEL)
+/turf/wall/untinted/onestar
+	icon_state = "onestar_wall"
+	wall_style = "minimalistic"
+	wall_type = "onestar_wall"
+
+/turf/wall/untinted/onestar_reinforced
+	name = "reinforced wall"
```
vode-code, line 97: "reinforced"... but has no more health than standard.

TheShown911:
Test merging this.

PR #8456, opened 2024-03-10
New mech weapons: club, bow & BFG

githubuser4141:
## About The Pull Request

Adds 3 new mech weapons, the club, a bow & arrow, and a BFG.
## Why It's Good For The Game

More variety, more fun.
## Testing
(Not tested yet)
Spawned a mech, gave it the weapons, shot mobs with them. Built a mech via R&D, printed the modules, attached them and shot mobs with it.

## Changelog
:cl: by Doster-D
add: Adds a mech club, mech bow & arrow and a mech BFG (plasma weapon) to the game.
/:cl:

PR #8486, opened 2024-05-12
[COMMISSION] New mech weapons: club, bow & BFG

cheesePizza2:
## About The Pull Request

Remake of #8456, with permission from the author.

Adds 3 new mech weapons:
- MACE: has two modes: the mace does loads of penetrating damage and can quickly break structures, whereas the flail is somewhat weaker but has a longer range and applies heavy knockback and weaken.
- BFG: Heavy, slowly-charging energy gun that shoots heavy blasts of AoE plasma.
- CROSSBOW: Accepts materials and shoots them as crossbow bolts. Bolt stats depend on the material - damage is based on weight whereas armor pen is based on hardness.

## Why It's Good For The Game

More weapon variety, as well as some integration with the materials system.

## Testing

Mace can be used to attack others. Mace can be swapped to and from flail mode.
BFG charging is properly handled.
Crossbow charging is properly handled. Crossbows can be refilled and used properly. Filling a half-filled crossbow takes fewer sheets (as intended) and using fewer sheets than possible grants fewer shots (as intended).

POSSIBLE IMPROVEMENTS:
Mace flail sprite is a duplicate of the mace sprite.
BFG and crossbow charging can sometimes trigger a drag-click.
BFG and crossbow charging don't play a sound or make a light.

## Changelog
:cl: Cheesepizza2, Doster-D
add: Added a new mech weapon: the NT \"Warhead\" mace! In its mace mode, it will absolutely obliterate structures and people. Using it in hand will convert it to a flail that does less damage but has a longer range and weakens + knocks back targets.
add: Added a new mech weapon: the BFG! After a lengthy charging period, this massive cannon shoots a massive bolt of plasma energy that does loads of damage in a wide radius.
add: Added a new mech weapon: the mounted crossbow! Insert in materials to load as crossbow bolts - the harder and heavier the material, the more armor penetration and damage it has respectively.
/:cl:


SirRichardFrancis:
Screenshots?

code/modules/mechs/equipment/combat.dm
```diff
@@ -802,6 +805,221 @@
...
+	if(ishuman(target))
+		var/mob/living/carbon/human/targ = target
+		if(targ.stats.getStat(STAT_VIG) > STAT_LEVEL_EXPERT)
+			targ.visible_message(SPAN_DANGER("[targ] dodges the [holding] swing!"), "You dodge [loc]'s [holding] swing!", "You hear a woosh.", 6)
+			return
+	. = ..()
+	if(. && ismech(loc) && istype(target) && target != owner)
+		if(flail_mode)
+			target.visible_message(SPAN_DANGER("[target] gets slammed by [src]'s [holding]!"), SPAN_NOTICE("You get slammed by [src]'s [holding]!"), "You hear something soft hit a metal plate!", 6)
+			target.Weaken(5)
```
Humonitarian, line 842: Weaken is a stun that slows movement to a crawl and prevents fighting back, which we might have problems with if it can be applied repeatedly. A different system like recoil can apply slowdown, while also having a capacity to disarm someone. It might be preferable here.

code/modules/mechs/equipment/combat.dm
```diff
@@ -802,6 +805,221 @@
...
+	if(.)
+		var/obj/item/tool/hammer/mace/mech/holdin = holding
+		holdin.update_flail_mode(flail_mode)
+		flail_mode = !flail_mode
+		owner.update_icon()
+
+/obj/item/mech_equipment/mounted_system/mace/resolve_attackby(mob/living/target, mob/user, params)
+	if(ishuman(target))
+		var/mob/living/carbon/human/targ = target
+		if(targ.stats.getStat(STAT_VIG) > STAT_LEVEL_EXPERT)
```
Humonitarian, line 835: Flat stat check? Shouldn't there be a Prob()?

SirRichardFrancis: Probn't

Firefox13: I would *usually* agree with it, but we should avoid any kind of RNG in PvP (it is PvP). This is fine currently, or the damage can just start to decrease after certain VIG (so if godlike is 80 (iirc?), as example at 40 VIG it starts to decrease, at 60 VIG damage is halved, at 79 it, like 2 brute; *The mace barely grazes you!*)

Humonitarian: Making the weapon have no effect though is definitely not fine.

cheesePizza2: Rewrote this proc to use an intensity variable instead of a flat skip. Current formula is `STAT_LEVEL_ADEPT / targ.stats.getStat(STAT_VIG)`, meaning expert gives you 0.625x damage,  professional gives you 0.416x damage, and godlike gives you 0.3125x damage. If that feels too light at the higher-tier levels, I could mess with the formula a bit.

code/modules/mechs/equipment/combat.dm
cheesePizza2: Couldn't find any recoil system. However, I reduced this value from 5 to 3, and the weaken with the mace mode is gone entirely.

cheesePizza2:
Alright, all three modules are functional, although in need of polishing (and a missing sprite).

cheesePizza2:
Tested, everything is in order except:

- Mace is currently missing a flail mode sprite in-hand.
- BFG and crossbow charging can sometimes trigger a drag-click.
- BFG and crossbow charging don't play a sound or make a light.

Working with my commissioner to correct the first. Latter two aren't breaking issues and would take some further code fuckery to fix.

cheesePizza2:
I don't think he has a sprite for flail in-hand so I'm just gonna duplicate the regular one. This is good enough for review.

P.S. I kinda want this in sooner rather than later, so uh, maintainers have full permission to make changes without asking if it'll get this through faster.

TheShown911 (approved):
Code doesn't look bad.

Humonitarian:
If there is any balance issues, we can patch it once we find it.

PR #8488, opened 2024-05-15
chaos level now automatically increases hourly

vode-code:
## About The Pull Request
This PR increases the Chaos Level once per hour.

## Why It's Good For The Game
Destination? Death.

## Testing
tested the chaos level increase timing- it does not increase the level roundstart, but does increase it an appropriate cycle.

## Changelog
:cl:
balance: Bluespace Chaos now infuses the ship over time.
/:cl:

PR #8489, opened 2024-05-15
Revert "chaos level now automatically increases hourly"

MLGTASTICa:
Reverts discordia-space/CEV-Eris#8488

This PR won't fix the issue of greenshifts on Eris and it will help guarantee a boring TDM-fest later on in the round. (aka 4 people roll traitor and decide to buy all the default gear they can through the uplink instead of actually prepping)
Greenshifts are a player issue created by lack of gimmicks , some gameplay aspects contribute to this , most notably the lack of content relating to non-combat , research and exploration.

### DIFF
--- a/code/game/gamemodes/storyteller.dm
+++ b/code/game/gamemodes/storyteller.dm
@@ -67,7 +67,7 @@ GLOBAL_VAR_INIT(chaos_level, 1) //Works as global multiplier for all storyteller
 
 	var/votable = TRUE
 	//whether or not the players can vote for it. If this is set to false, it can only be activated by being forced by admins.
-	var/last_chaos_hour = 0
+
 
 
 /********************************
@@ -111,7 +111,6 @@ GLOBAL_VAR_INIT(chaos_level, 1) //Works as global multiplier for all storyteller
 
 /datum/storyteller/proc/set_up()
 	build_event_pools()
-	last_chaos_hour = REALTIMEOFDAY
 	set_timer()
 	set_up_events()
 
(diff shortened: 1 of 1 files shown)


Humonitarian (changes requested):
If you provide an alternative for exponential point ramp-up that doesn't increase antag role points, I'll approve it. More destructive events and the likes as the round goes on would be good.

MLGTASTICa:
> If you provide an alternative for exponential point ramp-up that doesn't increase antag role points, I'll approve it. More destructive events and the likes as the round goes on would be good.

if you're not gonna merge the revert then close it.

TheShown911:
I actually agree with the revert, chaos vote was a simple bandaid fix, as storyteller was going to be fixed "in a week". That was 2 years ago and I don't think chaos vote does much of anything except break the storyteller. It multiplies storyteller points generated but when the events that cost those points are called they mostly get refunded because a) Antags can't be generated anymore as there is no preference for them or b) no events can be called because they were called recently.
This will most likely just break the storyteller even more if the chaos level gets too high. The storyteller is fundamentally broken and I don't think any amount of bandaids can fix it.
Anyways humon is a gangster communist computer god for approving and merging the PR (who even does that) and thanks for coming to my Ted talk.

Humonitarian:
> I actually agree with the revert, chaos vote was a simple bandaid fix, as storyteller was going to be fixed "in a week". That was 2 years ago and I don't think chaos vote does much of anything except break the storyteller. It multiplies storyteller points generated but when the events that cost those points are called they mostly get refunded because a) Antags can't be generated anymore as there is no preference for them or b) no events can be called because they were called recently. This will most likely just break the storyteller even more if the chaos level gets too high. The storyteller is fundamentally broken and I don't think any amount of bandaids can fix it. Anyways humon is a gangster communist computer god for approving and merging the PR (who even does that) and thanks for coming to my Ted talk.

I am the villain of game design.

Humonitarian (approved):
We are working on a better fix to storyteller.

PR #8492, opened 2024-05-24
Disable client auth checks

SirRichardFrancis:
## About The Pull Request

Basically a mirror of  https://github.com/tgstation/tgstation/pull/83435

## Why It's Good For The Game

BYOND hub experienced technical issues recently, and players were unable to join due to that.
This change should prevent hub issues from affecting Eris.

## Testing

Not possible to test currently (the hub is back online), but this compiles and runs fine.

## Changelog
:cl:
server: Disabled BYOND hub-based authentication.
/:cl:

### DIFF
--- a/code/modules/client/client_defines.dm
+++ b/code/modules/client/client_defines.dm
@@ -68,6 +68,14 @@
 	/// Disables default BYOND verb panel in favor of Stat Panel™
 	show_verb_panel = FALSE
 
+	// BYOND documentation has this to say about "authenticate" variable:
+	// This value may be set to 0 at compile-time to disable BYOND hub-based authentication of users. The default value is 1, which enables authentication. Hub authentication provides an additional level of assurance that the user is really the owner of the BYOND key in question.
+	// When a world requests certification, Dream Seeker generates a random password and passes it through the hub (for certification) to the world. The certificate is saved for faster access in the future and for protection against possible hub outages.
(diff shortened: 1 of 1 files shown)


MLGTASTICa:
i feel like this is a double-edged knife scenario , yes this will allow a bunch of people to connect when byond is having issues , but this also permits the people collecting hardware ID's and IP's (using fake servers) to join under the ACTUAL name of someone who joined any of their servers. (So if a admin joins a fake Eris and has his Hardware ID , IP and Certificate duplicated , someone could fake his ckey without authorization from byond hub with just his certificate). I suggest making this a setting that is on by default and that can be configured to be off perhaps.

SirRichardFrancis:
While there might be some security implications, I wouldn't go as far as to state theoretically possible issues as facts.

For you see, when spoofing happened in SS13 in the past, it was due to hub's own vulnerabilities, and no amount of mandatory hub verification helped.

That said, I'm not a cybersecurity guy, and this is not something I personally test. Then again, neither do you.
I'd say we do what TG does.
If they end up accepting the change, then it's probably kosher ¯\\\_(ツ)_/¯

MLGTASTICa:
> While there might be some security implications, I wouldn't go as far as to state theoretically possible issues as facts.
> 
> For you see, when spoofing happened in SS13 in the past, it was due to hub's own vulnerabilities, and no amount of mandatory hub verification helped.
> 
> That said, I'm not a cybersecurity guy, and this is not something I personally test. Then again, neither do you. I'd say we do what TG does. If they end up accepting the change, then it's probably kosher ¯\_(ツ)_/¯

Even TG mentions they don't know how secure their authentification checks are and the reason that they are doing it is because they force admins to log in through a website ontop to get acces to admin powers.

PR #8494, opened 2024-06-04
Fixes missing insight when fulfilling a goal while already inspired

the-soy-coder:
## About The Pull Request
You lose ALL insight when you finish focusing, no matter if it goes over 100. This fixes the edge cases when you fulfill a goal right before finishing your current insight, and thus lose an additional 100 insight.

## Changelog
:cl: Buttcrack Berry Picker
fix: fixed losing all insight when you focus, instead of just your max cap
/:cl:

### DIFF
--- a/code/game/machinery/autolathe/artist_bench.dm
+++ b/code/game/machinery/autolathe/artist_bench.dm
@@ -62,7 +62,7 @@
 			if(H.stats.getPerk(PERK_ARTIST) && H.sanity.insight > 40)
 				ins_used = input("How much of your insight will you dedicate to this work? 40-[H.sanity.insight > 100 ? 100 : H.sanity.insight]","Insight Used") as null|num
 			else
-				ins_used = H.sanity.insight
+				ins_used = H.sanity.insight > 100 ? 100 : H.sanity.insight
 			create_art(ins_used, H)
 			return TRUE
 		return FALSE
--- a/code/modules/sanity/sanity_mob.dm
+++ b/code/modules/sanity/sanity_mob.dm
@@ -352,7 +352,7 @@ GLOBAL_VAR_INIT(GLOBAL_INSIGHT_MOD, 1)
 
 	owner.pick_individual_objective()
 	resting = 0
-	insight = 0
+	insight -= 100
 
 /datum/sanity/proc/onDamage(amount)
 	changeLevel(-SANITY_DAMAGE_HURT(amount, owner.stats.getStat(STAT_VIG)))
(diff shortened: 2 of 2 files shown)


dingleberrypicker:
@TheShown911 @ACCount12 Hi~

Humonitarian:
There was a bug that allowed insight to rise over 100 on artist, might want to check if its still present, as that would allow some potential exploits.

dingleberrypicker:
artist_bench now capped at 100 insight maximum investment even if you evade the prompt, should be fine

code/game/machinery/autolathe/artist_bench.dm
```diff
@@ -62,7 +62,7 @@
 			if(H.stats.getPerk(PERK_ARTIST) && H.sanity.insight > 40)
 				ins_used = input("How much of your insight will you dedicate to this work? 40-[H.sanity.insight > 100 ? 100 : H.sanity.insight]","Insight Used") as null|num
 			else
-				ins_used = H.sanity.insight
+				ins_used = H.sanity.insight > 100 ? 100 : H.sanity.insight
```
intercepti0n, line 65: Maybe clamp function would fit better here? Ternary operator looks kinda messy.

SirRichardFrancis: Factually incorrect, ternary operator is the best operator, and hardly worse than a function call.

dingleberrypicker:
why dis shit failing doe?

SirRichardFrancis:
> why dis shit failing doe?

`[19:18:37] Runtime in utility.dm,914: Cannot execute null.forceMove().`

Proc `ejectLifting()` got called without a prerequisite check (which should probably be a part of the proc itself, but that's another story).
Known bug that is thought to be fixed on up-to-date Master branch.
Either it wasn't actually fixed, or you've used slightly outdated branch for this PR.

MLGTASTICa:
> > why dis shit failing doe?
> 
> `[19:18:37] Runtime in utility.dm,914: Cannot execute null.forceMove().`
> 
> Proc `ejectLifting()` got called without a prerequisite check (which should probably be a part of the proc itself, but that's another story). Known bug that is thought to be fixed on up-to-date Master branch. Either it wasn't actually fixed, or you've used slightly outdated branch for this PR.

its actually a bit more unusual than that.. apparently it tries to out-lift a turf (which can't move), so its either null because 1)the turf is marked for deletion and so are all its contents , or 2)theres a massive fuck up somewhere thats allowing turfs to be picked up.

PR #8504, opened 2024-08-04
Increases oxygen damage from blood-loss

sgtmind:
## About The Pull Request

If a human drops below 40% blood, they take oxygen damage over time at a rate high enough to outpace natural recovery.
This leads to unconsciousness followed by death. 

## Why It's Good For The Game

Increases lethality of bleeding wounds. This allows bullets to kill, while still allowing medical attention to delay the target's demise.

## Testing
Increased damage due to bloodloss was tested and timed on multiple dummies. A set amount of blood was drained, with me continuously looking at the amount of oxygen damage through VV, and time to die compared to without the increased damage measured using a stopwatch.  
Blood loss wounds was tested by draining 420 units of blood from a human alongside giving the human a dose of dexalin. The dexalin would prevent death from suffocation, allowing enough time for the wounds to appear on a med bodyscanner.

## Changelog
:cl:
balance: Humans will now receive increased oxygen damage if one loses a large portion of their blood
add: Unconsciousness due to oxygen damage will now play a unique message
fix: Blood loss wounds have been fixed to appear. If a human has lost a large amount of blood and is on dexalin, they can receive wounds on their organs, eventually causing the organ to die
/:cl:

### DIFF
--- a/code/modules/organs/internal/internal_organ_processes.dm
+++ b/code/modules/organs/internal/internal_organ_processes.dm
@@ -148,6 +148,7 @@
 
 	if(blood_volume < total_blood_req)
 		status_flags |= BLEEDOUT
+		adjustOxyLoss(ROUND_PROB(2.5))
 		if(prob(15))
 			to_chat(src, SPAN_WARNING("Your organs feel extremely heavy."))
 	else
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -813,6 +813,8 @@
 
 		//UNCONSCIOUS. NO-ONE IS HOME
 		if(getOxyLoss() > (species.total_health/2))
+			if(stat == CONSCIOUS)
+				to_chat(src, SPAN_DANGER("You feel your consciousness become increasingly distant from the physical world."))
 			Paralyse(3)
 
 		if(hallucination_power)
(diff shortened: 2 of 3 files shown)


githubuser4141:
So... how much is the lethal amount now? Does it scale with increasing amounts of bloss? Is it affected by organ efficiency/organ amount?

sgtmind:
> So... how much is the lethal amount now? Does it scale with increasing amounts of bloss? Is it affected by organ efficiency/organ amount?

Lethal threshold should be in the PR already, while the only efficiency that affects dying from oxy is the lungs. Dying from blood loss is at the end of the day dying from oxy damage.

TheShown911:
@Humonitarian this one is up to you mr headcoder

code/modules/organs/internal/internal_organ_processes.dm
```diff
@@ -148,6 +148,7 @@
 
 	if(blood_volume < total_blood_req)
```
Humonitarian, line 149: Does this not scale with how much oxygen organs use? So someone with modded organs would die from less bloodloss

Humonitarian: Okay, I codedived. So what this means is that if blood is below blood required + 20, you receive 2 oxyloss each tick. But if your blood is below required, you receive 4.5 oxyloss?

Is there math behind why this number is so specific? If there is, can you include it as a comment?

sgtmind:
I fucked something up in VSC, I am redoing this on a new attempt

PR #8516, opened 2024-08-15
Fixing runtimes, part 1

Ganza9991:
## About The Pull Request

Fixes some runtimes. From Celadon with love.
This pr was pulled from the fork of a fork of this repository, some changes may cause questions.

## Why It's Good For The Game

Less runtimes = less runtime logs = more sanity to coders.

## Testing

Recieved no runtimes after changes. In much hope that they wont even on server.

## Changelog
:cl:
Fixed die() from organs to handle efficiency, when there is no owner
Fixed runtime when reagent container was empty
Fixed runtime from breakdown attempt to a creature w/o sanity
Fixed random runtime when boltaction guns could not handle bullet queue
Fixed runtime from level_up() attempt to no mob w/o client
And other random runtimes
:cl:

### DIFF
--- a/code/modules/sanity/breakdown.dm
+++ b/code/modules/sanity/breakdown.dm
@@ -49,6 +49,11 @@
 	flick(icon_state, img)
 
 /datum/breakdown/proc/occur()
+	if(!holder)
+		return FALSE
+	if(!holder.owner)
+		return FALSE
+
 	occur_animation()
 	holder.owner.playsound_local(get_turf(holder.owner), breakdown_sound, 100)
 	if(holder.owner.head && istype(holder.owner.head, /obj/item/clothing/head/mindreader))
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -369,7 +369,7 @@ its easier to just keep the beam vertical.
 
 	if(reagents)
 		if(reagent_flags & TRANSPARENT)
-			if(LAZYLEN(reagents.reagent_list) > 1)
+			if(LAZYLEN(reagents.reagent_list))
 				if(user.can_see_reagents())
 					output += SPAN_NOTICE("\nIt contains:")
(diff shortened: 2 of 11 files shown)


code/modules/sanity/breakdown.dm
```diff
@@ -49,6 +49,11 @@
 	flick(icon_state, img)
 
 /datum/breakdown/proc/occur()
+	if(!holder)
+		return FALSE
+	if(!holder.owner)
+		return FALSE
+
```
AltHit, line 56: Again, we might never know why in the Hell breakdown is trying to occur where there is no sanity or where there's no mob who should hold this sanity datum. Just hides away the problem

AltHit (commented):
Whole PR contains fixes to a bunch of different problems yet in many cases those fixes only fix half of the problem or solve bugs rooting from a much deeper problem.
To conclude all that - yes, if it will be merged we will see less runtimes in round but it does not mean that there will be any less runtimes actually happening in code unseen and unfixed.

code/modules/sanity/breakdown.dm
Ganza9991: If a mob receives breakdown w/o sanity, it's probably an admincall, rather than actual bug. Since holder is sanity and owner is a mob. If sanity exists w/o mob, then... i dont really know why it exists then.

AltHit: I'm pretty sure there was no case of manual admincalling breakdowns of people in a long time already. Moreover why would admin call a breakdown on a deleted mob?
Runtimes should be used to trace the reason why this even happens and without those runtimes we might end up with improper deletion of mobs leaving sanity module behind and still processing or some other obscure bug.
Leaving this error unattended and silencing it with a check might just fuck things up even further!

PR #8526, opened 2024-09-30
Eris Spooktober update

AltHit:
## About The Pull Request

Joke PR that turns the whole crew to spooky skeletons, returns the funny bones (turns off small bone sprites) and also gives every crewman a rubber duck on the round start
(and yeah, also proper quacking ducks)

## Why It's Good For The Game

изображение_2024-09-30_222747343

It's spooky month already and Eris already had a fair share of funny halloween updates in the past so why shouldn't we add some more fun

## Testing

Playtested with two people (gunfights, arson, a bit of surgery) and it was terrible! But nothing gamebreaking or crushing server so it should be fun!

## Changelog
:cl:
add: Bonus ducks!
tweak: Now CEV Eris has a real skeleton crew
/:cl:


code/game/objects/structures/watercloset.dm
```diff
@@ -357,6 +357,22 @@
...
+		UPGRADE_ITEMFLAGPLUS = DUCKING
+	)
+	I.weapon_upgrades = list(
+		GUN_UPGRADE_DUCK = TRUE
+	)
+	I.gun_loc_tag = GUN_MECHANISM
+
+/obj/item/bikehorn/attack_self(mob/user)
+		playsound(src.loc, 'sound/items/duck.ogg', 50, 1)
+		src.add_fingerprint(user)
```
SirRichardFrancis, line 375:
```suggestion
/obj/item/bikehorn/rubberducky/New()
	..()
	var/datum/component/item_upgrade/I = GetComponent(/datum/component/item_upgrade)
	I.tool_upgrades[UPGRADE_ITEMFLAGPLUS] = DUCKING
	I.weapon_upgrades[GUN_UPGRADE_DUCK] = TRUE

/obj/item/bikehorn/rubberducky/attack_self(mob/user)
	playsound(loc, 'sound/items/duck.ogg', 50, 1)
	add_fingerprint(user)
```
Overriding the parent item procs makes no sense here, should use those of `/rubberduck`.
Don't create a new item_upgrade component, parent object already got one of these, just fetch and modify it.
Using `src.` should be avoided unless absolutely necessary (it never is).

SirRichardFrancis (changes requested):
Imo "skeletonization" is not something that should happen with literally everyone.

Having some players remain flesh and blood humans will lead to ~skeleton racism~ potentially more interesting gameplay, as opposed to "business as usual, but everyone looks like a skeleton".

PR #8539, opened 2024-12-21
Optimize action buttons

SirRichardFrancis:
## About The Pull Request

Makes sure action buttons are added and removed only when it is actually necessary, not every single `Life()` tick.

## Why It's Good For The Game

Better performance.

## Testing

Equipped, dropped, and activated items with action buttons (voidsuit and a flashlight) to see if they work correctly.
Hosted via Dream Daemon, reconnected after equipping voidsuit and a flashlight, observed action buttons being visible on login.

## Changelog
:cl:
code: Optimized action buttons
/:cl:

### DIFF
--- a/code/game/objects/items/_inventory.dm
+++ b/code/game/objects/items/_inventory.dm
@@ -53,6 +53,7 @@
 			equipped.forceMove(src)
 	return FALSE
 
+
 /obj/item/proc/equipped(mob/user, slot)
 	equip_slot = slot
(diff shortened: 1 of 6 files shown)


MLGTASTICa:
You need to also hook in to Mob/LogIn and Mob/LogOut else someone reconnects and is missing their action buttons.

SirRichardFrancis:
Touche.

PR #8543, opened 2025-01-19
Superglue is printable with an advanced tools disk + superglue buffs

VmpOS2NHSkhkejA9:
## About The Pull Request

Add superglue to 'Technomancers IJIRO-451 Advanced Tools' disk.

Change superglue's materials from 30 biomatter to 20 plastic, the same as fiber tape, making it printable by all fabricators.

Change superglue's adhesive stat from 40 to 50. This makes superglue contribute 3 mechanical stat instead of 2 to the techno-tribalism enforcer, allowing technomancers to create 15-mechanical oddities at roundstart.

I specifically chose superglue and not fiber tape because fiber tape's current stats would let technomancers create 30-mechanical oddities which I think is too drastic of a buff.

## Why It's Good For The Game

Mechanical is an important stat to technomancers, yet it's the only one they have no reliable access to. This PR would finally let them create oddities with mechanical stats without relying on finding some of the rarest junkpile items in the game.

## Testing

image

## Changelog
:cl:
balance: Superglue can now be printed with an advanced tool disk
balance: Superglue is now made of 20 plastic instead of 30 biomatter
balance: Superglue now has 50 adhesive stat instead of 40.
/:cl:

### DIFF
--- a/code/datums/autolathe/tools.dm
+++ b/code/datums/autolathe/tools.dm
@@ -159,3 +159,6 @@
 /datum/design/autolathe/tool/rpd
 	name = "rapid piping device"
 	build_path = /obj/item/rpd
+
+/datum/design/autolathe/tool/superglue
+	build_path = /obj/item/tool/tape_roll/glue
--- a/code/game/objects/items/weapons/design_disks/technomancers.dm
+++ b/code/game/objects/items/weapons/design_disks/technomancers.dm
@@ -50,6 +50,7 @@
 		/datum/design/autolathe/tool/combi_driver,
 		/datum/design/autolathe/tool/armature_cutter,
 		/datum/design/autolathe/tool/weldertool_adv,
+		/datum/design/autolathe/tool/superglue,
 		/datum/design/autolathe/part/diamondblade,
 		/datum/design/autolathe/tool/rpd,
 		/datum/design/autolathe/container/hcase_engi,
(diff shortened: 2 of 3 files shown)


SirRichardFrancis (approved):
Not sure why technomancers would want MEC stat oddities to begin with.
Between roundstart focus on MEC stat, MEC boosting perk, and superior tools they have access to, there is little use in getting even more MEC.
It's only practical to raise up to a point, after that there is no difference to speak of.

Then again, I don't see how this change is harmful either, so unless there is reasonable objection before Friday, I think it's good to go.

VmpOS2NHSkhkejA9:
> Not sure why technomancers would want MEC stat oddities to begin with. Between roundstart focus on MEC stat, MEC boosting perk, and superior tools they have access to, there is little use in getting even more MEC. It's only practical to raise up to a point, after that there is no difference to speak of.
> 
> Then again, I don't see how this change is harmful either, so unless there is reasonable objection before Friday, I think it's good to go.

Technomancers may not want to go all-in on roundstart mechanical, maybe they want other stats or to vary up with different perks (that might even be important to their character?). Maybe they'd prefer higher base stats to temporary buffs, etc. In all of those cases they've got very few options if they want to work for better mechanical stats.

SirRichardFrancis:
Kept you waiting, huh?

PR #8546, opened 2025-01-21
Remove nonfunctional, unused code for golem waves and burrows + misc fixes and adjustments

VmpOS2NHSkhkejA9:
## About The Pull Request

Remove golem waves, burrows and their associated code from the code. They are already impossible to trigger, so the only player-facing changes here should be the bugfixes, a couple changes to texts and diamond golems no longer being lobotomized.

## Why It's Good For The Game

- Make the deep drill and golems easier to work with and update codewise.
- Fixes are self explanatory.
- Diamond golems being nonhostile to miners meant they simply stood still while providing zero challenge. Arguably, diamond golems simultaneously are the most valuable and least dangerous golem. This is made worse by the fact coal golems become diamond golems when lit on fire (usually by a plasma golem) because it made the next weakest golem even weaker.

## Testing

- Generated a cave. Collapsed the cave by de-activating the drill. Found no issues.
- Shot the drill with a variety of weapons. Repaired it. Found no issues.
- Built walls adjacent to an active drill. Shut down correctly and collapsed the attached cave. 
- TODO: Make sure floors under an active drill are destroyed
- TODO: Make sure parts can still be swapped (even if they're nonfunctional)
- TODO: Make sure the drill can still be activated adjacent to space tiles
- TODO: Make sure the drill can be unanchored from space tiles and plating

## Changelog
:cl:
del: Golem waves and burrows
tweak: Welding the deep drill now heals it by 500 health, instead of jumping to the next 1/3rd of its health.
tweak: The deep drill no longer accepts power cells, as it doesn't require power anyway.
tweak: Structures and machinery no longer block the deep drill's operation; It can be activated with ore boxes adjacent to it.
balance: diamond golems are now hostile
fix: Deep drill can now be unanchored from undrillable tiles
fix: Deep drill properly shuts down when non-floor tiles are too close
/:cl:


VmpOS2NHSkhkejA9:
at last, it is complete (probably)

hyperioo:
I will look into it 🙏

code/modules/mining/drilling/deep_drill.dm
```diff
@@ -206,176 +126,79 @@
...
+	// Check if there are no walls around the drill
 	for(var/turf/F in block(locate(x - 1, y - 1, z), locate(x + 1, y + 1, z)))
-		if(F != loc)
-			if(F.density)
-				return TRUE
-			for(var/atom/A in F)
-				if(A.density && !(A.flags & ON_BORDER) && !ismob(A))
-					return TRUE
+		if(!istype(F,/turf/floor) && !istype(F,/turf/space)) //if it's not a floor and not space it's probably an obstacle.
+			return TRUE
```
hyperioo, line 201: The idea with the atom check was to also detect dense structures like girders, crates and so on so that the drills could not be surrounded by impassible stuff. But I guess it's alright to simplify it like that since there is no burrows and golem waves anymore.

hyperioo (approved):
Nice cleaning 👍

code/modules/mining/drilling/deep_drill.dm
VmpOS2NHSkhkejA9: The old checks would return true on ore boxes or crates, which is the main reason I changed this part.

PR #8548, opened 2025-01-30
Small erismed fix

vode-code:
## About The Pull Request
This PR fixes wound calculations and tumor child structure being out of date.

## Why It's Good For The Game
Runtimes bad, Bugs bad

## Testing
shot disciples with .20 Kadmin and found good results, with and without armor.
shot disciple with Miller and found good results without armor, and appropriately weak results with armor.
shot disciple with decloner until tumor successfully appeared

## Changelog
:cl: Chickenish
fix: increased danger of radiation and war
/:cl:

### DIFF
--- a/code/modules/organs/external/damage.dm
+++ b/code/modules/organs/external/damage.dm
@@ -28,9 +28,9 @@
 		var/transferred_damage_amount
 		switch(damage_type)
 			if(BRUTE)
-				transferred_damage_amount = amount - (max_damage - brute_dam) / armor_divisor / 2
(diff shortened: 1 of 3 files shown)


SirRichardFrancis (commented):
Does `name = "tumor"` have to repeat three times in derived types, instead of being on the parent type?

vode-code:
technically no, but it's really a placeholder.

vode-code:
I expect to change those names later, and them being subtypes is not optional.

SirRichardFrancis (approved):
Ah, so it's the case of "I'll change it later probably"? Understandable. Been there, done that.
Not like Erismed can get any worse at this point and fix is a fix...

PR #8558, opened 2025-03-06
Seuche Redux

vode-code:
## About The Pull Request
This PR improves Seuche roaches, Fuhrer roaches, Kaiser roaches, Burrows, and Superior Animals.
How Fuhrers are improved:
- Fuhrers now become the leader of Overminds, datums that transmit data between cooperative members.
- When a Fuhrer targets a mob, all idle attacker type roaches in the Roach Overmind it is a member of will also target it.
- When a Fuhrer summons reinforcements, said Reinforcements will join their Overmind as long as they are not a fellow Fuhrer with a Overmind.

How Seuches are improved:
1. Seuches now do not trigger smoke passively when covered in seuche smoke, and do not trigger smoke when both they and their target are covered in seuche smoke.
2. Seuches now attempt to heal wounded Fuhrers, Panzers, and Kaisers as well as revive dead Roachmind Co-operatives when one of the above takes damage.
3.  If the Seuche is further than minimum range for all potential targets the Seuche will instead attempt to enter minimum range before smoking.

How Kaisers are improved:
1.  Kaisers also follow the no-overkill logic of Seuches, although they are classed as attackers, not healers.
2. Kaisers regenerate smoke twice as fast as before to compensate for their extreme size.
3. Kaisers are higher tier than Fuhrers, and have higher tier Roachminds as well, with similar ability to command idle roaches.

How Burrows are improved:
- Burrows forcefully summoning a mob will now check if the mob is in the same faction as the summoner, and in the case they are not, the summoned mob will treat the summoner as enemy number one.

How Superior Animals are improved:
- Superior Animals now wake up when health changes. In the case of roaches, this is transmitted through Overmind.

## Why It's Good For The Game
Roaches weren't scary enough, Seuches in particular.

## Testing
Roach Warred with Spider Army and Roach Army until code was good.

## Changelog
:cl: Chickenish
add: Improved Roach AI
balance: Kaiser now smokes more often
fix: Sleeping Creatures should now wake up when they take damage.
/:cl:


Humonitarian (approved):
Exactly the kind of PRs we need to make maint more dangerous and interesting - improvements to AI combat tactics

TheShown911:
This should really be test merged but pop too low

PR #8561, opened 2025-03-13
Make it harder to decapitate people with rods and ammo boxes

VmpOS2NHSkhkejA9:
## About The Pull Request

A collection of adjustments to stacks and ammo boxes to make it harder to deal hundreds of damage by dropping rods and emptying ammo boxes onto people's heads:
- Dumping ammo from a box stacks it instead of splitting it into hundreds of separate items
- Ammo piles do not deal damage when falling on people unless there are at least 10 casings in the pile.
- Stacks of rods or material sheets will merge with other stacks on their tile when created and when crossing them.
- Damage from stacks falling on someone scales with the amount in the stack
Of note:
- Osmium deals more damage than other materials when falling on people.

What this PR does NOT do is fix the root problems with erismed that make this change necessary. Erismed runs into serious issues when dealing with lots of small chip damage and ends up dealing far more damage than it should. This simply makes it more difficult to abuse such issues.

## Why It's Good For The Game

Being able to decapitate people instantly with a stack of steel and a conveyor belt is probably bad.
image

You couldn't really decapitate people with ammo boxes, but you could still deal upwards of 600 damage in a single click. Which also happens to be not a good thing.
image

## Testing

- Dumped ammoboxes and magazines. Casings were properly stacked. People only took marginal damage when boxes were dumped on their heads.
- Dragged sheets of materials on top of other sheets. They merged properly.
- Crafted rods individually, they all merged into a single stack.

## Changelog
:cl:
tweak: Ammo is compressed into stacks when dumped from boxes and magazines. 
tweak: Material stacks merge when moving over each other and when created.
balance: Ammo piles deal no fall damage unless there's 10 or more casings in the pile.
balance: Stacks deal fall damage based on the number of sheets in the stack.
balance: Osmium deals 2.5x more fall damage than other stack types.
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/stack.dm
+++ b/code/game/objects/items/stacks/stack.dm
@@ -50,6 +50,10 @@
 		amount = rand(rand_min, rand_max)
 		amount = round(amount, 1) //Just in case
 	update_icon()
+
+//do this a little later because trying to merge with uninitialized stacks is an easy way to cause runtimes
+/obj/item/stack/LateInitialize()
+	. = ..()
 	if(automerge)
 		merge_loc_stacks()
 
--- a/code/game/objects/items/stacks/stack.dm
+++ b/code/game/objects/items/stacks/stack.dm
@@ -42,7 +42,7 @@
 		src.amount = amount
 
 /obj/item/stack/Initialize()
-	.=..()
+	. = ..()
 	if (!stacktype)
 		stacktype = type
 
@@ -50,6 +50,8 @@
 		amount = rand(rand_min, rand_max)
 		amount = round(amount, 1) //Just in case
 	update_icon()
+	if(automerge)
+		return INITIALIZE_HINT_LATELOAD
 
 //do this a little later because trying to merge with uninitialized stacks is an easy way to cause runtimes
 /obj/item/stack/LateInitialize()
--- a/code/game/objects/items/stacks/stack.dm
+++ b/code/game/objects/items/stacks/stack.dm
@@ -26,14 +26,15 @@
 	var/consumable = TRUE	// Will the stack disappear entirely once the amount is used up?
 	var/splittable = TRUE	// Is the stack capable of being splitted?
 	var/novariants = TRUE //Determines whether the item should update it's sprites based on amount.
+	var/automerge = FALSE // Automatically merge with stacks on the same tile?
 
 	//If either of these two are set to nonzero values, the stack will have randomised quantity on spawn
 	//Used for the /random subtypes of material stacks. any stack works
 	var/rand_min = 0
 	var/rand_max = 0
-
-
-
+	//Damage dealt to something when falling on it, per amount in the stack.
+	//IE, if this is 0.2 a stack of 120 will deal 24 base damage when falling on a mob
+	var/fall_damage_per_amount = 0.2
 
 /obj/item/stack/New(var/loc, var/amount=null)
 	.=..()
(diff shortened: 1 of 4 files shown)


code/game/objects/items/stacks/stack.dm
```diff
@@ -50,6 +51,12 @@
 		amount = round(amount, 1) //Just in case
 	update_icon()
 
+//do this a little later because trying to merge with uninitialized stacks is an easy way to cause runtimes
+/obj/item/stack/LateInitialize()
+	. = ..()
+	if(automerge)
+		merge_loc_stacks()
+
```
SirRichardFrancis, line 59: Are you sure this proc is being called at all?
It would seem that `/obj/item/stack/Initialize()` returns `INITIALIZE_HINT_NORMAL` because so does it's parent proc invoked by `.=..()`.
But for `LateInitialize()` to be called, `Initialize()` must return `INITIALIZE_HINT_LATELOAD` instead.

code/modules/projectiles/ammunition.dm
```diff
@@ -366,10 +369,25 @@
 		to_chat(usr, SPAN_NOTICE("[src] is already empty!"))
 		return
 	to_chat(usr, SPAN_NOTICE("You take out ammo from [src]."))
-	for(var/i=1 to stored_ammo.len)
-		var/obj/item/ammo_casing/C = removeCasing()
-		C.forceMove(target)
-		C.set_dir(pick(cardinal))
+
+	while(stored_ammo.len)
```
SirRichardFrancis, line 373:
```suggestion
	while(LAZYLEN(stored_ammo))
```

code/modules/projectiles/ammunition.dm
```diff
@@ -366,10 +369,25 @@
...
+
+		var/obj/item/ammo_casing/stack = removeCasing()
+		stack.forceMove(target)
+		stack.set_dir(pick(cardinal))
+
+		if(stored_ammo.len)
```
SirRichardFrancis, line 379:
```suggestion
		if(LAZYLEN(stored_ammo))
```

code/modules/projectiles/ammunition.dm
```diff
@@ -366,10 +369,25 @@
...
+			// We end on -1 since we already removed one
+			for(var/i = 1, i <= stack.maxamount - 1, i++)
+				if(!stored_ammo.len)
+					break
+					stack.update_icon()
```
SirRichardFrancis, line 384:
```suggestion
				if(!LAZYLEN(stored_ammo))
					stack.update_icon()
					break
```

code/game/objects/items/stacks/stack.dm
VmpOS2NHSkhkejA9: That's odd, it worked fine when I tested it.

VmpOS2NHSkhkejA9: Nevermind, tested again and it doesn't work. I'll investigate.

SirRichardFrancis: We have a number of `LateInitialize()` procs in the codebase that aren't seeing use because their object doesn't call for late initialization at all.
Same story with many procs using `.=..()` and `..()` when it makes no sense, just because coders had no idea what these funny dots even do.
So copypasting code under assumption that previous coder knew what they were doing isn't a safe bet on Eris.

VmpOS2NHSkhkejA9: I assumed it was good practice to call . = ..() so that if the parent proc is edited the child proc doesn't also need to be edited. If that's not the case I'll try to make note not to do that.

VmpOS2NHSkhkejA9: Fixed. Tested by spawning 5 iron ingots at once, on the same tile. Merged properly.

SirRichardFrancis (approved):

TheShown911:
image

PR #8573, opened 2025-04-10
qdel optimizations

flleeppyy:
## Changelog
:cl:
server: Qdel now yells at you when you delete something that shouldn't be deleted
/:cl:


SirRichardFrancis (changes requested):
Doesn't compile.

flleeppyy:
> Doesn't compile.

augh
and i know why.
I'll fix it In the morning

flleeppyy:
Looking back at this, i see no reason why it's failing. `refcount` is a real proc in Byond

flleeppyy:
oh. maybe if you guys weren't on 514 it would work

PR #8578, opened 2025-04-19
Fix PDA ID insert sound not playing; Normalize PDA spans

flleeppyy:
## Changelog
:cl:
fix: Inserting ID into PDA should make a sound now.
/:cl:


flleeppyy:
oh maybe i should make it so that you guys can actually uh. compile. on 516.

Humonitarian (commented):
Code seems good.

PR #8584, opened 2025-05-03
builds on 516

Arturlang:
## About The Pull Request
fixes builds on 516, also removes some dupe vars, does not fix any HTML/JS stuff (there are likely bugs)

## Why It's Good For The Game
I don't want to downgrade to 515

## Changelog

not player facing

### DIFF
--- a/code/__HELPERS/unsorted.dm
+++ b/code/__HELPERS/unsorted.dm
@@ -177,7 +177,7 @@ Turf and target are seperate in case you want to teleport some distance from a t
 			return 1
 	return 0
 
-/proc/sign(x)
+/proc/sign_r(x)
 	return x!=0?x/abs(x):0
 
 /proc/getline(atom/M, atom/N)//Ultra-Fast Bresenham Line-Drawing Algorithm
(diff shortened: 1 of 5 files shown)


SirRichardFrancis (changes requested):
Code deleted on lines starting with `//DO NOT FUCKING CHANGE THIS` and `var/obj/item/organ/external/E` don't seem to be necessary and break stuff (these variables are still expected in the code below).

Mycah142:
> Deletion of code starting with `var/obj/item/organ/external/E` doesn't seem to be necessary and going to break stuff (these variables are still expected in the code below their declaration). This probably doesn't compile as-is, please do test that.

You mean the one that errors on compile and says it's a var not used? It's copy-pasted from earlier in the code. I've already removed it.

r4d6:
Fun fact : The reason that `sign()` conflict is because it became defined in 516!
You can just remove it straight-up, or if you want it to be still compatible with 515, you can do the following : 

#if DM_VERSION < 516
/proc/sign(x)
	return x!=0?x/abs(x):0
#endif

flleeppyy:
You're going to need a lot more changes than that if you want it to be functional. sure it'll build but all html will be broken. You need to wrap all sent html in an html skeleton.

r4d6:
> You're going to need a lot more changes than that if you want it to be functional. sure it'll build but all html will be broken. You need to wrap all sent html in an html skeleton.

Huh. First time I'm hearing about an html skeleton. How does that work in 516?

flleeppyy:
Any html you send must be wrapped in the usual HTML, HEAD, BODY format.

for reference, this is what i implemented in Monkeris

https://github.com/Monkestation/Monkestation2.0/blob/master/code/__DEFINES/html_assistant.dm

Everything needs that when you're doing `<< browse(`
also you NEED to update all hrefs (there's a lot, a LOT a lot) that do <a href="?blahblah"> and replace it with <a href="byond://?blahblah">

PR #8591, opened 2025-06-10
Alternative CEV Eris map

SirRichardFrancis:
## About The Pull Request

It's funny how the bigger a PR is, the less there is to say about it…

Anyway, this PR makes it so our map files are not linked at compile time, but instead could be loaded at runtime without every other system in the codebase breaking.
This allows players to vote on which map they want to play in the next round (one alternative map is included), and removes the need to keep every single "expedition" map loaded and running in the background at all times.
So places like deep maintenance only load when someone is about to discover an entrance.
Junk tractor beam did not survive this transition, for better or worse.

This PR also includes fixes for ship engines not drawing gas from their pipe network, and airlocks not pressurizing properly a lot of the time.
Bundling up unrelated bugfixes wasn't my intention, but these bugs interfered with testing quite a lot.

## Why It's Good For The Game

Future is now, old man.

## Testing

Hours and hours of manual testing, way too many things to describe, plus new map was on the test server for a week.

## Changelog
:cl:
add: Added an alternative, smaller map of CEV Eris
del: Removed junk tractor beam
del: Removed unused shuttles such as Skipjack
fix: Fixed airlocks not pressurizing sometimes
fix: Fixed engine nozzles not using gas from connected pipes
code: Made it possible to load maps at runtime
config: Added .json config files for all maps, these govern job slots and more
/:cl:


TheShown911 (approved):
Effectively unreviewable, will be test merged.

